PR #1, opened 2020-08-06
Adding C FFI to rust prio lib and python implementations

winstrom:


tgeoghegan:
@winstrom: This change is generally useful, but I think I'm missing some context: is there a plan for some client out there to use either the C or Python bindings here or is this mainly useful so folks can experiment with this library without needing to get up to speed on Rust?

winstrom:
> @winstrom: This change is generally useful, but I think I'm missing some context: is there a plan for some client out there to use either the C or Python bindings here or is this mainly useful so folks can experiment with this library without needing to get up to speed on Rust?

This is a quick and dirty implementation so that people can get up to speed on the data flow and test that packets serialized on devices are able to be properly aggregated.  This is not meant as a production implementation.

This also allows for experimentation in a python framework which may allow for faster iteration than in Rust.  E.g. for different serialization strategies.

tgeoghegan:
Thanks for that context, Luke. My concern is that this stuff doesn't have any tests on it and we won't use this stuff in prod, so we can't be sure that changes we make to libprio_rs in the future won't break the Python bindings or the notebooks built on them. If we put something in a public repo we own, the implication is that we own that code and attest to its functionality. But clearly this stuff is useful because folks at several partners have been using these Jupyter notebooks, so we should have them somewhere that's easy to access, though longer term I don't think we should have this code in this repo if we don't really plan to support it.

winstrom:
I'd be fine with this in an "experimental" or some other such branch.  Or in a different repo expressly for the purpose of demonstration.

src/c_vec.rs
```diff
@@ -0,0 +1,181 @@
+// Copyright (c) 2020 Apple Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+//! Slice C FFI
+
+extern crate libc;
```
yuriks, line 6: External crates should only ever be declared in the root module of the library, and added to the Cargo manifest so they're pulled in as a dependency when necessary.

yuriks (changes requested):
Sorry for how long it took to get this done. I think it's definitely valuable to have a C API for the library, and this is a starting point for that. While I think I did find a bug in some of the Python library definitions, most of the comments are about structure and organization of the code.

The Rust and Python code had several formatting issues, but I didn't point them out explicitly. Rather I'd recommend running `rustfmt` on the files to make sure they follow the usual Rust standards. I don't think we have a preferred formatting tool for Python, so if you have any particular preferences for that it's probably fine to just use that for now and we can standardize on something later.

Apart from the comments in specific files, I think we should have a better separation in terms of what's an API and what's example code, especially on the Python side. On the Rust code I think we should move all of the FFI-related modules into a submodule so that they're separate from the rest of the code implementing the algorithm. Similarly on the Python side, I think keeping the function wrappers and the rest of the code+ipy notebook would be useful.

I wasn't really able to test the Python notebook, but I checked that the Rust part of it at least builds, though it needs some modifications aroudn the crate setup for that. To make this work as an `.so`, the library type needs to be set to `cdylib` in `Cargo.toml`. However, given that this is going to be a dual Rust-C library, we probably want the C compilation target to be a separate library in a subdirectory, so that both can be built as needed.

 We should probably add some tests in the future to make sure the interfaces all work correctly when called.

I know we're all busy, so, instead of going back on forth on comments in this review, if you'd like instead I can just apply my suggestions to the PR, either on a separate commit or rebasing yours, so you don't have to spent time following up on it. Let me know if you'd prefer that.

winstrom:
Yuri,

Thank you for such a great and in depth review.  I think you're spot on with the edits.  Please do just apply them.  And if it's better to create a new repo for the FFI stuff and experiments, that sounds good to me.  Please go ahead and fix all my mistakes ;).

yuriks:
Just to clarify status: I started cleaning up the branch locally, but setting up the build system for the Python stuff turned out to be quite time-consuming, so I paused this for now in lieu of working on the prio-server code for a while. I'll pick this back up in the future when I have some time.

winstrom:
Sounds good to me.  I agree concentrating on server is more important.

tgeoghegan:
It's not clear that we want to commit to maintaining a Python interface to crate `prio`, and if we ever did we'd probably make one that exposes the PCP or PPM interfaces rather than the Prio v2 stuff, so I'm going to close this PR.

PR #3, opened 2020-09-22
Configure GitHub actions

tgeoghegan:
Adds a GitHub Actions workflow that lints, builds and tests on every
push to master and every pull request. Also adds a readme so we can
display the badge.

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -1,0 +1,22 @@
+name: ci-build
+
+on:
+  push:
+    branches: [ $default-branch ]
+  pull_request:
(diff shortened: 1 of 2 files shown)


yuriks (approved):
Looks like `ubuntu-latest` [has Rust 1.46.0](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu1804-README.md), which is good. I'm guessing they just keep it up to date with the latest?
 
LGTM

tgeoghegan:
I'm not certain, but it seems that if we need specific toolchains, we can use https://github.com/marketplace/actions/rust-toolchain

PR #6, opened 2020-09-23
workflows: publish to crates.io on release

tgeoghegan:
Adds a workflow triggered by the publishing of a GitHub release which
publishes a crate to crates.io. While the repo is named `libprio-rs`,
Rust naming conventions frown on both the prefix and the suffix: `lib`
is implied by it being a library crate and `-rs` is implied by it being
a Rust crate at all
(https://rust-lang.github.io/api-guidelines/naming.html,
https://doc.rust-lang.org/stable/rust-by-example/crates/lib.html).
Accordingly, the crate is simply named `prio`.

This also adds instructions to the README for cutting releases in GitHub
and also fixes the existing status icons.

### DIFF
--- a/.github/workflows/publish-crate.yml
+++ b/.github/workflows/publish-crate.yml
@@ -1,0 +1,14 @@
+name: publish-crate
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  crate:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - run: cargo publish
+      env:
+        CARGO_REGISTRY_TOKEN: ${{ secrets.LE_AUTOMATON_CRATES_IO_API_TOKEN }}
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,10 +1,11 @@
 [package]
-name = "libprio-rs"
-version = "0.1.0"
+name = "prio"
+version = "0.1.1"
 authors = ["Josh Aas <jaas@kflag.net>", "Karl Tarbe <tarbe@apple.com>"]
 edition = "2018"
 description = "Implementation of the Prio aggregation system core: https://crypto.stanford.edu/prio/"
 license = "MPL-2.0"
+repository = "https://github.com/abetterinternet/libprio-rs"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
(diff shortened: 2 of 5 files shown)


.github/workflows/publish-crate.yml
```diff
@@ -0,0 +1,15 @@
...
+    types: [published]
+
+jobs:
+  crate:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - run: cargo login ${LE_AUTOMATON_CRATES_IO_API_TOKEN}
+      env:
+        CRATES_IO_TOKEN: ${{ secrets.LE_AUTOMATON_CRATES_IO_API_TOKEN }}
```
yuriks, line 14: Why do these use two different styles of substitution? I'm guessing the `${ }` one is just shell expansion and that env var is automatically set by Github, but the second with `${{ secrets.FOO }}` is a yaml-level template substitution?

yuriks (approved):

.github/workflows/publish-crate.yml
tgeoghegan: Yeah, the `secrets.` one fetches the token from the secrets configured on the (if you have the right privs on this repo you can [take a look](https://github.com/abetterinternet/libprio-rs/settings/secrets)) and injects its value into the environment of the `run` action. However I think I have this wrong because the environment variable will be `CRATES_IO_TOKEN`. But I need to test this before merging.

yuriks: Ah, I had assumed that all variables were added to the environment by default, but you were also aliasing it to `CRATES_IO_TOKEN` because it was also required by something else.

Looks like you don't actually need to call `cargo login` though: that just sets the token in the persistent configuration. For CI we can either pass `--token` to `publish`, or just set [`CARGO_REGISTRY_TOKEN`](https://doc.rust-lang.org/cargo/reference/config.html#registrytoken) and that's equivalent to calling `login` first (but non-persistent).

yuriks: Ah, I had assumed that all variables were added to the environment by default, but you were also aliasing it to `CRATES_IO_TOKEN` because it was also required by something else.

Looks like you don't actually need to call `cargo login` though: that just sets the token in the persistent configuration. For CI we can either pass `--token` to `publish`, or just set [`CARGO_REGISTRY_TOKEN`](https://doc.rust-lang.org/cargo/reference/config.html#registrytoken) and that's equivalent to calling `login` first (but non-persistent).

yuriks (changes requested):

tgeoghegan:
I published crate 0.1.0 while testing this change (and then `cargo yank`ed it to be safe), so I bumped the version to 0.1.1 in this PR so we can publish a new version from main after merging.

yuriks (approved):

PR #12, opened 2021-03-09
Add summation example

jvsg:
This adds an example where two clients wish to add their bit vectors using two servers.


examples/sum.rs
```diff
@@ -0,0 +1,70 @@
...
+    let data2 = data2_u32
+        .iter()
+        .map(|x| Field::from(*x))
+        .collect::<Vec<Field>>();
+
+    let (share1_1, share1_2) = client1.encode_simple(&data1).unwrap();
+    let (share2_1, share2_2) = client2.encode_simple(&data2).unwrap();
+    let eval_at = Field::from(12313);
+
+    let mut server1 = Server::new(dim, true, priv_key.clone());
```
tlepoint, line 45: It would be preferable to use different private keys on the server to better illustrate the example.

jvsg: Thanks for the suggestion. Latest commit should fix that.

tgeoghegan (approved):

tgeoghegan:
Thank you for your contribution!

PR #13, opened 2021-03-12
add accumulate_vector utility function

tgeoghegan:
Adds a function `accumulate_vector` which sums one `&[Field]` into
another independently from `Server`. Since this changes the public API
`Server::merge_total_shares` (which now returns an error if the vector
lengths do not match), this also bumps the crate version to 0.3.0 to
avoid breaking existing clients.

### DIFF
--- a/src/encrypt.rs
+++ b/src/encrypt.rs
@@ -21,19 +21,19 @@ const KEY_LENGTH: usize = 16;
 pub enum EncryptError {
     /// Base64 decoding error
     #[error("base64 decoding error")]
-    DecodeBase64Error(#[from] base64::DecodeError),
+    DecodeBase64(#[from] base64::DecodeError),
     /// Error in ECDH
     #[error("error in ECDH")]
-    KeyAgreementError,
+    KeyAgreement,
     /// Buffer for ciphertext was not large enough
     #[error("buffer for ciphertext was not large enough")]
-    EncryptionError,
+    Encryption,
     /// Authentication tags did not match.
     #[error("authentication tags did not match")]
-    DecryptionError,
+    Decryption,
     /// Input ciphertext was too small
     #[error("input ciphertext was too small")]
-    DecryptionLengthError,
+    DecryptionLength,
 }
 
 /// NIST P-256, public key in X9.62 uncompressed format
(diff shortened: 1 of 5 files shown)


tgeoghegan:
I added this function to make it easier to implement https://github.com/abetterinternet/prio-server/pull/486

jsha:
I see conflicts 👁️

tgeoghegan:
> I see conflicts eye

Whoops, I had to rebase after merging #12.

jsha (approved):
Code looks good. I have a naming nit: The fn `acccumulate_vector` should be something like `add_vector` or `merge_vector`. Accumulate is an operation that does multiple rounds of addition (or whatever); this fn does a single round.

tlepoint:
Suggestion/nit: this error is not really an 'EncryptError'. Maybe it would be worth defining another Enum?

tgeoghegan:
>The fn acccumulate_vector should be something like add_vector or merge_vector. Accumulate is an operation that does multiple rounds of addition (or whatever); this fn does a single round.

Good call: `merge_vector` would be more consistent with the existing `Server::merge_total_shares` method.

> Suggestion/nit: this error is not really an 'EncryptError'. Maybe it would be worth defining another Enum?

I agree in principle, but the difficulty is that if you add a new enum like `FiniteFieldError` to `finite_field.rs` with a variant `FiniteFieldError::InputSizeMismatchError`, then a method like `Server::aggregate` or `Server::merge_total_shares`, which returns `Result<_, EncryptError>`, has to be able to somehow convert `FiniteFieldError` into `EncryptError`. I think there's two ways to resolve that:

- Add `enum FiniteFieldError` to `finite_field.rs` and add a variant `EncryptError::FiniteFieldError(finite_field::FiniteFieldError)`. Advantages: it's obvious based on the returned error type what can go wrong in a given function or method and errors that are propagated to clients of `libprio-rs` are highly structured and easy to pattern match on. Disadvantages: Any module within `libprio-rs` that uses any other module must define error enum variants to wrap errors from other modules.
- Have a single, crate-wide Error enum that all modules use. Advantages: fewer enum cases that just wrap other errors. Disadvantages: errors are less structured and it's less obvious which errors can be expected from which methods/functions.

I was grappling with a similar problem in https://github.com/abetterinternet/prio-server/pull/416, though that situation was a little worse because of pervasive use of `anyhow::Error`. Here, since we have a smaller crate, and I'm bumping the minor version to break compat anyway, I'm going to try adding `finite_field::FiniteFieldError` and add the corresponding wrapper variant to `EncryptError`.

src/encrypt.rs
```diff
@@ -21,19 +21,22 @@ const KEY_LENGTH: usize = 16;
...
     #[error("authentication tags did not match")]
-    DecryptionError,
+    Decryption,
     /// Input ciphertext was too small
     #[error("input ciphertext was too small")]
-    DecryptionLengthError,
+    DecryptionLength,
+    /// Input sizes do not match
+    #[error("input sizes do not match")]
+    InputSizeMismatch,
```
tlepoint, line 39: Should be deleted?

tgeoghegan: Good catch, thank you!

tgeoghegan:
I'm going to rebase+merge this change tomorrow morning (March 16) unless one of the reviewers gives me a good reason not to.

jsha (approved):

Issue #10, opened 2021-03-04
Add support for more fields


cjpatton:
This implementation of Prio supports a single, 32 bit field.The [design of "Prio v3"](https://github.com/abetterinternet/prio-documents/blob/main/design-document.md) calls for the implementation of more, larger fields, where the size of the field may depend on the application. This issue is intended to track the development of finite field arithmetic for additional fields.

There are two approaches we may consider at this stage:
(1) Add support for generic FF arithmetic that is suitable for all parameters we might consider for Prio.
(2) Pick a handful of fields and develop optimized code for each.

Option (2) will eventually be needed in the long term. However, option (1) offers several advantages for the medium term. First and foremost, developing optimized code now will incur a high cost later if we end up needing to change the parameters.

My coworker, @armfazh, has developed a generic implementation for fields with a prime modulus of size < 2^126. This code is written in Go; we hope to be able to release it soon. In the meantime, I'd like us to consider whether porting his code into Rust is the best path forward for the medium term.

Here are some benchmarks for that code.
```
goos: linux
goarch: amd64
BenchmarkFp/1/Add-8         	302270450	         3.89 ns/op
BenchmarkFp/1/Sub-8         	368558247	         3.29 ns/op
BenchmarkFp/1/Mul-8         	148488744	         8.00 ns/op
BenchmarkFp/1/Inv-8         	  783758	      1492 ns/op
BenchmarkFp/9/Add-8         	308265258	         3.90 ns/op
BenchmarkFp/9/Sub-8         	365978109	         3.27 ns/op
BenchmarkFp/9/Mul-8         	149422360	         7.93 ns/op
BenchmarkFp/9/Inv-8         	  521503	      2262 ns/op
```
The first set of benchmarks are for GF(3 * 2^30  + 1) and the second set of benchmarks are for GF(7 * 2^120 + 1). Inversion is slow because we just do x^-1 = x^(p-2) (mod p) instead of using extended Euclid as libprio-rs does now. (We do this so that inversion is constant time.) I don't think this is much of a problem though, since modular inversion is rarely on the critical path for Prio. (In any case, we will eventually do better with optimized code.) Note that the runtime for the other operations is about the same, regardless of the field size. This is because the code represents field elements the same way, using a pair of `uint64`'s.

Running this code with my own Prio implementation, I get the following proof generation/verification time for a length-1000 bit-vector (the proof is based on the SIMD circuit described in Theorem 5.3 of the CRYPTO 2019 paper):
```
[...]

PR #14, opened 2021-03-19
Implement generic FF arithmetic

cjpatton:
Partially solves #10.

This change replaces the existing finite field (FF) arithmetic with a generic, constant time implementation that is suitable for any modulus p < 2^126. The code was designed by @armfazh and ported from Go into Rust by me.

The goal of this PR is to make no API changes. In particular, the same FF parameters are used for `Field`. The next step will be to rework the code to be compatible with any field, which may or may not involve breaking the API.

On a personal note, I would appreciate *all* levels of feedback, as this is the first time I've ever tried to contribute to a real Rust project. Feel free to tear it apart :)

### DIFF
--- a/src/finite_field.rs
+++ b/src/finite_field.rs
@@ -3,6 +3,8 @@
 
 //! Finite field arithmetic over a prime field using a 32bit prime.
 
+use crate::fp::FieldParameters;
+
 /// Possible errors from finite field operations.
 #[derive(Debug, thiserror::Error)]
 pub enum FiniteFieldError {
@@ -11,14 +13,22 @@ pub enum FiniteFieldError {
     InputSizeMismatch,
 }
 
-/// Newtype wrapper over u32
+/// Newtype wrapper over u128
 ///
 /// Implements the arithmetic over the finite prime field
-#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Default)]
-pub struct Field(u32);
+#[derive(Clone, Copy, Debug, PartialOrd, Ord, Hash, Default)]
+pub struct Field(u128);
+
+/// Parameters for GF(2^32 - 2^20 + 1).
+pub(crate) const SMALL_FP: FieldParameters = FieldParameters {
+    p: 4293918721,
+    p2: 8587837442,
+    mu: 17302828673139736575,
+    r2: 1676699750,
+};
 
 /// Modulus for the field, a FFT friendly prime: 2^32 - 2^20 + 1
-pub const MODULUS: u32 = 4293918721;
+pub const MODULUS: u32 = SMALL_FP.p as u32;
 /// Generator for the multiplicative subgroup
 pub(crate) const GENERATOR: u32 = 3925978153;
 /// Number of primitive roots
@@ -28,7 +38,7 @@ impl std::ops::Add for Field {
     type Output = Field;
 
     fn add(self, rhs: Self) -> Self {
-        self - Field(MODULUS - rhs.0)
+        Self(SMALL_FP.add(self.0, rhs.0))
     }
 }
 
@@ -42,14 +52,7 @@ impl std::ops::Sub for Field {
     type Output = Field;
 
     fn sub(self, rhs: Self) -> Self {
-        let l = self.0;
-        let r = rhs.0;
-
-        if l >= r {
-            Field(l - r)
-        } else {
-            Field(MODULUS - r + l)
-        }
+        Self(SMALL_FP.sub(self.0, rhs.0))
     }
 }
 
@@ -62,12 +65,8 @@ impl std::ops::SubAssign for Field {
 impl std::ops::Mul for Field {
     type Output = Field;
 
-    #[allow(clippy::suspicious_arithmetic_impl)]
     fn mul(self, rhs: Self) -> Self {
-        let l = self.0 as u64;
-        let r = rhs.0 as u64;
-        let mul = l * r;
-        Field((mul % (MODULUS as u64)) as u32)
+        Self(SMALL_FP.mul(self.0, rhs.0))
     }
 }
 
@@ -80,7 +79,6 @@ impl std::ops::MulAssign for Field {
 impl std::ops::Div for Field {
     type Output = Field;
 
-    #[allow(clippy::suspicious_arithmetic_impl)]
     fn div(self, rhs: Self) -> Self {
         self * rhs.inv()
     }
@@ -92,95 +90,76 @@ impl std::ops::DivAssign for Field {
     }
 }
 
+impl PartialEq for Field {
+    fn eq(&self, rhs: &Self) -> bool {
+        SMALL_FP.from_elem(self.0) == SMALL_FP.from_elem(rhs.0)
+    }
+}
+
+impl Eq for Field {}
+
 impl Field {
     /// Modular exponentation
     pub fn pow(self, exp: Self) -> Self {
-        // repeated squaring
-        let mut base = self;
-        let mut exp = exp.0;
-        let mut result: Field = Field(1);
-        while exp > 0 {
-            while (exp & 1) == 0 {
-                exp /= 2;
-                base *= base;
-            }
-            exp -= 1;
-            result *= base;
-        }
-        result
+        Self(SMALL_FP.pow(self.0, SMALL_FP.from_elem(exp.0)))
     }
 
     /// Modular inverse
     ///
     /// Note: inverse of 0 is defined as 0.
     pub fn inv(self) -> Self {
-        // extended Euclidean
-        let mut x1: i32 = 1;
-        let mut a1: u32 = self.0;
-        let mut x0: i32 = 0;
-        let mut a2: u32 = MODULUS;
-        let mut q: u32 = 0;
-
-        while a2 != 0 {
-            let x2 = x0 - (q as i32) * x1;
-            x0 = x1;
-            let a0 = a1;
-            x1 = x2;
-            a1 = a2;
-            q = a0 / a1;
-            a2 = a0 - q * a1;
-        }
-        if x1 < 0 {
-            let (r, _) = MODULUS.overflowing_add(x1 as u32);
-            Field(r)
-        } else {
-            Field(x1 as u32)
-        }
+        Self(SMALL_FP.inv(self.0))
     }
 }
 
 impl From<u32> for Field {
     fn from(x: u32) -> Self {
-        Field(x % MODULUS)
+        Field(SMALL_FP.elem(x as u128))
     }
 }
 
 impl From<Field> for u32 {
     fn from(x: Field) -> Self {
-        x.0
+        SMALL_FP.from_elem(x.0) as u32
     }
 }
 
 impl PartialEq<u32> for Field {
     fn eq(&self, rhs: &u32) -> bool {
-        self.0 == *rhs
+        SMALL_FP.from_elem(self.0) == *rhs as u128
     }
 }
 
 impl std::fmt::Display for Field {
     fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
-        write!(f, "{}", self.0)
+        write!(f, "{}", SMALL_FP.from_elem(self.0))
     }
+}
+
+#[test]
+fn test_small_fp() {
+    assert_eq!(SMALL_FP.check(), Ok(()));
 }
 
 #[test]
 fn test_arithmetic() {
     use rand::prelude::*;
+
     // add
-    assert_eq!(Field(MODULUS - 1) + Field(1), 0);
-    assert_eq!(Field(MODULUS - 2) + Field(2), 0);
-    assert_eq!(Field(MODULUS - 2) + Field(3), 1);
-    assert_eq!(Field(1) + Field(1), 2);
-    assert_eq!(Field(2) + Field(MODULUS), 2);
-    assert_eq!(Field(3) + Field(MODULUS - 1), 2);
+    assert_eq!(Field::from(MODULUS - 1) + Field::from(1), 0);
+    assert_eq!(Field::from(MODULUS - 2) + Field::from(2), 0);
+    assert_eq!(Field::from(MODULUS - 2) + Field::from(3), 1);
+    assert_eq!(Field::from(1) + Field::from(1), 2);
+    assert_eq!(Field::from(2) + Field::from(MODULUS), 2);
+    assert_eq!(Field::from(3) + Field::from(MODULUS - 1), 2);
 
     // sub
-    assert_eq!(Field(0) - Field(1), MODULUS - 1);
-    assert_eq!(Field(1) - Field(2), MODULUS - 1);
-    assert_eq!(Field(15) - Field(3), 12);
-    assert_eq!(Field(1) - Field(1), 0);
-    assert_eq!(Field(2) - Field(MODULUS), 2);
-    assert_eq!(Field(3) - Field(MODULUS - 1), 4);
+    assert_eq!(Field::from(0) - Field::from(1), MODULUS - 1);
+    assert_eq!(Field::from(1) - Field::from(2), MODULUS - 1);
+    assert_eq!(Field::from(15) - Field::from(3), 12);
+    assert_eq!(Field::from(1) - Field::from(1), 0);
+    assert_eq!(Field::from(2) - Field::from(MODULUS), 2);
+    assert_eq!(Field::from(3) - Field::from(MODULUS - 1), 4);
 
     // add + sub
     for _ in 0..100 {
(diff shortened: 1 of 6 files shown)


acmiyaguchi (commented):
For what it's worth, I merged these changes into a wasm-compatible fork ([link](https://github.com/abetterinternet/libprio-rs/compare/main...acmiyaguchi:ring-mods-fp?expand=1), rebasing was ugly...) and tried out a test web-app using the library ([link](https://github.com/acmiyaguchi/libprio-rs-wasm-demo/tree/fp)). For an input of `[1,1,0,1]`, I used to get `[1,1,0,1]` out. After switching to the new branch, I get `[2564090464,6858009185,4293918721,2564090464]`.  I'm guessing that this might be caused by the new field representation of `u128` vs `u32` and wasm-bindgen only supporting 64-bit integers.

cjpatton:
> For what it's worth, I merged these changes into a wasm-compatible fork ([link](https://github.com/abetterinternet/libprio-rs/compare/main...acmiyaguchi:ring-mods-fp?expand=1), rebasing was ugly...) and tried out a test web-app using the library ([link](https://github.com/acmiyaguchi/libprio-rs-wasm-demo/tree/fp)). For an input of `[1,1,0,1]`, I used to get `[1,1,0,1]` out. After switching to the new branch, I get `[2564090464,6858009185,4293918721,2564090464]`. I'm guessing that this might be caused by the new field representation of `u128` vs `u32` and wasm-bindgen only supporting 64-bit integers.

This likely has to do with the way the new code represents field elements. You need to cast field elements to their integer representation by doing `SMALL_FP.from_elem(x)`.

armfazh:
> Not sure why you're seeing the internal representations in your context. The Into and From traits do the conversion properly...

This need to be enforced with the type system, thus, the external u128 is not mapped to the u128 used internally.

acmiyaguchi:
Just to clarify, I'm using the default Serialize/Deserialize traits on the VerificationMessage struct.

https://github.com/abetterinternet/libprio-rs/blob/e0247d036716f3bdbe8702147568a7e107dc5d28/src/server.rs#L164-L172

The above is serialized in a wrapper:

```rust
    pub fn generate_verification_message(&mut self, eval_at: u32, share: &[u8]) -> JsValue {
        set_panic_hook();
        let verification = self
            .this
            .generate_verification_message(eval_at.into(), share)
            .unwrap();
        JsValue::from_serde(&verification).unwrap()
    }
```

Serde doesn't use the Into and From traits afaik, so I may need to write a custom implementation of Serialize and Deserialize if I can't rely on serializing the Field elements directly. It sounds like I can just use `::from` to go from Field to u32 and vice-versa during that process?

cjpatton:
To get from `Field` to `u32`, do `SMALL_FP.from_elem(x)`.
To get from `u32` to `Field`, do `SMALL_FP.elem(x as u128)`.
This is what the Into/From traits do.

acmiyaguchi:
Aha, I see now. I'm using `reconstruct_shares` which returns a vector of Field elements:

https://github.com/abetterinternet/libprio-rs/blob/e0247d036716f3bdbe8702147568a7e107dc5d28/src/util.rs#L129

I was just showing the serialized form of this directly, but casting the elements into integers was what I needed.

```
    let reconstructed: Vec<u32> = util::reconstruct_shares(share1.as_ref(), share2.as_ref())
        .unwrap()
        .iter()
        .map(|el| u32::from(*el))
        .collect();
```

Would this be considered an API compatible change from the perspective of https://github.com/abetterinternet/prio-server? I don't have much context from that perspective.

cjpatton:
> Would this be considered an API compatible change from the perspective of https://github.com/abetterinternet/prio-server? I don't have much context from that perspective.

Golly, I hope so. That's a question for @tgeoghegan

src/util.rs
```diff
@@ -110,7 +110,7 @@ pub fn serialize(data: &[Field]) -> Vec<u8> {
 
 /// Get a vector of field elements from a byte slice
 pub fn deserialize(data: &[u8]) -> Vec<Field> {
-    let field_size = std::mem::size_of::<Field>();
+    let field_size = 4;
```
tgeoghegan, line 113: Was `std::mem::size_of` not working? While I don't expect the width of `u32` to change I'd still prefer not to hard code this.

src/prng.rs
```diff
@@ -61,7 +61,7 @@ fn random_field_from_seed(seed: &[u8], length: usize) -> Vec<Field> {
         cipher.apply_keystream(&mut buffer);
 
         // rejection sampling
-        for chunk in buffer.chunks_exact(std::mem::size_of::<Field>()) {
+        for chunk in buffer.chunks_exact(4) {
```
tgeoghegan, line 64: Same comment as in `src/util.rs`: let's put this back to `size_of` if we can.

src/fp.rs
```diff
@@ -0,0 +1,295 @@
+// Copyright (c) 2021 The Authors
+// SPDX-License-Identifier: MPL-2.0
+
+use rand::prelude::*;
+use rand::Rng;
+
+// Parameters of a finite field with prime p < 2^126. Field elements are represented as u128's.
+#[derive(Debug)]
+pub struct Fp {
```
tgeoghegan, line 9: What does `Fp` stand for? Can we use a more descriptive module and struct name than `Fp`?

src/fp.rs
```diff
@@ -0,0 +1,295 @@
+// Copyright (c) 2021 The Authors
+// SPDX-License-Identifier: MPL-2.0
+
+use rand::prelude::*;
+use rand::Rng;
+
+// Parameters of a finite field with prime p < 2^126. Field elements are represented as u128's.
```
tgeoghegan, line 7: Rust doccomments should start with `///` so that various documentation tools (like docs.rs) can parse them: https://doc.rust-lang.org/beta/rust-by-example/meta/doc.html

Similarly, the comments next to the fields on `Fp` can be converted into doccomments

src/fp.rs
```diff
@@ -0,0 +1,295 @@
...
+#[derive(Debug)]
+pub struct Fp {
+    pub p: u128,  // the prime modulus p
+    pub p2: u128, // p * 2
+    pub mu: u64,  // mu = -p^(-1) mod 2^64
+    pub r2: u128, // r2 = (2^128)^2 mod p
+}
+
+impl Fp {
```
tgeoghegan, line 16: Could there be an associated function `Fp::new(p: u128)` that computes `p2`, `mu` and `r2`, reducing the risk that a caller will compute them wrong?

edit: I see there is such a function implemented in the tests. Were you trying to avoid pulling `BigNum` into the regular dependencies?

src/finite_field.rs
```diff
@@ -3,6 +3,8 @@
 
 //! Finite field arithmetic over a prime field using a 32bit prime.
```
tgeoghegan, line 4:
```suggestion
//! Finite field arithmetic over a prime field using a 128 bit prime.
```

src/finite_field.rs
```diff
@@ -28,7 +37,7 @@ impl std::ops::Add for Field {
     type Output = Field;
 
     fn add(self, rhs: Self) -> Self {
-        self - Field(MODULUS - rhs.0)
+        Self(SMALL_FP.add(self.0, rhs.0))
```
tgeoghegan, line 40: This doesn't need to be `Self::from` because `Fp` already guarantees everything is done `mod p`, right?

src/finite_field.rs
```diff
@@ -92,95 +89,71 @@ impl std::ops::DivAssign for Field {
...
     fn from(x: Field) -> Self {
-        x.0
+        SMALL_FP.from_elem(x.0) as u32
     }
 }
 
 impl PartialEq<u32> for Field {
     fn eq(&self, rhs: &u32) -> bool {
-        self.0 == *rhs
+        SMALL_FP.from_elem(self.0) as u32 == *rhs
```
tgeoghegan, line 128: [Casting `u128` to `u32` will truncate](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=60c255ef47135400efb9a971b10f3cd5). I think you could instead promote `rhs` to `u128` to safely compare.

src/fp.rs
```diff
@@ -0,0 +1,295 @@
...
+        z.wrapping_add(m & self.p2)
+    }
+
+    pub fn sub(&self, x: u128, y: u128) -> u128 {
+        let (z, carry) = x.overflowing_sub(y);
+        let m = 0u128.wrapping_sub(carry as u128);
+        z.wrapping_add(m & self.p2)
+    }
+
+    pub fn mul(&self, x: u128, y: u128) -> u128 {
```
tgeoghegan, line 29: It'd be good for all the `pub` functions on `Fp` to have doccomments, especially `mul` which merits some more explanation. At least a link to a reference describing how it works.

tgeoghegan (changes requested):

tgeoghegan:
re: `from` and API compatibility: I'm not really worried about breaking `prio-server`, for a couple reasons. First, we would have to explicitly adopt a new `libprio-rs` version which would give us a chance to fix tests that break, and I assume we will eventually make breaking API changes to this crate. Second, I don't believe the usage of `reconstruct_shares` in `prio-server` should be a problem. We only ever do comparisons on `Vec<Field>` and I don't think we ever convert them to primitives.

Now, on the subject of primitive->`Field` conversions: I'm worried about unsafe conversions from `Field(u128)` to `u32`. Right now, we have `impl From<Field> for u32`, which can't be safe because `u128` is bigger than `u32`. As implemented, it'll silently truncate values larger than `u32::MAX`. I think what we should do is instead implement `TryFrom<Field> for u32` (and also u16 and u64 and all the integer sizes smaller than `u128`; [see for example the implementations of `From` and `TryFrom` on `u32` itself](https://doc.rust-lang.org/std/primitive.u32.html)). I know @cjpatton is aiming not to break the API with this particular PR, so for now, rather than removing `impl From<Field> for u32`, I think it should use `u32::try_from<u128>` and `panic!` if it fails. This won't break clients who are converting back and forth from `u32` to `Field`, but it also won't silently truncate if someone is doing something weird. Providing all the `TryFrom` impls is a bit of a chore so feel free to punt that to a future PR (please record an issue if you do).

src/util.rs
cjpatton: I added a function `size()` to `Fp` that returns the number of bytes needed to encode field elements as integers and replaced the hard-coded value with `SMALL_FP.size()`. Here and in prng.rs.

src/prng.rs
cjpatton: Oh, is there a trait for this that `Fp` could implement? (See comment above.)

src/fp.rs
cjpatton: "Field parameters". How about `FieldParam`?

src/fp.rs
cjpatton: It seems sensible to me to avoid adding new dependencies to the library wherever possible, in order to keep it as small possible. Though TBH I don't know enough about Rust internals to know how much this actually matters.

I'd be happy to implement `Fp::new(p : u128)`. Note that this would actually add `modinverse` as a dependency in addition to `num-bigint`. (The latter doesn't do modular inverse AFACIT.)

armfazh: Take a look on Alg.3, it's easy to derive `mu` so you can remove such a dependency.
https://eprint.iacr.org/2017/1057.pdf

src/fp.rs
cjpatton: Done. Do let me know if there's anything else I can do to improve documentation.

cjpatton: I changed `mod fp` to `pub mod fp` in `lib.rs` in order to make example code in the docucomments compile. I'm not sure we should actually expose this module in the public API.

src/finite_field.rs
cjpatton: The prime hasn't changed, so this comment is correct as-is.

src/finite_field.rs
cjpatton: I'm not sure what your question is about. My understanding of what this line does: grab the `u128` underlying `self` and `rhs`, wrap the sum in a `Field`, and returns the result.

src/finite_field.rs
cjpatton: Done.

src/fp.rs
cjpatton: I added docucomments to all of the methods. @armfazh, can you provide a pointer that describes Montgomery multiplication? I found https://en.wikipedia.org/wiki/Montgomery_modular_multiplication ... not sure of the quality of the description there.

cjpatton:
> I know @cjpatton is aiming not to break the API with this particular PR, so for now, rather than removing `impl From<Field> for u32`, I think it should use `u32::try_from<u128>` and `panic!` if it fails. This won't break clients who are converting back and forth from `u32` to `Field`, but it also won't silently truncate if someone is doing something weird. Providing all the `TryFrom` impls is a bit of a chore so feel free to punt that to a future PR (please record an issue if you do).

My main concern about breaking the API is its impact on prio-server, If you're not concerned about this, then I'm fine breaking the API.

However, I think we should punt to a future PR for the following reason. The conversion between `Field` and `u32` is *currently* safe because the modulus is hard-coded to a 32-bit prime. This will change once we allow larger fields in a future PR. In that PR we'll need to carefully consider how to safely convert between field elements and primitive types. For example, we may define conversion only to the primitive type that contains the modulus. For example, for a 64-bit prime field, we may only define conversion to u64s.

src/fp.rs
cjpatton: Neat! This leaves the `num-bigint` dependency. Anyway we can get rid of this with some mathmagic?

src/finite_field.rs
tgeoghegan: Oh of course, you are correct!

src/finite_field.rs
tgeoghegan: Sorry, I was being unnecessarily terse. I was trying to confirm that this intentionally avoids invoking `Field::from()` to convert the result of `Fp:add()` into `Field`. In any case, I see now my question is a non sequitur, because there is no `Field::from<u128>()` that could be used here.

src/fp.rs
tgeoghegan: Why not `FieldParameters`? I think we can afford the extra five characters.

cjpatton: Works for me.

src/fp.rs
tgeoghegan: One option would be to provide a method `FieldParameters::validate(&self) -> bool`, tagged with `#[cfg(test)]`, which would take a `FieldParameters` and return true if the four values check out against each other. Then `FieldParameters::validate` could be used in the `Field` module's own tests to verify that `SMALL_FP` is valid.

cjpatton: That's a great idea! Working..

cjpatton: Done.

src/fp.rs
armfazh: Montgomery REDC
 https://www.ams.org/journals/mcom/1985-44-170/S0025-5718-1985-0777282-X/S0025-5718-1985-0777282-X.pdf

src/fp.rs
```diff
@@ -0,0 +1,367 @@
...
+    pub fn size(&self) -> usize {
+        (16 - (self.p.leading_zeros() / 8)) as usize
+    }
+
+    #[cfg(test)]
+    pub fn check(&self) -> Result<(), &'static str> {
+        use modinverse::modinverse;
+        use num_bigint::{BigInt, ToBigInt};
+
+        let err_modulus_too_large = "p > 2^126";
```
tgeoghegan, line 216: nit: this is fine for now since the only caller of this method is only going to print any error they encounter, but if we end up with more errors from `fp`, it'd be best to define an `fp::Error` enum with variants for specific error cases (like [finite_field does](https://github.com/abetterinternet/libprio-rs/blob/e0247d036716f3bdbe8702147568a7e107dc5d28/src/finite_field.rs#L8)).

tgeoghegan (approved):
I agree with your point about `u32` <-> `Field` being safe, even if `Field` wraps `u128`, because everything is modulo 4293918721, and that we can make this safer when we expand the `Field` object model. LGTM, thank you for this contribution!

src/fp.rs
cjpatton: Added the reference.

src/fp.rs
cjpatton: Ack. Yeah, this seems like it will be handy. I'll keep this in mind for the future.

src/prng.rs
cjpatton: Resolving this, since size_of doesn't seem appropriate here.

cjpatton:
Squashed.

src/fp.rs
cjpatton: Reverted this change. `prio::fp` is now private.

PR #17, opened 2021-03-26
fp: Fix copyright

cjpatton:


tgeoghegan:
After some offline discussion, and checking the precedent set in [Boulder](https://github.com/letsencrypt/boulder/blob/main/LICENSE.txt), here's what I'd like to do:

- Add a line to the top of `libprio-rs/LICENSE.txt` which reads `Copyright 2021 ISRG, except where otherwise noted.  All rights reserved.`
- Put no copyright attribution on any files that don't already have one (that is, leave the Apple copyright notices wherever they currently occur, but delete the first line from `fp.rs`)

The intent here is that any new files we add to the project will default to ISRG's copyright, and perhaps someday we can ask Apple nicely if they want to assign copyright over the existing files to ISRG. My naive understanding here is that copyright ownership will probably only matter if we ever want to change the license on `libprio-rs`, so I'm not especially concerned about what entity owns the copyright, since MPL guarantees that everyone can do pretty much whatever they want anyway, but it seems to me like we should aim to have as few copyright holders as possible, just for simplicity's sake.

Does that seem acceptable @cjpatton? @bdaehlie? I am not an IP lawyer so it's likely I've missed something important.

cjpatton:
This should work for me! So what should be header of `src/fp.rs`? Should it simply be:
```
```

with no Copyright line?

tgeoghegan:
> This should work for me! So what should be header of `src/fp.rs`? Should it simply be:
> 
> ```
> // SPDX-License-Identifier: MPL-2.0
> ```
> 
> with no Copyright line?

Yep, that should work. I'm not certain we need the SPDX thing (if we think of the crate as the atom of software in the Rust world the `license` line in `Cargo.toml` is presumably authoritative) but it's good for consistency with other files. Thanks for being accommodating on this matter!

cjpatton:
Removed the line, should be ready to merge.

tgeoghegan (approved):

PR #22, opened 2021-04-05
Use any FieldElement and clean up field API

### LINKED ISSUE #10: Add support for more fields

cjpatton:
Closes #10.

This PR includes the following changes:
1. Generalizes the code for generating and validating proofs to work any implementation of the `FieldElement` trait.
2. Renames `Field` to `Field32` in order to signify that it implements a 32-bit field.
3. Renames the `finite_field` module to simply `field`.
4. Renames enum `FiniteFieldError` to `FieldError`.


src/server.rs
```diff
@@ -143,17 +143,18 @@ impl Server {
...
     /// Choose a random point for polynomial evaluation
     ///
     /// The point returned is not one of the roots used for polynomial
     /// evaluation.
-    pub fn choose_eval_at(&self) -> Field {
+    pub fn choose_eval_at(&self) -> F {
+        let mut rng = rand::thread_rng();
         loop {
-            let eval_at = Field::from(rand::random::<u32>());
+            let eval_at = F::rand(&mut rng);
```
tgeoghegan, line 157: It seems like everywhere `F::rand` is used, the caller has to provide a `rand::Rng`, which is a bit of a hassle. I see the value in exposing a variant that lets the caller provide an RNG, but could we perhaps have two variants, one that takes no arguments and defaults to `rand::thread_rng()` and another that lets the caller provide a `rand::Rng`?

src/field.rs
```diff
@@ -226,6 +245,31 @@ macro_rules! make_field {
...
+                let mut int = 0;
+                for i in 0..Self::BYTES {
+                    int |= (bytes[i] as u128) << (i << 3);
+                }
+
+                if int >= $fp.p {
+                    return Err(FieldError::FromBytesModulusOverflow);
+                }
+                Ok(Self($fp.elem(int)))
+            }
```
tgeoghegan, line 271: It bugs me slightly that these methods aren't symmetric. It also seems unfortunate that `append_to` is only capable of appending to a `Vec<u8>`. It would be more generally useful to provide routines that allow transforming a `FieldElement` to and from `[u8]`, and then you could do anything you want with the byte array, whether it's appending to a `Vec` or [writing to an `std::io::Write`](https://doc.rust-lang.org/std/io/trait.Write.html#tymethod.write).

I wonder: what if instead of these two methods on `FieldElement`, you had `FieldElement` require `TryFrom<[u8]>`, and provided `From $elem for [u8]`?

tgeoghegan (changes requested):
Looks great! While it's a lot of lines changed, I'm pleased at how gracefully `F: FieldElement` replaces `Field` all over the place. All I have is some obtuse gripes about error handling and interfaces on `FieldElement`.

src/client.rs
```diff
@@ -4,31 +4,33 @@
...
 }
 
-impl Client {
+impl<F: FieldElement> Client<F> {
     /// Construct a new Prio client
     pub fn new(dimension: usize, public_key1: PublicKey, public_key2: PublicKey) -> Option<Self> {
         let n = (dimension + 1).next_power_of_two();
 
-        if 2 * n > Field::generator_order() as usize {
+        if F::Integer::try_from(2 * n).unwrap() > F::generator_order() {
```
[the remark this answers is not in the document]
cjpatton: Doine. (Please take a look at the diff ... I had to refactor things a bit.)

src/util.rs
```diff
@@ -96,46 +99,32 @@ pub fn unpack_proof_mut(proof: &mut [Field], dimension: usize) -> Option<Unpacke
...
-    let mut vec = Vec::with_capacity(data.len() / field_size);
-    use std::convert::TryInto;
-
-    for chunk in data.chunks_exact(field_size) {
-        // TODO(cjpatton) Implement FieldElement::from_bytes() that decodes field elements from a
-        // string with FieldParameters::size() bytes.
-        let integer = <Field as FieldElement>::Integer::from_le_bytes(chunk.try_into().unwrap());
-        vec.push(Field::from(integer));
+pub fn deserialize<F: FieldElement>(data: &[u8]) -> Result<Vec<F>, FieldError> {
+    let mut vec = Vec::<F>::with_capacity(data.len() / F::BYTES);
```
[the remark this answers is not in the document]
cjpatton: Done. (Needed to refactor the code to propagate this error.)

src/util.rs
```diff
@@ -96,46 +99,32 @@ pub fn unpack_proof_mut(proof: &mut [Field], dimension: usize) -> Option<Unpacke
...
 ///
 /// Returns None, when array lengths are not equal.
-pub fn reconstruct_shares(share1: &[Field], share2: &[Field]) -> Option<Vec<Field>> {
+pub fn reconstruct_shares<F: FieldElement>(share1: &[F], share2: &[F]) -> Option<Vec<F>> {
     if share1.len() != share2.len() {
         return None;
     }
 
-    let mut reconstructed = vector_with_length(share1.len());
+    let mut reconstructed = vec![F::zero(); share1.len()];
```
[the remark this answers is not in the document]
cjpatton: Done.

src/server.rs
```diff
@@ -189,12 +190,13 @@ pub fn generate_verification_message(
     mem.points_h[0] = *unpacked.h0;
 
     // set points_f and points_g
+    let one = F::root(0).unwrap();
```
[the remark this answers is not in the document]
cjpatton: I'll go with `F::one()`, since this is always the multiplicative identity.

src/server.rs
cjpatton: Hmm, well already have  `random_field_from_seed` which maps a seed to a field element. The client and aggregators need to agree on how to do this procedure. This is basically doing the same thing as the current `fn rand<R: Rng + ?Sized>(rng: &mut R) -> Self` API, which is a little messy. I think we should just nix `fn rand<R: Rng + ?Sized>(rng: &mut R) -> Self;`  call and just have `fn rand() -> Self`.

src/field.rs
cjpatton: As discussed on the design call, @acmiyaguchi will look into replacing these functions with implementations of whatever traits are needed by serde. See https://github.com/abetterinternet/libprio-rs/issues/24. I've added TODOs in the code to remind us.

src/client.rs
```diff
@@ -46,20 +71,21 @@ impl Client {
...
 
     /// Construct a pair of encrypted shares based on the input data.
-    pub fn encode_simple(&mut self, data: &[Field]) -> Result<(Vec<u8>, Vec<u8>), EncryptError> {
-        let copy_data = |share_data: &mut [Field]| {
+    pub fn encode_simple(&mut self, data: &[F]) -> Result<(Vec<u8>, Vec<u8>), ClientError> {
+        let copy_data = |share_data: &mut [F]| {
             share_data[..].clone_from_slice(data);
         };
         self.encode_with(copy_data)
+            .or_else(|err| Err(ClientError::Encrypt(err)))
```
tgeoghegan, line 79:
```suggestion
        Ok(self.encode_with(copy_data)?)
```
Because we have `Encrypt(#[from] EncryptError)`, we get an appropriate `From<EncryptError> for ClientError` for free, [and the `?` operator will automatically use that `From` to convert the error](https://doc.rust-lang.org/stable/reference/expressions/operator-expr.html#the-question-mark-operator).

src/server.rs
```diff
@@ -74,10 +77,10 @@ impl Server {
     }
 
     /// Decrypt and deserialize
-    fn deserialize_share(&self, encrypted_share: &[u8]) -> Result<Vec<Field>, ServerError> {
+    fn deserialize_share(&self, encrypted_share: &[u8]) -> Result<Vec<F>, ServerError> {
         let share = decrypt_share(encrypted_share, &self.private_key)?;
         Ok(if self.is_first_server {
-            deserialize(&share)
+            deserialize(&share).or_else(|err| Err(ServerError::Serialize(err)))?
```
tgeoghegan, line 83:
```suggestion
            deserialize(&share)?
```
[`?` will automatically use an appropriate `From` implementation to convert errors to the appropriate type](https://doc.rust-lang.org/stable/reference/expressions/operator-expr.html#the-question-mark-operator) and `thiserror` made you a `From` due to `Serialize(#[from] SerializeError)` so you don't need to wrap the error yourself.

tgeoghegan (changes requested):
A couple nits about the `?` operator but this is essentially good to go. Please address those, then squash as you see fit, and I'll merge. Thanks!

src/client.rs
tgeoghegan: Oh, and you might wonder why we have to do `Ok(thing_that_fails()?))`: I believe it's because while we get `From<EncryptError> for ClientError` for free, we don't get `from<Result<T, ClientError>> for Result<T, EncryptError>`, or at least `?` doesn't automagically invoke that conversion, so we have to do `?` to unwrap the `Result` from `self.encode_with`, and then wrap it back in `Ok` to be returned by this method.

cjpatton: Damn, that's just magic. Done.

src/server.rs
cjpatton: Also magic. Also done.

cjpatton:
Squashed and ready to merge.

tgeoghegan (approved):

Issue #24, opened 2021-04-07
Unify and clean up serialization logic


cjpatton:
Data serialization is currently ad-hoc throughout this library. For example, The `FieldElement` trait specifies methods for writing field elements to or reading field elements from a buffer. We would like to replace this with whatever serialization logic is idiomatic for Rust. (See https://github.com/abetterinternet/libprio-rs/pull/22#discussion_r608236929.)

When thinking about how to rework this, we should also think about how protocol messages will be encoded on the wire.

cc/ @acmiyaguchi

tgeoghegan:
We also have this messy business where `util::serialize` and `util::deserialize` convert back and forth between `Vec<F: FieldElement>` and `Vec<u8>`. We then apply a further deserialization step via `unpack_proof[_mut]` to deserialize `[F: Field]` into a Rust struct `UnpackedProof[Mut]`, which somewhat perversely also contains the input share.

I wonder now if all of `util.rs` could be replaced by one or more structs with a better name than `UnpackedProof` (it's obviously unpacked by virtue of being a struct and not just `[u8]`, and it doesn't just contain the proof) that are appropriate decorated with `serde` tags so that we can convert back and forth from byte arrays.

acmiyaguchi:
I've spent some time trying out bincode with structs that implement the serde traits in #29. I'm not sure it's worth pursuing (but I could be convinced otherwise).

It would be nice to use an off-the-shelf data format like bincode, but it performs poorly with the serialization of FieldElements from a u128 into a packed bytearray. It treats arrays as a tuple of (u64, [u8]) which ends up being wider than just storing the u128 alone. One way to overcome this would be to pack the values into an anonymous tuple with an entry for each byte (but there isn't a way to do this without it looking like a big hack). 

The serialize/deserialize methods that exist in utils _could_ adopt traits for Serializer and Deserializer, but it doesn't add any value (e.g. make it easier write language bindings). It's also verbose to boot.

In any case, #27 adds the minimum amount of code for needed for the [serde-serialize feature in wasm-bindgen](https://rustwasm.github.io/docs/wasm-bindgen/reference/arbitrary-data-with-serde.html).

cjpatton:
My main concern is this serialization function in `field.rs` (and its corresponding deserializer, `read_from`):
```rust
fn append_to(&self, bytes: &mut Vec<u8>) {
   // Convert field element to its integer representation
   let int = $fp.from_elem(self.0); 

   // Encode integer as bytes
[...]

PR #25, opened 2021-04-07
Return `Result` from `Server` methods

tgeoghegan:
The remaining methods on `Server` which returned `Option<Something>` now return `Result<Something, ServerError>`, to make it easier for `prio-server` to tell errors apart ([#550](https://github.com/abetterinternet/prio-server/issues/550)).

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -12,6 +12,9 @@ pub enum SerializeError {
     /// element of the field.
     #[error("last chunk of bytes is incomplete")]
     IncompleteChunk,
+    /// Emitted by `unpack_proof[_mut]` if the serialized share+proof has the wrong length
+    #[error("serialized input has wrong length")]
+    UnpackInputSizeMismatch,
     /// Finite field operation error.
     #[error("finite field operation error")]
     Field(#[from] FieldError),
@@ -34,6 +37,7 @@ pub fn vector_with_length<F: FieldElement>(len: usize) -> Vec<F> {
 }
 
 /// Unpacked proof with subcomponents
+#[derive(Debug)]
 pub struct UnpackedProof<'a, F: FieldElement> {
     /// Data
     pub data: &'a [F],
@@ -48,6 +52,7 @@ pub struct UnpackedProof<'a, F: FieldElement> {
 }
 
 /// Unpacked proof with mutable subcomponents
+#[derive(Debug)]
 pub struct UnpackedProofMut<'a, F: FieldElement> {
     /// Data
     pub data: &'a mut [F],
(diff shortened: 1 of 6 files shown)


jsha:
I'm excited about this change! It looks like it incorporates moving a big chunk of code besides the change from `Option` to `Result`. Would you mind splitting up the "move code" and "Option -> Result" into two separate PRs? Makes them much easier to review.

tgeoghegan:
> I'm excited about this change! It looks like it incorporates moving a big chunk of code besides the change from `Option` to `Result`. Would you mind splitting up the "move code" and "Option -> Result" into two separate PRs? Makes them much easier to review.

Yes, good idea!

tgeoghegan:
There you are @jsha: I will move the structs and functions from `util.rs` into `proof.rs` in a subsequent pull request.

jsha (approved):

src/util.rs
```diff
@@ -5,6 +5,14 @@
 
 use crate::field::{FieldElement, FieldError};
 
+/// Possible errors from proofs
+#[derive(Debug, thiserror::Error)]
+pub enum ProofError {
+    /// Input sizes do not match
+    #[error("input sizes do not match")]
+    InputSizeMismatch,
+}
```
cjpatton, line 14: The name "ProofError" sound as if this might be returned if the input-validation protocol failed. Looking at the error that's propagated, this seems to happen only when "unpacking" a proof. Perhaps this is better cast as another "SerializeError" type error? Something like "SerializeError::UnpackInputSizeMismatch"?

src/util.rs
```diff
@@ -176,14 +189,35 @@ pub mod tests {
...
+        let len = proof_length(dim);
+
+        let share = vec![Field64::from(0); len];
+        unpack_proof(&share, dim).unwrap();
+
+        let short_share = vec![Field64::from(0); len - 1];
+        assert_matches!(
+            unpack_proof(&short_share, dim),
+            Err(ProofError::InputSizeMismatch)
+        );
```
cjpatton, line 220: Could you write this assertion without `assert_matches!()`? Something like
```rust
assert_eq!(unpack_proof(...), Err(...));
```

cjpatton (commented):
Minor quip about naming. Otherwise this looks good.

src/util.rs
tgeoghegan: My compiler won't let me: it complains that:
```
error[E0369]: binary operation `==` cannot be applied to type `std::result::Result<UnpackedProofMut<'_, field::Field32>, util::ProofError>`
   --> src/util.rs:202:9
    |
202 | /         assert_eq!(
203 | |             unpack_proof_mut(&mut short_share, dim),
204 | |             Err(ProofError::InputSizeMismatch)
205 | |         );
    | |          ^
    | |          |
    | |__________std::result::Result<UnpackedProofMut<'_, field::Field32>, util::ProofError>
    |            std::result::Result<_, util::ProofError>
    |
    = note: this error originates in a macro (in Nightly builds, run with -Z macro-backtrace for more info)
```

We could do:

```
    let err = unpack_proof_mut(&mut short_share, dim).unwrap_err();
    assert_eq!(err, ProofError::InputSizeMismatch);
```
... but for that to work I'd have to provide a `PartialEq` impl on `ProofError` and I think it's easier to use `assert_matches!`.

cjpatton: Works for me.

src/util.rs
tgeoghegan: That's a good point. The tricky thing is that I want to make a follow up PR where I move this enum, `struct UnpackedProof`, `struct UnpackedProofMut`, `unpack_proof` and `unpack_proof_mut` to a new module `proof.rs` (the idea is to remove stuff from `util.rs` with the eventual goal of deleting that module). But arguably, all the stuff here in `util.rs` has to do with serialization, so maybe instead of `proof.rs` my later PR should just rename `util.rs` to `serialize.rs`...

OK, given that we want to revisit serialization in the near future anyway (#24), I think I'm just going to punt on moving anything from `util.rs` into a new module, and for now I will add a case to `SerializeError` as you suggest.

cjpatton: IMO `serialize` is a better name for the stuff in this package than `proof`.

cjpatton (approved):

PR #27, opened 2021-04-10
Add serde for VerificationMessage

acmiyaguchi:
This adds some basic serialization for the VerificationMessage. I need to take a deeper dive to tackle #24 (for example, why do field elements serialize to a variable number of bytes when it's derived from a u128 and the field parameters [here](https://github.com/abetterinternet/libprio-rs/blob/b68a16b04cfa0d62bc3ba02ae264685a2930116e/src/field.rs#L275) and [here](https://github.com/abetterinternet/libprio-rs/blob/b68a16b04cfa0d62bc3ba02ae264685a2930116e/src/fp.rs#L187)? edit: nevermind, I see now). 

This should be enough to allow usage of the public api in an application that wants to use serde for transporting messages. I used json_serde in the tests here.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,12 +16,14 @@ base64 = "0.12.3"
 rand = "0.7"
 ring = "0.16.15"
 thiserror = "1.0"
+serde = { version = "1.0", features = ["derive"] }
 
 [dev-dependencies]
 assert_matches = "1.5.0"
 criterion = "0.3"
 modinverse = "0.1.0"
 num-bigint = "0.4.0"
+serde_json = "1.0"
 
 [[bench]]
 name = "fft"
--- a/examples/sum.rs
+++ b/examples/sum.rs
@@ -8,12 +8,12 @@ use prio::server::*;
 fn main() {
     let priv_key1 = PrivateKey::from_base64(
         "BIl6j+J6dYttxALdjISDv6ZI4/VWVEhUzaS05LgrsfswmbLOgN\
-        t9HUC2E0w+9RqZx3XMkdEHBHfNuCSMpOwofVSq3TfyKwn0NrftKisKKVSaTOt5seJ67P5QL4hxgPWvxw==",
+         t9HUC2E0w+9RqZx3XMkdEHBHfNuCSMpOwofVSq3TfyKwn0NrftKisKKVSaTOt5seJ67P5QL4hxgPWvxw==",
     )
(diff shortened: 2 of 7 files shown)


acmiyaguchi:
@tgeoghegan do you mind reviewing this?

examples/sum.rs
```diff
@@ -8,12 +8,12 @@ use prio::server::*;
 fn main() {
     let priv_key1 = PrivateKey::from_base64(
         "BIl6j+J6dYttxALdjISDv6ZI4/VWVEhUzaS05LgrsfswmbLOgN\
-        t9HUC2E0w+9RqZx3XMkdEHBHfNuCSMpOwofVSq3TfyKwn0NrftKisKKVSaTOt5seJ67P5QL4hxgPWvxw==",
+         t9HUC2E0w+9RqZx3XMkdEHBHfNuCSMpOwofVSq3TfyKwn0NrftKisKKVSaTOt5seJ67P5QL4hxgPWvxw==",
```
cjpatton, line 11: What accounts for this extra space?

acmiyaguchi: This is just the output of `cargo fmt` on my machine. I wonder if this is different between versions of rust? As for the leading whitespace, it's ignored as per https://doc.rust-lang.org/reference/tokens.html#string-literals.

cjpatton: Ah interesting. I guess that's why `cargo fmt` feels comfortable modifying this string :)

Cargo.toml
```diff
@@ -16,12 +16,15 @@ base64 = "0.12.3"
 rand = "0.7"
 ring = "0.16.15"
 thiserror = "1.0"
+serde = { version = "1.0", features = ["derive"] }
+
```
cjpatton, line 20: nit: Remove this `\n`, since the other headers in this file have a single new line after them.

cjpatton (approved):
LGTM modulo a nit.

tgeoghegan (commented):
LGTM; thanks! Could I ask you to rewrite the commit message "Add serialization for VerificationMessage" so it also discusses deriving `Serialize` and `Deserialize` for all the `FieldElement` impls? Let me know when you're happy with this and I can merge it.

acmiyaguchi:
If the latest commit message looks good to you, feel free to merge anytime.

tgeoghegan (approved):

PR #29, opened 2021-04-13
wip: Replace FieldElement trait methods with Serde traits

acmiyaguchi:
This is a work in progress of replacing serialization internally with serde traits for consistency. This will be using bincode to pack proofs before they are encrypted.

### helpful links
* https://serde.rs/custom-serialization.html
* https://docs.serde.rs/serde/de/trait.Visitor.html#method.visit_bytes
* https://docs.rs/bincode/1.3.3/bincode/enum.ErrorKind.html


src/prng.rs
```diff
@@ -60,15 +61,14 @@ fn random_field_from_seed<F: FieldElement>(seed: &[u8], length: usize) -> Vec<F>
...
-            match F::read_from(chunk) {
+            match bincode::deserialize(chunk) {
                 Ok(x) => {
                     output[output_written] = x;
                     output_written += 1;
                     if output_written == length {
                         break;
                     }
                 }
-                Err(FieldError::FromBytesModulusOverflow) => (), // reject this sample
```
acmiyaguchi, line 71: Handling errors is kind of ugly. Since bincode [implements it's own ErrorKind](https://docs.rs/bincode/1.3.3/bincode/enum.ErrorKind.html), errors propagate from serde into the `Custom(String)` kind after formatting the error via the `Display` trait. This means that pattern matching on the specific deserialization error can't be done through the guard.

src/util.rs
```diff
@@ -37,57 +40,43 @@ pub fn vector_with_length<F: FieldElement>(len: usize) -> Vec<F> {
...
-            f0,
-            g0,
-            h0,
-            points_h_packed,
-        })
-    } else {
-        Err(SerializeError::UnpackInputSizeMismatch)
-    }
+) -> Result<UnpackedProof<F>, Box<bincode::ErrorKind>> {
+    let bytes = bincode::serialize(proof).unwrap();
```
acmiyaguchi, line 78: It's less than ideal to serialize the FieldElements into bytes (since they may take up a different amount of space depending on the field size), and then to deserialize into the unpacked structure.

src/util.rs
```diff
@@ -37,57 +40,43 @@ pub fn vector_with_length<F: FieldElement>(len: usize) -> Vec<F> {
...
 
 /// Unpacked proof with mutable subcomponents
-#[derive(Debug)]
-pub struct UnpackedProofMut<'a, F: FieldElement> {
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(bound(deserialize = ""))]
+pub struct UnpackedProofMut<F: FieldElement> {
     /// Data
-    pub data: &'a mut [F],
+    pub data: Vec<F>,
```
acmiyaguchi, line 63: I'm not sure if this is the right way to go, but trying to set the correct deserialization trait here with the struct lifetime and the deserialization lifetime was a bit much for me. I had to annotate each of these fields as so: `[serde(bound(deserialize="&'a mut F: Deserialize<'a>"))]`.  I thought it would be would be nice in general to replace serialization helpers with bincode wherever possible.

acmiyaguchi:
I'm going to close this; I don't think any off-the-shelf serializer will out do the existing serialization of proofs in terms of size. In particular, bincode represents an array of bytes as (u64; [u8]). MessagePack would use a single byte to encode the array length, while a format like CBOR would use a start/end delimiter instead.

 It could be useful to implement the serialization of the `FieldElement` into an anonymous tuple where each entry is a byte in the encoding of the u128. This would serialize with 0 overhead, but would likely involve heavy usage of macros.

Issue #23, opened 2021-04-05
Add support for more data types


cjpatton:
This library currently supports a single input type, a vector of 0s and 1s. In [Prio v3](https://github.com/abetterinternet/prio-documents) we will need support for a much richer set of input types. In my opinion, we ought to aim for the full range of input types that [[BBG+19](https://eprint.iacr.org/2021/017), Theorem 4.3] supports. In order to maximize code reuse, what I suggest we do is implement the core logic of the fully linear PCP system and delegate the implementation of the validity circuit to each data type. Something like this (e.g., `pcp.rs`):
```rust

//! This module implements the fully linear PCP ("Probabilistically Checkable Proof") system
//! decsribed in [[BBG+19](https://eprint.iacr.org/2021/017), Theorem 4.3]. It is the core
//! component of Prio's input-validation protocol.
//!
//! A PCP asserts that a given input is a member of a formal language, e.g., that each integer in
//! an input vector is within a certain range. The proof system is comprised of a proover and a
//! verifier. The prover generates a PCP for its input, and the verifier checks the PCP to
//! determine if the input is valid. This procedure should always succeed when the input is valid;
//! if the input is invalid, then the validatin procedure should fail, except with small
//! probability.
//!
//! A "fully linear" PCP system is a PCP system that can operate on secret shared data. In this
//! setting, the prover splits its input and proof into shares and distributes the shares among a
//! set of verifiers. Th verifiers then run a protocol in which they learn whether the input is
//! valid, but without learning the input itself.
//!
//! The fully linear PCP system of [BBG+19, Theorem 4.3] applies languages recognized by arithmetic
//! circuits over finite fields that have a particular structure. Namely, all gates in the circuit
//! are either affine (i.e., addition or scalar multiplication) or invoke a special sub-circuit,
//! called the gadget, which may contain non-affice operations (i.e., multiplication).

use crate::field::FieldElement;

/// This type represents an input to be validated.
pub trait Datum<F: FieldElement, G: Gadget<F>>: Sized {
    /// Evalauates the arithmetic circuit on the given input (i.e., `self`) and returns the output.
    /// `rand` is the random input of the validity circuit.
    fn valid(&self, g: &mut dyn Gadget<F>, rand: &[F]) -> F;

    /// Returns an instance of gadget associated with the validity circuit. The length of the proof
    /// generated for this data type is linear in the number of times the gadget is invoked.
    fn gadget(&self) -> G;

[...]

PR #30, opened 2021-04-14
Add the PolyCheckedVector type

cjpatton:
Partially addresses #23.

This change adds a new type called `PolyCheckedVector`. This type represents vectors for which `p(x) == 0` for some polynomial `p` and each vector element `x`. It can be used to construct higher level types, such as boolean vectors. The construction is based on [BBG+19, Theorem 5.3]. Its performance is comparable to the proof system of Prio v2, though proof verification is generally faster. 

Here are the numbers for proof generation, on my laptop (sizes are in number of field elements):

| proof system (input size) | average prove time | average query time | proof size |
|-------------------|------------|-----------|--------------|
| Prio v2 (10) | 7.1 us | 8.2 us | 29 |
| PolyCheckedVector (10) | 6.7 us | 5.9 us | 32 | 
| Prio v2 (100) | 47.3 us | 39.5 us | 231 |
| PolyCheckedVector (100) | 45.9 us | 30.7 us | 256 |
| Prio v2 (1,000) | 0.46 ms | 0.35 ms | 2,027 |
| PolyCheckedVector (1,000) | 0.45 ms | 0.26 ms | 2,048 |
| Prio v2 (10,000) | 10.4 ms | 7.8 ms | 26,387 |
| PolyCheckedVector (10,000) | 11.4 ms | 6.2 ms | 32,768 | 

You can run these benchmarks (all except input size of 1mil) with `cargo bench`. Note that I refactored the code in `client.rs` a bit in order to expose a suitable benchmark for v2.

[[UPDATE: Originally the `PolyCheckedVector` type employed the "parallel sum" technique of Corollary 4.9 in order to reduce the proof size to `O(sqrt(n))`. I now realize that this technique can't be used here directly.]]


cjpatton:
It occurs to me that there might be a way for me to reduce the overhead of my code somewhat. I'm moving this PR back to WIP.

cjpatton:
Managed to shave a bit off. That's enough.

src/pcp/gadgets.rs
```diff
@@ -2,42 +2,69 @@
...
+use std::marker::PhantomData;
+
+const FFT_THRESHOLD_MUL: usize = 60;
+const FFT_THRESHOLD_POLY_EVAL: usize = 8;
 
 /// An arity-2 gadget that multiples its inputs.
 pub struct Mul<F: FieldElement> {
-    buf: Vec<F>,
+    n: usize, // Size of buffer for FFT operations
+    n_inv: F, // Inverse of n in F
```
tgeoghegan, line 19:
```suggestion
    /// Size of buffer for FFT operations
    n: usize,
    // Inverse of n in F
    n_inv: F,
```

src/pcp/gadgets.rs
```diff
@@ -2,42 +2,69 @@
...
-use crate::fft::{discrete_fourier_transform, discrete_fourier_transform_inv};
+use crate::fft::{discrete_fourier_transform, discrete_fourier_transform_inv_finish};
 use crate::field::FieldElement;
 use crate::pcp::{Gadget, GadgetCallOnly, PcpError};
+use crate::polynomial::{poly_deg, poly_eval, poly_mul};
+
+use std::convert::TryFrom;
+use std::marker::PhantomData;
+
+const FFT_THRESHOLD_MUL: usize = 60;
```
tgeoghegan, line 13: I saw your comment in `speed_tests.rs` explaining how these values were derived. I think we should have a similar comment here.

src/pcp/gadgets.rs
```diff
@@ -49,25 +76,186 @@ impl<F: FieldElement> GadgetCallOnly<F> for Mul<F> {
...
 
+/// An arity-1 gadget that evaluates its input on some polynomial.
+pub struct PolyEval<F: FieldElement> {
+    poly: Vec<F>,
+    n: usize, // Size of buffer for FFT operations
+    n_inv: F, // Inverse of n in F
+}
+
+impl<F: FieldElement> PolyEval<F> {
+    /// Returns a gadget that evaluates its input on `poly`. The prameter `in_len` denotes the
```
tgeoghegan, line 99:
```suggestion
    /// Returns a gadget that evaluates its input on `poly`. The parameter `in_len` denotes the
```

src/pcp/gadgets.rs
```diff
@@ -49,25 +76,186 @@ impl<F: FieldElement> GadgetCallOnly<F> for Mul<F> {
...
+/// outputs. The arity is equal to the arity of the inner gadget times the number of chunks.
+pub struct ParallelSum<F: FieldElement, G: Gadget<F>> {
+    inner: G,
+    chunks: usize,
+    phantom: PhantomData<F>,
+}
+
+impl<F: FieldElement, G: Gadget<F>> ParallelSum<F, G> {
+    /// Wraps `inner` into a parallel sum gadget with `chunks` chunks.
+    pub fn new(inner: G, chunks: usize, _in_len: usize) -> Self {
```
tgeoghegan, line 201: If `in_len` is unused, why not remove it from the prototype?

src/pcp/gadgets.rs
```diff
@@ -49,25 +76,186 @@ impl<F: FieldElement> GadgetCallOnly<F> for Mul<F> {
...
+    }
+}
+
+impl<F: FieldElement, G: Gadget<F>> GadgetCallOnly<F> for ParallelSum<F, G> {
+    fn call(&mut self, inp: &[F]) -> Result<F, PcpError> {
+        gadget_call_check(self, inp.len())?;
+        let mut outp = F::zero();
+        for chunk in inp.chunks(self.inner.call_in_len()) {
+            outp += self.inner.call(chunk)?;
+        }
```
tgeoghegan, line 216: It seems like the computation on each chunk is totally independent. Would it be worthwhile to actually make this parallel? Perhaps we should write up an issue for that future performance optimization.

src/pcp/types.rs
```diff
@@ -74,6 +74,107 @@ impl<F: FieldElement> Value<F, Mul<F>> for Boolean<F> {
...
+    fn valid(&self, g: &mut dyn GadgetCallOnly<F>, rand: &[F]) -> Result<F, PcpError> {
+        if rand.len() < 1 {
+            return Err(PcpError::CircuitIn("random input not long enough"));
+        }
+
+        let mut r = rand[0];
+        let mut outp = F::zero();
+        for chunk in self.data.chunks(self.chunk_len) {
+            let n = chunk.len();
+            let v = match n == self.chunk_len {
```
tgeoghegan, line 124: This works fine if you want to keep it but you could make this a little less indented if you do 
```
let v = if n == self.chunk_len {
    g.call(chunk)?
} else {
    // the other thing
};
```

tgeoghegan (changes requested):

src/pcp/gadgets.rs
cjpatton: Ugh, good catch. That's just left over from an earlier rewrite.

src/pcp/gadgets.rs
cjpatton: Done: https://github.com/abetterinternet/libprio-rs/issues/34

src/pcp/types.rs
cjpatton: Done.

tgeoghegan (approved):

PR #39, opened 2021-05-04
pcp: Distinguish between joint and query randomness

cjpatton:
The "query randomness" is the random point at which the verifier evaluates the proof polynomial; the "joint randomness" is whatever randomness is used by both the prover and verifier. This allows us to implement proof systems like [BCG+19, Theorem 5.3] in which the verifier sends the prover a random challenge to use for proof generation.

Along the way, this PR cleans up the unit tests and benchmarks a bit.

### DIFF
--- a/src/pcp.rs
+++ b/src/pcp.rs
@@ -16,16 +16,25 @@
 //! use prio::pcp::{decide, prove, query};
 //! use prio::field::{FieldElement, Field64};
 //!
+//! // The randomness shared by the prover and verifier. In proof systems like [BCG+19, Theorem
+//! // 5.3], the verifier sends the prover a random challenge in the first round, which the
+//! // prover uses to construct the proof. (This is not used for this example, so `joint_rand'
+//! // is empty.)
+//! let joint_rand = [];
+//!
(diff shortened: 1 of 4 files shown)


src/pcp.rs
```diff
@@ -343,17 +353,31 @@ impl<F: FieldElement> From<Vec<F>> for Proof<F> {
...
+    pf: &Proof<F>,
+    query_rand: &[F],
+    joint_rand: &[F],
+) -> Result<Verifier<F>, PcpError>
 where
     F: FieldElement,
     G: Gadget<F>,
     V: Value<F, G>,
 {
+    if query_rand.len() != 1 {
```
tgeoghegan, line 373: If the query randomness can only ever have one element, why not make it `F` instead of `&[F]`? Is the idea that in the future, we might support vectors of query randomness in some other PCP?

tgeoghegan (approved):

src/pcp.rs
cjpatton: Yeah, that's the idea. I don't foresee it happening, but I figured it's a good idea to be general in the API.

PR #40, opened 2021-05-06
[DO NOT MERGE] Prototype the upload/verify flows

cjpatton:
This PR prototypes a potential API for processing messages in the upload/verify flow. It adds the MeanVarUnsignedVector type, which represents a vector of L-bit integers. The collector wants to compute the mean and variance of each element of the vector.


cjpatton:
Moving this to my fork.

PR #44, opened 2021-05-24
pcp: Translate between Verifier<F> and &[F]

cjpatton:
Based on #39.

Changes the internal representation of Verifier<F> in a way that allows it to be efficiently serialized into a slice of field elements.

### DIFF
--- a/src/pcp.rs
+++ b/src/pcp.rs
@@ -165,8 +165,12 @@ pub enum PcpError {
     GadgetPolyOutLen,
 
     /// Calling `query` returned an error.
-    #[error("query error")]
+    #[error("query error: {0}")]
     Query(&'static str),
+
+    /// Calling `decide` returned an error.
+    #[error("decide error: {0}")]
+    Decide(&'static str),
 
     /// The validity circuit was called with the wrong amount of randomness.
     #[error("incorrect amount of randomness")]
@@ -379,6 +383,11 @@ where
     let g = x.gadget(m);
     let l = g.call_in_len();
 
+    // Allocate space for the verifier data. This includes the output of the validity circuit, the
+    // intermediate polynomials evaluated at the query randomness `r`, and the proof polynomial
+    // evaluated at `r`.
+    let mut verifier_data = Vec::with_capacity(2 + l);
+
     // Run the validity circuit with a "shim" gadget that records each input to each gadget.
     // Record the output of the circuit.
     //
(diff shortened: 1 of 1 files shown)


src/pcp.rs
```diff
@@ -379,6 +383,8 @@ where
     let g = x.gadget(m);
     let l = g.call_in_len();
 
+    let mut verifier_data = Vec::with_capacity(2 + l);
```
tgeoghegan, line 386:
```suggestion
    // Allocate space for the value of `l` polynomials at `r`, the result of the validity circuit and the value of `p` at `r`.
    let mut verifier_data = Vec::with_capacity(2 + l);
```

src/pcp.rs
```diff
@@ -476,12 +483,20 @@ impl<F: FieldElement> GadgetCallOnly<F> for QueryShimGadget<F> {
 /// The output of `query`, the verifier message generated for a proof.
 #[derive(Debug)]
 pub struct Verifier<F: FieldElement> {
-    /// Output of the validity circuit.
-    v: F,
-    /// The proof polynomial evluated at `r`.
-    p_at_r: F,
-    /// The intermediate proof polynomials evaluated at `r`.
-    f_at_r: Vec<F>,
+    data: Vec<F>,
```
tgeoghegan, line 486: nit: this could use a doccomment explaining the order in which `v`, `p_at_r` and `f_at_r` are expected to appear in the `Vec<F>`

src/pcp.rs
```diff
@@ -523,8 +533,17 @@ where
...
-    if e == vf.p_at_r && vf.v == F::zero() {
+    let mut g = x.gadget(0);
+    let vf_len = vf.data.len();
+    if vf_len < 2 + g.call_in_len() {
+        return Err(PcpError::Decide("short verifier message"));
+    }
+
+    let v = vf.data[0];
+    let f_at_r = &vf.data[1..vf_len - 1];
+    let p_at_r = vf.data[vf_len - 1];
```
tgeoghegan, line 544: I'd prefer to have methods `Verifier::validity(&self) -> F`, `Verifier::p_at_r(&self) -> F` and `Verifier::f_at_r(&self) -> &[F]` so that the implementation detail of how `Verifier` stores its data is abstracted away. Similarly, `Verifier::new(v: F, p_at_r: F, f_at_r: &[F]) -> Verifier<F>` would be better than making `query()` construct the `Vec<F>`.

tgeoghegan (changes requested):

src/pcp.rs
cjpatton: Changed to
```
// Allocate space for the verifier data. This includes the output of the validity circuit, the
// intermediate polynomials evaluated at `query_rand[0]`, and the proof polynomial evaluated at                                                                                                                                    
// `query_rand[0]`.
```
but the idea is the same.

src/pcp.rs
cjpatton: Done.

src/pcp.rs
cjpatton: - Added `validity()`, `f_at_r()`, and `p_at_r()` to Verifier implementation.
- I'm not sure what you mean for the second part. `query()` returns a `Verifier<F>`, not a `Vec<F>`.

tgeoghegan: >     * I'm not sure what you mean for the second part. `query()` returns a `Verifier<F>`, not a `Vec<F>`.

What I mean is that instead of 
```
    Ok(Verifier {
            data: verifier_data,
   })
```
`query()` could do
```
Ok(Verifier::new(v, p_at_r, f_at_r))
```
And `Verifier::new()` would be responsible for allocating an appropriately sized `Vec<F>` and constructing a `Verifier` from it. This would also absolve `decide()` of the need to check that the `data` member of the `Verifier` it is passed has an appropriate length.

cjpatton: I see. My one argument against is that it's slightly less efficient because it requires re-allocating an length-`(2+l)` vector and copying data into it. Usually `l` will be small, but not always.

In any case, IMO we're talking about an implementation detail that (hopefully) isn't exposed to the user. From the user's perspective, a `Verifier<F>` is constructed from a `Vec<F>` of any length. (Of course, not all lengths are valid, which `decide()` needs to check for.) I'm not sure I see the value in exposing more detail than that in the API.

tgeoghegan (approved):

src/pcp.rs
tgeoghegan: True, it is all internal to the `pcp` module.

cjpatton:
Squashed!

PR #45, opened 2021-05-24
pcp: Add TryFrom for Value from field sequence

cjpatton:
Based on #44.

The new_with() method on Value allows a new instance of Value to be constructed from an existing instance of Value and a sequence of field elements. This is used in order construct a Value from a secret share of the underlying data. However, it requires the caller to already have an instance of Value on hand, which is usually not the case. 

This change drops new_with() in favor of a proper TryFrom supertrait on Value. The caller provides a vector of field elements as well as a "parameter", the type of which is an associated type of the particular implementation of Value.

### DIFF
--- a/src/pcp.rs
+++ b/src/pcp.rs
@@ -75,9 +75,10 @@
 //! // The prover encodes its input and splits it into two secret shares. It
 //! // sends each share to two aggregators.
 //! let x: Boolean<Field64>= Boolean::new(true);
+//! let x_par = x.param();
 //! let x_shares: Vec<Boolean<Field64>> = split(x.as_slice(), 2)
 //!     .into_iter()
-//!     .map(|data| x.new_with(data))
+//!     .map(|data| Boolean::try_from((x_par, data)).unwrap())
 //!     .collect();
 //!
 //! // The prover generates a proof of its input's validity and splits the proof
@@ -179,12 +180,23 @@ pub enum PcpError {
 
 /// A value of a certain type. Implementations of this trait specify an arithmetic circuit that
 /// determines whether a given value is valid.
-pub trait Value<F, G>: Sized + PartialEq + Eq + Debug
+pub trait Value<F, G>:
+    Sized
+    + PartialEq
+    + Eq
+    + Debug
+    + TryFrom<(<Self as Value<F, G>>::Param, Vec<F>), Error = <Self as Value<F, G>>::TryFromError>
 where
     F: FieldElement,
     G: Gadget<F>,
 {
-    /// Evalauates the validity circuit on the given input (i.e., `self`) and returns the output.
+    /// Parameters used to construct a value of this type from a vector of field elements.
+    type Param;
+
+    /// Error returned when converting a `(Vec<F>, Param)` to a `Value<F, G>` fails.
+    type TryFromError: Debug;
+
+    /// Evaluates the validity circuit on the given input (i.e., `self`) and returns the output.
     /// Slice `rand` is the random input consumed by the validity circuit.
     ///
     /// ```
@@ -210,10 +222,6 @@ where
     /// Returns a reference to the underlying data.
     fn as_slice(&self) -> &[F];
 
-    /// Constructs a value of this type from the given data without checking whether the data is
-    /// valid. This method takes ownership of `data`.
-    fn new_with(&self, data: Vec<F>) -> Self;
-
     /// The length of the random input expected by the validity circuit.
     fn valid_rand_len(&self) -> usize;
 
@@ -224,6 +232,9 @@ where
     /// maximum degree of each of the polynomials passed into `call_poly`. If `call_poly` is never
     /// used by the caller, then it is safe to set `in_len == 0`.
     fn gadget(&self, in_len: usize) -> G;
+
+    /// Returns a copy of the associated type parameters for this value.
+    fn param(&self) -> Self::Param;
 }
 
 /// The gadget functionality required for evaluating a validity circuit. The `Gadget` trait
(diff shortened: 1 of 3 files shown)


src/pcp.rs
```diff
@@ -179,11 +181,15 @@ pub enum PcpError {
 
 /// A value of a certain type. Implementations of this trait specify an arithmetic circuit that
 /// determines whether a given value is valid.
-pub trait Value<F, G>: Sized + PartialEq + Eq + Debug
+pub trait Value<F, G>:
+    Sized + PartialEq + Eq + Debug + TryFrom<(<Self as Value<F, G>>::Param, Vec<F>), Error = TypeError>
```
tgeoghegan, line 185: `TryFrom<T>` where `T` is a tuple is interesting. How does the [implied `TryInto`](https://doc.rust-lang.org/std/convert/trait.TryFrom.html#generic-implementations) implementation work? You get back the `Vec<F>` and the `Param`?

src/pcp/types.rs
```diff
@@ -7,6 +7,12 @@ use crate::pcp::gadgets::{Mul, PolyEval};
 use crate::pcp::{GadgetCallOnly, PcpError, Value};
 use crate::polynomial::poly_range_check;
 
+use std::convert::TryFrom;
+
+/// Errors propagagted by methods in this module.
+#[derive(Debug, PartialEq, thiserror::Error)]
+pub enum TypeError {}
```
tgeoghegan, line 14: It looks like there are currently no failure modes for any of the `Value` implementations. Maybe we could use [`std::convert::Infallible`](https://doc.rust-lang.org/std/convert/enum.Infallible.html) here? Each `Value` implementation's `TryFrom` implementation can use a different associated error type, so if/when a `Value` comes along whose `TryFrom` can fail, it can provide an error enum.

src/pcp/types.rs
```diff
@@ -112,6 +128,8 @@ impl<F: FieldElement> PolyCheckedVector<F> {
 }
 
 impl<F: FieldElement> Value<F, PolyEval<F>> for PolyCheckedVector<F> {
+    type Param = Vec<F>; // A polynomial
```
tgeoghegan, line 131: Ah, wouldn't it be nice if we had a `Polynomial` type so we wouldn't have to use comments to indicate when a `[F]` or `Vec<F>` is actually a polynomial 😄

src/pcp.rs
```diff
@@ -179,11 +181,15 @@ pub enum PcpError {
...
 where
     F: FieldElement,
     G: Gadget<F>,
 {
+    /// Parameters used to construct a value of this type from a vector of field elements.
+    type Param;
```
tgeoghegan, line 191: I wish we had a better name for this than `Param`, though I'll admit I'm struggling to come up with one.

src/pcp.rs
```diff
@@ -179,11 +181,15 @@ pub enum PcpError {
...
+
     /// Evalauates the validity circuit on the given input (i.e., `self`) and returns the output.
```
tgeoghegan, line 193:
```suggestion
    /// Evaluates the validity circuit on the given input (i.e., `self`) and returns the output.
```
This typo was already there, but why not fix it while we're in here

src/pcp/types.rs
```diff
@@ -7,6 +7,12 @@ use crate::pcp::gadgets::{Mul, PolyEval};
 use crate::pcp::{GadgetCallOnly, PcpError, Value};
 use crate::polynomial::poly_range_check;
 
+use std::convert::TryFrom;
+
+/// Errors propagagted by methods in this module.
```
tgeoghegan, line 12:
```suggestion
/// Errors propagated by methods in this module.
```

src/pcp/types.rs
```diff
@@ -143,11 +161,19 @@ impl<F: FieldElement> Value<F, PolyEval<F>> for PolyCheckedVector<F> {
...
-        Self {
-            data,
-            poly: self.poly.clone(),
-        }
+    fn param(&self) -> Self::Param {
+        self.poly.clone()
+    }
+}
+
+impl<F: FieldElement> TryFrom<(Vec<F>, Vec<F>)> for PolyCheckedVector<F> {
```
tgeoghegan, line 169: It seems like it'd be easy to accidentally confuse the vector of `FieldElement` and the `Vec<F>` that's actually a polynomial. Absent a real `Polynomial` type, could this be written as `TryFrom<(Vec<F>, PolyCheckedVector<F>::Param)>` (not sure if that compiles, it's just a guess)?

tgeoghegan (changes requested):

src/pcp.rs
cjpatton: Hmm, I hadn't considered that. Yes, ideally you'd get the `(Vec<F>, Param)`, but I haven't tried this.

src/pcp/types.rs
cjpatton: Done. (There is a failure case for a data type that will be coming in a future PR.)

src/pcp/types.rs
cjpatton: Hahaha. Yeah.

To make this explicit, we could do something like
```rust
type Param = PolyCheckedVector<F>;
```
where
```rust
pub struct PolyCheckedVectorParam<F: FieldElement> {
   pub poly: Vec<F>,
}
```
I just figure this was overkill here. (Though I'm not sure if the associated type declaration would need a type parameter?)

src/pcp/types.rs
cjpatton: Hmmm, I can't seem to spell this in a way that compiles.

tgeoghegan (approved):

cjpatton:
Squashed!

PR #47, opened 2021-06-08
pcp: Distinguish leader share from helper shares

cjpatton:
Adds a method set_leader() on an instance of Value that can be used to
designate the value as the leader's input share. This is useful when the
leader's share needs to be handled differently than the others during
proof verification.


tgeoghegan (approved):
LGTM. If `client.rs` and `server.rs` get migrated to an implementation of `pcp::Value`, then they would be a data type where the leader distinction is meaningful, right?

PR #48, opened 2021-06-08
Add the MeanVarUnsginedVector type

cjpatton:
Based on #47.

This type is uses to compute the mean and variance of each integer in a sequence of integers.

### DIFF
--- a/src/pcp.rs
+++ b/src/pcp.rs
@@ -86,7 +86,7 @@
 //! let pf = prove(&x, &joint_rand).unwrap();
 //! let pf_shares: Vec<Proof<Field64>> = split(pf.as_slice(), 2)
 //!     .into_iter()
-//!     .map(|data| Proof::from(data))
+//!     .map(Proof::from)
 //!     .collect();
 //!
 //! // Each verifier queries its shares of the input and proof and sends its
@@ -123,7 +123,7 @@ use std::convert::TryFrom;
 use std::fmt::Debug;
 
 use crate::fft::{discrete_fourier_transform, discrete_fourier_transform_inv_finish, FftError};
-use crate::field::FieldElement;
+use crate::field::{FieldElement, FieldError};
 use crate::fp::log2;
 use crate::polynomial::{poly_deg, poly_eval};
 
@@ -176,6 +176,14 @@ pub enum PcpError {
     /// The validity circuit was called with the wrong amount of randomness.
     #[error("incorrect amount of randomness")]
     ValidRandLen,
+
+    /// Encountered an error while evaluating a validity circuit.
+    #[error("failed to run validity circuit: {0}")]
+    Valid(&'static str),
+
+    /// Returned if a field operation encountered an error.
+    #[error("Field error")]
+    Field(#[from] FieldError),
 }
 
 /// A value of a certain type. Implementations of this trait specify an arithmetic circuit that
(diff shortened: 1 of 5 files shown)


src/pcp.rs
```diff
@@ -240,7 +248,45 @@ where
     /// "leader" share from the others. This is useful, for example, when some of the gadget inputs
     /// are constants used for both proof generation and verification.
     ///
-    /// TODO(cjpatton) Add an example once we have implemented a data type that uses this method.
+    /// ```
+    /// use prio::pcp::types::MeanVarUnsignedVector as MvuVec;
```
tgeoghegan, line 252: nit: using an alias like this in a doc example makes it a little harder to understand because the reader might then grep through the crate looking for the `MvuVec` type when what they really want is `MeanVarUnsignedVector`.

src/pcp.rs
```diff
@@ -240,7 +248,45 @@ where
...
+    ///         let mut share = MvuVec::try_from((x.param(), data)).unwrap();
+    ///         share.set_leader(i == 0);
+    ///         share
+    ///     })
+    ///     .collect();
+    ///
+    /// let pf = prove(&x, &joint_rand).unwrap();
+    /// let pf_shares: Vec<Proof<Field64>> = split(pf.as_slice(), 2)
+    ///     .into_iter()
+    ///     .map(|data| Proof::from(data))
```
tgeoghegan, line 278: nit: I'm not sure but I think this works, because `Proof::from` itself is already a function that takes `Vec<Field64>` and returns `Proof<Field64>`
```suggestion
    ///     .map(Proof::from)
```

src/pcp/gadgets.rs
```diff
@@ -192,6 +192,176 @@ impl<F: FieldElement> Gadget<F> for PolyEval<F> {
...
+/// The gadget is designed to output `x^2` *as long as* `x_vec` is a vector comprised of 0s and 1s.
+/// It first computes `w[l] = p(x_vec[l]) * r_vec[l]` for each `0 <= l < bits`, where `p(x) =
+/// x(x-1)`. It then computes `w = w[0] + ... + w[bits-1]` and returns `w + x^2`.
+///
+/// If `vec_x[l] == 1` or `vec_x[l] == 0`, then `p(vec_x[l]) == 0`; otherwise, `p(vec_x[l]) != 0`.
+/// The validity circuit for MeanVarUnsignedVector sets `r_vec[l] = r^(l+1)`, where `r` is a
+/// uniform random field element. This is ensures that, if `p(vec_x[l]) != 0` for some `l`, then
+/// `w != 0` with high probability.
+pub struct MeanVarUnsigned<F: FieldElement> {
+    poly: [F; 3],
```
tgeoghegan, line 212:
```suggestion
    /// Polynomial used to check that values are in range
    poly: [F; 3],
```

src/pcp/gadgets.rs
```diff
@@ -192,6 +192,176 @@ impl<F: FieldElement> Gadget<F> for PolyEval<F> {
...
+
+    pub(crate) fn call_poly_direct<V: AsRef<[F]>>(
+        &mut self,
+        outp: &mut [F],
+        inp: &[V],
+    ) -> Result<(), PcpError> {
+        let bits = self.bits;
+        let r_vec = &inp[..bits];
+        let x_vec = &inp[bits..2 * bits];
+        let x = inp[2 * bits].as_ref();
```
tgeoghegan, line 244: Should check that `inp.len() == 2 * bits + 1`

src/pcp/gadgets.rs
```diff
@@ -192,6 +192,176 @@ impl<F: FieldElement> Gadget<F> for PolyEval<F> {
...
+
+        let z = poly_mul(x, x);
+        for i in 0..z.len() {
+            outp[i] = z[i];
+        }
+        for i in z.len()..outp.len() {
+            outp[i] = F::zero();
+        }
```
tgeoghegan, line 252: This should check that `outp.len() >= z.len()`

src/pcp/gadgets.rs
```diff
@@ -192,6 +192,176 @@ impl<F: FieldElement> Gadget<F> for PolyEval<F> {
...
+    pub(crate) fn call_poly_fft<V: AsRef<[F]>>(
+        &mut self,
+        outp: &mut [F],
+        inp: &[V],
+    ) -> Result<(), PcpError> {
+        let bits = self.bits;
+        let n = self.n;
+        let r_vec = &inp[..bits];
+        let x_vec = &inp[bits..2 * bits];
+        let x = inp[2 * bits].as_ref();
```
tgeoghegan, line 279: As in `call_poly_direct`, this should check that `inp.len() == 2 * bits + 1` and that `outp.len >= z.len`

src/pcp/gadgets.rs
```diff
@@ -192,6 +192,176 @@ impl<F: FieldElement> Gadget<F> for PolyEval<F> {
...
+    }
+}
+
+impl<F: FieldElement> GadgetCallOnly<F> for MeanVarUnsigned<F> {
+    fn call(&mut self, inp: &[F]) -> Result<F, PcpError> {
+        gadget_call_check(self, inp.len())?;
+        let bits = self.bits;
+        let r_vec = &inp[..bits];
+        let x_vec = &inp[bits..2 * bits];
+        let x = inp[2 * bits];
```
tgeoghegan, line 333: This should check that `inp.len() == 2 * bits + 1`

tgeoghegan (changes requested):

src/pcp.rs
cjpatton: Oooh nice! Fixed here and elsewhere.

src/pcp/gadgets.rs
cjpatton: Added a comment that's a bit more precise.

src/pcp/gadgets.rs
cjpatton: This check is performed by the caller (`call_poly`): see https://github.com/abetterinternet/libprio-rs/pull/48/files#diff-e01a59bccdbfaacea57d3cf97d4ad2fc0b9d8f1d333aeaa0098b150ad60a7a02R352.

src/pcp/gadgets.rs
cjpatton: Here again, this check is performed by the caller: see https://github.com/abetterinternet/libprio-rs/pull/48/files#diff-e01a59bccdbfaacea57d3cf97d4ad2fc0b9d8f1d333aeaa0098b150ad60a7a02R352.

src/pcp/gadgets.rs
cjpatton: Hmm. The idea is to avoid a panic, right? IIUC, If `outp.len() < z.len()`, the result would be a no-op, since the range would be effectively `0..0`?

cjpatton: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=056ccc5e24f694249bac579d1c659abb

src/pcp/gadgets.rs
cjpatton: This is checked by `gadget_call_check` above: https://github.com/abetterinternet/libprio-rs/pull/48/files#diff-e01a59bccdbfaacea57d3cf97d4ad2fc0b9d8f1d333aeaa0098b150ad60a7a02R329.

cjpatton (commented):
Rebased and addressed some comments.

tgeoghegan (approved):

cjpatton (commented):
Squashed.

PR #54, opened 2021-06-10
prng: Fix block alignment bug

tgeoghegan:
The Prng structure has a buffer that stores a segment of the keystream.
When it's time to emit the next field element, it reads chunks of this
buffer until a chunk is found that is smaller than the field modulus.
Once the end of the buffer is reached, the buffer is refreshed with the
next segment of the keystream.

The size of the buffer is a multiple of the block size for the
underlying blockcipher (16 bytes). If the size of each field element
chunk does not divide the length of the buffer evenly, then the current
implementation will read passed the end of the buffer.

This changes the length of the buffer so that it's guaranteed to be a
multiple of the length of each field element.

### DIFF
--- a/src/prng.rs
+++ b/src/prng.rs
@@ -3,9 +3,7 @@
 
 use super::field::{FieldElement, FieldError};
 use aes::{
-    cipher::{
-        generic_array::GenericArray, BlockCipher, FromBlockCipher, NewBlockCipher, StreamCipher,
-    },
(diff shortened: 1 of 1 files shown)


tgeoghegan:
This change was reviewed in #50, but this time we are going to merge it to `main` instead of into a different PR's branch.

cjpatton (approved):

PR #55, opened 2021-06-10
Replace rand dependency with getrandom

cjpatton:
We currently use rand=0.7.3. Support for WASM was dropped in rand=0.8, so users won't be able to compile libprio-rs if we upgrade this dependency: see https://github.com/rust-random/rand/blob/HEAD/CHANGELOG.md#080---2020-12-18.
    
This change aims to nip this problem in the bud by dropping the rand dependency and replacing it with getrandom, which will support WASM in the future.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@ aes = { version = "0.7.3", features = ["ctr"] }
 cipher = "0.3.0"
 aes-gcm = "0.6.0"
 base64 = "0.12.3"
-rand = "0.7"
+getrandom = { version = "0.2.3", features = ["std"] }
 ring = "0.16.15"
 thiserror = "1.0"
 serde = { version = "1.0", features = ["derive"] }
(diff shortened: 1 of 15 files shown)


tgeoghegan (changes requested):
It seems like we `unwrap()` everywhere we call `Prng::new()` or `Prng::new_with_length()`. The only way those functions can fail is if `getrandom` fails, and [its](https://docs.rs/getrandom/0.2.3/getrandom/) [docs](https://docs.rs/getrandom/0.2.3/getrandom/fn.getrandom.html) suggest that's extremely unlikely. I'm ok with the notion that short reads from /dev/urandom are so unlikely that it's not worth plumbing errors and `Result<>`s all over the place, but if that's the case, then let's make `Prng::new()` and its variants return `Prng` instead of `Result<Prng>`. Then we can document our chosen behavior (i.e. panic if there's insufficient entropy) in one place.

cjpatton:
Per our offline discussion, we decided it's best to propagate getrandom::Error and let the application handle it. The reason is that there's just too many ways an error might occur here.

tgeoghegan (approved):

cjpatton:
Squashed.

PR #56, opened 2021-06-10
prng: Clear useless bits during rejection sampling

cjpatton:
Based on #55.

To map a random byte string to a field element, we divide the byte
string into `n`-bit chunks, where `n` is the number of bits needed to
encode the field element. We interpret each chunk as an `n`-bit,
unsigned integer. For each integer, we check if the integer is smaller
than the field modulus: if so, we accept it; otherwise, we reject it and
try the next integer.

This algorithm is fine if `n` is equal to the number of bits needed to
encode the modulus. This isn't always the case, however. In particular,
for Field126, the modulus is a 126-bit prime, but we use 128 bits to
encode field elements. As was pointed out
[here](https://github.com/abetterinternet/prio-documents/issues/14),
this results in rejection happening more frequently than it needs to.

This change clears the extra bits in order to reduce the rate of
ejection. Note that this change is a wire-breaking because impacted
fields will derive a different sequence of field elements for the same
seed. In particular, this impacts Field126, but not Field32, Field64, or
Field80.


cjpatton:
cc/ @tlepoint (feel free to leave a review).

src/fp.rs
```diff
@@ -257,6 +259,10 @@ impl FieldParameters {
         }
         assert_eq!(&self.roots, &roots[..MAX_ROOTS + 1], "roots mismatch");
         assert_eq!(self.from_elem(self.roots[0]), 1, "first root is not one");
+
+        let bit_mask: u128 = (1 << (129 - p.leading_zeros())) - 1;
```
tlepoint, line 263: This will not work for a 127-bit or 128-bit prime.

cjpatton: That's true, but the current implementation only allows up to 126. This is enforced by a check above:
```rust
if let Some(x) = p.checked_next_power_of_two() {
    if x > 1 << 126 {
        panic!("p >= 2^126");
    }
} else {
    panic!("p >= 2^126");                                                                                                                                        
}
```

tgeoghegan (approved):

cjpatton:
Rebased.

cjpatton:
Rebased.

PR #60, opened 2021-06-17
pcp: Use less joint randomness for MeanVarUnsignedVector

cjpatton:
The circuit now does precisely what Theorem 5.3 says to do.

### DIFF
--- a/src/pcp/types.rs
+++ b/src/pcp/types.rs
@@ -252,14 +252,15 @@ impl<F: FieldElement> Value<F, MeanVarUnsigned<F>> for MeanVarUnsignedVector<F> 
         let mut inp = vec![F::zero(); 2 * bits + 1];
         let mut outp = F::zero();
 
-        for (i, chunk) in self.data.chunks(bits + 1).enumerate() {
+        let r = rand[0];
+        let mut pr = r;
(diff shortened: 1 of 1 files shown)


tgeoghegan (changes requested):
The change looks good and is consistent with my understanding of how MeanVarUnsignedVector works, but why isn't there a new test or updates to existing tests to go with this fix? Put another way, why didn't the previous, incorrect version fail?

edit: I guess if I understand right, the version that used more randomness was just as correct, but this version is more efficient because it requires less randomness?

cjpatton:
> The change looks good and is consistent with my understanding of how MeanVarUnsignedVector works, but why isn't there a new test or updates to existing tests to go with this fix? Put another way, why didn't the previous, incorrect version fail?

The circuit has changed, but the tests aren't design to check for the change. (Eventually we will need proper test vectors to check for these things, but I would argue that we will need more security analysis before we settle on test vectors.)

> edit: I guess if I understand right, the version that used more randomness was just as correct, but this version is more efficient because it requires less randomness?

I'm fairly certain the current code is OK from a security perspective, but this change brings it closer to what Theorem 5.3 says to do. That makes it safer, IMO.

tgeoghegan (approved):

PR #63, opened 2021-07-01
Add support for multiple gadgets

cjpatton:
This extends the PCP system to support types whose validity circuits have multiple distinct gadgets.

### DIFF
--- a/src/pcp.rs
+++ b/src/pcp.rs
@@ -1,39 +1,41 @@
 // SPDX-License-Identifier: MPL-2.0
 
-//! **(NOTE: This module is experimental. Applications should not use it yet.)** This
-//! module implements the fully linear PCP ("Probabilistically Checkable Proof") system described
-//! in \[[BBC+19](https://eprint.iacr.org/2019/188), Theorem 4.3\].
+//! **(NOTE: This module is experimental. Applications should not use it yet.)** This module
+//! implements a fully linear PCP ("Probabilistically Checkable Proof") system based on
+//! \[[BBC+19](https://eprint.iacr.org/2019/188), Theorem 4.3\].
 //!
 //! # Overview
 //!
 //! The proof system is comprised of three algorithms. The first, `prove`, is run by the prover in
 //! order to generate a proof of a statement's validity. The second and third, `query` and
-//! `decide`, are run by the verifier in order to check the proof. In our setting, the proof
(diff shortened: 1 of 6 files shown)


src/pcp.rs
```diff
@@ -299,114 +304,152 @@ where
...
+        .enumerate()
+        .map(|(idx, g)| ProveShimGadget::new(g, g_calls[idx]))
+        .collect::<Result<Vec<_>, _>>()?;
+
+    // Create a buffer for storing the proof. The buffer is longer than the proof itself; the extra
+    // length is to accommodate the computation of each of the proof polynomials.
+    let data_len = (0..shim.len())
+        .map(|idx| shim[idx].arity() + shim[idx].deg() * (1 + g_calls[idx]).next_power_of_two())
+        .reduce(|a, b| a + b)
+        .unwrap();
```
tgeoghegan, line 351:
```suggestion
        .sum();
```
`Iterator::sum()` doesn't return an `Option` so you can remove the unsightly `unwrap()`

src/pcp.rs
```diff
@@ -429,167 +472,219 @@ impl<F: FieldElement> From<Vec<F>> for Proof<F> {
...
+    }
+
+    // Create a buffer for the verifier data. This includes the output of the validity circuit as
+    // well as the intermediate polynomials evaluated at the query randomness `query_rand[idx]`,
+    // and the proof polynomial evaluated at `query_rand[idx]` for the `idx`-th gadget.
+    let data_len = 1
+        + (0..shim.len())
+            .map(|idx| shim[idx].arity() + 1)
+            .reduce(|a, b| a + b)
+            .unwrap();
```
tgeoghegan, line 537:
```suggestion
            .sum();
```
as in `prove()`, to avoid, `unwrap()`

tgeoghegan (changes requested):

src/pcp.rs
cjpatton: Nice!

src/pcp.rs
cjpatton: Done. (This also needed a type declaration, i.e., `.sum::<usize>()`.)

tgeoghegan (approved):

cjpatton:
Fixed up comments a bit and squashed.

PR #65, opened 2021-07-21
Reject invalid query randomness

cjpatton:
The query randomness is invalid if it coincides with any of the points at which the wire polynomials were interpolated. In our case,  these are roots of unity for the underlying field.

Along the way, clean up some of the internal comments.

### DIFF
--- a/src/pcp.rs
+++ b/src/pcp.rs
@@ -174,6 +174,14 @@ pub enum PcpError {
     #[error("query error: {0}")]
     Query(&'static str),
 
+    /// Returned by `query` if one of the elements of the query randomness vector is invalid. An
+    /// element is invalid if using it to generate the verification message would result in a
+    /// privacy violation.
+    ///
+    /// If this error is returned, the caller may generate fresh randomness and retry.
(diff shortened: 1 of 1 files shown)


src/pcp.rs
```diff
@@ -513,6 +518,13 @@ where
             let proof_data = &pf.data[proof_len..proof_len + next_len];
             proof_len += next_len;
 
+            // Make sure the query randomness isn't a root of unity. Evaluating the gadget
+            // polynomial at any of these points would be a privacy violation, since these points
+            // were used by the prover to construct the wire polynomials.
+            if query_rand[idx].pow(F::Integer::try_from(m).unwrap()) == F::one() {
+                return Err(PcpError::QueryRandInvalid);
+            }
```
tgeoghegan, line 526: nit: we could move this check up a few lines so it comes right after `m` is initially computed.

tgeoghegan (approved):

src/pcp.rs
cjpatton: Done.

PR #66, opened 2021-09-01
update crypto dependencies

tgeoghegan:
`prio` pinned `aes-gcm` v0.6.0, which pulls in `aes` v0.4.0, which
depends on `aesni`, which has been yanked by its author, having been
merged into newer versions of crate `aes`. This commit updates a few
crypto dependencies to resolve this.


cjpatton (approved):
Nice. Will it be possible to eventually consolidate `aes`, `aes-gcm`, and `ring` into just `ring`? I know for a fact that `ring` implements AES-GCM, but I'm not sure about AES-CTR, which we need for our PRNG.

PR #67, opened 2021-09-02
Run clippy in CI and configure Dependabot

tgeoghegan:
- Fixes all existing clippy warnings
- Adds a CI job to run clippy against PRs
- Configures dependabot to manage Rust dependencies for us

### DIFF
--- a/src/fft.rs
+++ b/src/fft.rs
@@ -27,12 +27,13 @@ pub enum FftError {
 /// Interpreting the input as the coefficients of a polynomial, the output is equal to the input
 /// evaluated at points `p^0, p^1, ... p^(size-1)`, where `p` is the `2^size`-th principal root of
 /// unity.
+#[allow(clippy::many_single_char_names)]
 pub fn discrete_fourier_transform<F: FieldElement>(
     outp: &mut [F],
     inp: &[F],
     size: usize,
 ) -> Result<(), FftError> {
-    let d = usize::try_from(log2(size as u128)).unwrap();
+    let d = usize::try_from(log2(size as u128)).map_err(|_| FftError::SizeTooLarge)?;
 
     if size > outp.len() {
         return Err(FftError::OutputTooSmall);
@@ -46,13 +47,10 @@ pub fn discrete_fourier_transform<F: FieldElement>(
         return Err(FftError::SizeInvalid);
     }
 
+    #[allow(clippy::needless_range_loop)]
     for i in 0..size {
         let j = bitrev(d, i);
-        if j < inp.len() {
-            outp[i] = inp[j];
-        } else {
-            outp[i] = F::zero();
-        }
+        outp[i] = if j < inp.len() { inp[j] } else { F::zero() }
     }
 
     let mut w: F;
@@ -191,6 +189,7 @@ mod tests {
         // Just for fun, let's do something different with a subset of the inputs. For the first
         // share, every odd element is set to the plaintext value. For all shares but the first,
         // every odd element is set to 0.
+        #[allow(clippy::needless_range_loop)]
         for i in 0..len {
             if i % 2 != 0 {
                 x_shares[0][i] = x[i];
@@ -204,8 +203,8 @@ mod tests {
 
         let mut got = vec![Field64::zero(); len];
         let mut buf = vec![Field64::zero(); len];
-        for j in 0..num_shares {
-            discrete_fourier_transform_inv(&mut buf, &x_shares[j], len).unwrap();
+        for share in x_shares {
+            discrete_fourier_transform_inv(&mut buf, &share, len).unwrap();
             for i in 0..len {
                 got[i] += buf[i];
             }
--- a/src/field.rs
+++ b/src/field.rs
@@ -13,6 +13,7 @@ use std::{
     cmp::min,
     convert::TryFrom,
     fmt::{Debug, Display, Formatter},
+    hash::{Hash, Hasher},
     ops::{Add, AddAssign, BitAnd, Div, DivAssign, Mul, MulAssign, Neg, Shr, Sub, SubAssign},
 };
 
@@ -118,7 +119,7 @@ macro_rules! make_field {
         $elem:ident, $int:ident, $fp:ident, $bytes:literal
     ) => {
         $(#[$meta])*
-        #[derive(Clone, Copy, PartialOrd, Ord, Hash, Default, Deserialize, Serialize)]
+        #[derive(Clone, Copy, PartialOrd, Ord, Default, Deserialize, Serialize)]
         pub struct $elem(u128);
 
         impl $elem {
@@ -143,7 +144,19 @@ macro_rules! make_field {
 
         impl PartialEq for $elem {
             fn eq(&self, rhs: &Self) -> bool {
+                // The fields included in this comparison MUST match the fields
+                // used in Hash::hash
+                // https://doc.rust-lang.org/std/hash/trait.Hash.html#hash-and-eq
                 $fp.from_elem(self.0) == $fp.from_elem(rhs.0)
+            }
+        }
+
+        impl Hash for $elem {
+            fn hash<H: Hasher>(&self, state: &mut H) {
+                // The fields included in this hash MUST match the fields used
+                // in PartialEq::eq
+                // https://doc.rust-lang.org/std/hash/trait.Hash.html#hash-and-eq
+                $fp.from_elem(self.0).hash(state);
             }
         }
 
@@ -211,6 +224,7 @@ macro_rules! make_field {
 
         impl Div for $elem {
             type Output = $elem;
+            #[allow(clippy::suspicious_arithmetic_impl)]
             fn div(self, rhs: Self) -> Self {
                 self * rhs.inv()
             }
@@ -404,9 +418,7 @@ pub fn split<F: FieldElement>(
     }
 
     let mut outp = vec![vec![F::zero(); inp.len()]; num_shares];
-    for j in 0..inp.len() {
-        outp[0][j] = inp[j];
-    }
+    outp[0][..inp.len()].clone_from_slice(inp);
 
     let mut prng = Prng::new()?;
     for i in 1..num_shares {
@@ -443,6 +455,12 @@ mod tests {
         assert_matches!(result, Err(FieldError::InputSizeMismatch));
     }
 
+    // Some of the checks in this function, like `assert_eq!(one - one, zero)`
+    // or `assert_eq!(two / two, one)` trip this clippy lint for tautological
+    // comparisons, but we have a legitimate need to verify these basics. We put
+    // the #[allow] on the whole function since "attributes on expressions are
+    // experimental" https://github.com/rust-lang/rust/issues/15701
+    #[allow(clippy::eq_op)]
     fn field_element_test<F: FieldElement>() {
         let mut prng: Prng<F> = Prng::new().unwrap();
         let int_modulus = F::modulus();
--- a/src/pcp.rs
+++ b/src/pcp.rs
@@ -75,7 +75,7 @@
 //!
 //! // The prover encodes its input and splits it into two secret shares. It
 //! // sends each share to two aggregators.
-//! let x: Boolean<Field64>= Boolean::new(true);
+//! let x: Boolean<Field64> = Boolean::new(true);
 //! let x_par = x.param();
 //! let x_shares: Vec<Boolean<Field64>> = split(x.as_slice(), 2)
 //!     .unwrap()
@@ -232,7 +232,8 @@ pub trait Value<F: FieldElement>:
     /// let v = x.valid(&mut x.gadget(), &joint_rand).unwrap();
     /// assert_eq!(v, Field64::zero());
     /// ```
-    fn valid(&self, g: &mut Vec<Box<dyn Gadget<F>>>, joint_rand: &[F]) -> Result<F, PcpError>;
+    fn valid(&self, gadgets: &mut Vec<Box<dyn Gadget<F>>>, joint_rand: &[F])
+        -> Result<F, PcpError>;
 
     /// Returns a reference to the underlying data.
     fn as_slice(&self) -> &[F];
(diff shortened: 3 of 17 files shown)


src/field.rs
```diff
@@ -143,10 +144,22 @@ macro_rules! make_field {
 
         impl PartialEq for $elem {
             fn eq(&self, rhs: &Self) -> bool {
+                // The fields included in this comparison MUST match the fields
+                // used in Hash::hash
+                // https://doc.rust-lang.org/std/hash/trait.Hash.html#hash-and-eq
                 $fp.from_elem(self.0) == $fp.from_elem(rhs.0)
             }
         }
```
cjpatton, line 153: What's going on here? Why do we need to implement this?

src/field.rs
```diff
@@ -443,6 +455,7 @@ mod tests {
         assert_matches!(result, Err(FieldError::InputSizeMismatch));
     }
 
+    #[allow(clippy::eq_op)]
```
cjpatton, line 458: What's the clippy warning that's beeing suppressed here?

src/pcp.rs
```diff
@@ -412,16 +412,17 @@ struct ProveShimGadget<F: FieldElement> {
...
         g_calls: usize,
     ) -> Result<Box<dyn Gadget<F>>, getrandom::Error> {
         let mut f_vals = vec![vec![F::zero(); 1 + g_calls]; inner.arity()];
         let mut prng = Prng::new_with_length(f_vals.len())?;
 
-        for wire in 0..f_vals.len() {
+        for wire in &mut f_vals {
             // Choose a random field element as the first point on the wire polynomial.
-            f_vals[wire][0] = prng.next().unwrap();
+            wire[0] = prng.next().unwrap();
```
cjpatton, line 425: I don't like this change. `wire` is a good name for an index into the `f_vals`, but not a good name for a pointer to one of the values.

src/pcp.rs
```diff
@@ -608,6 +610,7 @@ struct QueryShimGadget<F: FieldElement> {
 }
 
 impl<F: FieldElement> QueryShimGadget<F> {
+    #[allow(clippy::new_ret_no_self)]
```
cjpatton, line 613: What's the clippy warning?

src/pcp.rs
```diff
@@ -746,13 +750,13 @@ where
 
     // Check that each of the proof polynomials are well-formed.
     let mut verifier_len = 1;
-    for idx in 0..g.len() {
-        let next_len = 1 + g[idx].arity();
+    for gadget in &mut gadgets {
+        let next_len = 1 + gadget.arity();
```
cjpatton, line 754: Here again I prefer to use an index `idx`. The reason is that the comments refer to the `idx`-th gadget.

src/pcp/types.rs
```diff
@@ -265,22 +263,22 @@ impl<F: FieldElement> Value<F> for MeanVarUnsignedVector<F> {
 
             // Sets `x` to the `bits`-bit integer encoded by `x_vec`.
             let mut x = F::zero();
-            for l in 0..bits {
-                let w = F::from(F::Integer::try_from(1 << l).unwrap());
-                x += w * x_vec[l];
+            for (position, bit) in x_vec.iter().enumerate() {
```
cjpatton, line 266: s/position/i/

src/pcp/types.rs
```diff
@@ -265,22 +263,22 @@ impl<F: FieldElement> Value<F> for MeanVarUnsignedVector<F> {
...
+                let w = F::from(F::Integer::try_from(1 << position).unwrap());
+                x += w * *bit;
             }
 
             // The first `bits` inputs to the gadget are are `r, r^2, ..., r^bits`.
-            for l in 0..bits {
+            for item in inp.iter_mut().take(bits) {
```
cjpatton, line 272: Here again I prefer the index `l` for readability.

src/pcp/types.rs
```diff
@@ -506,8 +504,8 @@ mod tests {
             let xx = chunk[bits];
 
             let mut x = TestField::zero();
-            for l in 0..bits {
-                x += TestField::from(1 << l) * x_vec[l];
+            for (position, bit) in x_vec.iter().enumerate() {
```
cjpatton, line 507: s/position/l/

src/polynomial.rs
```diff
@@ -370,6 +370,7 @@ fn test_fft() {
         false,
         &mut mem.fft_memory,
     );
```
cjpatton, line 372: nit: Add newline

src/polynomial.rs
```diff
@@ -147,8 +147,8 @@ fn fft_interpolate_raw<F: FieldElement>(
...
     if invert {
         let n_inverse = F::from(F::Integer::try_from(n_points).unwrap()).inv();
-        for i in 0..n_points {
-            out[i] *= n_inverse;
+        for item in out.iter_mut().take(n_points) {
```
cjpatton, line 150: Not my code, but here again I think the index `i` is clearer.

cjpatton (changes requested):
Clippy you magnificent bastard. I'd like to request a few changes, but I'm glad we're doing this.

src/field.rs
tgeoghegan: As noted in the docs on trait `Hash`, it must hold that `k1 == k2 -> hash(k1) == hash(k2)` or bad things will happen if a `Field` is ever used as a key in a `HashMap`. IIUC, the derived `Hash` implementation that I replaced with the new impl would do `self.0.hash(state)`, meaning it wasn't hashing the value modulo the field prime, and this would violate the relation between `Eq` and `Hash` that Rust wants us to enforce.

The lint in question:  https://rust-lang.github.io/rust-clippy/master/index.html#derive_hash_xor_eq

src/field.rs
tgeoghegan: [eq_op](https://rust-lang.github.io/rust-clippy/master/index.html#eq_op): clippy warns about seemingly tautological comparisons, like `assert_eq!(one - one, zero);` and `assert_eq!(two / two, one);` in this function. In our case, it makes sense to check that such obvious seeming expressions evaluate as expected.

Unfortunately you can't put the `#[allow]` directly on the offending lines because ["`attributes on expressions are experimental`"](https://github.com/rust-lang/rust/issues/15701). I'll add a comment explaining what we're silencing.

src/pcp.rs
tgeoghegan: It is unhappy because this returns `Box<dyn Gadget<F>>` instead of `Self` or `Box<Self>` or some other type that references `Self`. I'll see if I can fix this better by changing the function signature.

src/pcp.rs
tgeoghegan: I was on the fence about a lot of these, hence all the cases where I silenced the linter instead of introducing an iterator. I think I agree with you: the use of iterators is presumably natural for C++ programmers but I find them awkward coming from C, and agree that using indices is clearer for this usage. I'll flip this and the others you flagged back to the previous form and silence clippy where necessary.

src/field.rs
cjpatton: Thanks for clarifying.

src/pcp.rs
cjpatton: I'd be happy with a signature change.

src/field.rs
cjpatton: looks good.

src/pcp.rs
```diff
@@ -412,24 +412,21 @@ struct ProveShimGadget<F: FieldElement> {
 }
 
 impl<F: FieldElement> ProveShimGadget<F> {
-    #[allow(clippy::new_ret_no_self)]
-    fn new(
-        inner: Box<dyn Gadget<F>>,
-        g_calls: usize,
-    ) -> Result<Box<dyn Gadget<F>>, getrandom::Error> {
+    fn new(inner: Box<dyn Gadget<F>>, g_calls: usize) -> Result<Self, PcpError> {
```
cjpatton, line 415: Change `g_calls` to `gadget_calls` for consistency with related changes elsewhere.

src/pcp.rs
```diff
@@ -412,24 +412,21 @@ struct ProveShimGadget<F: FieldElement> {
...
 
-        Ok(Box::new(Self {
+        Ok(Self {
             inner,
             f_vals,
             ct: 1,
-        }))
+        })
     }
 }
```
cjpatton, line 432: (GitHub won't let me reach below here.) Several lines down from here we declare two variables, `g_deg` and `g_arity`. We should change these to `gadget_deg` and `gadget_arity` for consistency with related changes elsewhere.

src/fft.rs
```diff
@@ -27,12 +27,15 @@ pub enum FftError {
 /// Interpreting the input as the coefficients of a polynomial, the output is equal to the input
 /// evaluated at points `p^0, p^1, ... p^(size-1)`, where `p` is the `2^size`-th principal root of
 /// unity.
```
cjpatton, line 29: nit: Add blank comment line
```rust
//
```

src/fft.rs
```diff
@@ -46,13 +49,9 @@ pub fn discrete_fourier_transform<F: FieldElement>(
         return Err(FftError::SizeInvalid);
     }
 
-    for i in 0..size {
+    for (i, item) in outp.iter_mut().enumerate().take(size) {
```
cjpatton, line 52: I think here i'd prefer a needless range loop :)

src/pcp.rs
```diff
@@ -705,21 +710,21 @@ impl<F: FieldElement> TryFrom<&[Verifier<F>]> for Verifier<F> {
 
     /// Returns the verifier corresponding to a sequence of verifier shares.
     fn try_from(vf_shares: &[Verifier<F>]) -> Result<Verifier<F>, PcpError> {
-        if vf_shares.len() == 0 {
```
cjpatton, line 708: How about s/vf/verifier/ for consistency with related changes in this PR?

src/pcp/gadgets.rs
```diff
@@ -274,11 +266,10 @@ impl<F: FieldElement> MeanVarUnsigned<F> {
         Ok(())
     }
 
-    pub(crate) fn call_poly_fft(
-        &mut self,
-        outp: &mut [F],
-        inp: &Vec<Vec<F>>,
-    ) -> Result<(), PcpError> {
+    // The single character variable names used in this function are consistent
+    // with the literature on FFTs
```
cjpatton, line 270:
```suggestion
    // with the literature on FFTs.
```

src/pcp/gadgets.rs
```diff
@@ -274,11 +266,10 @@ impl<F: FieldElement> MeanVarUnsigned<F> {
         Ok(())
     }
 
-    pub(crate) fn call_poly_fft(
-        &mut self,
-        outp: &mut [F],
-        inp: &Vec<Vec<F>>,
-    ) -> Result<(), PcpError> {
+    // The single character variable names used in this function are consistent
```
cjpatton, line 269: I don't think this comment is necessary, but I don't object to it.

src/pcp/types.rs
```diff
@@ -265,12 +263,13 @@ impl<F: FieldElement> Value<F> for MeanVarUnsignedVector<F> {
 
             // Sets `x` to the `bits`-bit integer encoded by `x_vec`.
             let mut x = F::zero();
-            for l in 0..bits {
-                let w = F::from(F::Integer::try_from(1 << l).unwrap());
-                x += w * x_vec[l];
+            for (i, bit) in x_vec.iter().enumerate() {
```
cjpatton, line 266: Can we change `i` back to `l`?

src/fft.rs
```diff
@@ -27,12 +27,15 @@ pub enum FftError {
 /// Interpreting the input as the coefficients of a polynomial, the output is equal to the input
 /// evaluated at points `p^0, p^1, ... p^(size-1)`, where `p` is the `2^size`-th principal root of
 /// unity.
+// The single character variable names used here are consistent with much of the
```
cjpatton, line 30: I don't think this comment is necessary, but I don't object to it.

cjpatton (changes requested):

src/pcp.rs
tgeoghegan: With pleasure! I expanded a few more terse variable names while I was in there.

src/pcp/gadgets.rs
tgeoghegan: My thought is that any time we opt out of any kind of static analysis we should explain why, but I haven't done that consistently, and at some point it's more noise than it's worth, so I deleted the comment.

src/pcp.rs
```diff
@@ -329,69 +330,73 @@ pub trait Gadget<F: FieldElement> {
...
+    let gadget_calls = input.valid_gadget_calls();
 
-    let mut shim = x
+    let mut shim = input
         .gadget()
         .into_iter()
         .enumerate()
-        .map(|(idx, g)| Ok(Box::new(ProveShimGadget::new(g, g_calls[idx])?) as Box<dyn Gadget<F>>))
+        .map(|(idx, g)| {
+            Ok(Box::new(ProveShimGadget::new(g, gadget_calls[idx])?) as Box<dyn Gadget<F>>)
```
cjpatton, line 352: You could s/g/gadget/ here as well

src/pcp.rs
```diff
@@ -329,69 +330,73 @@ pub trait Gadget<F: FieldElement> {
...
     // polynomials for each gadget in the next step.
-    let _ = x.valid(&mut shim, joint_rand);
+    let _ = input.valid(&mut shim, joint_rand);
 
     // Fill the buffer with the proof. `proof_len` keeps track of the amount of data written to the
     // buffer so far.
     let mut proof_len = 0;
     for idx in 0..shim.len() {
-        let g = shim[idx]
+        let proof_gadget = shim[idx]
```
cjpatton, line 374: `gadget` is fine IMO. If you want to give this a more specific name than `gadget`, I would go with `prove_shim_gadget`, since it's snake case version of `ProveShimGadget`. `proof_gadget` is a new word, and it's a little confusing. You would also want to make a similar change in `verify()`.

src/pcp.rs
```diff
@@ -329,69 +330,73 @@ pub trait Gadget<F: FieldElement> {
...
             .as_any()
             .downcast_mut::<ProveShimGadget<F>>()
             .unwrap();
 
-        let g_deg = g.deg();
-        let g_arity = g.arity();
+        let proof_gadget_degree = proof_gadget.degree();
```
cjpatton, line 379: Same here. I would just say `gadget_degree` and `gadget_arity`

cjpatton (changes requested):
Just some gripes about variable names.

src/pcp.rs
tgeoghegan: There's no `fn verify` but I double checked `fn prove` and `fn query` for consistency.

src/pcp.rs
tgeoghegan: I did you one better and eliminated the variables and to instead just call `gadget.degree()` and `gadget.arity()`. All the implementations of those methods should be quite amenable to inlining by the compiler.

cjpatton (approved):

PR #74, opened 2021-09-11
prng: Define the StreamCipher trait

cjpatton:
This PR does two things:
- Refactors `Prng` to add support for stream ciphers besides AES128-CTR.
- Adds support for BLAKE3.

Notes:
- AES128-CTR is still hard-coded for `rand()` and `split()`. Generalizing these will require making the `prng` module public, but we can do this in a later PR.
- The use of const generics makes instantiating a `Prng` a bit ugly. I think it would be nice to wrap this with a macro per implementation, e.g., `prng_aes128ctr!()` and `prng_blake3!()`. However, I'm not quite sure how to make Rust do what I want it to do.
- BLAKE3 might be slower than AES128-CTR for this application since it lacks hardware support. We should add some benchmarks. The reason I wanted to consider adding it is that we will need a hash function and it's nice to able to use the same primitive for two purposes. But maybe this isn't a good enough reason to add it.
- Another implementation we might consider adding is ChaCha20.

### DIFF
--- a/src/prng.rs
+++ b/src/prng.rs
@@ -9,7 +9,6 @@ use aes::{
     },
     Aes128, Aes128Ctr,
 };
-use blake3;
 use getrandom::getrandom;
 
 use std::marker::PhantomData;
--- a/src/prng.rs
+++ b/src/prng.rs
@@ -3,42 +3,95 @@
 
 use super::field::{FieldElement, FieldError};
 use aes::{
-    cipher::{generic_array::GenericArray, FromBlockCipher, NewBlockCipher, StreamCipher},
+    cipher::{
+        generic_array::GenericArray, FromBlockCipher, NewBlockCipher,
+        StreamCipher as AesStreamCipher,
+    },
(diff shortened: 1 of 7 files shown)


src/prng.rs
```diff
@@ -3,42 +3,95 @@
...
+            Some(length),
+        )
+        .collect(),
+    )
+}
+
+/// A stream cipher maps a short key to a stream of pseudorandom bytes.
+pub(crate) trait StreamCipher<const KEY_LENGTH: usize> {
+    /// Returns a new stream ciphered keyed by `key`.
+    fn new(key: [u8; KEY_LENGTH]) -> Self;
```
cjpatton, line 62: Instead of a generic const, ideally `KEY_LENGTH` would be a trait parameter. However, we can't do
```rust
fn new(key: [u8; Self::KEY_LENGTH]) -> Self;
```

cjpatton:
I'm gonna try something a little different.

PR #76, opened 2021-09-13
Implement prototype of the core PPM protocol

cjpatton:
Initial design for the core PPM scheme for Prio. This has not undergone
significant analysis, but is nonetheless useful for prototyping the full
protocol.

This commit also exposes the prng::PrngError in the public API. While
not strictly necessary, it simplifies the task of converting this
internal error. The assumption is that we're going to need this module
in the public API at some point anyway.

### DIFF
--- a/src/ppm.rs
+++ b/src/ppm.rs
@@ -1,0 +1,497 @@
+// SPDX-License-Identifier: MPL-2.0
+
+//! This module implements core functionality of the PPM protocol for Prio. It consists of the
+//! following functions:
+//!
+//! * `upload` is run by the client to generate the input and proof shares.
+//!
+//! * `verify_start` is run by each aggregator upon receiving its input and proof share. The
+//! output is broadcast to the other aggregators.
+//!
+//! * `verify_finish` is run by each aggregator to decide if the input was valid.
+//!
+//! In the special case of two aggregators (one leader, one helper), these calls map to HTTP
+//! requests as follows (assuming they agree on the query randomness in advance):
+//!
+//! ```none
+//! Client                 Leader                    Helper
+//!   | upload()             |                         |
+//!   |                      |                         |
+//!   |    upload request    |                         |
+//!   |--------------------->| verify_start()          |
+//!   |                      |                         |
+//!   |                      |    aggregate request    |
+//!   |                      |------------------------>| verify_start()
+//!   |                      |                         |
+//!   |                      |                         |
+//!   |                      |<------------------------|
+//!   |                      | verify_finish()         | verify_finish()
+//!   v                      v                         v
+//!                        output share              output share
+//! ```
+//!
+//! NOTE: This protocol implemented here is a prototype and has not undergone security analysis.
+//! Use at your own risk.
+
+use std::convert::TryFrom;
+
+use getrandom::getrandom;
+
+use crate::field::{rand, FieldElement};
+use crate::pcp::types::TypeError;
+use crate::pcp::{decide, prove, query, PcpError, Proof, Value, Verifier};
+use crate::prng::{Prng, PrngError, SEED_LENGTH};
+
+/// Errors emitted by this module.
+#[derive(Debug, thiserror::Error)]
+pub enum PpmError {
+    /// An error occurred.
+    #[error("ppm error: {0}")]
+    Ppm(String),
+
+    /// The distributed input was deemed invalid.
+    #[error("ppm error: invalid distributed input: {0}")]
+    Validity(&'static str),
+
+    /// PCP error.
+    #[error("ppm error: pcp error: {0}")]
+    Pcp(#[from] PcpError),
+
+    /// Type error.
+    #[error("ppm error: type error: {0}")]
+    Type(#[from] TypeError),
+
+    /// PRNG error.
+    #[error("ppm error: prng error: {0}")]
+    Prng(#[from] PrngError),
+
+    /// Calling get_random() returned an error.
+    #[error("ppm error: getrandom error: {0}")]
+    GetRandom(#[from] getrandom::Error),
+}
+
+/// A share of an input or proof for Prio.
+#[derive(Clone, Debug)]
+pub enum Share<F: FieldElement> {
+    /// An uncompressed share, typically sent to the leader.
+    Leader(Vec<F>),
+
+    /// A compressed share, typically sent to the helper.
+    Helper {
+        /// The seed for the pseudorandom generator.
+        seed: Vec<u8>,
+        /// The length of the uncompressed share.
+        length: usize,
+    },
+}
+
+impl<F: FieldElement> TryFrom<Share<F>> for Vec<F> {
+    type Error = PpmError;
+
+    fn try_from(share: Share<F>) -> Result<Self, PpmError> {
+        match share {
+            Share::Leader(data) => Ok(data),
+            Share::Helper { seed, length } => {
+                let mut prng = Prng::try_from_seed(&seed)?;
+                Ok((0..length).map(|_| prng.next().unwrap()).collect())
+            }
+        }
+    }
+}
+
+/// The message sent by the client to each aggregator. This includes the client's input share and
+/// the initial message of the input-validation protocol.
+#[derive(Clone, Debug)]
+pub struct UploadMessage<F: FieldElement> {
+    /// The input share.
+    pub input_share: Share<F>,
+
+    /// The proof share.
+    pub proof_share: Share<F>,
+
+    /// The sum of the joint randomness seed shares sent to the other aggregators.
+    pub joint_rand_seed_hint: Vec<u8>,
+
+    /// The blinding factor, used to derive the aggregator's joint randomness seed share.
+    pub blind: Vec<u8>,
+}
+
+fn assert_safe() {
+    // TODO(cjpatton) Replace this hack with a more consistent API for the Prio ciphersuite.Prng
+    // should be constructed with an implementation of the following trait:
+    //
+    // pub trait Cipher {
+    //    // Panics if `seed.len() != StreamCipher::KEY_LEN`.
+    //    fn new_stream_cipher(seed: &[u8]) -> StreamCipher;
+    //
+    //    // Panics if `Hasher::OUT_LEN != StreamCipher::KEY_LEN`.
+    //    fn new_hasher() -> Hasher;
+    // }
+    //
+    // pub trait StreamCipher {
+    //    const KEY_LEN: usize;
+    //
+    //    /// Fill `out` with the next `out.len()` bytes of the stream cipher.
+    //    fn fill(out: &mut [u8]);
+    // }
+    //
+    // pub trait Hasher<const OUT_LEN: usize> {
+    //    const OUT_LEN: usize;
+    //
+    //    /// Appends `input` to the hash input.
+    //    fn update(input: &[u8]);
+    //
+    //    /// Panics if `out.len() != Self::OUT_LEN`.
+    //    fn finalize(out: &mut [u8]);
+    // }
+    const_assert_eq!(blake3::KEY_LEN, SEED_LENGTH);
+    const_assert_eq!(blake3::OUT_LEN, SEED_LENGTH);
+}
+
+#[derive(Clone)]
+struct HelperShare {
+    input_share: Vec<u8>,
+    proof_share: Vec<u8>,
+    joint_rand_seed_hint: Vec<u8>,
+    blind: Vec<u8>,
+}
+
+impl HelperShare {
+    fn new() -> Self {
+        HelperShare {
+            input_share: vec![0; SEED_LENGTH],
+            proof_share: vec![0; SEED_LENGTH],
+            joint_rand_seed_hint: vec![0; SEED_LENGTH],
+            blind: vec![0; SEED_LENGTH],
+        }
+    }
+}
+
+/// Run by the client, this generates the sequence of [`UploadMessage`] messages to send to the
+/// aggregators.
+///
+/// # Parameters
+///
+/// * `input` is the input to be secret shared.
+/// * `num_shares` is the number of input shares (i.e., aggregators) to generate.
+pub fn upload<F, V>(input: &V, num_shares: u8) -> Result<Vec<UploadMessage<F>>, PpmError>
+where
+    F: FieldElement,
+    V: Value<F>,
+{
+    assert_safe();
+
+    if num_shares < 2 {
+        return Err(PpmError::Ppm(format!(
+            "upload(): at least 2 shares are required; got {}",
+            num_shares
+        )));
+    }
+
+    let input_len = input.as_slice().len();
+    let num_shares = num_shares as usize;
+
+    // Generate the input shares and compute the joint randomness.
+    let mut helper_shares = vec![HelperShare::new(); num_shares - 1];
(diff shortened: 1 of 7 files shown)


cjpatton:
Rebased.

tgeoghegan:
@cjpatton  This PR still has commit f683ddc404da6e501676a45d9a5aca1baf218fa0 in it. Could you please rebase? Also, clippy is very, very mad at you.

src/ppm.rs
```diff
@@ -0,0 +1,473 @@
...
+    #[error("ppm error: prng error: {0}")]
+    Prng(#[from] PrngError),
+
+    /// Calling get_random() returned an error.
+    #[error("ppm error: getrandom error: {0}")]
+    GetRandom(#[from] getrandom::Error),
+}
+
+/// A share of an input or proof for Prio.
+pub enum PrioShare<F: FieldElement> {
```
tgeoghegan, line 69: I suggest deriving at least `Clone` and `Debug` on this, and all other structs in the crate. I always end up wanting everything to be `Clone` and `Debug`.

src/ppm.rs
```diff
@@ -0,0 +1,473 @@
...
+            blind: col.next().unwrap(),
+        });
+    }
+
+    Ok(msgs)
+}
+
+/// The message sent by an aggregator to every other aggregator. This is the final message of the
+/// input-validation protocol.
+#[derive(Clone)]
```
tgeoghegan, line 277:
```suggestion
#[derive(Clone, Debug)]
```

src/ppm.rs
```diff
@@ -0,0 +1,473 @@
...
+pub struct PrioVerifier<F: FieldElement> {
+    /// The aggregator's share of the verifier message.
+    pub verifier_share: Vec<F>,
```
tgeoghegan, line 280:
```suggestion
    pub verifier_share: Verifier<F>,
```
`Verifier<F>` is basically just `Vec<F>` anyway, but it's more clear -- so many different things are represented as `Vec<F>` or `&[F]` in this library that it's easy to get confused or pass the wrong variable somewhere.

src/ppm.rs
tgeoghegan, line 69: I don't think we need every item in this module to start with `Prio`, because we're already in crate `prio` and any client who is confused can use this as `prio::ppm::Share`.

src/ppm.rs
```diff
@@ -0,0 +1,473 @@
...
+/// # Parameters
+///
+/// * `input` is the input to be secret shared.
+/// * `num_shares` is the number of input shares (i.e., aggregators) to generate.
+pub fn prio_upload<F, V>(input: &V, num_shares: usize) -> Result<Vec<PrioUpload<F>>, PpmError>
+where
+    F: FieldElement,
+    V: Value<F>,
+{
+    assert_safe();
```
tgeoghegan, line 156: [Rust doesn't have static assertions yet](https://github.com/rust-lang/rfcs/issues/2790), but the [`static_assertions`](https://docs.rs/static_assertions/1.1.0/static_assertions/index.html) crate looks like it'll work nicely.

src/ppm.rs
```diff
@@ -0,0 +1,473 @@
...
+    query_rand_seed: &[u8],
+) -> Result<(PrioAggregatorState<F, V>, PrioVerifier<F>), PpmError>
+where
+    F: FieldElement,
+    V: Value<F>,
+{
+    assert_safe();
+
+    let input_share_data: Vec<F> = Vec::try_from(msg.input_share)?;
+    let input_share = V::try_from((param, &input_share_data)).map_err(|err| {
```
tgeoghegan, line 319: It's unfortunate to throw away the structure of the underlying error. I assume the problem here is that the various `Value::try_from` implementations could yield all sorts of different errors? We could provide a variant on `PpmError` like `Conversion(&'static str, #[source] Box<dyn Debug>)`, and then do 

```
let input_share = V::try_from(...).map_err(|err| {
    PpmError::Conversion("can't convert the thing to a thing!", Box::new(err))
})
```
Then other crates like [`eyre_report`](https://docs.rs/eyre/0.6.5/eyre/struct.Report.html) can still print good error chains.

tgeoghegan (commented):
First cut of review comments. I want to try using this in `ppm-prototype` to see how it fits together and may come back with more!

src/ppm.rs
```diff
@@ -0,0 +1,473 @@
...
+        )));
+    }
+
+    let input_len = input.as_slice().to_vec().len();
+
+    let mut helper = vec![vec![vec![0; SEED_LENGTH]; 4]; num_shares - 1];
+    const INPUT_SHARE: usize = 0;
+    const PROOF_SHARE: usize = 1;
+    const JOINT_RAND_HINT: usize = 2;
+    const BLIND: usize = 3;
```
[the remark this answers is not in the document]
cjpatton: *MUCH* better. Done.

src/ppm.rs
```diff
@@ -0,0 +1,473 @@
...
+
+    for row in helper.into_iter() {
+        aggregator_id += 1;
+        let mut col = row.into_iter();
+        msgs.push(PrioUpload {
+            aggregator_id,
+            input_share: PrioShare::Helper(col.next().unwrap(), input_len),
+            proof_share: PrioShare::Helper(col.next().unwrap(), proof_len),
+            joint_rand_seed_hint: col.next().unwrap(),
+            blind: col.next().unwrap(),
```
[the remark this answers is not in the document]
cjpatton: Done.

src/ppm.rs
cjpatton: Done.

src/ppm.rs
```diff
@@ -0,0 +1,473 @@
...
+    pub input_share: PrioShare<F>,
+
+    /// The proof share.
+    pub proof_share: PrioShare<F>,
+
+    /// The sum of the joint randomness seed shares sent to the other aggregators.
+    pub joint_rand_seed_hint: Vec<u8>,
+
+    /// The blinding factor, used to derive the aggregator's joint randomness seed share.
+    pub blind: Vec<u8>,
```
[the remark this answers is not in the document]
cjpatton: That would allow the client to choose *any* joint rand it wants, which violates soundness. Deriving joint rand from the input shares removes the client's influence on the distribution of the joint rand, at least when we model the hash function as a random oracle. The purpose of `blind`, at a very high level, is to prevent the joint rand shares from leaking information about the input shares to other aggregators.

Bare in mind that all of this needs a proof, but what we're following the lead of the Crypto 2019 paper. This technique is suggested, though not spelled out completely, in Section 6.

src/ppm.rs
```diff
@@ -0,0 +1,499 @@
...
+    hasher.update(&leader_blind);
+    for x in leader_input_share.iter() {
+        hasher.update(&(*x).into());
+    }
+
+    let mut leader_joint_rand_seed_hint = vec![0; SEED_LENGTH];
+    leader_joint_rand_seed_hint.copy_from_slice(hasher.finalize().as_bytes());
+    for j in 0..SEED_LENGTH {
+        joint_rand_seed[j] ^= leader_joint_rand_seed_hint[j];
+    }
```
tgeoghegan, line 235: nit: I'm fine with this, but for consistency with the rest of the function:
```
for (x, y) in joint_rand_seed.iter_mut().zip(leader_joint_rand_seed_hint.iter()) {
    *x ^= y;
}
```

src/ppm.rs
tgeoghegan: I agree in the case of `Share` because it's useful for it to be flexible, but here we can gain some clarity without loss of usefulness of this struct. That said we then end up with a collision between `ppm::Verifier` and `pcp::Verifier`. Maybe this should be `ppm::VerifierMessage`?

src/ppm.rs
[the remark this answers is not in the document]
tgeoghegan: I see what I was missing now -- in Prio v2, we trusted the ingestion servers to provide joint randomness, not the client, but we do not assume any such actor in PPM. This makes sense to me now, thank you!

src/ppm.rs
```diff
@@ -0,0 +1,499 @@
...
+    //
+    //    /// Panics if `out.len() != Self::OUT_LEN`.
+    //    fn finalize(out: &mut [u8]);
+    // }
+    const_assert_eq!(blake3::KEY_LEN, SEED_LENGTH);
+    const_assert_eq!(blake3::OUT_LEN, SEED_LENGTH);
+}
+
+#[derive(Clone)]
+struct HelperShare {
```
tgeoghegan, line 149: nit: A module-private struct is perfectly adequate, but if you want to hide it even harder you could put the definition inside `fn upload`.

src/ppm.rs
```diff
@@ -0,0 +1,473 @@
...
+}
+
+/// A share of an input or proof for Prio.
+pub enum PrioShare<F: FieldElement> {
+    /// An uncompressed share, typically sent to the leader.
+    Leader(Vec<F>),
+
+    /// A compressed share, typically sent to the helper. The first parameter is a seed for a
+    /// pseudorandom generator
+    Helper(Vec<u8>, usize),
```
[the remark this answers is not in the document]
tgeoghegan: I'm happy to defer discussion of `SEED_LENGTH` until the `Prng` API solidifies, but I think this would still be better:
```suggestion
    Helper {
        seed: Vec<u8>,
        uncompressed_length: usize,
    },
```

src/ppm.rs
```diff
@@ -0,0 +1,499 @@
...
+    };
+
+    Ok((state, out_msg))
+}
+
+/// Run by each aggregator, this consumes the [`Verifier`] messages broadcast by each of the
+/// other aggregators. It returns the aggregator's input share only if the input is valid.
+pub fn verify_finish<F, V>(
+    state: AggregatorState<F, V>,
+    msgs: Vec<Verifier<F>>,
```
tgeoghegan, line 393:
```suggestion
    verifier_messages: Vec<Verifier<F>>,
```

src/ppm.rs
```diff
@@ -0,0 +1,499 @@
...
+/// returns the aggregator's state.
+///
+/// # Parameters
+///
+/// * `msg` is the message sent by the client. It contains, among other things, an input and proof
+/// share.
+/// * `param` is used to reconstruct the input share from the raw message.
+/// * `query_rand_seed` is used to derive the query randomness shared by all the aggregators.
+pub fn verify_start<F, V>(
+    msg: Upload<F>,
```
tgeoghegan, line 327:
```suggestion
    upload_message: Upload<F>,
```

src/ppm.rs
```diff
@@ -0,0 +1,499 @@
...
+    for x in input_share.as_slice() {
+        hasher.update(&(*x).into());
+    }
+    joint_rand_seed_share.copy_from_slice(hasher.finalize().as_bytes());
+
+    let mut joint_rand_seed = vec![0; SEED_LENGTH];
+    for j in 0..SEED_LENGTH {
+        joint_rand_seed[j] = msg.joint_rand_seed_hint[j] ^ joint_rand_seed_share[j];
+    }
+
```
tgeoghegan, line 358: nit: use the `for (x, y) in joint_rand_seed.iter_mut().zip(joint_rand_seed_share.iter()) {}` form

src/ppm.rs
```diff
@@ -0,0 +1,499 @@
...
+    let mut prng = Prng::try_from_seed(&joint_rand_seed)?;
+    let joint_rand: Vec<F> = (0..input_share.joint_rand_len())
+        .map(|_| prng.next().unwrap())
+        .collect();
+
+    // Compute the query randomness.
+    let mut prng = Prng::try_from_seed(query_rand_seed)?;
+    let query_rand: Vec<F> = (0..input_share.query_rand_len())
+        .map(|_| prng.next().unwrap())
+        .collect();
```
tgeoghegan, line 368: nit: just because I find `unwrap()` and closures that discard their arguments unsightly:
```suggestion
    let prng = Prng::try_from_seed(query_rand_seed)?;
    let query_rand: Vec<F> = prng.take(input_share.query_rand_len()).collect();
```

src/ppm.rs
```diff
@@ -0,0 +1,473 @@
...
+}
+
+/// Run by the client, this generates the sequence of [`PrioUpload`] messages to send to the
+/// aggregators.
+///
+/// # Parameters
+///
+/// * `input` is the input to be secret shared.
+/// * `num_shares` is the number of input shares (i.e., aggregators) to generate.
+pub fn prio_upload<F, V>(input: &V, num_shares: usize) -> Result<Vec<PrioUpload<F>>, PpmError>
```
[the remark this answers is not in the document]
tgeoghegan: I agree with you that it's difficult to imagine having deployments with more than a single digit number of helpers, let alone more than 255. IIUC this is important because we need to define the length of the aggregator ID that gets hashed into the seed hint. We haven't settled on boundaries for this at the PPM layer yet (which I think should be covered by [#138](https://github.com/abetterinternet/ppm-specification/issues/138), so I think we should settle that matter in PPM and then reflect the decision here.

src/ppm.rs
```diff
@@ -0,0 +1,473 @@
...
+    leader_joint_rand_seed_hint.copy_from_slice(hasher.finalize().as_bytes());
+    for j in 0..SEED_LENGTH {
+        joint_rand_seed[j] ^= leader_joint_rand_seed_hint[j];
+    }
+
+    // Run the proof-generation algorithm.
+    let mut prng = Prng::from_seed(&joint_rand_seed)?;
+    let joint_rand: Vec<F> = (0..input.joint_rand_len())
+        .map(|_| prng.next().unwrap())
+        .collect();
```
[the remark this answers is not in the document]
tgeoghegan: I think [this](https://github.com/abetterinternet/libprio-rs/pull/76#discussion_r708529562) is a better suggestion than exposing that construtor.

src/ppm.rs
cjpatton: I have a weak preference for a short and generic name for messages?

src/ppm.rs
cjpatton: Doesn't quite work here because we have three variables instead of just two: joint_rand_seed, msg.joint_rand_seed_hint, and joint_rand_seed_share. I changed the iterator to `joint_rand_seed.iter_mut().enumerate()`, which is a bit more consistent.

src/ppm.rs
cjpatton: Oooh pretty!

src/ppm.rs
cjpatton: Again, I have a weak preference for a short, generic name here.

src/ppm.rs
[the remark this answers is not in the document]
cjpatton: SGTM.

src/ppm.rs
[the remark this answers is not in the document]
tgeoghegan: Yes, that's correct, thanks for the clarification.

src/ppm.rs
tgeoghegan: Nope, I'm fine with this being defined at the module level.

src/ppm.rs
tgeoghegan: It's clear enough as a local variable name,

tgeoghegan (approved):

cjpatton:
Squashed and rebased.

PR #77, opened 2021-09-16
prng: Define the StreamCipher trait

cjpatton:
Struct Prng uses AES128-CTR as a stream cipher. This commit defines a
trait StreamCipher for stream ciphers and generalizes Prng to use any
implementation of this trait. It also adds an implementation of
StreamCipher for BLAKE3 (https://docs.rs/blake3/1.0.0/blake3). While at
it, this commit reduces the internal API surface for Prng.

### DIFF
--- a/src/field.rs
+++ b/src/field.rs
@@ -7,7 +7,7 @@
 //! subgroup of order `2^n` for some `n`.
 
 use crate::fp::{FP126, FP32, FP64, FP80};
-use crate::prng::Prng;
+use crate::prng::{random_vector, PrngError};
 use serde::{Deserialize, Serialize};
 use std::{
     cmp::min,
@@ -531,30 +531,21 @@ pub fn merge_vector<F: FieldElement>(
     Ok(())
 }
 
-/// Generate a vector of uniform random field elements.
-pub fn rand<F: FieldElement>(len: usize) -> Result<Vec<F>, getrandom::Error> {
-    Ok(Prng::new_with_length(len)?.collect())
-}
-
 /// Outputs an additive secret sharing of the input.
-pub fn split<F: FieldElement>(
-    inp: &[F],
-    num_shares: usize,
-) -> Result<Vec<Vec<F>>, getrandom::Error> {
+pub fn split<F: FieldElement>(inp: &[F], num_shares: usize) -> Result<Vec<Vec<F>>, PrngError> {
     if num_shares == 0 {
         return Ok(vec![]);
     }
 
-    let mut outp = vec![vec![F::zero(); inp.len()]; num_shares];
-    outp[0][..inp.len()].clone_from_slice(inp);
+    let mut outp = Vec::with_capacity(num_shares);
+    outp.push(inp.to_vec());
 
-    let mut prng = Prng::new()?;
-    for i in 1..num_shares {
-        for j in 0..inp.len() {
-            let r = prng.next().unwrap();
-            outp[i][j] = r;
-            outp[0][j] -= r;
+    for _ in 1..num_shares {
+        let share: Vec<F> = random_vector(inp.len())?;
+        for (x, y) in outp[0].iter_mut().zip(&share) {
+            *x -= *y;
         }
+        outp.push(share);
     }
 
     Ok(outp)
@@ -564,6 +555,8 @@ pub fn split<F: FieldElement>(
 mod tests {
     use super::*;
     use crate::fp::MAX_ROOTS;
+    use crate::prng::Prng;
+    use aes::Aes128Ctr;
     use assert_matches::assert_matches;
     use std::io::{Cursor, Write};
 
@@ -600,7 +593,7 @@ mod tests {
     // experimental" https://github.com/rust-lang/rust/issues/15701
     #[allow(clippy::eq_op)]
     fn field_element_test<F: FieldElement>() {
-        let mut prng: Prng<F> = Prng::new().unwrap();
+        let mut prng: Prng<F, Aes128Ctr> = Prng::new().unwrap();
         let int_modulus = F::modulus();
         let int_one = F::Integer::try_from(1).unwrap();
         let zero = F::zero();
--- a/src/prng.rs
+++ b/src/prng.rs
@@ -7,43 +7,102 @@
 
 use super::field::{FieldElement, FieldError};
 use aes::{
-    cipher::{generic_array::GenericArray, FromBlockCipher, NewBlockCipher, StreamCipher},
+    cipher::{
+        generic_array::GenericArray, FromBlockCipher, NewBlockCipher,
+        StreamCipher as AesStreamCipher,
+    },
     Aes128, Aes128Ctr,
 };
 use getrandom::getrandom;
 
 use std::marker::PhantomData;
 
-const BLOCK_SIZE: usize = 16;
-const DEFAULT_BUFFER_SIZE_IN_ELEMENTS: usize = 128;
-const MAXIMUM_BUFFER_SIZE_IN_ELEMENTS: usize = 4096;
+const BUFFER_SIZE_IN_ELEMENTS: usize = 128;
 
-pub(crate) const SEED_LENGTH: usize = 2 * BLOCK_SIZE;
-
-pub(crate) fn secret_share<F: FieldElement>(share1: &mut [F]) -> Result<Vec<u8>, getrandom::Error> {
-    let mut seed = vec![0; SEED_LENGTH];
-    getrandom(seed.as_mut_slice())?;
+pub(crate) fn secret_share<F: FieldElement>(share1: &mut [F]) -> Result<Vec<u8>, PrngError> {
+    let mut seed = [0; Aes128Ctr::STREAM_CIPHER_SEED_SIZE];
+    getrandom(&mut seed)?;
 
     // get prng array
-    let data: Vec<F> = Prng::new_with_seed_and_optional_length(&seed, Some(share1.len())).collect();
+    let data: Vec<F> = Prng::<F, Aes128Ctr>::new_with_seed(&seed)?
+        .take(share1.len())
+        .collect();
 
     // secret share
     for (s1, d) in share1.iter_mut().zip(data.iter()) {
         *s1 -= *d;
     }
 
-    Ok(seed)
+    Ok(seed.to_vec())
 }
 
 pub(crate) fn extract_share_from_seed<F: FieldElement>(
     length: usize,
     seed: &[u8],
 ) -> Result<Vec<F>, PrngError> {
-    if seed.len() != 2 * BLOCK_SIZE {
-        return Err(PrngError::SeedLen);
+    Ok(Prng::<F, Aes128Ctr>::new_with_seed(seed)?
+        .take(length)
+        .collect())
+}
+
+/// Generate a vector of uniform random field elements.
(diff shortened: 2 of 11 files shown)


src/prng.rs
```diff
@@ -7,51 +7,109 @@
...
+pub fn random_vector<F: FieldElement>(len: usize) -> Result<Vec<F>, PrngError> {
+    Ok(Prng::<_, Aes128Ctr>::new()?.take(len).collect())
+}
+
+/// A stream cipher maps a short key to a stream of pseudorandom bytes.
+pub(crate) trait StreamCipher: Sized {
+    const STREAM_CIPHER_SEED_SIZE: usize;
+
+    /// Returns a new stream ciphered keyed by `key`.
+    fn new(key: &[u8]) -> Result<Self, PrngError>;
```
tgeoghegan, line 58:
```suggestion
    fn new(key: &[u8; Self::STREAM_CIPHER_SEED_SIZE]) -> Result<Self, PrngError>;
```
Then you can skip the seed length checks in impls.

src/prng.rs
```diff
@@ -7,51 +7,109 @@
...
+impl StreamCipher for blake3::OutputReader {
+    const STREAM_CIPHER_SEED_SIZE: usize = 32;
+
+    fn new(key: &[u8]) -> Result<Self, PrngError> {
+        if key.len() != Self::STREAM_CIPHER_SEED_SIZE {
+            return Err(PrngError::SeedSize);
+        }
+
+        let mut key_copy = [0; Self::STREAM_CIPHER_SEED_SIZE];
+        key_copy.clone_from_slice(key);
```
tgeoghegan, line 96: Is this copy needed just to go from `&[u8]` to `&[u8; 32]`? I think you could skip this if you had `StreamCipher::new(key: &[u8; Self::STREAM_CIPHER_SEED_SIZE]`.

src/prng.rs
```diff
@@ -7,51 +7,109 @@
...
+}
+
+/// Generate a vector of uniform random field elements.
+pub fn random_vector<F: FieldElement>(len: usize) -> Result<Vec<F>, PrngError> {
+    Ok(Prng::<_, Aes128Ctr>::new()?.take(len).collect())
+}
+
+/// A stream cipher maps a short key to a stream of pseudorandom bytes.
+pub(crate) trait StreamCipher: Sized {
+    const STREAM_CIPHER_SEED_SIZE: usize;
```
tgeoghegan, line 55: `StreamCipher::STREAM_CIPHER_SEED_SIZE` repeats itself. I think `const SEED_SIZE: usize` suffices.

src/prng.rs
```diff
@@ -61,74 +119,42 @@ pub enum PrngError {
 /// This type implements an iterator that generates a pseudorandom sequence of field elements. The
 /// sequence is derived from the key stream of AES-128 in CTR mode with a random IV.
 #[derive(Debug)]
-pub(crate) struct Prng<F: FieldElement> {
+pub(crate) struct Prng<F: FieldElement, X: StreamCipher> {
```
tgeoghegan, line 122: `C` or `SC` would be better than `X`.

src/prng.rs
```diff
@@ -61,74 +119,42 @@ pub enum PrngError {
...
 }
 
-impl<F: FieldElement> Prng<F> {
+impl<F: FieldElement, X: StreamCipher> Prng<F, X> {
     /// Generates a seed and constructs an iterator over an infinite sequence of pseudorandom field
     /// elements.
-    pub(crate) fn new() -> Result<Self, getrandom::Error> {
-        let mut seed = [0; SEED_LENGTH];
+    pub(crate) fn new() -> Result<Self, PrngError> {
+        let mut seed = vec![0; X::STREAM_CIPHER_SEED_SIZE];
```
tgeoghegan, line 134: nit: make this a byte slice on the stack instead of a heap allocation that never escapes this function.

src/prng.rs
```diff
@@ -61,74 +119,42 @@ pub enum PrngError {
...
-
-        let buf_len_in_elems = match length {
-            Some(length) => std::cmp::min(length + 1, MAXIMUM_BUFFER_SIZE_IN_ELEMENTS),
-            None => DEFAULT_BUFFER_SIZE_IN_ELEMENTS,
-        };
-        let mut buffer = vec![0; buf_len_in_elems * F::ENCODED_SIZE];
-        cipher.apply_keystream(&mut buffer);
+    pub(crate) fn new_with_seed(seed: &[u8]) -> Result<Self, PrngError> {
+        let mut key_stream = X::new(seed)?;
+        let mut buffer = vec![0; BUFFER_SIZE_IN_ELEMENTS * F::ENCODED_SIZE];
```
tgeoghegan, line 141: nit: since the size of the buffer depends upon constants, `Prng.buffer` could be a byte slice... doesn't need to be done in this PR though.

src/prng.rs
```diff
@@ -61,74 +119,42 @@ pub enum PrngError {
...
-    pub(crate) fn try_from_seed(seed: &[u8]) -> Result<Self, PrngError> {
-        if seed.len() != SEED_LENGTH {
-            return Err(PrngError::SeedLen);
-        }
-
-        Ok(Self::new_with_seed_and_optional_length(seed, None))
+        Self::new_with_seed(&seed)
     }
 
-    fn new_with_seed_and_optional_length(seed: &[u8], length: Option<usize>) -> Self {
```
tgeoghegan, line 99: 👍🏻 I'm glad to see this gone.

tgeoghegan (changes requested):
This looks great. I notice that `Prng<_, AesCtr128>` is hard coded in lots of places. What's the plan for making that generic everywhere?

src/field.rs
```diff
@@ -531,29 +531,20 @@ pub fn merge_vector<F: FieldElement>(
...
     for i in 1..num_shares {
-        for j in 0..inp.len() {
-            let r = prng.next().unwrap();
-            outp[i][j] = r;
-            outp[0][j] -= r;
+        let r: Vec<F> = random_vector(inp.len())?;
+        outp[i] = r.clone();
+        for (x, y) in outp[0].iter_mut().zip(r) {
+            *x -= y;
         }
```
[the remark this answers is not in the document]
cjpatton:
```
error[E0502]: cannot borrow `outp` as immutable because it is also borrowed as mutable
   --> src/field.rs:545:47
    |
545 |         for (x, y) in outp[0].iter_mut().zip(&outp[i]) {
    |                       ----               ---  ^^^^ immutable borrow occurs here
    |                       |                  |
    |                       |                  mutable borrow later used by call
    |                       mutable borrow occurs here

```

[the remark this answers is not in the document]
cjpatton: Found a different way, but I'm not sure it avoids a copy.

src/prng.rs
```diff
@@ -7,51 +7,109 @@
...
-const DEFAULT_BUFFER_SIZE_IN_ELEMENTS: usize = 128;
-const MAXIMUM_BUFFER_SIZE_IN_ELEMENTS: usize = 4096;
+const BUFFER_SIZE_IN_ELEMENTS: usize = 128;
 
-pub(crate) const SEED_LENGTH: usize = 2 * BLOCK_SIZE;
-
-pub(crate) fn secret_share<F: FieldElement>(share1: &mut [F]) -> Result<Vec<u8>, getrandom::Error> {
-    let mut seed = vec![0; SEED_LENGTH];
-    getrandom(seed.as_mut_slice())?;
+pub(crate) fn secret_share<F: FieldElement>(share1: &mut [F]) -> Result<Vec<u8>, PrngError> {
```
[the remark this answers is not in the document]
cjpatton: `split` is really only used for tests, I think if we kill anything we should kill that :) The only reason `split` is `pub` is because it's used for examples in the `pcp` module.

That said, both `secret_share` and `extract_share_from_seed` are remnants from the original v2 code. I'd be happy to remove them, but can it wait for a later pr?

src/prng.rs
cjpatton: If only :(
```
error: generic parameters may not be used in const operations
  --> src/prng.rs:61:25
   |
61 |     fn fella(key: &[u8; Self::STREAM_CIPHER_SEED_SIZE]) -> Self;
   |                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ cannot perform const operation using `Self`
   |
   = note: type parameters may not be used in const expressions

```

src/prng.rs
cjpatton: Sadly this doesn't work (see comment above)

src/prng.rs
cjpatton: `SEED_SIZE` is a bit ambiguous for `Aes128Ctr`, hence the prefix `STREAM_CIPHER`. Another way to disambiguate this is to define a StreamCipherAes128Ctr that implements StreamCipher, rather than implementing StreamCipher for Aes128Ctr directly.

src/prng.rs
cjpatton: Ah, good call. Done.

src/prng.rs
cjpatton: :(
```
error: generic parameters may not be used in const operations
   --> src/prng.rs:126:44
    |
126 |     buffer: [u8; BUFFER_SIZE_IN_ELEMENTS * F::ENCODED_SIZE],
    |                                            ^^^^^^^^^^^^^^^ cannot perform const operation using `F`
    |
    = note: type parameters may not be used in const expressions
```

cjpatton (commented):
> This looks great. I notice that `Prng<_, AesCtr128>` is hard coded in lots of places. What's the plan for making that generic everywhere?

* I think the v2 code can be stuck on Aes128Ctr, there's no need to generalize it I think.
* For v3 we're going to want to extend `StreamCipher` to provide a way to derive keys. This will end up looking quite a lot like HKDF (in fact, HKDF might be one instantiation). One idea I'm considering is using dynamic dispatch to get an implementation of this trait.

src/prng.rs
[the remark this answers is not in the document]
tgeoghegan: Yes, like I said, I think we should avoid any changes that are risky Priov2 until we have sufficient test coverage in place to ensure we don't regress. Can you please file an issue so we don't forget to clean this up?

src/prng.rs
tgeoghegan: Bummer! This is fine as-is then.

src/prng.rs
tgeoghegan: That's a good point, I hadn't thought about how it reads when accessed through an implementation of `StreamCipher`.

src/field.rs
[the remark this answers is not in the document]
tgeoghegan: What you've got now avoids the `clone()` so the unnecessary copy is gone.

tgeoghegan (approved):

src/prng.rs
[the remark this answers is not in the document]
cjpatton: you got it: https://github.com/abetterinternet/libprio-rs/issues/78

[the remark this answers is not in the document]
cjpatton: We have at least one test vector for `extract_share_from_seed()` in prng, but I agree that regression testing would be useful. It might also be worth expanding test vectors for v2.

cjpatton:
Squashed.

PR #79, opened 2021-09-17
Adopt serde for various types

tgeoghegan:
Adds derivations of `serde::Serialize` and `serde::Deserialize` for a
number of types. We provide custom implementations for trait
`FieldElement` for the reasons noted in the source comments. We may yet
want to provide more specialized serialization for `[F: FieldElement]`
or `Vec<F: FieldElement>`.

These custom impls replace the existing derived impls. This should not
present any backward compatibility problems because the existing client
of crate `prio` doesn't use `serde` to encode values.


cjpatton (approved):
Stamped!

PR #80, opened 2021-09-18
ppm: Have verify_finish() consume the output of verify_start()

cjpatton:
This aligns better with the spec and also simplifies the code a bit.

### DIFF
--- a/src/vdaf.rs
+++ b/src/vdaf.rs
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: MPL-2.0
 
-//! This module implements core functionality of the PPM protocol for Prio. It consists of the
-//! following functions:
+//! This module constructs  a Verifiable Distributed Aggregation Function (VDAF) from a
+//! [`pcp::Value`]. It consists of the following functions:
 //!
 //! * `upload` is run by the client to generate the input and proof shares.
 //!
@@ -41,6 +41,7 @@ use aes::Aes128Ctr;
 use getrandom::getrandom;
 use serde::{Deserialize, Serialize};
 use std::convert::TryFrom;
+use std::marker::PhantomData;
 
 /// Errors emitted by this module.
 #[derive(Debug, thiserror::Error)]
@@ -174,6 +175,7 @@ impl HelperShare {
 ///
 /// * `input` is the input to be secret shared.
 /// * `num_shares` is the number of input shares (i.e., aggregators) to generate.
+// TODO(cjpatton) Rename to dist_input to align with spec
 pub fn upload<F, V>(input: &V, num_shares: u8) -> Result<Vec<UploadMessage<F>>, PpmError>
 where
     F: FieldElement,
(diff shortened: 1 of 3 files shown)


src/vdaf.rs
```diff
@@ -405,9 +406,15 @@ where
...
+        return Err(PpmError::Ppm(
+            "verify_finish(): expected at least one inbound messages; got none".to_string(),
+        ));
+    }
+
     // Combine the verifier messages.
-    let mut verifier_data = state.verifier_share.data;
-    let mut joint_rand_seed_check = vec![0; Aes128Ctr::STREAM_CIPHER_SEED_SIZE];
+    let mut joint_rand_seed = state.joint_rand_seed;
+    let mut verifier_data = vec![F::zero(); msgs[0].verifier_share.as_slice().len()];
```
tgeoghegan, line 417: meganit: it's a little odd that we use `verifier_share.as_slice()` here, but `verfier_share.data` below, in the iterator. Both achieve the same thing, but I feel like we should either make `Verifier`'s `data` field public, or have a public method allowing access to it, but not both.

src/vdaf.rs
```diff
@@ -476,18 +480,8 @@ mod tests {
...
-                Vec::with_capacity(num_shares - 1);
-            for (j, verifier) in verifiers.iter().enumerate() {
-                if i != j {
-                    verifiers_for_aggregator.push(verifier.clone());
-                }
-            }
-
-            // Run aggregator i.
-            let output_share = verify_finish(state, verifiers_for_aggregator).unwrap();
+        for state in states.into_iter() {
```
tgeoghegan, line 483: nit
```suggestion
        for state in states {
```

tgeoghegan (changes requested):

src/vdaf.rs
```diff
@@ -41,6 +41,7 @@ use aes::Aes128Ctr;
 use getrandom::getrandom;
```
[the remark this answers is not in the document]
cjpatton: Yup, done.

src/vdaf.rs
cjpatton: Good call. Replaced `verifier_share.data` below with `verifier_share.as_slice()` and made `data` private.

src/vdaf.rs
```diff
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: MPL-2.0
 
-//! This module implements core functionality of the PPM protocol for Prio. It consists of the
-//! following functions:
+//! This module constructs  a Verifiable Distributed Aggregation Function (VDAF) from a
+//! [`pcp::Value``]. It consists of the following functions:
```
tgeoghegan, line 4:
```suggestion
//! [`pcp::Value`]. It consists of the following functions:
```

tgeoghegan (approved):
LGTM!

cjpatton:
Squashed.

PR #84, opened 2021-09-22
vdaf: Improve alignment with VDAF spec

cjpatton:
Based on #83 (merge that first).

### DIFF
--- a/src/vdaf.rs
+++ b/src/vdaf.rs
@@ -1,34 +1,7 @@
 // SPDX-License-Identifier: MPL-2.0
 
-//! This module constructs  a Verifiable Distributed Aggregation Function (VDAF) from a
-//! [`pcp::Value`]. It consists of the following functions:
-//!
-//! * `upload` is run by the client to generate the input and proof shares.
-//!
-//! * `verify_start` is run by each aggregator upon receiving its input and proof share. The
-//! output is broadcast to the other aggregators.
-//!
-//! * `verify_finish` is run by each aggregator to decide if the input was valid.
-//!
-//! In the special case of two aggregators (one leader, one helper), these calls map to HTTP
-//! requests as follows (assuming they agree on the query randomness in advance):
-//!
-//! ```none
-//! Client                 Leader                    Helper
-//!   | upload()             |                         |
-//!   |                      |                         |
-//!   |    upload request    |                         |
-//!   |--------------------->| verify_start()          |
-//!   |                      |                         |
-//!   |                      |    aggregate request    |
-//!   |                      |------------------------>| verify_start()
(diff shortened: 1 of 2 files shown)


cjpatton:
Rebased.

src/vdaf.rs
```diff
@@ -272,43 +247,86 @@ pub struct VerifierMessage<F: FieldElement> {
...
-    input_share: V,
-    joint_rand_seed: Key,
+pub enum AggregatorState<V: Value> {
+    /// Aggregator is waiting for its input share.
+    Ready {
+        /// Aggregator ID.
+        aggregator_id: u8,
+
+        /// Input parameter, needed to reconstruct a [`Value`] from a vector of field elements.
+        input_param: V::Param,
```
tgeoghegan, line 265: Does this need to be carried around in this state object? Isn't it always re-derivable from the PPM task definition?

src/vdaf.rs
```diff
@@ -272,43 +247,87 @@ pub struct VerifierMessage<F: FieldElement> {
...
+        AggregatorState::Ready {
+            aggregator_id,
+            input_param,
+            query_rand_seed,
+        } => (aggregator_id, input_param, query_rand_seed),
+        _ => {
+            return Err(VdafError::Uncategorized(
+                "verify_start(): called on incorrect state".to_string(),
+            ));
+        }
```
tgeoghegan, line 326: nit: You could capture the invalid state in the error, if `AggregatorState` implemented `Debug`
```suggestion
        state => {
            return Err(VdafError::Uncategorized(
                format!("verify_start(): called on incorrect state {:?}", state),
            ));
        }
```

tgeoghegan (approved):

src/vdaf.rs
cjpatton: Nice

src/vdaf.rs
cjpatton: Yes,  in theory it's one of the task parameters. From a code standpoint It's required for verify_start(), and I thought it would be cleanest to pass it through the initial state rather than pass it as a function parameter there. Purely a matter of taste.

cjpatton:
Squashed.

cjpatton:
nit: Changed `Waiting` to `Wait`.

src/vdaf.rs
tgeoghegan: It's likely that the opaque `helper_state` blob in a PPM `AggregateResp` will contain a serialized `struct AggregatorState`, so besides code esthetics, including this means bigger messages exchanged between the aggregators and more storage required by the leader. I'm happy to move forward with this and optimize message sizes later, though.

PR #85, opened 2021-09-22
make `vdaf::verify_finish` take a slice

tgeoghegan:
`verify_finish` wants to consume a collection of `VerifierMessage`.
Accomplishing this with `Vec<VerifierMessage>` forces allocation and
copying to the heap. We now instead use const generics and take `M:
TryInto<[VerifierMessage; const N: usize]>`, so that callers that hold
either a `[VerifierMessage]` on the stack or a `Vec<VerifierMessage>` on
the heap, or anything else that can be converted to a slice of
`VerifierMessage`, can use `verify_finish`.


src/vdaf.rs
```diff
@@ -374,10 +374,22 @@ pub fn verify_start<V: Value>(
...
+pub fn verify_finish<M, E, V, const N: usize>(
     state: AggregatorState<V>,
-    msgs: Vec<VerifierMessage<V::Field>>,
-) -> Result<V, VdafError> {
+    msgs: M,
+) -> Result<V, VdafError>
+where
+    V: Value,
+    M: TryInto<[VerifierMessage<V::Field>; N], Error = E>,
+    E: std::fmt::Debug,
```
cjpatton, line 384: It would be nice to reduce the number of type parameters. Is this possible?
```suggestion
    M: TryInto<[VerifierMessage<V::Field>; N], Error: std::fmt::Debug>,
```
It would avoid the additional type parameter `E`. (Also, I think `Debug` is imported by default these days.)

Alternatively, I would be comfortable with a `panic!()` here instead of propagating the error. This is less of an input validation issue than it is a programmer error.

src/vdaf.rs
```diff
@@ -374,10 +374,22 @@ pub fn verify_start<V: Value>(
...
+{
+    let msgs: [VerifierMessage<V::Field>; N] = msgs.try_into().map_err(|e| {
+        VdafError::Uncategorized(format!(
+            "input could not be converted to VerifierMessage slice: {:?}",
```
cjpatton, line 388:
```suggestion
            "verify_finish(): input could not be converted to VerifierMessage slice: {:?}",
```
For consistency

src/vdaf.rs
```diff
@@ -468,7 +480,8 @@ mod tests {
         // Aggregators decide whether the input is valid based on the verifier messages.
         let mut output = vec![Field64::zero(); input.as_slice().len()];
         for state in states {
-            let output_share = verify_finish(state, verifiers.clone()).unwrap();
+            let output_share =
+                verify_finish::<_, _, _, NUM_SHARES>(state, verifiers.clone()).unwrap();
```
cjpatton, line 484: I'm curious if all of the type parameters, including `NUM_SHARES`, could be inferred if `verifiers` had type `[VerifierMessage<Boolean<Field64>>; NUM_SHARES]`.

src/vdaf.rs
```diff
@@ -374,10 +374,22 @@ pub fn verify_start<V: Value>(
 /// [`VerifierMessage`] messages broadcast by all of the aggregators and produces the aggregator's
 /// input share.
 // TODO(cjpatton) Check for ciphersuite mismatch between `state` and `msgs` and among `msgs`.
-pub fn verify_finish<V: Value>(
+pub fn verify_finish<M, E, V, const N: usize>(
```
cjpatton, line 377: How about `SHARES` instead of `N`? This would align nicely with the spec.

cjpatton (changes requested):
Dope! While at it, do you think we should do the same thing for `dist_input()` and `dist_init()`, i.e., make these output a vector of constant size?

src/vdaf.rs
tgeoghegan: I tried doing that but you can't do it yet. There's a [Rust RFC](https://rust-lang.github.io/rfcs/2289-associated-type-bounds.html), [it's implemented](https://github.com/rust-lang/rust/pull/57428), but it's [not yet stabilized](https://github.com/rust-lang/rust/issues/52662).

src/vdaf.rs
tgeoghegan: Yes, you could instead do:
```
let verifiers_slice: [VerifierMessage<Field64>; NUM_SHARES] = verifiers.try_into().unwrap();
for state in states {
    let output_share = verify_finish(state, verifiers.clone()).unwrap();
<...>
```
I wanted to prove that passing in a `Vec<VerifierMessage<F>>` works, though. What would be cool is if Rust had named trait bounds, so that instead of specifying `NUM_SHARES` positionally, I could do `let output_share = verify_finish::<N = NUM_SHARES>(...);`, since the `_, _, _,` is rather ugly.

tgeoghegan:
> Dope! While at it, do you think we should do the same thing for `dist_input()` and `dist_init()`, i.e., make these output a vector of constant size?

Yeah, that is a good idea, if we're prepared to commit to making the number of shares a parameter that's known at compile time. I think that's fine because `num_shares` will be 2 in all realistic deployments of `libprio-rs` for as far out as I can see. I'm going to refactor `dist_input` and `dist_init` as you suggest and update `ppm-prototype` to use the resulting interfaces. It will take me until Friday to update this PR with those changes because I'm flying tomorrow, though.

src/vdaf.rs
cjpatton: I'm not concerened about the ugliness here, I was just curious.

src/vdaf.rs
cjpatton: In that case my preference would be to panic instead of propagating the error.

cjpatton:
> > Dope! While at it, do you think we should do the same thing for `dist_input()` and `dist_init()`, i.e., make these output a vector of constant size?
> 
> Yeah, that is a good idea, if we're prepared to commit to making the number of shares a parameter that's known at compile time. I think that's fine because `num_shares` will be 2 in all realistic deployments of `libprio-rs` for as far out as I can see. I'm going to refactor `dist_input` and `dist_init` as you suggest and update `ppm-prototype` to use the resulting interfaces. It will take me until Friday to update this PR with those changes because I'm flying tomorrow, though.

If we're going to constrain the number of inputs to verify_finish, then it makes sense to constrain the other functions the same way, since they would be effectively constrained at the protocol level anyway.

I think `SHARES > 2` is going to be needed eventually. I think the more important underlying assumption s that a PPM task is always going to have a constant number of aggregators, so hard-coding this at compile time is fine.

src/vdaf.rs
tgeoghegan: Sure, I struggle to imagine this failing for any reason beyond whatever collection `msg` is having the wrong number of elements, so I think we can just `panic!` and get rid of the `Debug` trait bound.

tgeoghegan: cjpatton and I chatted about this and got extremely galaxy brained about it. I'm going to try changing this so that `verify_finish` takes `IntoIterator<Item = VerifierMessage<V::Field>>`, which ✅ allows callers to provide any kind of slice or a vec or whatever and ✅ is infallible so we don't have to deal with any type bounds on errors.

tgeoghegan:
I tried using const generics to make `dist_input` and `dist_init` return constant size vectors, but I found that it ends up being quite awkward. Right now we do something like:

```
let mut out: Vec<Value> = Vec::with_capacity(num_shares);
for foo in foos {
    out.push(Value { ... });
}
out
```

The equivalent construction with an array would be

```
let mut out = [Value; NUM_SHARES]::default();
for (i, foo) in foos.enumerate() {
    out[i] = Value { ... }
}
out
```

The problem is that we need to provide an `std::default::Default` implementation for types like `InputShareMessage` and `VerifierMessage`, and in turn for `Key`, which isn't obvious. Since @cjpatton wants to change the VDAF syntax early next week, I think I'd rather defer that change until other stuff stabilizes some more.

tgeoghegan:
Tagging @BranLwyd for visibility -- don't worry if none of this makes sense, I just want you to be in the loop on libprio-rs changes.

Issue #71, opened 2021-09-09
Set up integration test with `libprio-cc`


tgeoghegan:
Our distinguished colleagues at Google have written a [Prio implementation designed to be compatible with this implementation](https://github.com/google/libprio-cc) (hi @tlepoint!). As we add more features to `libprio-rs` to support PPM, we want to make sure we don't break ENPA as we introduce new encodings or API surfaces. A good way to achieve this would be to set up an integration test where we use `libprio-cc` to construct inputs and `libprio-rs` to aggregate them.

cjpatton:
Good idea. FWIW there are some test vectors here that should ensure compatibility for the current parameters. Don't know what Google's plans are for their library though.

tgeoghegan:
I spent a little time trying to get `libprio-cc` to compile yesterday, and I concluded that it' nots worthwhile to figure out how to build `libprio-cc` in `libprio-rs`, since we'd need a C++ toolchain and also apparently an Android SDK. This would already be a lot of work to set up in GitHub actions, and a huge pain in the butt for developers to run locally.

Instead, I want to test backward compatibility with ENPA (which is what testing against `libprio-cc` was really meant to achieve) by testing against some test vectors, recorded with a version of `libprio-rs` known empirically to work in `prio-server` and hence ENPA. Pull request #88 achieves this with a `FieldPriov2` based test vector.

Once we take that change, I think we should release `prio` 0.5.0 to crates.io, and then empirically verify that a `prio-server` built against `prio` 0.5.0 and using `FieldPriov2` can still perform aggregations with a `prio-server` using `prio` 0.4.0. If that works, then we will know that 0.5.0 is compatible with ENPA, and thus that test vectors generated from 0.5.0 can be used to maintain backward compatibility going forward.

The other "obvious" thing to do would be to generate test vectors from the already shipped `prio` 0.4.0 which we know to work just fine in `prio-server`. However that version depends on some crypto crates that have been yanked from crates.io, and so building it is surprisingly difficult!

cjpatton:
SGTM. Once we've generated the test vectors we should send a PR to `libprio-cc` that runs them against their server.

PR #88, opened 2021-09-29
add backward compatibility test

tgeoghegan:
We extend `tests/accumulating.rs` so that generated inputs and the
expected aggregation output are recorded and output to stdout, base64
encoded. We then add a new test, `tests/backward_compatibility.rs` which
uses such a generated test vector for `FieldPriov2` to ensure that we
remain compatible with ENPA/Prio v2 regardless of whatever future
changes we make to support PPM or VDAF.


tgeoghegan:
See #71 for discussion of the plan this change fits into.

tests/README.md
```diff
@@ -0,0 +1,12 @@
+# Backward compatibility and test vector generation
+
+Crate `prio` supports Prio v2, used in [Exposure Notification Private Analytics](enpa-whitepaper), as implemented in [`prio-server`](https://github.com/abetterinternet/prio-server), as well as the Prio v3 Verifiable Distributed Aggregation Function used in [Privacy Preserving Measurements](ppm-spec).
```
cjpatton, line 3: nit: I prefer breaking lines at 80 characters in markdown files. Feel free to override my preference, of course :)

tests/README.md
```diff
@@ -0,0 +1,12 @@
...
+
+To ensure backward compatibility with older versions of crate `prio` and [other implementations of Prio v2](https://github.com/google/libprio-cc), we continuously test `prio`'s Prio v2 implementation against a test vector recorded with `prio` 0.4.0. That test is implemented in `tests/backward_compatibility.rs`. The test vector was generated using `tests/accumulating.rs`. To generate a new set of test vectors, run:
+
+    cargo test --test accumulating -- --nocapture
+
+...and the test will print base64 inputs and an expected sum to stdout, which may then be plugged into a backward compatibility test as illustrated in `tests/backward_compatibility.rs`. Adjust `number_of_clients` to generate longer or shorter test vectors.
```
cjpatton, line 9: Best practice would be to dump a JSON blob, which can easily be consumed by multiple test vector runners. In Go I would put the JSON blob in a directory called `testdata/` located in the module with the test vector runners. I'm not sure what the right thing would be to do for Rust, but I think it makes perfect sense to just stick it in `tests/testdata`.

tests/README.md
```diff
@@ -0,0 +1,12 @@
+# Backward compatibility and test vector generation
+
+Crate `prio` supports Prio v2, used in [Exposure Notification Private Analytics](enpa-whitepaper), as implemented in [`prio-server`](https://github.com/abetterinternet/prio-server), as well as the Prio v3 Verifiable Distributed Aggregation Function used in [Privacy Preserving Measurements](ppm-spec).
+
+To ensure backward compatibility with older versions of crate `prio` and [other implementations of Prio v2](https://github.com/google/libprio-cc), we continuously test `prio`'s Prio v2 implementation against a test vector recorded with `prio` 0.4.0. That test is implemented in `tests/backward_compatibility.rs`. The test vector was generated using `tests/accumulating.rs`. To generate a new set of test vectors, run:
+
+    cargo test --test accumulating -- --nocapture
```
cjpatton, line 7: What about creating a binary for generating test vectors? Maybe call it `prio_v2_test_vector_gen`?

tests/backward_compatibility.rs
```diff
@@ -0,0 +1,70 @@
+pub mod test_common;
```
cjpatton, line 1: Don't forget to add copyright.

tests/test_common/mod.rs
```diff
@@ -0,0 +1,25 @@
+use prio::encrypt::{PrivateKey, PublicKey};
```
cjpatton, line 1: Copyright

cjpatton (commented):
Left a few comments, none of which should be considered blocking. They're mostly suggestions I have about best practices for producing/consuming test vectors I've observed from other crypto libraries.

tests/README.md
tgeoghegan: You caught me, I was getting lazy here. I'll add a binary target to the crate and do this properly. I bet we'll want to generate lots of more test vectors in the future so having a more robust tool to build on for this will probably pay off. Plus, we'll end up with a `Cargo.lock` in `libprio-rs` so that we can go back and rebuild old versions of the crate even if they use yanked dependencies.

tests/backward_compatibility.rs
tgeoghegan: We put a default copyright assignation at the top of `LICENSE`, but I'll add the SPDX-License-Identifier line for consistency.

tgeoghegan:
Attempt #2 here. This is now based on #90, so please review that first.

tgeoghegan:
- [ ] make sure that `generate-test-vector` gets built by CI

src/bin/generate_test_vector.rs
```diff
@@ -0,0 +1,75 @@
...
+    dimension: usize,
+    /// Number of inputs to generate
+    #[structopt(short, long, required = true)]
+    number_of_inputs: usize,
+    /// Field to use to encode inputs
+    #[structopt(short, long, required = true)]
+    field: FieldArgument,
+}
+
+fn generate_and_print_vector<F: FieldElement + Serialize>(options: Options) -> Result<()> {
```
tgeoghegan, line 53:
```suggestion
fn generate_and_print_vector<F: FieldElement>(options: Options) -> Result<()> {
```

- [ ] delete redundant trait bound

src/test_vector.rs
```diff
@@ -0,0 +1,229 @@
...
+        let mut server_2_shares = Vec::with_capacity(number_of_clients);
+
+        let mut rng = rand::thread_rng();
+
+        for _ in 0..number_of_clients {
+            // Generate a random vector of booleans
+            let data: Vec<F> = (0..dimension)
+                .map(|_| Ok(F::from(F::Integer::try_from(rng.gen_range(0..2))?)))
+                .collect::<Result<_, F::IntegerTryFromError>>()
+                .map_err(|e| TestVectorError::IntegerConversion(format!("{:?}", e)))?;
```
tgeoghegan, line 102: - [ ] figure out if we can put a tighter bound on `FieldElement::IntegerTryFromError` to make this error conversion less ugly... or just do `F::Integer::try_from(...).unwrap()` since it's never really going to fail.

src/test_vector.rs
```diff
@@ -0,0 +1,229 @@
...
+    /// may be used to decrypt `server_1_shares`.
+    pub server_1_private_key: String,
+    /// Base64 encoded private key for the non-"first" a.k.a. "facilitator"
+    /// server, which may be used to decrypt `server_2_shares`.
+    pub server_2_private_key: String,
+    /// Dimension (number of buckets) of the inputs
+    pub dimension: usize,
+    /// Encrypted shares of ENPA format inputs for the "first" a.k.a. "PHA"
+    /// server. The inner `Vec`s are encrypted bytes.
+    pub server_1_shares: Vec<Vec<u8>>,
```
tgeoghegan, line 74: - [ ] `serde_json` serialization of a `Vec[u8]` is terrible; see if we can get it to serialize to a base64 string instead

src/bin/generate_test_vector.rs
```diff
@@ -0,0 +1,75 @@
...
+impl FromStr for FieldArgument {
+    type Err = &'static str;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s.to_lowercase().as_str() {
+            "fieldpriov2" => Ok(Self::FieldPriov2),
+            "field32" => Ok(Self::Field32),
+            "field64" => Ok(Self::Field64),
+            "field80" => Ok(Self::Field80),
+            "field126" => Ok(Self::Field126),
```
cjpatton, line 28: Let's remove these until the parameters stabilize.

src/bin/generate_test_vector.rs
```diff
@@ -0,0 +1,75 @@
...
+    #[structopt(short, long, required = true)]
+    number_of_inputs: usize,
+    /// Field to use to encode inputs
+    #[structopt(short, long, required = true)]
+    field: FieldArgument,
+}
+
+fn generate_and_print_vector<F: FieldElement + Serialize>(options: Options) -> Result<()> {
+    let test_vector: EnpaTestVector<F> =
+        EnpaTestVector::new(options.dimension, options.number_of_inputs)
```
cjpatton, line 55: If we're adding the noun "ENPA" to this library, maybe change "FieldPriov2" to "FieldEnpa"?

src/bin/generate_test_vector.rs
```diff
@@ -0,0 +1,75 @@
+use color_eyre::eyre::{Result, WrapErr};
```
cjpatton, line 1: Missing copyringht

src/bin/generate_test_vector.rs
```diff
@@ -0,0 +1,75 @@
...
+fn main() -> Result<()> {
+    color_eyre::install()?;
+    let options = Options::from_args();
+
+    match options.field {
+        FieldArgument::FieldPriov2 => generate_and_print_vector::<FieldPriov2>(options),
+        FieldArgument::Field32 => generate_and_print_vector::<Field32>(options),
+        FieldArgument::Field64 => generate_and_print_vector::<Field64>(options),
+        FieldArgument::Field80 => generate_and_print_vector::<Field80>(options),
+        FieldArgument::Field126 => generate_and_print_vector::<Field126>(options),
```
cjpatton, line 73: Test vectors are mighty big. Is there a need to generate them for each field?

tgeoghegan: I only generated a `FieldPriov2` vector, but I wanted to support more types. However, this tool only supports emitting test vectors in the Priov2 specific encoding and encryption scheme, which will only ever use `FieldPriov2`, so I can simplify all this code by removing the generic `FieldElement` parameter.

src/bin/generate_test_vector.rs
tgeoghegan: Good point. I'd prefer to avoid introducing the notion of ENPA here, so I will align all the names around `Priov2`.

src/bin/generate_test_vector.rs
cjpatton: I see, in theory would could use other fields for ENPA. Do you know if there are plans to do so? If this isn't planned, then I would hard-code FieldPriov2.

src/bin/generate_test_vector.rs
```diff
@@ -0,0 +1,52 @@
+// SPDX-License-Identifier: MPL-2.0
+
+use color_eyre::eyre::{Result, WrapErr};
+use prio::test_vector::Priov2TestVector;
+use structopt::StructOpt;
+
+#[derive(Debug, StructOpt)]
+#[structopt(about = "Generate Priov2 test vector", rename_all = "kebab-case")]
+enum Subcommand {
+    FieldPriov2 {
```
cjpatton, line 10: `Priov2` might be a better name.

src/server.rs
```diff
@@ -64,7 +64,8 @@ pub struct Server<F> {
     is_first_server: bool,
     accumulator: Vec<F>,
     validation_mem: ValidationMemory<F>,
-    private_key: PrivateKey,
+    /// The private key used by the server to decrypt inputs
```
cjpatton, line 67: Can the code be written without making this public? As a general rule it's a good idea to keep secret values private in order to avoid API misuse.

.gitignore
```diff
@@ -2,9 +2,5 @@
 # will have compiled files and executables
 /target/
 
-# Remove Cargo.lock from gitignore if creating an executable, leave it for libraries
-# More information here https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html
-Cargo.lock
```
cjpatton, line 7: I'm a bit confused about the architecture of the repository. (1) Does it implement two crates, one a library and the other a binary for generating test vectors? (2) Or does it build one crate, either the library or binary, depending on the build flag? Which approach is "better"?

I ask because the linked documentation does not say what to do for (2).

tgeoghegan: Cargo allows for crates that have both library and binary targets. However there is a more formally supported way to do this which is a [workspace](https://doc.rust-lang.org/cargo/reference/workspaces.html). I think the idea is that we would have two crates in the workspace: one that builds the library and another that builds binaries that depend on the library crate.

I was trying to avoid doing that to keep this change simple, but I agree that `Cargo.lock` becomes ambiguous (since the recommendation is to avoid checking a library's `Cargo.lock`) and also a workspace might be more clear and better supported than the optional features I rigged up to separate binary dependencies from library dependencies. I think I'll try reorganizing this into a workspace.

cjpatton: Up to you, you're much more in tune with best practice here. Just wanted to point out a potential weirdness that we may end up needing to fix one way or another.

src/server.rs
tgeoghegan: Good call. This is needed so that `tests/tweaks.rs` can use it, and integration tests have to use public API. But clearly `tweaks.rs` wants to poke at library internals, so I think it'd be appropriate to make it a unit test and not expose this externally.

.gitignore
tgeoghegan: I'm also figuring out the best practices as I go here, so no worries. This has been a good opportunity to read up on Rust packaging. It looks like even with a workspace, you still have one big `Cargo.lock` for all the crates, but it did make dependency management cleaner, so I think the workspace is a good move.

.github/workflows/ci-build.yml
```diff
@@ -18,10 +18,10 @@ jobs:
     - name: Lint
       run: cargo fmt --message-format human -- --check
     - name: Build
-      run: cargo build --verbose
+      run: cargo build --verbose --workspace
     - name: Run tests
       run: cargo test --verbose
```
cjpatton, line 23: `cargo test` no longer runs tests. It appears you also need to pass the `--workspace` flag here.

cjpatton (changes requested):

.github/workflows/ci-build.yml
tgeoghegan: Good catch! I added `"."` to `workspace.default-members` in `Cargo.toml`, and now `cargo test` will run the tests in the root crate (`prio`) without requiring `--all` or `--workspace`.

cjpatton (approved):

tgeoghegan:
Checks aren't running on this PR, but there appears to be an [ongoing incident with GH Actions](https://www.githubstatus.com/incidents/bdbzpz7qxmbx). I'll rebase and merge once the tests pass.

tgeoghegan:
I ran `ci-build` manually against the latest commit in this PR and it passed: https://github.com/abetterinternet/libprio-rs/actions/runs/1308793171

PR #92, opened 2021-10-11
add test vector generator and backward compat test

tgeoghegan:
Adds a module for generating and working with test vectors, a utility
for generating test vectors and printing their JSON representation and a
test that ensures backward compatibility with the ENPA system (i.e.,
`FieldPriov2` and the ENPA encryption scheme). Along the way, we
introduce a Cargo workspace and a new package from which we will build
binaries like `generate-test-vector`.

Related to #71


tgeoghegan:
This is exactly the same change as was reviewed in #88, but that pull request accidentally targeted the wrong branch, and this one targets `main`.

PR #98, opened 2021-10-16
Implement the VDAF for heavy hitters

cjpatton:
Based on #97 (merge that first).

Adds a new module, vdaf::hits, with a VDAF for heavy hitters. The
function is compiled from an IDPF with the API specified by the Idpf
trait. A toy implementation of an IDPF is provided for testing purposes.


src/vdaf/hits.rs
```diff
@@ -0,0 +1,741 @@
...
+    //
+    // NOTE(cjpatton) The IDPF of [BBCG+21] might use different fields for different levels of the
+    // prefix tree.
+    type Field: FieldElement;
+
+    /// Generate a new sequence of IDPF shares for `input`. Parameter `output` is an iterator that
+    /// is invoked to get the output value for each successive level of the prefix tree.
+    fn gen<O: IntoIterator<Item = Self::Field>>(
+        input: &IdpfInput,
+        output: O,
```
tgeoghegan, line 104: It's a bit confusing having an input parameter named `output` and then some other value returned from the function. I need to read the HH paper to better understand how all this fits together, but at least a doccomment explaining what value is returned from this function would help.

src/vdaf/hits.rs
```diff
@@ -0,0 +1,741 @@
...
+    fn eval(&self, prefix: &IdpfInput) -> Result<[Self::Field; OUT_LEN], VdafError>;
+}
+
+/// A "toy" IPDF used for demonstration purposes. The space consumed by each share is `O(2^n)`,
+/// where `n` is the length of the input. The size of each share is restricted to 1MB, so this IDPF
+/// is only suitable for very short inputs.
+//
+// NOTE(cjpatton) It would be straight-forward to generalize this construction to any `KEY_LEN` and
+// `OUT_LEN`.
+pub struct ToyIdpf<F: FieldElement> {
```
tgeoghegan, line 117: [Put the trait bounds on the impl, not the struct](https://rust-lang.github.io/api-guidelines/future-proofing.html#data-structures-do-not-duplicate-derived-trait-bounds-c-struct-bounds)
```suggestion
pub struct ToyIdpf<F> {
```

src/vdaf/hits.rs
```diff
@@ -0,0 +1,741 @@
...
+    len: usize,
+}
+
+impl<F: FieldElement> Idpf<2, 1> for ToyIdpf<F> {
+    type Field = F;
+
+    fn gen<M: IntoIterator<Item = Self::Field>>(
+        input: &IdpfInput,
+        output: M,
+    ) -> Result<[Self; 2], VdafError> {
```
tgeoghegan, line 128: Do you have to duplicate the literal `2` here or would something like `Self::KEY_LEN` work?

src/vdaf/hits.rs
```diff
@@ -0,0 +1,741 @@
...
+            )));
+        }
+
+        Ok([self.data[prefix.index()]])
+    }
+}
+
+/// An input share for the heavy hitters VDAF.
+pub struct InputShareMessage<I: Idpf<2, 1>> {
+    /// IDPF share of inpuot
```
tgeoghegan, line 176:
```suggestion
    /// IDPF share of input
```

src/vdaf/hits.rs
```diff
@@ -0,0 +1,741 @@
...
+    let mut helper_sketch_start_prng: Prng<I::Field> =
+        Prng::from_key_stream(KeyStream::from_key(&helper_sketch_start_seed));
+    let mut helper_sketch_next_prng: Prng<I::Field> =
+        Prng::from_key_stream(KeyStream::from_key(&helper_sketch_next_seed));
+    let mut leader_sketch_next: Vec<I::Field> = Vec::with_capacity(2 * input.len);
+    for k in auth_rand.iter() {
+        // [BBCG+21, Appendix C.4]
+        //
+        // $(a, b, c)$
+        let a = leader_sketch_start_prng.next().unwrap() + helper_sketch_start_prng.next().unwrap();
```
tgeoghegan, line 214: This is a lot of `unwrap()`. Since we are always dealing with a concrete instance of `Prng` rather than generically using some `Iterator`, maybe we should put a method on `Prng` that infallibly returns a value.

src/vdaf/hits.rs
```diff
@@ -0,0 +1,741 @@
...
+                "prefix length ({}) exceeds input length ({})",
+                prefix.len, self.len
+            )));
+        }
+
+        Ok([self.data[prefix.index()]])
+    }
+}
+
+/// An input share for the heavy hitters VDAF.
```
tgeoghegan, line 174:
```suggestion
/// An input share for the heavy hitters VDAF.
#[derive(Clone, Debug, Serialize, Deserialize)]
```

src/vdaf/hits.rs
```diff
@@ -0,0 +1,741 @@
...
+            sketch_start_seed: helper_sketch_start_seed,
+            sketch_next: Share::Helper {
+                seed: helper_sketch_next_seed,
+                length: 2 * input.len,
+            },
+        },
+    ])
+}
+
+/// The verification parameter used by the aggregators to evaluate the VDAF on a distributed input.
```
tgeoghegan, line 249:
```suggestion
/// The verification parameter used by the aggregators to evaluate the VDAF on a distributed input.
#[derive(Clone, Debug, Serialize, Deserialize)]
```

src/vdaf/hits.rs
```diff
@@ -0,0 +1,741 @@
...
+/// The VDAF aggregation parameter, a sequence of equal-length candidate prefixes.
+#[derive(Debug, PartialEq)]
+pub struct AggregationParam<'a> {
+    prefixes: Vec<IdpfInput<'a>>,
+    len: usize,
+}
+
+impl<'a> AggregationParam<'a> {
+    /// Constructs an aggregation parameter from a sequence of prefixes. An error is returned if
+    /// the prefixes aren't equal length, a prefix appears twice in the set of prefixesx, or the
```
tgeoghegan, line 278:
```suggestion
    /// the prefixes aren't equal length, a prefix appears twice in the set of prefixes, or the
```

src/vdaf/hits.rs
```diff
@@ -0,0 +1,741 @@
...
+                if let Some(l) = len {
+                    if prefix.len != l {
+                        return Err(VdafError::Uncategorized(
+                            "prefixes must all have the same length".to_string(),
+                        ));
+                    }
+                } else {
+                    len = Some(prefix.len);
+                }
+                unique_prefixes.insert(prefix.clone());
```
tgeoghegan, line 304: nit: does `unique_prefixes.insert(&prefix);` not work? You might need `P: IntoIterator(<Item = &IdpfInput<'a>>`, which may be a good idea anyway to avoid consuming `prefixes`.

src/vdaf/hits.rs
```diff
@@ -0,0 +1,741 @@
...
+    verify_param: &VerifyParam,
+    agg_param: &AggregationParam,
+    nonce: &[u8],
+    input_share: &InputShareMessage<I>,
+) -> Result<(VerifyState<I::Field>, VerifyMessage<I::Field>), VdafError> {
+    let i = agg_param.len;
+
+    // Derive the verification randomness.
+    let mut deriver = KeyDeriver::from_key(&verify_param.verify_rand_init);
+    deriver.update(nonce);
```
tgeoghegan, line 368: No aggregator ID, as in the prio3?

src/vdaf/hits.rs
```diff
@@ -0,0 +1,741 @@
...
+        .reduce(|z_left, z_right| {
+            let z_left = z_left?;
+            let z_right = z_right?;
+            Ok([
+                z_left[0] + z_right[0],
+                z_left[1] + z_right[1],
+                z_left[2] + z_right[2],
+            ])
+        })
+        .unwrap()?;
```
tgeoghegan, line 449: nit: maybe something like
```suggestion
        .unwrap_or_else(Err(VdafError::Uncategorized("empty verify message iterator".to_string()))?;
```
IIUC the `unwrap()` handles the case where the iterator passed to `.reduce()` is empty, and `reduce()` will return `None`.

src/vdaf/hits.rs
tgeoghegan: No, I think it's reasonable to expect readers or editors of this code to have read the paper and be familiar with IDPFs.

src/vdaf/hits.rs
tgeoghegan: Let's hold off -- at some point, I will try to integrate this VDAF into ppm-prototype, and will probably have to make a few changes to make it all fit together. Tacking on these traits can happen then.

src/vdaf/hits.rs
```diff
@@ -0,0 +1,749 @@
...
+    /// caller.
+    fn prefix(&self, len: usize) -> Self {
+        Self {
+            data: self.data,
+            len,
+        }
+    }
+}
+
+impl<'a> PartialEq<IdpfInput<'a>> for IdpfInput<'a> {
```
tgeoghegan, line 67: We provide a `PartialEq` that only considers the first `self.len` bytes of `self.data`. We should provide an implementation of `Hash` that works the same way, or data structures will behave surprisingly!

tgeoghegan: On second thought, the derived `Hash` impl is sufficient. Given `IdpfInput::new(b"hello", 40)` and `IdpfInput::new(b"hello", 39)`, the hash will include the entire string "hello" in both cases, but the different `len` values will cause the resulting hash to differ, satisfying the requirements of the relationship between `Hash` and `Eq`.

src/vdaf/hits.rs
tgeoghegan: I'll do this in a PR after this one to avoid touching loads more files in this change.

PR #107, opened 2021-10-28
Define traits for VDAFs and implement hits

cjpatton:
Adds traits to the `vdaf` module that collectively define the functionalities required to execute a VDAF. To demonstrate how the API is implemented, this PR also adds an implementation of hits.

The API is based on an idea from @bifurcation (https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=bbf89d42971b3c068d4b13acbb56b3b1) The main thing that has changed is that I wanted to find a way to use the type system to express that, at the end of the Prepare process, the Aggregators should return output shares compatible with the VDAF. @bifurcation's sketch doesn't quite accomplish this.

Something I'm not totally satisfied with here is the use of the verb "Preapre" to describe the process of mapping input shares and an aggregation parameter to output shares. ("Prepare" as in "to prepare a sequence of input shares corresponding to a single measurement for aggregation".) @bifurcation and I settled on this verb for [the first VDAF draft](https://datatracker.ietf.org/doc/html/draft-patton-cfrg-vdaf-00), but we wondered if we could do better. I'm hoping that looking at a concrete API will inspire folks to think of a better term. (@chris-wood, your feedback would be helpful here as well!)

First, some context. We consider the aggregation function to be the composition of two functions: the "outer function" `f`, and the "inner function" `g`. The inner function `g` is used to map a measurement and an aggregation parameter to an *(inner) output*, and the outer function `f` is used to map a set of inner outputs to the *aggregate result* That is, we decompose the aggregation function `F(agg_param, measurement_1, ..., measurement_N)` into `f(g(agg_param, measurement_1), ..., g(agg_param, measurement_N))`. 

Next, it is important to make a distinction between the inner function `g` and the the *distributed computation of `g`*. The function `g` takes a measurement and aggregation parameter and returns an (inner) output. On the other hand, the distributed computation of `g` starts with input shares and an aggregation parameter and ends with each Aggregator having an output share (i.e., a share of an (inner) output). We make a similar distinction for the outer function `f`. 

Overall, there are two processes going on in the distributed computation of `F`:
1. The distributed evaluation of the inner function `g`, i.e., interactive process of mapping an aggregation parameter and set of input shares to the corresponding set of (verified) output shares.
2. The distributed evaluation of the outer function `f`. Here, each Aggregator aggregates its output shares locally, then they combine their aggregate shares to get the final aggregate result.

Alternatives to "Prepare" that we considered:

* "Evaluate" --- as in to distributively evaluate the inner function `g`. The problem here is that "Evaluate" is ambiguous, since it could refer to either the inner function or the outer function.

* "Verify" --- as in to verify the input shares and return output shares. This is confusing because, in general, it's not possible to split up the process into "input verification" and "output computation". In general it is the output shares that need to be verified. (Hits is an example of a protocol for which this is true. On the other hand, in Prio it is the inputs that are validated, but in this case, a valid input implies a valid output.)

* "Recover" --- as in to recover (verified) output shares from input shares and an aggregation parameter. The problem here is that the word "recover" make it sound as if something's gone wrong.

### DIFF
--- a/src/field.rs
+++ b/src/field.rs
@@ -17,7 +17,7 @@ use serde::{
 };
 use std::{
     cmp::min,
-    convert::TryFrom,
+    convert::{TryFrom, TryInto},
     fmt::{self, Debug, Display, Formatter},
     hash::{Hash, Hasher},
     io::{Cursor, Read},
@@ -83,8 +83,11 @@ pub trait FieldElement:
     /// Size in bytes of the encoding of a value.
     const ENCODED_SIZE: usize;
 
-    /// The error returned if converting `usize` to an `Int` fails.
-    type IntegerTryFromError: Debug;
+    /// The error returned if converting `usize` to an `Integer` fails.
+    type IntegerTryFromError: std::error::Error;
+
+    /// The error returend if converting an `Integer` to a `u64` fails.
+    type TryIntoU64Error: std::error::Error;
 
     /// The integer representation of the field element.
     type Integer: Copy
@@ -94,7 +97,9 @@ pub trait FieldElement:
         + Div<Output = <Self as FieldElement>::Integer>
         + Shr<Output = <Self as FieldElement>::Integer>
         + Sub<Output = <Self as FieldElement>::Integer>
-        + TryFrom<usize, Error = Self::IntegerTryFromError>;
+        + From<Self>
+        + TryFrom<usize, Error = Self::IntegerTryFromError>
+        + TryInto<u64, Error = Self::TryIntoU64Error>;
 
     /// Modular exponentation, i.e., `self^exp (mod p)`.
     fn pow(&self, exp: Self::Integer) -> Self;
@@ -472,6 +477,7 @@ macro_rules! make_field {
             const ENCODED_SIZE: usize = $encoding_size;
             type Integer = $int;
             type IntegerTryFromError = <Self::Integer as TryFrom<usize>>::Error;
+            type TryIntoU64Error = <Self::Integer as TryInto<u64>>::Error;
 
             fn pow(&self, exp: Self::Integer) -> Self {
                 Self($fp.pow(self.0, u128::try_from(exp).unwrap()))
--- a/src/vdaf.rs
+++ b/src/vdaf.rs
@@ -1,9 +1,226 @@
 // SPDX-License-Identifier: MPL-2.0
 
-//! **(NOTE: This module is experimental. Applications should not use it yet.)** This module
-//! implements the prio3 Verifiable Distributed Aggregation Function specified in
-//! [[VDAF](https://cjpatton.github.io/vdaf/draft-patton-cfrg-vdaf.html)]. It is constructed from a
-//! [`crate::pcp::Value`].
+//! **(NOTE: This module is experimental. Applications should not use it yet.)** This module defines an
+//! API for Verifiable Distributed Aggregation Functions (VDAFs) and implements two constructions
+//! described in [[VDAF]].
+//!
+//! [BBCG+19]: https://ia.cr/2019/188
+//! [BBCG+21]: https://ia.cr/2021/017
+//! [VDAF]: https://datatracker.ietf.org/doc/draft-patton-cfrg-vdaf/
 
+use std::convert::TryFrom;
+
+use crate::field::FieldElement;
+use crate::pcp::types::TypeError;
+use crate::pcp::PcpError;
+use crate::prng::{Prng, PrngError};
+use crate::vdaf::suite::{Key, KeyStream, SuiteError};
+use serde::{Deserialize, Serialize};
+
+/// Errors emitted by this module.
+#[derive(Debug, thiserror::Error)]
+pub enum VdafError {
+    /// An error occurred.
+    #[error("vdaf error: {0}")]
+    Uncategorized(String),
+
+    /// PCP error.
+    #[error("pcp error: {0}")]
+    Pcp(#[from] PcpError),
+
+    /// Type error.
+    #[error("type error: {0}")]
+    Type(#[from] TypeError),
+
+    /// PRNG error.
+    #[error("prng error: {0}")]
+    Prng(#[from] PrngError),
+
+    /// Suite error.
+    #[error("suite error: {0}")]
+    Suite(#[from] SuiteError),
+}
+
+/// An additive share of a vector of field elements.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub enum Share<F> {
+    /// An uncompressed share, typically sent to the leader.
+    Leader(Vec<F>),
+
+    /// A compressed share, typically sent to the helper.
+    Helper {
+        /// The seed for the pseudorandom generator.
+        seed: Key,
+        /// The length of the uncompressed share.
+        //
+        // TODO(cjpatton) Avoid encoding the length of the uncompressed share on the wire. The VDAF
+        // should always provide a way to compute this.
+        length: usize,
+    },
+}
+
+impl<F: FieldElement> TryFrom<Share<F>> for Vec<F> {
+    type Error = VdafError;
+
+    fn try_from(share: Share<F>) -> Result<Self, VdafError> {
+        match share {
+            Share::Leader(data) => Ok(data),
+            Share::Helper { seed, length } => {
+                let prng: Prng<F> = Prng::from_key_stream(KeyStream::from_key(&seed));
+                Ok(prng.take(length).collect())
+            }
+        }
+    }
+}
+
+/// The base trait for VDAF schemes. This trait is inherited by traits [`Client`], [`Aggregator`],
+/// and [`Collector`], which define the roles of the various parties involved in the execution of
+/// the VDAF.
+pub trait Vdaf {
+    /// The type of Client measurement to be aggregated.
+    type Measurement;
+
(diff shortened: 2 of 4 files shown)


src/vdaf/hits.rs
```diff
@@ -0,0 +1,897 @@
...
+//
+// TODO Make this an associated type of `Ipdf`.
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+pub struct IdpfInput {
+    index: usize,
+    level: usize,
+}
+
+impl IdpfInput {
+    /// Constructs an IDPF input using the first `len` bits of `data`.
```
tgeoghegan, line 44:
```suggestion
    /// Constructs an IDPF input using the first `level` bits of `data`.
```

src/vdaf/hits.rs
```diff
@@ -0,0 +1,897 @@
...
+    fn data_index(&self) -> usize {
+        self.index | (1 << self.level)
+    }
+}
+
+impl PartialOrd for IdpfInput {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
```
tgeoghegan, line 86: nit: Since we provide `Ord`, can we just `#[derive(PartialOrd)]` on `IdpfInput`?

src/vdaf/hits.rs
```diff
@@ -0,0 +1,897 @@
...
+        }
+    }
+}
+
+/// An Incremental Distributed Point Function (IDPF), as defined by [[BBCG+21]].
+///
+/// [BBCG+21]: https://eprint.iacr.org/2021/017
+//
+// NOTE(cjpatton) The real IDPF API probably needs to be stateful.
+pub trait Idpf<const KEY_LEN: usize, const OUT_LEN: usize>: Sized {
```
tgeoghegan, line 102: Will we ever have more than one implementation of trait `Idpf`? IIUC `ToyIdpf` will eventually be replaced by a more efficient and scalable implementation, so could we just have `pub struct Idpf<const KEY_LEN: usize, const OUT_LEN: usize, F>`?

src/vdaf/hits.rs
```diff
@@ -0,0 +1,897 @@
...
+
+/// A "toy" IPDF used for demonstration purposes. The space consumed by each share is `O(2^n)`,
+/// where `n` is the length of the input. The size of each share is restricted to 1MB, so this IDPF
+/// is only suitable for very short inputs.
+//
+// NOTE(cjpatton) It would be straight-forward to generalize this construction to any `KEY_LEN` and
+// `OUT_LEN`.
+pub struct ToyIdpf<F> {
+    data0: Vec<F>,
+    data1: Vec<F>,
```
tgeoghegan, line 129: Can we have more descriptive names?

src/vdaf/hits.rs
```diff
@@ -0,0 +1,897 @@
...
+                        "unexpected message count: got {}; want 2",
+                        inputs.len()
+                    )));
+                }
+
+                // Compute the polynomial evaluation.
+                let mut y = F::zero();
+                for y_share in inputs.iter() {
+                    if y_share.len() != 1 {
+                        // XXX Tim- How do I print the state here?
```
tgeoghegan, line 554: Do you mean `self.state`? Is it not in scope here?

tgeoghegan (changes requested):

src/vdaf.rs
```diff
@@ -1,9 +1,229 @@
...
+
+    /// Generates the long-lived parameters used by the Clients and Aggregators.
+    fn setup(&self) -> Result<(Self::PublicParam, Vec<Self::VerifyParam>), VdafError>;
+
+    /// The number of rounds of communication among the Aggregators that are required before they
+    /// have recovered output shares.
+    fn rounds(&self) -> usize;
+
+    /// The number of input shares generated by a Client.
+    fn shares(&self) -> usize;
```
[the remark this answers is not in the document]
cjpatton: Changed `shares` to `num_aggregators`. While at it, I also changed `rounds` to `num_rounds` for consistency.

src/vdaf.rs
```diff
@@ -1,9 +1,229 @@
...
```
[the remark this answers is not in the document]
cjpatton: For `prio3`, the number of aggregators is configurable, i.e., we'll have a constructor that specifies, among other things, the number of aggregators.

src/vdaf.rs
```diff
@@ -1,9 +1,229 @@
...
+}
+
+/// The Client's role in the execution of a VDAF.
+pub trait Client: Vdaf {
```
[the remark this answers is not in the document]
cjpatton: Well, they'd have to be defined by the config object, which essentially reduces to Option (1). @bifurcation I'm gonna keep this as-is unless you have a strong objection. I still think the current PR is slightly tidier than Option (1).

src/vdaf.rs
```diff
@@ -1,9 +1,229 @@
...
+    Finish(O),
+
+    /// Fail and return an error.
+    Fail(VdafError),
+}
+
+/// An aggregate share that can merged with aggregate shares of the same type.
+pub trait Aggregatable {
+    /// Update an aggregate share by merging it with another (`agg_share`).
+    fn merge(&mut self, agg_share: &Self) -> Result<(), VdafError>;
```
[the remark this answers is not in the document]
cjpatton: I'd be cool with this. Just a small nit: I'd move the `lhs` into `merge` so that you can consume it and avoid a copy.

src/vdaf/hits.rs
cjpatton: Nice catch.

src/vdaf/hits.rs
cjpatton: `ToyIdpf` SHOULD be replaced eventually, but I think we want to keep things generic here. There's likely room for improvement over the IDPF from the paper, so we should keep the door open for "dropping in" future schemes.

src/vdaf/hits.rs
cjpatton: This an IDPF with an output size of two. Perhaps `data_left` and data_right`?

src/vdaf/hits.rs
cjpatton: What I'm asking is how to print the variant of `self.state` (an `enum`) in the error message. You once pointed out how to do this, but I forgot :/

src/field.rs
```diff
@@ -94,7 +97,9 @@ pub trait FieldElement:
         + Div<Output = <Self as FieldElement>::Integer>
         + Shr<Output = <Self as FieldElement>::Integer>
         + Sub<Output = <Self as FieldElement>::Integer>
-        + TryFrom<usize, Error = Self::IntegerTryFromError>;
+        + From<Self>
```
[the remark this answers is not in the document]
tgeoghegan: Oh, I misread the trait bound. This is fine then.

src/vdaf/hits.rs
tgeoghegan: I think you could just derive a `Debug` implementation on `SketchState` (none of its variants have fields so the implementation will just write the variant name) and then use `{:?}` in a format string.

src/vdaf/hits.rs
tgeoghegan: That's about the same. Let's keep this for now, and if I can think of a better name I'll change it in another PR.

src/vdaf/hits.rs
cjpatton: Oh, derp. Thanks.

src/vdaf.rs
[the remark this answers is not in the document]
tgeoghegan: Well, I might not want to move `lhs`. Anyway, this is premature because I can't show you the use case for either yet. Let's stick with `merge` for now, and I may make a PR on this when i adopt it.

src/vdaf.rs
```diff
@@ -1,9 +1,229 @@
...
+
+    /// An Aggregator's share of the aggregate result.
+    type AggregateShare: Aggregable;
+
+    /// Generates the long-lived parameters used by the Clients and Aggregators.
+    fn setup(&self) -> Result<(Self::PublicParam, Vec<Self::VerifyParam>), VdafError>;
+
+    /// The number of rounds of communication among the Aggregators that are required before they
+    /// have recovered output shares.
+    fn rounds(&self) -> usize;
```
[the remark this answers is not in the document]
tgeoghegan: To me, it's akin to an iterator. I don't need to know how many items are in the iterator to enumerate them, and if I do `for item in iterator`, I am trusting that `iterator.next()` will eventually return `None`.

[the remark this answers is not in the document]
cjpatton: Fair enough. I'd be fine with removing `num_rounds` if you'd like me to.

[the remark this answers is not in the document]
tgeoghegan: My 2c: let's remove it, since it'll be very easy to add it back in if we end up finding a need for it in some client.

tgeoghegan (approved):

src/vdaf.rs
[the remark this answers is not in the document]
cjpatton: Done.

cjpatton:
Squashed.

cjpatton:
Rebased.

PR #109, opened 2021-10-30
pcp::types: Rename `Boolean` to `Count`

### LINKED ISSUE #94: Naming conventions for `Value` implementations, plus new types

cjpatton:
Based on #108 (merge that first!).
Partially closes #94.

### DIFF
--- a/src/pcp/types.rs
+++ b/src/pcp/types.rs
@@ -18,14 +18,15 @@ pub enum TypeError {
     Instantiate(&'static str),
 }
 
-/// Values of this type encode a simple boolean (either `true` or `false`).
+/// The counter data type. Each measurement is `0` or `1` and the aggregate result is the sum of
+/// the measurements (i.e., the number of `1s`).
 #[derive(Clone, Debug, PartialEq, Eq)]
-pub struct Boolean<F: FieldElement> {
+pub struct Count<F: FieldElement> {
     data: Vec<F>,  // The encoded input
     range: Vec<F>, // A range check polynomial for [0, 2)
 }
 
-impl<F: FieldElement> Boolean<F> {
+impl<F: FieldElement> Count<F> {
     /// Encodes a boolean as a value of this type.
     pub fn new(b: bool) -> Self {
         Self {
@@ -38,7 +39,7 @@ impl<F: FieldElement> Boolean<F> {
     }
 }
 
-impl<F: FieldElement> Value for Boolean<F> {
+impl<F: FieldElement> Value for Count<F> {
     type Field = F;
     type Param = ();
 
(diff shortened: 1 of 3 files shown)


src/pcp/types.rs
```diff
@@ -18,14 +18,15 @@ pub enum TypeError {
...
+/// the measurements (i.e., the number of `1s`).
 #[derive(Clone, Debug, PartialEq, Eq)]
-pub struct Boolean<F: FieldElement> {
+pub struct Count<F: FieldElement> {
     data: Vec<F>,  // The encoded input
     range: Vec<F>, // A range check polynomial for [0, 2)
 }
 
-impl<F: FieldElement> Boolean<F> {
+impl<F: FieldElement> Count<F> {
```
tgeoghegan, line 29: Having a type `Count` that can only represent 0 or 1 feels wrong to me. First, a boolean is the wrong semantic: we're not representing a logical proposition, we're representing a count of something, so an integer is more natural. Second, I don't think this lines up with how it will be used. The analogy here would be a [Prometheus counter](https://prometheus.io/docs/concepts/metric_types/#counter), right? The [interface to a counter ](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#Counter) lets me increment the existing value by arbitrary steps, and I don't think this `Count` type does that naturally.

Let's say that I'm a client application and I've recorded something of interest happening 711 times. How do I represent that with `Count`? Do I make a vector of 711 `Count::new(true)` and secret share all of them? That doesn't feel right to me.

I wonder if what we want is something like

```impl<F: FieldElement> Count<F> {
   fn new(upper_bound: F::Integer, value: F::Integer) -> Self {
        Self {
            range: poly_range_check(0, upper_bound),
            data: F::from(value),
       }
   }
```

tgeoghegan: Having now read #112, I see that `PolyCheckedVector`/`Sum` is the type I was asking for. At that point, though, isn't `Count` just a special case of `Sum` where the upper bound is 1?

tgeoghegan (changes requested):

src/pcp/types.rs
cjpatton: Yes, from a logical standpoint `Count` is basically a special case of `Sum`. the proof systems are different, however. In particular, `Count` is simpler because it doesn't have a joint random input.

You're right that the analogy with Prometheus isn't perfect, but there's a lot of ways that the analogy breaks down. Another is that Prometheus provides time series data, whereas PPM will provide aggregates for time windows.

As far as `bool` being the wrong input type: I agree. I change it to `u64` later on in the queue.

cjpatton:
Rebased and changed base to main.

src/pcp/types.rs
tgeoghegan: OK, I buy that `Count` admits a simpler proof system than `Sum`.

tgeoghegan (approved):

PR #111, opened 2021-11-01
pcp::types: Specialize `PolyCheckedVector` into `Sum` type

### LINKED ISSUE #94: Naming conventions for `Value` implementations, plus new types

cjpatton:
Based on #110.
Partially closes #94.

### DIFF
--- a/src/pcp/types.rs
+++ b/src/pcp/types.rs
@@ -103,41 +103,58 @@ impl<F: FieldElement> TryFrom<((), &[F])> for Count<F> {
     }
 }
 
-/// This type represents vectors for which `poly(x) == 0` holds for some polynomial `poly` and each
-/// vector element `x`.
+/// This sum type. Each measurement is a integer in `[0, 2^bits)` and the aggregate is the sum of the measurements.
 ///
-/// This type is "generic" in that it can be used to construct high level types. For example, a
-/// boolean vector can be represented as follows:
+/// The validity circuit is based on the SIMD circuit construction of [[BBCG+19], Theorem 5.3].
 ///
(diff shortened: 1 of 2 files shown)


src/pcp/types.rs
```diff
@@ -103,41 +103,58 @@ impl<F: FieldElement> TryFrom<((), &[F])> for Count<F> {
...
+                "value of summand exceeds bit length".to_string(),
+            ));
+        }
+
+        let one = F::Integer::try_from(1).unwrap();
+        let mut data: Vec<F> = Vec::with_capacity(bits);
+        for l in 0..bits {
+            let l = F::Integer::try_from(l).unwrap();
+            let w = F::from((int >> l) & one);
+            data.push(w);
```
tgeoghegan, line 145: IIUC, we'll end up encoding a single `Field126` value as 128 `Field126` values, each of which is 0 or 1. Why not represent it as a single `Field126`, and use `poly_range_check(0, Field126::MAX)` as `range_checker`? Or use some caller provided upper bound (perhaps they know something about the data they're recording that lets them provide a tighter upper bound).

tgeoghegan (changes requested):
I think my question here is related to my questions on #109.

src/pcp/types.rs
tgeoghegan: OK, that all makes sense. I have no other comments on this PR, but we should resolve #109 so I'll refrain from approving this for now.

tgeoghegan (approved):

cjpatton:
Removed `PolyCheckedVector` benchmark.

PR #112, opened 2021-11-01
pcp::types: Remove `MeanVarUnsignedVector`

### LINKED ISSUE #94: Naming conventions for `Value` implementations, plus new types

cjpatton:
Based on #111.
Partially closes #94.


tgeoghegan (approved):

cjpatton:
Rebased and removed `MeanVarUnsignedVector` benchmark.

PR #113, opened 2021-11-01
pcp::type: Clean up unit test framing

cjpatton:
Based on #112.

### DIFF
--- a/src/pcp/types.rs
+++ b/src/pcp/types.rs
@@ -251,14 +251,17 @@ mod tests {
                 expect_valid: true,
                 expected_proof_len: 9,
             },
-        );
+        )
+        .unwrap();
+
         pcp_validity_test(
             &Count::<TestField>::new(0).unwrap(),
             &ValidityTestCase {
                 expect_valid: true,
                 expected_proof_len: 9,
             },
-        );
+        )
+        .unwrap();
 
         // Test PCP on invalid input.
         pcp_validity_test(
@@ -270,7 +273,8 @@ mod tests {
                 expect_valid: false,
                 expected_proof_len: 9,
             },
-        );
+        )
+        .unwrap();
 
         // Try running the validity circuit on an input that's too short.
         let malformed_x = Count::<TestField> {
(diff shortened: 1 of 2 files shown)


src/pcp/types.rs
```diff
@@ -245,23 +245,26 @@ mod tests {
...
-        pcp_validity_test(
+        let res = pcp_validity_test(
             &Count::<TestField>::new(1).unwrap(),
             &ValidityTestCase {
                 expect_valid: true,
                 expected_proof_len: 9,
             },
         );
-        pcp_validity_test(
+        assert!(res.is_ok(), "error: {:?}", res.unwrap_err());
```
tgeoghegan, line 255: You can just do `pcp_validity_test(...).unwrap();` and if `pcp_validity_test` returns an error, `unwrap()` will print it.

tgeoghegan (changes requested):

src/pcp/types.rs
cjpatton: Good call! Will fix when this PR is next up.

cjpatton:
Rebased.

src/pcp/types.rs
cjpatton: Done.

src/pcp/types.rs
```diff
@@ -356,59 +369,83 @@ mod tests {
...
                 expected_proof_len: 16,
             },
-        );
+        )
+        .unwrap();
     }
 
-    // TODO(cjpatton) Have this return an error and have the caller assert success (or failure).
-    fn pcp_validity_test<V: Value>(input: &V, t: &ValidityTestCase) {
+    fn pcp_validity_test<V: Value>(input: &V, t: &ValidityTestCase) -> Result<(), PcpError> {
```
tgeoghegan, line 376: Having a `cfg(test)`-only variant on `PcpError` is totally fine, but another option would be to have this return `Result<(), String>`, and then do `return Err(format!("whatever {}", something));`. That is, unless you also need to propagate other variants of `PcpError` from this function.

tgeoghegan (approved):

src/pcp/types.rs
cjpatton: Since the same function may propagate a `PcpError`, adding the variant seemed like the simplest option. Otherwise we'd have to convert `PcPError`s to strings explicitly. (This wouldn't be so bad...)

tgeoghegan: Since you need to handle other `PcpError` variants, let's keep what you have and which already works.

PR #114, opened 2021-11-01
add rust-toolchain.toml

tgeoghegan:
Adds a
[`rust-toolchain.toml`](https://rust-lang.github.io/rustup/overrides.html)
to specify a required compiler version, ensuring that (1) we build with
a recent toolchain that supports features we need like const generics
and (2) we use a toolchain that includes the mitigation for
[CVE-2021-42574](https://rust-lang.github.io/rustup/overrides.html).

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,0 +1,3 @@
+[toolchain]
+channel = "1.56.1"
+components = [ "rustc", "rustfmt", "clippy" ]


tgeoghegan:
The other neat thing about this is that we're no longer stuck with whatever toolchain the GitHub runner VMs happen to have installed. To their credit they keep it reasonably up to date, but [as of this writing, it's 1.55.0](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md).

cjpatton (approved):

PR #115, opened 2021-11-01
pcp: Clean up API for constructing input shares

cjpatton:
Based on #113.

* Remove the `TryFrom<(Value::Param, &'a [Value::Field])` trait bound
from `Value`.

* Remove `Value::set_leader()`, which was used to optionally specialize
the validity circuit depending on who owns the share (a leader or
helper).

* Add `Value::new_share()` method, which takes input the raw input
share, type parameters, and the total number of shares. The total number
of shares is used to optionally specialize the validity circuit.

### DIFF
--- a/src/pcp.rs
+++ b/src/pcp.rs
@@ -113,7 +113,7 @@ pub enum PcpError {
 
 /// A value of a certain type. Implementations of this trait specify an arithmetic circuit that
 /// determines whether a given value is valid.
-pub trait Value: Sized + PartialEq + Eq + Debug {
+pub trait Value: Sized + Eq + Debug {
     /// The finite field used for this type.
     type Field: FieldElement;
 
--- a/src/pcp.rs
+++ b/src/pcp.rs
@@ -113,18 +113,21 @@ pub enum PcpError {
 
 /// A value of a certain type. Implementations of this trait specify an arithmetic circuit that
 /// determines whether a given value is valid.
-pub trait Value:
-    Sized
-    + PartialEq
-    + Eq
-    + Debug
-    + for<'a> TryFrom<(<Self as Value>::Param, &'a [Self::Field]), Error = PcpError>
-{
+pub trait Value: Sized + PartialEq + Eq + Debug {
     /// The finite field used for this type.
     type Field: FieldElement;
 
     /// Parameters used to construct a value of this type from a vector of field elements.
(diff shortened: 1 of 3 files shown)


src/pcp.rs
```diff
@@ -113,19 +113,22 @@ pub enum PcpError {
...
 /// A value of a certain type. Implementations of this trait specify an arithmetic circuit that
 /// determines whether a given value is valid.
-pub trait Value:
-    Sized
-    + PartialEq
-    + Eq
-    + Debug
-    + for<'a> TryFrom<(<Self as Value>::Param, &'a [Self::Field]), Error = PcpError>
-{
+pub trait Value: Sized + PartialEq + Eq + Debug {
```
tgeoghegan, line 116:
```suggestion
pub trait Value: Sized + Eq + Debug {
```
[`Eq` implies `PartialEq`](https://doc.rust-lang.org/std/cmp/trait.Eq.html)

src/pcp.rs
```diff
@@ -113,19 +113,22 @@ pub enum PcpError {
...
     type Field: FieldElement;
 
     /// Parameters used to construct a value of this type from a vector of field elements.
     type Param: Clone + Debug;
 
+    /// Constructs an instance of a value from an additive share of the input (`data`) and type
+    /// parameter (`param`). For some types, the validity circuit may depend on the total number of
+    /// shares (`num_shares`).
+    fn new_share(
+        data: Vec<Self::Field>,
```
tgeoghegan, line 127:
```suggestion
        data: &[Self::Field],
```
More flexible than `Vec` and I don't think we need to consume the data.

tgeoghegan (changes requested):

src/pcp.rs
cjpatton: Actually, the data is always consumed. E.g., see the implementation of `Sum`.

src/pcp.rs
cjpatton: Done.

src/pcp.rs
tgeoghegan: Yeah, seems like you either have to force the caller to allocate a `Vec`, or you have to force a `to_vec` or a `clone` in the various constructors. Let's keep this as-is.

tgeoghegan (approved):

PR #116, opened 2021-11-01
pcp::types: Add `Histogram` type

### LINKED ISSUE #94: Naming conventions for `Value` implementations, plus new types

cjpatton:
Based on #115.
Partially closes #94.

### DIFF
--- a/src/pcp/types.rs
+++ b/src/pcp/types.rs
@@ -217,6 +217,131 @@ impl<F: FieldElement> Value for Sum<F> {
     }
 }
 
+/// The histogram type. Each measurement is a non-negative integer and the aggregate is a histogram
+/// approximating the distribution of the measurements.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct Histogram<F> {
+    data: Vec<F>,
+    range_checker: Vec<F>,
+    sum_check_share: F,
+}
+
+impl<F: FieldElement> Histogram<F> {
+    /// Constructs a new histogram input. The values of `buckets` must be strictly increasing.
+    pub fn new(measurement: u64, buckets: &[u64]) -> Result<Self, PcpError> {
+        let mut data = vec![F::zero(); buckets.len() + 1];
+
+        if buckets.len() >= u32::MAX as usize {
+            return Err(PcpError::Value(
+                "invalid buckets: number of buckets exceeds maximum permitted".to_string(),
+            ));
+        }
+
+        if !buckets.is_empty() {
+            for i in 0..buckets.len() - 1 {
+                if buckets[i + 1] <= buckets[i] {
+                    return Err(PcpError::Value(
+                        "invalid buckets: out-of-order boundary".to_string(),
+                    ));
+                }
+            }
+        }
+
+        let bucket = match buckets.binary_search(&measurement) {
+            Ok(i) => i,  // on a bucket boundary
+            Err(i) => i, // smaller than the i-th bucket boundary
+        };
+
+        data[bucket] = F::one();
+
+        Ok(Self {
+            data,
+            range_checker: poly_range_check(0, 2),
+            sum_check_share: F::one(),
+        })
+    }
+}
+
+impl<F: FieldElement> Value for Histogram<F> {
+    type Field = F;
+    type Param = u32; // Number of buckets
+
+    fn new_share(data: Vec<F>, num_buckets: &u32, num_shares: usize) -> Result<Self, PcpError> {
+        let expected_data_len = usize::try_from(*num_buckets).unwrap() + 1;
+        if data.len() != expected_data_len {
+            return Err(PcpError::Value(
+                "data length does not match buckets".to_string(),
+            ));
+        }
+
+        let sum_check_share = F::one() / F::from(F::Integer::try_from(num_shares).unwrap());
+        Ok(Self {
+            data: data.to_vec(),
+            range_checker: poly_range_check(0, 2),
+            sum_check_share,
+        })
+    }
+
+    fn valid(&self, g: &mut Vec<Box<dyn Gadget<F>>>, rand: &[F]) -> Result<F, PcpError> {
+        if rand.len() != self.joint_rand_len() {
+            return Err(PcpError::Valid(format!(
+                "unexpected joint randomness length: got {}; want {}",
+                rand.len(),
+                self.joint_rand_len()
+            )));
+        }
+
+        // Check that each element of `data` is a 0 or 1.
+        let mut range_check = F::zero();
+        let mut r = rand[0];
+        for chunk in self.data.chunks(1) {
+            range_check += r * g[0].call(chunk)?;
+            r *= rand[0];
+        }
+
+        // Check that the elements of `data` sum to 1.
+        let mut sum_check = -self.sum_check_share;
+        for val in self.data.iter() {
+            sum_check += *val;
+        }
+
+        // Take a random linear combination of both checks.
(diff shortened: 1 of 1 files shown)


src/pcp/types.rs
```diff
@@ -217,6 +217,133 @@ impl<F: FieldElement> Value for Sum<F> {
...
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct Histogram<F> {
+    data: Vec<F>,
+    range_checker: Vec<F>,
+    sum_check_share: F,
+}
+
+impl<F: FieldElement> Histogram<F> {
+    /// Constructs a new histogram input. The values of `buckets` must be strictly increasing.
+    pub fn new(measurement: u64, buckets: &[u64]) -> Result<Self, PcpError> {
```
tgeoghegan, line 231: I wonder if we could make `Histogram` generic over integer primitive types, and then allow construction of histograms with anything from `u32` to `u128`, provided that the chosen `FieldElement` is big enough to represent the chosen primitive integer? [trait `Unsigned`](https://docs.rs/num/0.4.0/num/trait.Unsigned.html) could work.

src/pcp/types.rs
```diff
@@ -217,6 +217,133 @@ impl<F: FieldElement> Value for Sum<F> {
...
+        let mut data = vec![F::zero(); buckets.len() + 1];
+
+        if buckets.len() == 0 {
```
tgeoghegan, line 234: nit: elsewhere, `num_buckets` is `u32`, so perhaps you should check that `buckets.len() <= u32.MAX`?

src/pcp/types.rs
```diff
@@ -217,6 +217,133 @@ impl<F: FieldElement> Value for Sum<F> {
...
+            sum_check_share: F::one(),
+        })
+    }
+}
+
+impl<F: FieldElement> Value for Histogram<F> {
+    type Field = F;
+    type Param = u32; // Number of buckets
+
+    fn new_share(data: Vec<F>, num_buckets: &u32, num_shares: usize) -> Result<Self, PcpError> {
```
tgeoghegan, line 271:
```suggestion
    fn new_share(data: Vec<F>, num_buckets: u32, num_shares: usize) -> Result<Self, PcpError> {
```

src/pcp/types.rs
```diff
@@ -217,6 +217,133 @@ impl<F: FieldElement> Value for Sum<F> {
     }
 }
 
+/// The histogram type. Each measurement is a non-negative integer and the aggregate is a histogram
+/// approximating the distribution of the measurements.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct Histogram<F> {
+    data: Vec<F>,
```
tgeoghegan, line 224: IIUC, we represent a single contribution to a histogram as a one-hot bit vector whose dimension is the number of buckets, where we use a field for each bit. Could we compress this into the index of the bucket that the value falls into, which would require a single field? That said, the range checker polynomial would have to be of degree equal to the number of buckets, but maybe that's tenable for most histograms which will have a small number of buckets.

tgeoghegan: If this is possible at all, we could/should save it for later. Maybe do something clever where we use either the bucket index only for histograms with few enough buckets, and the bit vector otherwise.

src/pcp/types.rs
```diff
@@ -217,6 +217,133 @@ impl<F: FieldElement> Value for Sum<F> {
...
+        }
+
+        for i in 0..buckets.len() {
+            if measurement <= buckets[i] {
+                data[i] = F::one();
+                break;
+            }
+        }
+
+        if measurement > buckets[buckets.len() - 1] {
```
tgeoghegan, line 255: nit: You could check this _before_ the for loop immediately above and short-circuit the linear search.

src/pcp/types.rs
```diff
@@ -217,6 +217,133 @@ impl<F: FieldElement> Value for Sum<F> {
...
+
+        for i in 0..buckets.len() - 1 {
+            if buckets[i + 1] <= buckets[i] {
+                return Err(PcpError::Value(
+                    "invalid buckets: out-of-order boundary".to_string(),
+                ));
+            }
+        }
+
+        for i in 0..buckets.len() {
```
tgeoghegan, line 248: nit: you could use a O(log n) search to find the right bucket but I don't think it'll matter for realistic histograms which will have relatively few buckets.

tgeoghegan (changes requested):

src/pcp/types.rs
tgeoghegan: Never mind, I see now this is necessary to conform to `Value`.

src/pcp/types.rs
```diff
@@ -362,6 +489,115 @@ mod tests {
...
+
+        // Test valid inputs.
+        let res = pcp_validity_test(
+            &Histogram::<TestField>::new(0, &buckets).unwrap(),
+            &ValidityTestCase {
+                expect_valid: true,
+                expected_proof_len: 8,
+            },
+        );
+        assert!(res.is_ok(), "error: {:?}", res.unwrap_err());
```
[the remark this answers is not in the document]
cjpatton: Done, here and below.

src/pcp/types.rs
cjpatton: Remember: the size of the proof is proportional to the degree of this polynomial, so something like this wouldn't yield any savings, at least asymptotically.

That said, we can definitely do better.
* By replacing the input share with a DPF key share, we can get the input size down from O(N) to O(lon N) where N is the number of buckets.
* For the proof we can do one of the following:
  * We can replace the Prio-style input validation step with the secure sketching protocol from the heavy hitters protocol. That would get the size of the proof down from O(N) to O(log N). The downside is that it requires two rounds instead of one.
  * Using something like #37 we can get the proof size down to O(N^0.5).

However, as you pointed out, N (the number of buckets) is going to be relatively small for most applications (on the order of 10 to 100 is what I'd imagine.) I definitely want to implement proof-size optimizations for the vectorized variants of things (see #94), but since these come at the cost of a certain amount of computational overhead, perhaps it makes sense to hold off here.

src/pcp/types.rs
cjpatton: `F` already provides `F::Integer`, which should be suitable here. It's slightly annoying to make the caller do the conversion from `u64` (or whatever) to `F::Integer`, but I don't it's too annoying. In any case, it probably makes sense at this (relatively low) level of the library.

If we do this, I think we should also do it for `Count` and `Sum`. Would you mind if I make a note to implement this change later, after everything else is done? (I've got a running list of TODOs.)

src/pcp/types.rs
cjpatton: Done.

src/pcp/types.rs
cjpatton: Done.

src/pcp/types.rs
cjpatton: No longer needed, now that we're doing binary search.

src/pcp/types.rs
tgeoghegan: > Would you mind if I make a note to implement this change later, after everything else is done? (I've got a running list of TODOs.)
Sounds good,

src/pcp/types.rs
tgeoghegan: Oh, right, I forgot about the proof size! OK, let's keep this as-is.

tgeoghegan (approved):

cjpatton:
Squashed.

cjpatton:
Solved clippy.

PR #120, opened 2021-11-02
pcp: Add functionality required for `prio3` VDAF

cjpatton:
Adds functionality required for the VDAF. While at it, this simplifies
the interface between `Value` and the PCP methods.

* Define the `ValueParam` trait, which determines the length (in field
elements) of the various inputs and outputs of PCP generation and
verification. This includes things like input length, proof length, and
verifier length that are required for the VDAF.

* Make `ValueParam` a trait bound on `Value::Param` and drop redundant
methods on `Value` (`joint_rand_len`, etc.).

* Add `Value::into_output()` method that transforms a `Value` instance
into an aggregatable output share. This functionality is required for
the VDAF.

* Add `Gadget::calls()` callback so that an instance of a gadget
describes how many times it expects to be called in an arithmetic
circuit. This information is needed to construct all current circuits
anyway.

* Remove `Value::valid_gadget_calls()` as this callback is no longer
needed.

### DIFF
--- a/src/pcp.rs
+++ b/src/pcp.rs
@@ -20,24 +20,25 @@
 //!
 //! ```
 //! use prio::pcp::types::Count;
-//! use prio::pcp::{decide, prove, query, Value};
+//! use prio::pcp::{decide, prove, query, Value, ValueParam};
 //! use prio::field::{random_vector, FieldElement, Field64};
 //!
 //! // The prover chooses a measurement.
 //! let input: Count<Field64> = Count::new(0).unwrap();
+//! let param = input.param();
 //!
 //! // The prover and verifier agree on "joint randomness" used to generate and
 //! // check the proof. The application needs to ensure that the prover
 //! // "commits" to the input before this point. In the `prio3` VDAF, the joint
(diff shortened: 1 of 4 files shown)


cjpatton:
Solved clippy.

src/pcp.rs
```diff
@@ -801,33 +780,66 @@ mod tests {
...
         }
 
-        fn query_rand_len(&self) -> usize {
+        fn into_output(self) -> Vec<F> {
+            self.data
+        }
+    }
+
+    #[derive(Clone, Debug)]
+    struct TestValueParam();
```
tgeoghegan, line 804: nit: You can define an empty struct and still derive Clone and Debug: 
```suggestion
    struct TestValueParam;
```
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=667df556946b5fab68f8d6dfa59bb588

tgeoghegan (approved):

src/pcp.rs
cjpatton: Thanks for the tip :)

PR #121, opened 2021-11-02
vdaf: Change `Helper` variant of `Share`

cjpatton:
Removes the length parameter from the helper share. This parameter can
always be computed in context.

### DIFF
--- a/src/vdaf/prio3.rs
+++ b/src/vdaf/prio3.rs
@@ -17,7 +17,6 @@ use crate::prng::Prng;
 use crate::vdaf::suite::{Key, KeyDeriver, KeyStream, Suite};
 use crate::vdaf::{Share, VdafError};
 use serde::{Deserialize, Serialize};
-use std::convert::TryFrom;
 use std::iter::IntoIterator;
 
 /// The message sent by the client to each aggregator. This includes the client's input share and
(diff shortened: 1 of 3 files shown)


src/vdaf/prio3.rs
```diff
@@ -266,10 +259,10 @@ pub fn prio3_start<V: Value>(
     deriver.update(nonce);
     let query_rand_seed = deriver.finish();
 
-    let input_share_data: Vec<V::Field> = Vec::try_from(msg.input_share)?;
+    let input_share_data: Vec<V::Field> = msg.input_share.into_vec(param.input_len())?;
     let input_share = V::new_share(input_share_data, param, verify_param.num_shares as usize)?;
 
-    let proof_share_data: Vec<V::Field> = Vec::try_from(msg.proof_share)?;
+    let proof_share_data: Vec<V::Field> = msg.proof_share.into_vec(param.proof_len())?;
```
tgeoghegan, line 265: Are the type annotations on `input_share_data` and `proof_share_data` needed anymore since we don't have to give the compiler a hint on which `try_from` to use?

tgeoghegan (approved):

src/vdaf/prio3.rs
cjpatton: Nope. Removed.

PR #122, opened 2021-11-02
pcp: Align API with draft-patton-cfrg-vdaf-00

cjpatton:
Replace `Value` with trait `Type`, implementations of which provide the
FLP functionality described in the spec. In particular, each `Type`
implementation specifies how measurements are encoded and validated. The
FLP system is now implemented as a set of derived methods on a generic
`Type`. However, the type itself does not subsume the input that is
validated.
    
Shifting the API boundary this way makes it easier to integrate the FLP
into `prio3`.

### DIFF
--- a/src/pcp.rs
+++ b/src/pcp.rs
@@ -1,14 +1,14 @@
 // SPDX-License-Identifier: MPL-2.0
 
 //! **(NOTE: This module is experimental. Applications should not use it yet.)** This module
-//! defines the [`Value`] trait, the main building block of the [`prio3`](crate::vdaf::prio3) VDAF.
+//! defines the [`Type`] trait, the main building block of the [`prio3`](crate::vdaf::prio3) VDAF.
 //! Implementations of this trait for various measurement types can be found in [`types`].
 //!
 //! This module also implements a fully linear PCP ("Probabilistically Checkable Proof") system
-//! based on [[BBCG+19], Theorem 4.3] suitable for implementations of [`Value`]. This proof system
-//! is used in `prio3` to ensure validity of the recovered output shares. Most applications will
-//! not need to use this module directly. Typically the VDAF will provide the functionality you
-//! need.
+//! based on [[BBCG+19], Theorem 4.3] suitable for implementations of [`Type`]. Implementations of
+//! [`Type`] provide the FLP functionality described in [[VDAF], Section 6.1] and used in `prio3`
+//! to ensure validity of the recovered output shares. Most applications will not need to use this
+//! module directly.
 //!
 //! # Overview
 //!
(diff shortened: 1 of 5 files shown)


src/pcp.rs
```diff
@@ -135,65 +199,269 @@ pub trait ValueParam: Clone + Debug {
...
+            // Construct the gadget polynomial `G(f[0], ..., f[g_arity-1])` and append it to `proof`.
+            gadget.call_poly(&mut proof[proof_len + gadget.arity()..], &f)?;
+            proof_len += gadget.arity() + gadget.degree() * (m - 1) + 1;
+        }
 
-    /// Evaluates the validity circuit on the given input (i.e., `self`) and returns the output.
-    /// `joint_rand` is the joint randomness shared by the prover and verifier. `g` is the sequence
-    /// of gadgets called by the circuit.
+        // Truncate the buffer to the size of the proof.
+        proof.truncate(proof_len);
```
tgeoghegan, line 287: I'm kind of surprised we never use `self.proof_len()` in this function. Should we at least check that the `proof_len` computed here is the expected value?

src/pcp.rs
```diff
@@ -135,65 +199,269 @@ pub trait ValueParam: Clone + Debug {
...
+        Ok(proof)
+    }
+
+    /// Query an input and proof and return the verifier message.
```
tgeoghegan, line 291:
```suggestion
    /// Query an input and proof and return the verifier message, encoded as [`Self::verifier_len`] field elements.
```

src/pcp.rs
```diff
@@ -135,65 +199,269 @@ pub trait ValueParam: Clone + Debug {
...
+            .into_iter()
+            .enumerate()
+            .map(|(idx, gadget)| {
+                if idx >= query_rand.len() {
+                    return Err(PcpError::Query(format!(
+                        "short query randomness: got {}; want {}",
+                        query_rand.len(),
+                        self.query_rand_len()
+                    )));
+                }
```
tgeoghegan, line 321: Would it be simpler to do `if self.gadget.len() == query_rand.len() {}` once, before iterating `self.gadget()`? Then I think you could also drop the check for `query_rand.len() > shim.len()` on line 364, since `shim.len()` should always be the same as `self.gadget.len()`.

 Additionally I think you could zip `self.gadget().into_iter()` with `query_rand.into_iter()` and avoid indexing into `query_rand`, and then you don't need to `.enumerate()`, but either way is fine with me.

src/pcp.rs
```diff
@@ -135,65 +199,269 @@ pub trait ValueParam: Clone + Debug {
...
+            let mut f = vec![Self::Field::zero(); m];
+            for wire in 0..gadget.arity() {
+                discrete_fourier_transform(&mut f, &gadget.f_vals[wire], m)?;
+                discrete_fourier_transform_inv_finish(&mut f, m, m_inv);
+                verifier.push(poly_eval(&f, r));
+            }
+
+            // Add the value of the gadget polynomial evaluated at `r`.
+            verifier.push(gadget.p_at_r);
+        }
```
tgeoghegan, line 415: In the unlikely event that `poly_eval` returns a value longer than `shim[idx].arity`, we will end up with more than `data_len` elements in `verifier`. Could we check that `verifier.len() == self.verifier_len()` and error out if so?

src/pcp.rs
```diff
@@ -135,65 +199,269 @@ pub trait ValueParam: Clone + Debug {
...
+
+        Ok(verifier)
+    }
+
+    /// Returns true if the verifier message indicates that the input from which it was generated is valid.
+    #[allow(clippy::needless_range_loop)]
+    fn decide(&self, verifier: &[Self::Field]) -> Result<bool, PcpError> {
```
tgeoghegan, line 422: Could check that `verifier.len() == self.verifier_len()`

src/pcp/types.rs
```diff
@@ -259,124 +216,101 @@ impl ValueParam for SumValueParam {
...
+            return Err(PcpError::Encode(
                 "invalid buckets: number of buckets exceeds maximum permitted".to_string(),
             ));
         }
 
-        if !buckets.is_empty() {
-            for i in 0..buckets.len() - 1 {
-                if buckets[i + 1] <= buckets[i] {
-                    return Err(PcpError::Value(
+        // XXX Move this check to constructor
```
tgeoghegan, line 248: Any reason not to do this now, and make `Histogram::new` fallible? Same goes for `Sum::new`.

tgeoghegan (changes requested):

src/pcp/types.rs
```diff
@@ -123,120 +104,96 @@ impl ValueParam for CountValueParam {
...
+}
 
-        if bits > (size_of::<F::Integer>() << 3) {
-            return Err(PcpError::Value(
+impl<F: FieldElement> Type for Sum<F> {
+    type Measurement = F::Integer;
+    type Field = F;
+
+    fn encode(&self, summand: &F::Integer) -> Result<Vec<F>, PcpError> {
+        // XXX Move this check to constructor
```
cjpatton, line 128: I meant to resolve these "XXX"s before starting review. I'll resolve them before we merge.

src/pcp.rs
cjpatton: Hmm, I'm not sure I understand the problem. Are you concerned about an overflow? `poly_eval` always returns exactly one field element. In any case, if a `.push()` would exceed the limits of the memory allocated for the vector, the vector's memory will be extended.

src/pcp.rs
cjpatton: This is done implicitly. Before parsing the next chunk from `verifier`, we check if `verifier` has enough memory (and fail if not). For the same reason I gave for `proof`, I think it's if we compute what `self.verifier_len()` is supposed to return to make it easier to catch incorrect implementations of this call.

Here again we could check that the computed verifier length matches `self.verifier_len()`.

src/pcp/types.rs
cjpatton: The reason to do this in the constructor is that we only need to validate that `self.buckets` is well-formed once. As it stands, we end up re-validating this every time we call `encode`.

src/pcp.rs
```diff
@@ -112,10 +118,68 @@ pub enum PcpError {
...
+    /// Evaluates the validity circuit on an input and returns the output.
+    ///
+    /// # Parameters
+    ///
+    /// * `gadgets` is the sequence of gadgets, presumably output by [`Type::gadget`].
+    /// * `input` is the input to be validated.
+    /// * `joint_rand` is the joint randomness shared by the prover and verifier.
+    /// * `num_shares` is the number of input shares. Typically its value is `1`, but
+    /// [`prio3`](crate::vdaf::prio3) sets this to the number of aggregators. This allows the
+    /// validity circuit to specialize the value of constants.
```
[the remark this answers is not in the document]
tgeoghegan: Yeah, I think that's correct in all cases.

src/pcp.rs
tgeoghegan: I agree with your observation that this is programmer error so a panic feels appropriate.

src/pcp.rs
```diff
@@ -135,65 +199,269 @@ pub trait ValueParam: Clone + Debug {
...
+            .map(|idx| {
+                shim[idx].arity() + shim[idx].degree() * (1 + shim[idx].calls()).next_power_of_two()
+            })
+            .sum();
+        let mut proof = vec![Self::Field::zero(); data_len];
+
+        // Run the validity circuit with a sequence of "shim" gadgets that record the value of each
+        // input wire of each gadget evaluation. These values are used to construct the wire
+        // polynomials for each gadget in the next step.
+        let _ = self.valid(&mut shim, input, joint_rand, 1);
```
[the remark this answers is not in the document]
tgeoghegan: You're right. I think it might have if you didn't have `let _ =` (which I like, because it explicitly tells the reader "this returns a value and we're intentionally ignoring it"). Might be worth filing an issue on clippy.

src/pcp.rs
tgeoghegan: Oh, good point. My thinking is that this module should defend itself against errors in implementation of a `Type`, but you're right, `poly_eval` can't be implemented wrong, so we have a reasonable guarantee that the vector will contain the right number of elements. It might still be nice to check that the `Type`'s `verifier_len()` implementation is correct and panic -- your call.

tgeoghegan (changes requested):
I think we still need the checks on `Type::{proof_len, verifier_len}` and maybe `input_len` as discussed and then this is good to go.

src/pcp.rs
cjpatton: Done.

src/pcp.rs
cjpatton: I'll add an assertion, like we do for proof length in `prove`.

src/pcp.rs
cjpatton: Done.

tgeoghegan (approved):

cjpatton:
Squashed.

PR #124, opened 2021-11-03
vdaf: Update associated types of `PrepareStep`

cjpatton:
Associated types `Input` and `Output` define the input type and output
type of each round respectively. It is impossible for these types to be
different, however, since the type of outputs in a given needs to match
the input types of the next round. Thus this commit simplifies the type
by merging `Input` and `Output` into a single `Message` type. Each round
having a different message type is expressed using an `enum`.


tgeoghegan (approved):

cjpatton:
Rebased.

PR #126, opened 2021-11-04
vdaf: Align prio3 with the VDAF traits

cjpatton:
While at it, this adds an `output_len` to `Type` for getting the length
of the truncated output. This is used for aggregation.

### DIFF
--- a/src/vdaf.rs
+++ b/src/vdaf.rs
@@ -10,8 +10,8 @@
 
 use crate::field::FieldElement;
 use crate::pcp::PcpError;
-use crate::prng::{Prng, PrngError};
-use crate::vdaf::suite::{Key, KeyStream, SuiteError};
+use crate::prng::PrngError;
+use crate::vdaf::suite::{Key, SuiteError};
 use serde::{Deserialize, Serialize};
 
 /// Errors emitted by this module.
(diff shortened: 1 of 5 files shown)


cjpatton:
Rebased.

src/vdaf/prio3.rs
```diff
@@ -39,396 +208,470 @@ pub struct InputShareMessage<F> {
...
+            Share::Leader(_) => None,
+            Share::Helper(ref seed) => {
+                let prng = Prng::from_key_stream(KeyStream::from_key(seed));
+                Some(prng.take(self.typ.input_len()).collect())
+            }
+        };
+        let input_share = match msg.input_share {
+            Share::Leader(ref data) => data,
+            Share::Helper(_) => expanded_input_share.as_ref().unwrap(),
+        };
```
tgeoghegan, line 374: I'm not crazy about this construction, but I can see the bind you're in (Prng output would get dropped within the `Share::Helper` case so you can't return a reference to it) and I can't think of how to do better just now.

tgeoghegan (approved):

src/vdaf.rs
```diff
@@ -219,6 +198,24 @@ pub trait Aggregatable {
     fn merge(&mut self, agg_share: &Self) -> Result<(), VdafError>;
 }
 
+impl<F: FieldElement> Aggregatable for Vec<F> {
```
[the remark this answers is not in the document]
cjpatton: Hmm, the problem with IntoIterator is that `merge()` expects a reference. I'd suggest we punt on this until we know we need it.

src/vdaf/prio3.rs
cjpatton: Ack. I'm wondering how common this problem is and whether there is a better (i.e., more common) pattern for it. We have to do it twice, so anything we can do to make this a bit more DRY would be great.

PR #128, opened 2021-11-05
vdaf::prio3: Use measurement count as aggregation parameter

cjpatton:
Changes the type of the aggregation parameter for `Prio3` from `()` to
`usize`. This allows the number of measurements to be used for the
aggregation parameter, which in turn allows this number to be used the
computation of the aggregate result. We can now use prio3 to directly
express things like arithmetic mean, standard deviation, and so on.

The mapping from the aggregation parameter and the finite field
representation of the result to the final result is expressed as a trait
bound. One consequence of this changes it that we will need a unique
result type for each type so that different types can compute this map
differently.

### DIFF
--- a/src/vdaf/prio3.rs
+++ b/src/vdaf/prio3.rs
@@ -21,7 +21,7 @@ use std::iter::IntoIterator;
 use std::marker::PhantomData;
 
 /// The count type. Each measurement is an integer in `[0,2)` and the aggregate is the sum.
-pub type Prio3Count64 = Prio3<Count<Field64>, Prio3Result<u64>>;
+pub type Prio3Count64 = Prio3<Count<Field64>, Prio3CountResult<u64>>;
 
 impl Prio3Count64 {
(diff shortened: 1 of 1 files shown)


tgeoghegan (approved):

cjpatton:
Moved this to draft so we can stew on this. It's unfortunate that this design requires you to have a different result type for each data type. The goal here is to allow for aggregation functions that take into account the number of measurements. A simpler alternative would be to move the needed functionality to the type's implementation.

For instance, say we have the following type for computing the average and standard deviation of the measurements.
```rust
type Prio3AvgStdDev64 = Prio3<MeanStdDev<Field96>, Prio3ResultVec<u64>>;
```

We could just do
```rust
impl Prio3AvgStdDev {
   pub fn average(agg: &Prio3ResultVec<u64>, num_measurements: u64) -> u64 {
       ...
   }

   pub fn standard_deviation(agg: &Prio3ResultVec<u64>, num_measurements: u64) -> u64 {
       ...
   } 
}
```

PR #134, opened 2021-11-09
pcp: Properly validate inputs of `Sum::encode()`

cjpatton:
If `self.bits == 64` then this method will panic. The behavior that we
want is that encoding should succeed if `2^self.bits` can be encoded as
a field element. This change has performs this check in the constructor.
Also, the constructor now computes the maximum representable integer
than can be represented so that the encoding method can check this
properly.

### DIFF
--- a/src/pcp/types.rs
+++ b/src/pcp/types.rs
@@ -8,7 +8,6 @@ use crate::pcp::{Gadget, PcpError, Type};
 use crate::polynomial::poly_range_check;
 
 use std::convert::TryFrom;
-use std::mem::size_of;
 
 /// The counter data type. Each measurement is `0` or `1` and the aggregate result is the sum of
 /// the measurements (i.e., the total number of `1s`).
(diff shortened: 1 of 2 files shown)


src/pcp/types.rs
```diff
@@ -108,25 +107,38 @@ impl<F: FieldElement> Type for Count<F> {
...
+        if F::modulus() >> bits_int == F::Integer::try_from(0).unwrap() {
             return Err(PcpError::Encode(format!(
-                "bits ({}) exceeds bit length of the field's integer representation ({})",
-                bits, max_bits,
+                "bit length ({}) exceeds field modulus",
+                bits,
             )));
         }
 
+        let one = F::Integer::try_from(1).unwrap();
```
tgeoghegan, line 135: I feel like we do this in a lot of places. Maybe we could clean up a bunch of unwraps and avoid caching the value of 1 in struct `Sum` if we added a bound to `FieldElement::Integer` that requires some trait that exposes infallible `one()` and `zero()` methods. I wouldn't do that in this PR, though.

tgeoghegan (approved):
This fix looks sound, but as you noted in Slack, we need more test coverage on these types. Could you write up an issue tracking that when you merge this?

src/pcp/types.rs
cjpatton: Ugh, I always forget about `F::Integer::from(F::one())` and `F::Integer::from(F::zero())`. I did this here, but will wait to fix it elsewhere for a future PR.

PR #135, opened 2021-11-09
Fixes to make Prio3 VDAF work in ppm-prototype

tgeoghegan:
These are the changes I needed on libprio to make it work in https://github.com/abetterinternet/ppm-prototype.

### DIFF
--- a/src/vdaf.rs
+++ b/src/vdaf.rs
@@ -12,7 +12,9 @@ use crate::field::FieldElement;
 use crate::pcp::PcpError;
 use crate::prng::PrngError;
 use crate::vdaf::suite::{Key, SuiteError};
+use serde::de::DeserializeOwned;
 use serde::{Deserialize, Serialize};
+use std::fmt::Debug;
 
 /// Errors emitted by this module.
 #[derive(Debug, thiserror::Error)]
@@ -47,7 +49,7 @@ pub enum Share<F> {
 /// The base trait for VDAF schemes. This trait is inherited by traits [`Client`], [`Aggregator`],
 /// and [`Collector`], which define the roles of the various parties involved in the execution of
 /// the VDAF.
-pub trait Vdaf {
+pub trait Vdaf: Clone + Debug {
     /// The type of Client measurement to be aggregated.
     type Measurement;
 
@@ -66,10 +68,10 @@ pub trait Vdaf {
     type VerifyParam;
 
     /// An input share sent by a Client.
-    type InputShare;
+    type InputShare: Clone + Debug + Serialize + DeserializeOwned;
 
     /// An output share recovered from an input share by an Aggregator.
-    type OutputShare;
+    type OutputShare: Clone + Debug + Serialize + DeserializeOwned;
 
     /// An Aggregator's share of the aggregate result.
     type AggregateShare: Aggregatable;
@@ -95,10 +97,10 @@ pub trait Client: Vdaf {
 /// The Aggregator's role in the execution of a VDAF.
 pub trait Aggregator: Vdaf {
     /// State of the Aggregator during the Prepare process.
-    type PrepareStep;
+    type PrepareStep: Clone + Debug;
 
     /// The type of messages exchanged among the Aggregators during the Prepare process.
-    type PrepareMessage;
+    type PrepareMessage: Clone + Debug + Serialize + DeserializeOwned;
 
     /// Begins the Prepare process with the other Aggregators. The result of this process is
     /// the Aggregator's output share.
@@ -193,7 +195,7 @@ pub enum PrepareTransition<S, M, O> {
 }
 
 /// An aggregate share that can merged with aggregate shares of the same type.
-pub trait Aggregatable {
+pub trait Aggregatable: Clone + Debug + Serialize + DeserializeOwned {
     /// Update an aggregate share by merging it with another (`agg_share`).
     fn merge(&mut self, agg_share: &Self) -> Result<(), VdafError>;
 }
--- a/src/vdaf.rs
+++ b/src/vdaf.rs
@@ -74,7 +74,7 @@ pub trait Vdaf: Clone + Debug {
     type OutputShare: Clone + Debug + Serialize + DeserializeOwned;
 
     /// An Aggregator's share of the aggregate result.
-    type AggregateShare: Aggregatable;
+    type AggregateShare: Aggregatable<OutputShare = Self::OutputShare>;
 
     /// Generates the long-lived parameters used by the Clients and Aggregators.
     fn setup(&self) -> Result<(Self::PublicParam, Vec<Self::VerifyParam>), VdafError>;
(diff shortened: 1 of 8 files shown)


src/vdaf.rs
```diff
@@ -168,6 +170,14 @@ pub trait Aggregator: Vdaf {
         agg_param: &Self::AggregationParam,
         output_shares: M,
     ) -> Result<Self::AggregateShare, VdafError>;
+
+    /// Accumulates the sequence of output shares into the accumulator.
+    fn accumulate(
```
tgeoghegan, line 175: This is what we were talking through in Slack. I think we were both leaning towards putting some kind of `aggregate` method on trait `Aggregatable` that would add an `OutputShare` to an `AggregateShare`. I want to try implementing the Hits VDAF to see what ergonomics I want in that case.

cjpatton: Ack. I think the trait the method would be more natural on `Aggregatable`, but we can leave this open for now. I'm mainly worried about introducing yet another term, `accumulator`. for `output_share`. We could avoid this if instead we extended `Aggregatable`. Something like:
```rust
pub trait Aggregatable {
+    /// NEW
+    type OutputShare;

    /// Update an aggregate share by merging it with another (`agg_share`).
    fn merge(&mut self, agg_share: &Self) -> Result<(), VdafError>;

+    /// NEW
+    fn accumulate(&mut self, output_share: &OutputShare) -> Result<(), VdafError>;   
} 
```

src/vdaf/prio3.rs
```diff
@@ -143,13 +144,27 @@ fn check_num_aggregators(num_aggregators: u8) -> Result<(), VdafError> {
...
+#[derivative(Debug)]
 pub struct Prio3<T: Type, A> {
     num_aggregators: u8,
     suite: Suite,
     typ: T,
+    #[derivative(Debug = "ignore")]
     phantom: PhantomData<A>,
 }
 
+impl<T: Type, A> Clone for Prio3<T, A> {
```
tgeoghegan, line 157: We can't derive Clone because of the `PhantomData<A>` member, since `A` is not required to be `Clone`, and evidently the derived `Clone` implementation doesn't understand that it doesn't need to work with an actual `A` value.

src/vdaf/hits.rs
```diff
@@ -202,6 +209,12 @@ impl<I> Hits<I> {
     }
 }
 
+impl<I> Clone for Hits<I> {
```
tgeoghegan, line 212: Similar to the `Prio3` case, we can't derive `Clone` on `Hits` because of the `PhantomData<I>`.

src/vdaf/hits.rs
```diff
@@ -229,8 +242,11 @@ impl<I: Idpf<2, 2>> Vdaf for Hits<I> {
 }
 
 /// An input share for the heavy hitters VDAF.
+#[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct HitsInputShare<I: Idpf<2, 2>> {
     /// IDPF share of input
```
cjpatton, line 247: nit: Add a blank line `//` under the docucomment. (I'm not sure if rust has a convention for docucomments followed immediately by a normal comment.)

src/vdaf/hits.rs
```diff
@@ -522,24 +538,37 @@ impl<I: Idpf<2, 2>> Aggregator for Hits<I> {
...
+        output_share: &Self::OutputShare,
+    ) -> Result<(), VdafError> {
+        if output_share.len() != agg_param.len() || accumulator.len() != agg_param.len() {
+            return Err(VdafError::Uncategorized(format!(
+                "unexpected accumulator or output share length: got {}/{}; want {}",
+                accumulator.len(),
+                output_share.len(),
+                agg_param.len()
+            )));
+        }
```
cjpatton, line 560: This error message is a bit confusing. I'd suggest validating the length of `accumulator` and `output_share` on separate lines.

src/vdaf/prio3.rs
cjpatton: We could avoid the `A` generic by merging `Prio3Result` and `Prio3ResultVec` into `Prio3ResultVec`, if that would simplify things. I thought flexibility would be useful here, but perhaps not if it comes at the cost of complexity.

src/vdaf/prio3.rs
```diff
@@ -521,29 +537,47 @@ impl<T: Type, A> Aggregator for Prio3<T, A> {
...
-            for (x, y) in agg_share.iter_mut().zip(output_share) {
-                *x += y;
-            }
+            self.accumulate(agg_param, &mut agg_share, &output_share)?;
         }
 
         Ok(agg_share)
     }
+
+    /// Accumulates the sequence of output shares into the accumulator.
```
cjpatton, line 551: nit: Add empty `//` line here.

src/vdaf/prio3.rs
```diff
@@ -521,29 +537,47 @@ impl<T: Type, A> Aggregator for Prio3<T, A> {
...
+    // TODO(timg) it would be nice if accumulator could be `&[T::Field]` and
+    // output_share could be `&[Field]`, but I'm not sure how to do that in
+    // trait Vdaf since the associated type OutputShare has no notion of a
+    // field.
```
cjpatton, line 555: Plus, there's no reason an OutputShare *should* always have a notion of field.

src/vdaf/prio3.rs
```diff
@@ -521,29 +537,47 @@ impl<T: Type, A> Aggregator for Prio3<T, A> {
...
+        output_share: &Vec<T::Field>,
+    ) -> Result<(), VdafError> {
+        if accumulator.len() != self.typ.output_len() || output_share.len() != self.typ.output_len()
+        {
+            return Err(VdafError::Uncategorized(format!(
+                "unexpected output share or accmulator length: got {}/{}; want {}",
+                output_share.len(),
+                accumulator.len(),
+                self.typ.output_len()
+            )));
```
cjpatton, line 569: Here gain: split input validation into `output_share` and `accumulator`.

src/vdaf/prio3.rs
```diff
@@ -143,13 +144,27 @@ fn check_num_aggregators(num_aggregators: u8) -> Result<(), VdafError> {
 }
 
 /// The base type for prio3.
+#[derive(Derivative)]
+#[derivative(Debug)]
```
cjpatton, line 148: It seems like what this line is doing is allowing us to ignore `phantom` for the purposes of deriving the `Debug` trait. Why is this necessary here, but not for hits?

cjpatton (changes requested):

src/vdaf/prio3.rs
tgeoghegan: In the hits case, we have `PhantomData<I>`, where `I: Idpf<2, 2>`, and trait `Idpf` requires `Debug`. The `A` type parameter on `Prio3` has no trait bound at all, not even Debug. I have to admit, I don't fully understand this, because [`PhantomData<T>` should be Debug for all `T`](https://doc.rust-lang.org/std/marker/struct.PhantomData.html#impl-Debug), but if I don't do this, the compiler is mad.

cjpatton: What about
```rust
pub struct Prio3<T: Type, A: Debug> {
   ...
}
```
In any case, we could solve this by taking https://github.com/abetterinternet/libprio-rs/pull/135#discussion_r746094366

src/vdaf/prio3.rs
```diff
@@ -109,7 +113,7 @@ impl<F: FieldElement> TryFrom<Vec<F>> for Prio3Result<u64> {
 
 /// Aggregate result for vector data types.
 #[derive(PartialEq, Eq)]
-pub struct Prio3ResultVec<T: Eq>(Vec<T>);
+pub struct Prio3ResultVec<T: Eq>(pub Vec<T>);
```
cjpatton, line 116: Instead of making the anonymous field `pub`, can we make this type "look" like a vector? Like, is there a Rusty way to overload `res[i]` and `res.len()`, etc.?

src/vdaf.rs
```diff
@@ -193,27 +203,52 @@ pub enum PrepareTransition<S, M, O> {
...
     fn merge(&mut self, agg_share: &Self) -> Result<(), VdafError>;
+
+    /// Update an aggregate share by merging it with an output share
+    fn accumulate(&mut self, output_share: &Self::OutputShare) -> Result<(), VdafError>;
 }
 
-impl<F: FieldElement> Aggregatable for Vec<F> {
-    fn merge(&mut self, agg_share: &Vec<F>) -> Result<(), VdafError> {
-        if self.len() != agg_share.len() {
+/// An output share suitable for VDAFs whose output shares are a vector of field
```
cjpatton, line 217: `F` doesn't have the `FieldElement` trait bound, so in theory this type is more general. I remember you saying that implementations should have the bound and not the type, so I'm not asking for a change to the type. However, it may be more valuable to make the comment more general, and maybe name the type accordingly.

What about

```rust
/// An output share comprised of a vector of `F` elements.
type OutputShareVec<F>(pub vec<F>);
```

Similarly for `AggregateShare` below.

cjpatton (changes requested):
A common pattern in this PR Is the following:

```rust
type Wrapper<F>(pub Vec<F>);
```
The reason we have the wrapper type is that we don't want users to be confused about how the data is supposed to be used. I think this make sense. However, the need to make the anonymous tuple field `pub` suggests that users actually need direct access to the underlying vector. I wonder if there's a cleaner way to express this.

What about implementing `From<Wrapper<F>> for Vec<F>` for each `Wrapper`? We could write a macro for this, and call it for each type we do this for.

src/vdaf.rs
tgeoghegan: This could be a case where it's appropriate to put the trait bound in the struct, since it informs what the struct contains and not just its behavior. Being explicit about `FieldElement` here would make this more readable.

tgeoghegan: I spoke too soon! This is exactly the case where we don't want to specify `F: FieldElement` because then we can't derive `Deserialize`. I'll amend the comment as you suggested.

src/vdaf.rs
```diff
@@ -192,23 +187,98 @@ pub enum PrepareTransition<S, M, O> {
...
+
+    /// Aggregates a sequence of output shares into a new Aggregatable
+    fn aggregate<M: IntoIterator<Item = Self::OutputShare>>(
+        output_shares: M,
+    ) -> Result<Self, VdafError>;
+
     /// Update an aggregate share by merging it with another (`agg_share`).
     fn merge(&mut self, agg_share: &Self) -> Result<(), VdafError>;
+
+    /// Update an aggregate share by merging it with an output share
```
cjpatton, line 204: nit (aligns the language with the comment above)
```suggestion
    /// Update an aggregate share by adding `output_share`.
```

src/vdaf/prio3.rs
tgeoghegan: I can no longer recall why I ever needed that to be `pub`, but it looks like I don't anymore, so I undid this change.

cjpatton (approved):

PR #137, opened 2021-11-12
prio3: Add a type for a vector of counters

cjpatton:
Adds the `Prio3CountVec64` type, which has comparable functionality to "Prio v2". Its main advantage is that the proof size is significantly smaller. The cost of this reduction in proof size that proof generation is somewhat more expensive. (See benchmarks in #37 for some rough numbers.)

While at it, this fixes the input-share size computation in the benchmarks.

### DIFF
--- a/src/pcp/gadgets.rs
+++ b/src/pcp/gadgets.rs
@@ -7,14 +7,20 @@ use crate::field::FieldElement;
 use crate::pcp::{Gadget, PcpError};
 use crate::polynomial::{poly_deg, poly_eval, poly_mul};
 
+#[cfg(feature = "multithreaded")]
+use rayon::prelude::*;
+
 use std::any::Any;
 use std::convert::TryFrom;
+use std::fmt::Debug;
+use std::marker::PhantomData;
 
 /// For input polynomials larger than or equal to this threshold, gadgets will use FFT for
 /// polynomial multiplication. Otherwise, the gadget uses direct multiplication.
 const FFT_THRESHOLD: usize = 60;
 
 /// An arity-2 gadget that multiples its inputs.
+#[derive(Clone, Debug, Eq, PartialEq)]
 pub struct Mul<F: FieldElement> {
     /// Size of buffer for FFT operations.
     n: usize,
@@ -104,6 +110,7 @@ impl<F: FieldElement> Gadget<F> for Mul<F> {
 /// An arity-1 gadget that evaluates its input on some polynomial.
 //
 // TODO Make `poly` an array of length determined by a const generic.
+#[derive(Clone, Debug, Eq, PartialEq)]
 pub struct PolyEval<F: FieldElement> {
     poly: Vec<F>,
     /// Size of buffer for FFT operations.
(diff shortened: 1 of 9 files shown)


cjpatton:
Rebased.

tgeoghegan:
I have to do more reading about blind polynomials and refresh myself on the relevant sections of the PCP paper, but I won't get to that until I'm back from vacation. Until then, I have a meta thought: this change introduces a dependency on `rayon` for parallel iteration. I like the iterator-based abstraction that `rayon` provides, but I wonder if it's appropriate for a library like `prio` to impose a threading model on an application. It might be quite surprising to some callers that an innocuous looking library call spins up <# logical cores> threads to do work.

Further, I wonder if problems could arise if you have a `tokio::runtime::Runtime` (which will be the case if the application linking `prio` is an HTTP server using something like `hyper` or `warp` to handle many incoming requests in parallel) and a `rayon` thread pool in the same process, both trying to efficiently schedule work onto a bunch of OS threads.

`rayon`'s parallel iterator abstraction seems appropriate to me, but I think we should look into what the recommended best practices are here. Maybe there is way for a `rayon` iterator to produce a bunch of `Future`s that crate `prio` could hand off to the application to be scheduled using whatever async executor it has created?

cjpatton:
> `rayon`'s parallel iterator abstraction seems appropriate to me, but I think we should look into what the recommended best practices are here. Maybe there is way for a `rayon` iterator to produce a bunch of `Future`s that crate `prio` could hand off to the application to be scheduled using whatever async executor it has created?

That's a great question, I'm not sure what the answer is. In any case, I'd be happy to take it out of this PR. IIUC `rayon` speeds things up by about of factor of `2` for large inputs. If this speed-up is useful we can always add it back later.

Let me give you a quick overview of that math. What we're doing here is similar to the `Sum` or `Histogram` type: For both types we have a vector of field elements and we want to check that each is equal to `0` or `1`. The validity circuit for this is
```
 r * p(input[0]) + (r^2) * p(input[1]) + ... + (r^n) * p(input[n-1])
```
where `r` is the joint randomness and `p(.)` is the "range check" polynomial for `[0, 2)`. This idea comes from [[BBCG+19], Section 5.2]. Basically, if `input` is a vector of `0`s and `1`s, then this circuit will always evaluate to 0. Otherwise, it evaluates to something other than 0 with high probability.

For the `Sum` and `Histogram` types, the gadget `G(.)` is equal to `p(.)`. So we rewrite this as
```
 r * G(input[0]) + (r^2) * p(input[1]) + ... + (r^n) * G(input[n-1])
```
This generates a proof of length `O(n)`. [[BBCG+19], Section 4.4] observes that we can reduce this to `O(sqrt(n))` for "parallel sum circuits", i.e., circuits of the form
```
H(input[0]) + H(input[1]) + ... + H(input[n-1])
```
where `H(.)` is a gadget. What we do is construct a "wrapper" gadget `P(.)` that simply computes `H(chunk[0]) + ... + H(chunk[c-1])` where `chunk` is a non-overlapping chunk of `input` of length `n/c`. Setting `n/c = sqrt(n)` is the optimal choice in terms of the proof size.

Putting this all together, what we for `CountVec` is construct each `chunk[i]` to be equal to `[r^i, input[i]]` and let `H(chunk) = chunk[0] * p(chunk[1])`.

[BBCG+19]: https://eprint.iacr.org/2019/188.pdf

tgeoghegan:
I've been chewing on the `rayon` question this morning. I'm currently leaning towards gating the use of `rayon` behind a cargo feature, so that clients can opt into using a threadpool but will get a conventional, serial iterator by default. I'm also interested in ways to make `rayon` and `tokio` (or other async runtimes) interoperate nicely (i.e., we want `rayon`'s parallel iterator abstraction, but have it schedule work into whatever ambient async runtime exists instead of its own thread pool). However the latter can be done in a later PR.

cjpatton:
I'd be good with that outcome. If you'd like we can rip out the `rayon` dependency from this PR, we can always add it back later as a feature.

Eventually we're going to want some form thread level parallelism for big inputs. I picked `rayon` because it seemed to solve my use-case out-of-the-box. There may be better ways.

tgeoghegan:
No, let's keep rayon. It should be really easy to add the crate feature I'm thinking of. If you'd like, I can tack a commit onto this PR when I get back around to reviewing it (sorry, I am still catching up from vacation).

tgeoghegan:
As discussed, I gated the `rayon` dependency behind a feature. I think I can do a bit better here and reduce the code duplication between `SerialSum` and `ParallelSum`.

tgeoghegan:
OK, I like this version much better. We now define a trait `SumGadget` so we can use generics to abstract over `SerialSum` and `ParallelSum`, and module `prio3` provides type aliases `Prio3CountVec64` and `Prio3CountVec64Parallel` to abstract that whole mess away from the caller.

src/pcp/gadgets.rs
```diff
@@ -214,7 +221,271 @@ impl<F: FieldElement> Gadget<F> for PolyEval<F> {
...
+/// A wrapper gadget that applies the inner gadget to chunks of input and returns the sum of the
+/// outputs. The arity is equal to the arity of the inner gadget times the number of chunks.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct SerialSum<F: FieldElement, G: Gadget<F>> {
+    inner: G,
+    chunks: usize,
+    phantom: PhantomData<F>,
+}
+
+impl<F: FieldElement, G: 'static + Gadget<F>> SumGadget<F, G> for SerialSum<F, G> {
```
cjpatton, line 352: Is `'static` required?

cjpatton (commented):
Really nice work, just a few inline comments and one high level one.

Here's what the benchmarks show for sharding time for various input lengths:

| length | serial shard time | parallel shard time |
| --- | ---- | ---- |
| 10,000 | 17.5ms | 7.5ms |
| 1,000 | 2.3ms | 1.0ms |
| 500 | 618.7us | 360.0us |
| 250 | 411.5us | 229.1us |
| 100 | 99.5us | 103.8us |
| 50 | 66.9us | 69.7us |
| 10 | 19.6us | 42.5us |

Thread-level parallelization is definitely buying us some time for very large inputs, but the overhead stops paying off when we get down to about 100. (The exact threshold is likely to be dependent on the environment.) I think it would make sense to add a warning in the documentation that `SerialSum` should be preferred for smaller inputs.

Incidentally, based on a conversation with @armfazh several months ago (he's an engineer on my team at Cloudflare), it's likely that a SIMD-style approach is going to get us a lot of parallelism without the overhead of spinning up threads.

src/pcp/gadgets.rs
tgeoghegan: It is, because of the method `fn as_any(&mut self) -> &mut dyn Any` on trait `Gadget`. That method hands out a mutable reference to the `Gadget`, and so absent any lifetime information, the compiler needs to be sure that the reference is valid independent of `self`. I think maybe we can modify the definition of `trait Gadget` to include a lifetime and then write `fn as_any(&'a mut self) -> &'a mut dyn Any`, meaning "the reference returned from `as_any` has the same lifetime as the gadget". But that would touch every existing `Gadget` impl so _if_ it works (I'm not sure yet) I want to do it in a later PR to avoid complicating this one, so let's stick with `'static` for now.

src/pcp/gadgets.rs
```diff
@@ -214,7 +221,271 @@ impl<F: FieldElement> Gadget<F> for PolyEval<F> {
...
+        Ok(outp)
+    }
+
+    fn call_poly(&mut self, outp: &mut [F], inp: &[Vec<F>]) -> Result<(), PcpError> {
+        gadget_call_poly_check(self, outp, inp)?;
+
+        let res = inp
+            .chunks(self.inner.arity())
+            .map(|chunk| {
+                let mut partial_outp = vec![F::zero(); outp.len()];
```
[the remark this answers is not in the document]
tgeoghegan: Per discussion I rewrote this as a `for` loop instead of using `.map().reduce()`, which lets us allocate `partial_outp` once and avoid awkward closure capture rules.

cjpatton:
LGTM, thanks for your help @tgeoghegan !

src/pcp/gadgets.rs
```diff
@@ -214,7 +221,265 @@ impl<F: FieldElement> Gadget<F> for PolyEval<F> {
...
+    }
+
+    fn as_any(&mut self) -> &mut dyn Any {
+        self
+    }
+}
+
+/// A wrapper gadget that applies the inner gadget to chunks of input and returns the sum of the
+/// outputs. The arity is equal to the arity of the inner gadget times the number of chunks. The sum
+/// is evaluated in parallel.
```
tgeoghegan, line 410:
```suggestion
/// outputs. The arity is equal to the arity of the inner gadget times the number of chunks. The sum
/// evaluation is multithreaded.
```

src/vdaf/prio3.rs
```diff
@@ -35,7 +38,39 @@ impl Prio3Count64 {
...
+    }
+}
+
+/// The count-vector type. Each measurement is a vector of integers in `[0,2)` and the aggregate is
+/// the element-wise sum.
+pub type Prio3CountVec64 =
+    Prio3<CountVec<Field96, ParallelSum<Field96, BlindPolyEval<Field96>>>, Prio3ResultVec<u64>>;
+
+/// The count-vector type. Each measurement is a vector of integers in `[0,2)` and the aggregate is
+/// the element-wise sum. Polynomial evaluations are parallelized.
```
tgeoghegan, line 53:
```suggestion
/// the element-wise sum. Polynomial evaluations are multithreaded.
```

tgeoghegan (approved):
LGTM with a few nits

src/pcp/gadgets.rs
```diff
@@ -214,7 +221,265 @@ impl<F: FieldElement> Gadget<F> for PolyEval<F> {
...
+#[cfg(feature = "multithreaded")]
```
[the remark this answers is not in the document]
cjpatton: I believe the only place to add this is `ParallelSumMultithreaded`. I did so, however I don't what effect this has on the documentation (try running `cargo doc --all-features`.)

cjpatton:
Squashed.

PR #147, opened 2021-12-28
vdaf: Add public preprocessing of prepare messages

cjpatton:
Aligns the API with https://github.com/cjpatton/vdaf/pull/14.

### DIFF
--- a/src/vdaf/prio3.rs
+++ b/src/vdaf/prio3.rs
@@ -473,87 +473,107 @@ impl<T: Type, A: Debug> Aggregator for Prio3<T, A> {
             input_share: msg.input_share.clone(),
             joint_rand_seed,
             verifier_msg: Prio3PrepareMessage {
-                verifier_share,
-                joint_rand_seed_share,
+                verifier: verifier_share,
+                joint_rand_seed: joint_rand_seed_share,
             },
+        })
+    }
+
+    fn prepare_preprocess<M: IntoIterator<Item = Prio3PrepareMessage<T::Field>>>(
+        &self,
+        inputs: M,
+    ) -> Result<Self::PrepareMessage, VdafError> {
+        let mut verifier = vec![T::Field::zero(); self.typ.verifier_len()];
+        let mut joint_rand_seed = Key::uninitialized(self.suite);
+        let mut count = 0;
+        for share in inputs.into_iter() {
+            count += 1;
+
+            if share.verifier.len() != verifier.len() {
+                return Err(VdafError::Uncategorized(format!(
+                    "unexpected verifier share length: got {}; want {}",
+                    share.verifier.len(),
+                    verifier.len(),
+                )));
+            }
+
+            if share.joint_rand_seed.suite() != self.suite {
+                return Err(VdafError::Uncategorized(format!(
+                    "unexpected suite for joint randomness seed share: got {:?}; want {:?}",
+                    share.joint_rand_seed.suite(),
+                    self.suite,
+                )));
+            }
+
+            for (x, y) in verifier.iter_mut().zip(share.verifier) {
+                *x += y;
+            }
+
+            for (x, y) in joint_rand_seed
+                .as_mut_slice()
+                .iter_mut()
+                .zip(share.joint_rand_seed.as_slice())
+            {
+                *x ^= y;
+            }
+        }
+
+        if count != self.num_aggregators {
+            return Err(VdafError::Uncategorized(format!(
+                "unexpected message count: got {}; want {}",
+                count, self.num_aggregators,
+            )));
+        }
+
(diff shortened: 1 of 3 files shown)


tgeoghegan:
Could you please rebase this? I don't think commit 4406001413075df2e38dc6d0a87a97bad1e57a09 should be in there.

src/vdaf/hits.rs
```diff
@@ -437,92 +437,87 @@ impl<I: Idpf<2, 2>> Aggregator for Hits<I> {
...
+    }
+
     // TODO Fix this clippy warning instead of bypassing it.
     #[allow(clippy::type_complexity)]
-    fn prepare_step<M: IntoIterator<Item = HitsPrepareMessage<I::Field>>>(
+    fn prepare_step(
         &self,
         mut state: HitsPrepareStep<I::Field>,
-        inputs: M,
+        input: Option<HitsPrepareMessage<I::Field>>,
```
tgeoghegan, line 480: I think we should be consistent about using either `HitsPrepareMessage` or `Self::PrepareMessage`. I lean toward `HitsPrepareMessage`.

src/vdaf/hits.rs
```diff
@@ -437,92 +437,87 @@ impl<I: Idpf<2, 2>> Aggregator for Hits<I> {
         })
     }
 
+    fn prepare_preprocess<M: IntoIterator<Item = Self::PrepareMessage>>(
```
tgeoghegan, line 440: I think we should be consistent about using either `HitsPrepareMessage` or `Self::PrepareMessage`. I lean toward `HitsPrepareMessage` because it makes it easier to jump to the concrete type definition.

src/vdaf/hits.rs
```diff
@@ -850,14 +848,16 @@ mod tests {
         let mut round2: Vec<HitsPrepareMessage<I::Field>> = Vec::with_capacity(2);
         let mut state2: Vec<HitsPrepareStep<I::Field>> = Vec::with_capacity(2);
         for state in state1.into_iter() {
-            let (state, msg) = hits.prepare_next(state, round1.clone())?;
+            let (state, msg) =
+                hits.prepare_next(state, hits.prepare_preprocess(round1.clone())?)?;
```
tgeoghegan, line 852: Is the `clone()` still needed or would a mutable reference to `round1` suffice?

cjpatton:
Rebased.

src/vdaf/prio3.rs
```diff
@@ -473,87 +473,107 @@ impl<T: Type, A: Debug> Aggregator for Prio3<T, A> {
...
             verifier_msg: Prio3PrepareMessage {
-                verifier_share,
-                joint_rand_seed_share,
+                verifier: verifier_share,
+                joint_rand_seed: joint_rand_seed_share,
             },
         })
     }
 
+    fn prepare_preprocess<M: IntoIterator<Item = Self::PrepareMessage>>(
```
tgeoghegan, line 482:
```suggestion
    fn prepare_preprocess<M: IntoIterator<Item = Prio3PrepareMessage>>(
```

src/vdaf/hits.rs
cjpatton: Done

src/vdaf/hits.rs
cjpatton: Done

src/vdaf/hits.rs
cjpatton: No because `prepare_preprocess` takes in an `IntoIterator`.

src/vdaf/prio3.rs
cjpatton: Done

src/vdaf/hits.rs
```diff
@@ -437,92 +437,87 @@ impl<I: Idpf<2, 2>> Aggregator for Hits<I> {
         })
     }
 
+    fn prepare_preprocess<M: IntoIterator<Item = Self::PrepareMessage>>(
+        &self,
+        inputs: M,
+    ) -> Result<Self::PrepareMessage, VdafError> {
```
tgeoghegan, line 443:
```suggestion
    fn prepare_preprocess<M: IntoIterator<Item = HitsPrepareMessage>>(
        &self,
        inputs: M,
    ) -> Result<HitsPrepareMessage, VdafError> {
```
...and then this is good to go

tgeoghegan (changes requested):

tgeoghegan:
Rebased on `main`, resolved conflicts introduced by the `Hits` -> `poplar1` rename, and addressed my last comment. I think this is OK to merge. @cjpatton can you review?

tgeoghegan (approved):

cjpatton:
Squashed.

PR #150, opened 2022-01-04
vdaf: Rename `hits` to `poplar1`

cjpatton:
The authors of [BBCGGI21] have given their protocol a proper name. Let's
update rename the hits module to match.

Note: We should wait to merge this until the VDAF spec is changed: https://github.com/cjpatton/vdaf/pull/16


tgeoghegan (approved):
GitHub won't let me leave a comment on the relevant line, but [this comment](https://github.com/abetterinternet/libprio-rs/blob/10e4d168c04546256916d940cde8b4ba9ac432a5/src/vdaf.rs#L240) should be updated to mention Poplar instead of Hits. That aside, LGTM so go ahead and merge when ready.

PR #160, opened 2022-01-25
make value inside `Prio3Result[Vec]` public

tgeoghegan:
Make the `T` or `Vec<T>` wrapped by `Prio3Result<T>` or
`Prio3ResultVec<T>` public so that clients of the crate can access the
value.

### DIFF
--- a/src/vdaf/prio3.rs
+++ b/src/vdaf/prio3.rs
@@ -126,7 +126,7 @@ impl Prio3Histogram64 {
 
 /// Aggregate result for singleton data types.
 #[derive(Clone, Debug, PartialEq, Eq)]
-pub struct Prio3Result<T: Eq>(T);
+pub struct Prio3Result<T: Eq>(pub T);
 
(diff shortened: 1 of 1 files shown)


tgeoghegan:
I encountered this while bringing ppm-prototype up to date with the latest `libprio`. We could also put methods on `Prio3Result` and `Prio3ResultVec` to get the wrapped value if we don't want to commit to the structure of these types in the crate API.

cjpatton (approved):
I came across the same issue. Looks good, but one question while we're at it: It may be worth doing away with the `Prio3Result` type and just treat it as a special case of `Prio3ResultVec`, i.e., `Prio3Result` is a lengh-1 `Prio3ResultVec`.

tgeoghegan:
I don't like forcing the (mild) friction of taking the 0-th element of a vector onto users of a type like `Prio3Count64`. Plus, making everything `Prio3ResultVec` would force heap allocations everywhere for the inner `Vec`, whereas `Prio3Result<u64>` can be instantiated on the stack. In any case, I think we'll learn more about the right ergonomics for these types as we add more and see how they get used.

PR #161, opened 2022-01-28
prio3: Add serialization for network messages

cjpatton:
Adds methods for encoding and decoding `prio3` input shares and preparation messages.

The methods are crate-private for now. While working on this I realized that we can simplify the API by having the client produce encoded input shares and having the aggregators produce encoded prepare messages. It's not necessary to expose types for these, as there will never be an application that needs to do anything but treat them as opaque strings. With this change, there would be no need to expose this serialization logic.

Anywho, food for thought. If we decide to keep the API as-is, then I'll make methods public. If we decide to simplify the API, I'd be happy to send a PR.

### DIFF
--- a/src/vdaf.rs
+++ b/src/vdaf.rs
@@ -8,13 +8,12 @@
 //! [BBCG+21]: https://ia.cr/2021/017
 //! [VDAF]: https://datatracker.ietf.org/doc/draft-patton-cfrg-vdaf/
 
-use crate::field::FieldElement;
+use crate::field::{FieldElement, FieldError};
 use crate::pcp::PcpError;
 use crate::prng::PrngError;
-use crate::vdaf::suite::{Key, SuiteError};
-use serde::de::DeserializeOwned;
-use serde::{Deserialize, Serialize};
+use crate::vdaf::suite::{Key, Suite, SuiteError};
 use std::fmt::Debug;
+use std::io::Read;
 
 /// Errors emitted by this module.
 #[derive(Debug, thiserror::Error)]
@@ -22,6 +21,14 @@ pub enum VdafError {
     /// An error occurred.
     #[error("vdaf error: {0}")]
     Uncategorized(String),
+
+    /// Field error.
+    #[error("field error: {0}")]
+    Field(#[from] FieldError),
+
+    /// An error occured while parsing a message.
+    #[error("io error: {0}")]
+    IoError(#[from] std::io::Error),
 
     /// PCP error.
     #[error("pcp error: {0}")]
@@ -37,13 +44,29 @@ pub enum VdafError {
 }
 
 /// An additive share of a vector of field elements.
-#[derive(Clone, Debug, Serialize, Deserialize)]
+#[derive(Clone, Debug, Eq, PartialEq)]
 pub enum Share<F> {
     /// An uncompressed share, typically sent to the leader.
     Leader(Vec<F>),
 
     /// A compressed share, typically sent to the helper.
     Helper(Key),
+}
+
+impl<F: FieldElement> Share<F> {
+    pub(crate) fn read_leader<R: Read>(length: usize, reader: &mut R) -> Result<Self, VdafError> {
+        let mut data = Vec::with_capacity(length);
+        for _ in 0..length {
+            data.push(F::try_from_reader(reader)?);
+        }
+        Ok(Share::Leader(data))
+    }
+
+    pub(crate) fn read_helper<R: Read>(suite: Suite, reader: &mut R) -> Result<Self, VdafError> {
+        let mut seed = Key::uninitialized(suite);
+        reader.read_exact(seed.as_mut_slice())?;
+        Ok(Share::Helper(seed))
+    }
 }
 
 /// The base trait for VDAF schemes. This trait is inherited by traits [`Client`], [`Aggregator`],
@@ -68,10 +91,10 @@ pub trait Vdaf: Clone + Debug {
     type VerifyParam;
 
     /// An input share sent by a Client.
-    type InputShare: Clone + Debug + Serialize + DeserializeOwned;
+    type InputShare: Clone + Debug;
 
     /// An output share recovered from an input share by an Aggregator.
-    type OutputShare: Clone + Debug + Serialize + DeserializeOwned;
+    type OutputShare: Clone + Debug;
 
     /// An Aggregator's share of the aggregate result.
     type AggregateShare: Aggregatable<OutputShare = Self::OutputShare>;
@@ -100,7 +123,7 @@ pub trait Aggregator: Vdaf {
     type PrepareStep: Clone + Debug;
 
     /// The type of messages exchanged among the Aggregators during the Prepare process.
-    type PrepareMessage: Clone + Debug + Serialize + DeserializeOwned;
+    type PrepareMessage: Clone + Debug;
 
     /// Begins the Prepare process with the other Aggregators. The result of this process is
     /// the Aggregator's output share.
@@ -202,7 +225,7 @@ pub enum PrepareTransition<S, M, O> {
 
 /// An aggregate share resulting from aggregating output shares together that
 /// can merged with aggregate shares of the same type.
-pub trait Aggregatable: Clone + Debug + Serialize + DeserializeOwned {
+pub trait Aggregatable: Clone + Debug {
     /// Type of output shares that can be accumulated into an aggregate share.
     type OutputShare;
 
@@ -214,7 +237,7 @@ pub trait Aggregatable: Clone + Debug + Serialize + DeserializeOwned {
 }
 
 /// An output share comprised of a vector of `F` elements.
-#[derive(Clone, Debug, Serialize, Deserialize)]
+#[derive(Clone, Debug)]
 pub struct OutputShare<F>(Vec<F>);
 
 impl<F> AsRef<[F]> for OutputShare<F> {
@@ -226,7 +249,7 @@ impl<F> AsRef<[F]> for OutputShare<F> {
 /// An aggregate share suitable for VDAFs whose output shares and aggregate
 /// shares are vectors of `F` elements, and an output share needs no special
 /// transformation to be merged into an aggregate share.
-#[derive(Clone, Debug, Serialize, Deserialize)]
+#[derive(Clone, Debug)]
 pub struct AggregateShare<F>(Vec<F>);
 
 impl<F> AsRef<[F]> for AggregateShare<F> {
--- a/src/vdaf/prio3.rs
+++ b/src/vdaf/prio3.rs
@@ -21,9 +21,9 @@ use crate::vdaf::{
     Aggregatable, AggregateShare, Aggregator, Client, Collector, OutputShare, PrepareTransition,
     Share, Vdaf, VdafError,
 };
-use serde::{Deserialize, Serialize};
 use std::convert::{TryFrom, TryInto};
 use std::fmt::Debug;
+use std::io::{Read, Write};
 use std::iter::IntoIterator;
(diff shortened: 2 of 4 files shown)


src/vdaf.rs
```diff
@@ -37,7 +38,7 @@ pub enum VdafError {
 }
 
 /// An additive share of a vector of field elements.
-#[derive(Clone, Debug, Serialize, Deserialize)]
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
```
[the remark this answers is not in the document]
tgeoghegan: Do you think we should take `Serialize, Deserialize` off of this and some other structs in the crate? On the one hand, we want clients to use the custom codec methods you've added, but on the other hand, it could be useful to store these messages in arbitrary formats. For instance we already store Priov2 test vectors as JSON, but we might be able to restrict the `serde` dependency to just the test targets, which could make the production config of the crate quite a bit smaller.

src/vdaf/prio3.rs
```diff
@@ -242,7 +243,7 @@ pub struct Prio3VerifyParam {
 
 /// The message sent by the client to each aggregator. This includes the client's input share and
 /// the initial message of the input-validation protocol.
-#[derive(Clone, Debug, Serialize, Deserialize)]
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
```
tgeoghegan, line 246:
```suggestion
#[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
```

src/vdaf/prio3.rs
```diff
@@ -252,14 +253,131 @@ pub struct Prio3InputShare<F> {
...
+    // randomness seed at all. See https://github.com/cjpatton/vdaf/issues/15.
     pub joint_rand_seed_hint: Key,
 
     /// The blinding factor, used to derive the aggregator's joint randomness seed share.
     pub blind: Key,
 }
 
+#[allow(dead_code)]
+impl<F: FieldElement> Prio3InputShare<F> {
+    pub(crate) fn append(&self, bytes: &mut Vec<u8>) {
```
tgeoghegan, line 266: Could this be `append<W>(&self, writer: &mut W) -> Result<(), VdafError>`? [`Vec<u8>` implements `Write`](https://doc.rust-lang.org/std/vec/struct.Vec.html#impl-Write) so it shouldn't be any harder to use for a caller that wants to append to a `Vec` while being more general (maybe I want to blast my `Prio3InputShare` directly into a socket instead of a heap buffer).

src/vdaf/prio3.rs
```diff
@@ -252,14 +253,131 @@ pub struct Prio3InputShare<F> {
...
+        let (input_share, proof_share) = match is_leader {
+            true => (
+                Share::read_leader(vdaf.typ.input_len(), reader)?,
+                Share::read_leader(vdaf.typ.proof_len(), reader)?,
+            ),
+            _ => (
+                Share::read_helper(vdaf.suite, reader)?,
+                Share::read_helper(vdaf.suite, reader)?,
+            ),
+        };
```
tgeoghegan, line 319:
```suggestion
        let (input_share, proof_share) = if is_leader {
            (
                Share::read_leader(vdaf.typ.input_len(), reader)?,
                Share::read_leader(vdaf.typ.proof_len(), reader)?,
            )
        } else {
            (
                Share::read_helper(vdaf.suite, reader)?,
                Share::read_helper(vdaf.suite, reader)?,
            ),
        };
```

src/vdaf/prio3.rs
```diff
@@ -252,14 +253,131 @@ pub struct Prio3InputShare<F> {
...
+                bytes.extend_from_slice(input_share_seed.as_slice());
+                bytes.extend_from_slice(proof_share_seed.as_slice());
+            }
+            _ => panic!("tried to encode input share with ambiguous encoding"),
+        };
+        bytes.extend_from_slice(self.joint_rand_seed_hint.as_slice());
+        bytes.extend_from_slice(self.blind.as_slice());
+    }
+
+    pub(crate) fn read_leader<R, A, T>(vdaf: &Prio3<T, A>, reader: &mut R) -> Option<Self>
```
tgeoghegan, line 286: This, `read_helper`, and `read` should all return `Result<Self, VdafError>`.

src/vdaf/prio3.rs
```diff
@@ -252,14 +253,131 @@ pub struct Prio3InputShare<F> {
...
+#[allow(dead_code)]
+impl<F: FieldElement> Prio3PrepareMessage<F> {
+    pub(crate) fn append(&self, bytes: &mut Vec<u8>) {
+        for x in self.verifier.iter() {
+            bytes.extend((*x).into());
+        }
+        bytes.extend_from_slice(self.joint_rand_seed.as_slice());
+    }
+
+    pub(crate) fn read<R, A, T>(vdaf: &Prio3<T, A>, reader: &mut R) -> Option<Self>
```
tgeoghegan, line 359: This should return `Result<Self, VdafError>`.

src/vdaf/prio3.rs
```diff
@@ -252,14 +253,131 @@ pub struct Prio3InputShare<F> {
...
+    /// (A share of) the joint randomness seed.
+    //
+    // TODO(cjpatton) If `input.joint_rand_len() == 0`, then we don't need to bother with the joint
+    // randomness seed at all. See https://github.com/cjpatton/vdaf/issues/15.
+    pub joint_rand_seed: Key,
+}
+
+#[allow(dead_code)]
+impl<F: FieldElement> Prio3PrepareMessage<F> {
+    pub(crate) fn append(&self, bytes: &mut Vec<u8>) {
```
tgeoghegan, line 352: As in the `Prio3InputShare` case, I think this should be `append<W: Write>(&self, &mut W) -> Result<(), VdafError> {}`

tgeoghegan (changes requested):
re: making this public API: we're going to have to iterate on this anyway, because I suspect these methods will have to be expressed generically over VDAFs, so that a server implementation can generically encode/decode messages for `prio3` or `poplar1`. I can take a stab at that as I continue working on bringing ppm-prototype up to date.

cjpatton:
> re: making this public API: we're going to have to iterate on this anyway, because I suspect these methods will have to be expressed generically over VDAFs, so that a server implementation can generically encode/decode messages for `prio3` or `poplar1`. I can take a stab at that as I continue working on bringing ppm-prototype up to date.

One option is to just handle the messages as opaque byte strings: see https://github.com/abetterinternet/libprio-rs/pull/162.

src/vdaf.rs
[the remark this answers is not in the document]
cjpatton: I don't see any reason to allow custom serialization of the messages. Test vectors for VDAFs will just be the messages encoded in hex, e.g.: https://www.ietf.org/archive/id/draft-irtf-cfrg-hpke-12.html#name-test-vectors

src/vdaf/prio3.rs
cjpatton: Yup!

src/vdaf.rs
```diff
@@ -46,6 +47,22 @@ pub enum Share<F> {
...
 }
 
+impl<F: FieldElement> Share<F> {
+    pub(crate) fn read_leader<R: Read>(length: usize, reader: &mut R) -> Option<Self> {
+        let mut data = Vec::with_capacity(length);
+        for _ in 0..length {
+            data.push(F::try_from_reader(reader).ok()?);
+        }
+        Some(Share::Leader(data))
+    }
```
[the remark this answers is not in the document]
tgeoghegan: Oh, I see the problem: there's a difference between passing in a `&[u8]` with a known size and passing in a `Read` implementation that may yield arbitrarily many bytes. OK, that's subtle enough that I don't think we should make One Method To Rule Them All.

src/vdaf.rs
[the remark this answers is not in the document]
tgeoghegan: Yes, you need `PartialEq` to implement `Eq`. I'm saying we should derive both.

Let's remove `Serialize, Deserialize`. I agree that there's no need for custom serialization.

src/vdaf.rs
```diff
@@ -46,6 +51,24 @@ pub enum Share<F> {
...
+            data.push(F::try_from_reader(reader).map_err(|_| VdafError::Parse)?);
+        }
+        Ok(Share::Leader(data))
+    }
+
+    pub(crate) fn read_helper<R: Read>(suite: Suite, reader: &mut R) -> Result<Self, VdafError> {
+        let mut seed = Key::uninitialized(suite);
+        reader
+            .read_exact(seed.as_mut_slice())
+            .map_err(|_| VdafError::Parse)?;
```
tgeoghegan, line 67: This throws away the underlying error. If you don't want to add a variant to `VdafError` to wrap each of `FieldError, `io::Error` and so on, you could put in a variant like:
```
#[error("Unspecified error {0}")]
Unspecified(#[from] Box<dyn Error>)
```

Then we don't get structure, but at least if something ever logs the error, the underlying error's description gets printed.

src/vdaf.rs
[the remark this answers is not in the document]
cjpatton: Done and done!

src/vdaf.rs
cjpatton: Ugh, I meant to do this but forgot! Done now.

src/vdaf/prio3.rs
```diff
@@ -252,14 +253,138 @@ pub struct Prio3InputShare<F> {
...
+        Ok(Prio3InputShare {
+            input_share,
+            proof_share,
+            joint_rand_seed_hint,
+            blind,
+        })
+    }
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
```
tgeoghegan, line 343:
```suggestion
#[derive(Clone, Debug)]
```

src/vdaf.rs
```diff
@@ -68,7 +93,7 @@ pub trait Vdaf: Clone + Debug {
     type VerifyParam;
 
     /// An input share sent by a Client.
-    type InputShare: Clone + Debug + Serialize + DeserializeOwned;
+    type InputShare: Clone + Debug;
 
     /// An output share recovered from an input share by an Aggregator.
     type OutputShare: Clone + Debug + Serialize + DeserializeOwned;
```
tgeoghegan, line 99: I think we should also remove the `Serialize + Deserialize` trait bound here, remove the derived traits from `struct OutputShare`, and provide codec routines. Same goes for `trait Aggregatable` and `struct AggregateShare`. I would expect neither `vdaf.rs` nor `vdaf/prio3.rs` to need to `use serde` anymore.

tgeoghegan (changes requested):

src/vdaf.rs
cjpatton: Removed the trait bounds. I'd prefer to hold off on adding serialization for this stuff until one of us actually needs it. The reason is that, in my implementation, `OutputShare` never gets serialized (though `AggregateShare` does).

tgeoghegan: `OutputShare` will get serialized in my implementation, because I run the collector separately from the leader and helper. However I can add that serialization myself later if you're very keen to get this merged.

tgeoghegan (approved):

src/vdaf.rs
cjpatton: Either way!

cjpatton:
Squashed.

src/vdaf.rs
cjpatton: I'll merge this now and we can add this later. Note that, by the end of next week, my hope is that the VDAF doc will have specified all of this stuff.

PR #162, opened 2022-01-28
WIP vdaf: Handle messages as opaque strings

cjpatton:


tgeoghegan (commented):
I don't like this approach because by defining every interface in terms of `Vec<u8>` or `&[u8]`, we reintroduce the possibility for type confusion bugs that we made impossible (or at least much much harder) to write by introducing the `InputShare`, `PrepareMessage`, etc. associated types. It also conflates the functions of executing the protocol with wire encoding of messages. So for instance a leader that also happens to be a collector is forced to deal with (de)serialization instead of just handing over a `Vdaf::AggregateShare`. Instead of changing the signatures of methods on traits `Vdaf`, `Collector`, and `Aggregator`, we should make the methods `append, read, etc.` on structs like `Prio3InputShare` public. I think we will need to define a `trait Codec` that `Prio3InputShare, Poplar1InputShare, Prio3PrepareMessage`, etc. implement so that server implementations can generically encode and decode messages. I can take a stab at this.

cjpatton:
Closing in favor of #166.

PR #165, opened 2022-02-01
Implement generic 128-bit prime field arithmetic.

armfazh:
Previous code considers only primes of at most 126 bits. Now, it adds support for 128-bit primes.

cc: @cjpatton

### DIFF
--- a/src/fp.rs
+++ b/src/fp.rs
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: MPL-2.0
 
-//! Finite field arithmetic for any field GF(p) for which p < 2^126.
+//! Finite field arithmetic for any field GF(p) for which p < 2^128.
 
 #[cfg(test)]
 use rand::{prelude::*, Rng};
@@ -10,13 +10,11 @@ use rand::{prelude::*, Rng};
 /// is the largest input size we would ever need for the cryptographic applications in this crate.
 pub(crate) const MAX_ROOTS: usize = 20;
 
-/// This structure represents the parameters of a finite field GF(p) for which p < 2^126.
+/// This structure represents the parameters of a finite field GF(p) for which p < 2^128.
 #[derive(Debug, PartialEq, Eq)]
 pub(crate) struct FieldParameters {
     /// The prime modulus `p`.
     pub p: u128,
-    /// `p * 2`.
-    pub p2: u128,
     /// `mu = -p^(-1) mod 2^64`.
     pub mu: u64,
     /// `r2 = (2^128)^2 mod p`.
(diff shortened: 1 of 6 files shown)


cjpatton (commented):
I'm not qualified to speak to the correctness of the new algorithm, however the changes to the tests in `fp.rs` look functionally equivalent, and I'm comfortable on leaning on the existing test coverage and @armfazh's expertise here. I think this change is ready for review.

A couple high level notes:
1. I think we ought to replace `Field126` with `Field128`. The former isn't used for anything yet, and the only reason we had a 126-bit field is because that was the largest field we could support.
2. The benchmarks show a performance regression resulting from the additional arithmetic, as much as 10% for some benchmarks. I think we can live with this regression as long as (a) @tgeoghegan et al find this doesn't over-burden `prio-server` (my guess is it won't, since we're talking about less than 10 microseconds additional time spent on proof verification) and (b) the regression would be addressed by optimized or fiat-generated arithmetic (@armfazh can confirm whether this is the case).

armfazh:
Current code is generic about the prime size (bounded to 128 bits) and the prime shape. 

On the one hand, we can use fiat-crypto to produce formally-verified implementation, but it requires to have a prime specified.

on the other hand, we can produce hand-crafted optimized code for primes that have a special shape. This has the advantage that code could run faster.

cjpatton (changes requested):
I have enabled CI for @armfazh. I expect the build to fail since Field126 has been removed but there are references to this elsewhere. Please replace references to Field126 with Field128.

src/fp.rs
```diff
@@ -36,16 +34,37 @@ pub(crate) struct FieldParameters {
 impl FieldParameters {
     /// Addition.
     pub fn add(&self, x: u128, y: u128) -> u128 {
-        let (z, carry) = x.wrapping_add(y).overflowing_sub(self.p2);
-        let m = 0u128.wrapping_sub(carry as u128);
-        z.wrapping_add(m & self.p2)
+        //   0,x
```
cjpatton, line 37: nit: If these inline comments about how the arithmetic works are useful to readers, then I would like to see them added wherever appropriate (in particular, to `mul`). If they're not especially useful, then let's remove them.

tgeoghegan (approved):
I'm going to trust the both of you that the math here is correct. Further, we have a backward compatibility test with `FieldPriov2` as used in ENPA, and that passed, so that gives me confidence that this change to the arithmetic doesn't break anything. Finally, changing `Field126` to `Field128` is risk-free because there's nothing deployed anywhere that used `Field126` (if there is, and you're reading this, WHAT ARE YOU DOING).

As Chris said, I'm not really concerned with the potential perf regression here because it won't meaningfully affect existing deployments of libprio which are (1) heavily I/O bound and (2) have tons of idle time anyway.

So with all that in mind, I think this change is good to go.

tgeoghegan:
And thank you @armfazh for this contribution! All of us at ISRG are very excited to be collaborating with you on this work.

PR #170, opened 2022-02-24
Introduce trait `PrepareStep`

tgeoghegan:
Adds a trait bound on the `vdaf::PrepareStep` associated type. This
allows clients of `prio` to generically determine whether they should
expect the next state transition to be terminal, and thus whether they
should call `prepare_next` or `prepare_finish`. We then implement trait
`PrepareStep` on `Prio3PrepareStep` and `Poplar1PrepareStep`.

### DIFF
--- a/src/vdaf.rs
+++ b/src/vdaf.rs
@@ -97,6 +97,14 @@ impl<F: FieldElement> Encode for Share<F> {
     }
 }
 
+/// PrepareStep is the state associated with the multi-round VDAF input preparation process.
+pub trait PrepareStep {
+    /// Returns true if the prepare process is on the last round, meaning clients should call
+    /// [`Vdaf::prepare_finish`] to recover an output share, or false otherwise, in which case
+    /// clients should call [`Vdaf::prepare_next`] to get the next prepare message share.
+    fn is_last_round(&self) -> bool;
+}
+
 /// The base trait for VDAF schemes. This trait is inherited by traits [`Client`], [`Aggregator`],
 /// and [`Collector`], which define the roles of the various parties involved in the execution of
 /// the VDAF.
@@ -148,7 +156,7 @@ pub trait Client: Vdaf {
 /// The Aggregator's role in the execution of a VDAF.
 pub trait Aggregator: Vdaf {
     /// State of the Aggregator during the Prepare process.
-    type PrepareStep: Clone + Debug;
+    type PrepareStep: Clone + Debug + PrepareStep;
 
     /// The type of messages exchanged among the Aggregators during the Prepare process.
     type PrepareMessage: Clone + Debug + Decode<Self::PrepareStep> + Encode;
--- a/src/vdaf/poplar1.rs
+++ b/src/vdaf/poplar1.rs
@@ -34,8 +34,8 @@ use crate::fp::log2;
 use crate::prng::Prng;
 use crate::vdaf::suite::{Key, KeyDeriver, KeyStream, Suite};
 use crate::vdaf::{
-    Aggregatable, AggregateShare, Aggregator, Client, Collector, OutputShare, PrepareTransition,
-    Share, ShareDecodingParameter, Vdaf, VdafError,
+    Aggregatable, AggregateShare, Aggregator, Client, Collector, OutputShare, PrepareStep,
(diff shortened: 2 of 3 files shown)


cjpatton (commented):
The implementation looks correct. However, I think there's a simpler alternative way to solve your problem.

First, let's stipulate that, if you know how many rounds there are going to be, then you'll always know whether to call `prepare_start` or `prepare_finish`. 

Thus this PR is useful only if you don't know in advance how many rounds there may be. In this case you could do something like this:
```rust
let mut inbound = None;
let mut state = vdaf.prepare_init(verify_param, agg_param, nonce, input_shares)?;
loop {
     state = match vdaf.prepare_next(state. inbound) {
          PrepareTransiton::Continue(new_state, outbound) => {
               // broadcast outbound and wait for next inbound
               new_state
          }
          PrepareTransition::Finish(out_share) => {
               return Ok(out_share);
          }
          PrepareTransition::Fail(err) => {
                return Err(err);
          }
     }
}
```

This seems simpler, to me at least. This is what I did for `run_vdaf()` in #171.

Incidentally, an alternative is to just have the `Vdaf` define how many rounds it needs for the preparation phase. For example:
```rust
pub trait Vdaf {
     fn num_rounds(&self) -> usize;
}
```

tgeoghegan:
The intent here is to enable PPM implementations that are generic over trait `vdaf::Aggregator`, and thus don't know how many rounds the concrete VDAF implementation needs, unless a generic method like this or your `num_rounds` is provided. For instance, my [leader implementation in the prototype](https://github.com/abetterinternet/ppm-prototype/blob/a0ed255336762ac1ef1ea08b459cb725996851d5/src/leader.rs#L261), which uses this method to determine if it's on the last round.

I agree that having clients use `Aggregator::prepare_next` would solve this problem. But then I argue we should remove `Aggregator::prepare_finish` and document the expectation that callers should keep calling `prepare_next` until it returns `PrepareTransition::Finish` or `PrepareTransition::Fail`, delegating the responsibility of managing the state of preparation to the implementation of `Vdaf`.

cjpatton:
Hahaha, yeah you're right. I meant `prepare_step`, not `prepare_next` :)

`prepare_start`, `prepare_next`, and `prepare_finish` are convenience functions, In order to use them, the user has to know what transition to expect next. Put another way, the user needs to know how many rounds of communication are required. If you don't know this, you need to use `prepare_step`. The documentation should be crystal clear about this.

I would be OK with dropping the convenience functions altogether, if you think that would be cleaner. In this case, it would probably be easier to merge #171 first, since it it removes usages of the convience functions from the tests.

tgeoghegan:
I agree, let's resolve #171, and then I can rebase this.

BranLwyd:
Popcorn gallery: I like the idea of the VDAF itself knowing the number of rounds it needs for the prepare step, and returning `PrepareTransition::Continue` or `PrepareTransition::Finish` -- I think that's what y'all are converging on with `prepare_step` over `prepare_{start,next,finish}`.

The reason I like this idea is that, from the perspective of a user of this library, this encapsulates more of the logic in the `Vdaf` rather than passing it up to the application, and the library ends up with a smaller, less error-prone API surface.

tgeoghegan:
I think I made this PR based on the misunderstanding that `prepare_step` is a "private" function that callers shouldn't be using. I am pretty confident that `prepare_init` and `prepare_step` satisfy the requirement of being able to generically execute a VDAF without introducing anything new like `trait PrepareStep`. @cjpatton already pointed out we should at a minimum document the two ways you can execute a VDAF. What I remain concerned about is whether having two ways to do this is too complicated. "Have one joint and keep it well oiled" suggests we should expose the minimal API surface that meets the client's needs. I'm coming around to the position of deleting the `prepare_start`, `prepare_next` and `prepare_finish`. I'm going to take a swing at that and see how it affects the API surface.

cjpatton:
> I'm going to take a swing at that and see how it affects the API surface.

I believe the only change will be in the tests in the `poplar1` module, but after #171 the changes should be straightforward.

tgeoghegan:
Yes, I'm going to take the opportunity to adopt `vdaf::run_vdaf` in `poplar1`'s tests.

tgeoghegan:
Closing this in favor of #174

PR #171, opened 2022-02-24
prio3: Add additional tests

cjpatton:
Adds aditional tests for prio3, including aggregating multiple
measurements. This also provides a generic implementation of
`run_vdaf()` from the VDAF spec.

### DIFF
--- a/src/vdaf/prio3.rs
+++ b/src/vdaf/prio3.rs
@@ -779,23 +779,103 @@ impl HelperShare {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::vdaf::{run_vdaf, run_vdaf_prepare};
+    use assert_matches::assert_matches;
 
     #[test]
     fn test_prio3_count64() {
-        let prio3 = Prio3Count64::new(Suite::Blake3, 23).unwrap();
+        let prio3 = Prio3Count64::new(Suite::Blake3, 2).unwrap();
+
+        assert_eq!(
+            run_vdaf(&prio3, &(), [1, 0, 0, 1, 1]).unwrap(),
+            Prio3Result(3)
+        );
+
         let (_, verify_params) = prio3.setup().unwrap();
         let nonce = b"This is a good nonce.";
-        let input_shares = prio3.shard(&(), &1).unwrap();
-        eval_vdaf(
-            &prio3,
-            &input_shares,
-            &verify_params,
-            nonce,
-            Some(Prio3Result(1)),
-        )
-        .unwrap();
 
-        // TODO(cjpatton) Add failure test cases.
+        let input_shares = prio3.shard(&(), &0).unwrap();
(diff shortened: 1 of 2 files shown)


src/vdaf/prio3.rs
```diff
@@ -779,23 +779,105 @@ impl HelperShare {
...
+
+        assert_eq!(
+            run_vdaf(&prio3, &(), [0, (1 << 16) - 1, 0, 1, 1]).unwrap(),
+            Prio3Result((1 << 16) + 1)
+        );
+
+        let (_, verify_params) = prio3.setup().unwrap();
+        let nonce = b"This is a good nonce.";
+
+        // TODO Check that the expected error is returned (requires `VdafError` to implement `Eq`).
```
tgeoghegan, line 815: If what you want to do is check that you got the expected variant of `VdafError`, then you don't need to compare by equality. You can use [`assert_matches!`](https://doc.rust-lang.org/std/assert_matches/macro.assert_matches.html) as in [`test_accumulate`](https://github.com/abetterinternet/libprio-rs/blob/fca3f26f572ede6e6f5f778d797661f1dbdb2566/src/field.rs#L660).

src/vdaf/prio3.rs
```diff
@@ -779,23 +779,105 @@ impl HelperShare {
...
+        input_shares[0].joint_rand_seed_hint.as_mut_slice()[0] ^= 255;
+        run_vdaf_prepare(&prio3, &verify_params, &(), nonce, input_shares).unwrap_err();
+
+        let mut input_shares = prio3.shard(&(), &1).unwrap();
+        match input_shares[0].input_share {
+            Share::Leader(ref mut data) => {
+                data[0] += Field96::one();
+            }
+            _ => panic!("expected leader share"),
+        }
```
tgeoghegan, line 830: `assert_matches!` would work well here. You can even get a scope where you can work with bindings of the destructured pattern, as seen [here](https://github.com/abetterinternet/ppm-prototype/blob/a0ed255336762ac1ef1ea08b459cb725996851d5/tests/tests.rs#L201).

src/vdaf/prio3.rs
```diff
@@ -779,23 +779,105 @@ impl HelperShare {
...
+        run_vdaf_prepare(&prio3, &verify_params, &(), nonce, input_shares).unwrap_err();
+
+        let mut input_shares = prio3.shard(&(), &1).unwrap();
+        match input_shares[0].proof_share {
+            Share::Leader(ref mut data) => {
+                data[0] += Field96::one();
+            }
+            _ => panic!("expected leader share"),
+        }
```
tgeoghegan, line 839: `assert_matches!` would work well here.

tgeoghegan (changes requested):
LGTM, I think it should just use `assert_matches!` in a handful of places.

tgeoghegan (approved):
LGTM with respect to `prio3`!

PR #172, opened 2022-02-24
prio3: Expose additional parameters via `Prio3`

cjpatton:
Adds methods to the `Prio3` implementation that provide access to the
underlying suite and the output and verifier lengths for the underlying
FLP.

### DIFF
--- a/src/vdaf/prio3.rs
+++ b/src/vdaf/prio3.rs
@@ -192,6 +192,23 @@ pub struct Prio3<T: Type, A: Clone + Debug> {
     phantom: PhantomData<A>,
 }
 
+impl<T: Type, A: Clone + Debug> Prio3<T, A> {
+    /// The suite used for this instance of prio3.
+    pub fn suite(&self) -> Suite {
+        self.suite
+    }
+
+    /// The output length of the underlying FLP.
+    pub fn output_len(&self) -> usize {
+        self.typ.output_len()
+    }
+
+    /// The verifier length of the underlying FLP.
+    pub fn verifier_len(&self) -> usize {
+        self.typ.verifier_len()
+    }
+}
+
 impl<T: Type, A: Clone + Debug + Sync + Send> Vdaf for Prio3<T, A> {
     type Measurement = T::Measurement;
     type AggregateResult = A;


src/vdaf/prio3.rs
```diff
@@ -192,6 +192,23 @@ pub struct Prio3<T: Type, A: Clone + Debug> {
     phantom: PhantomData<A>,
 }
 
+impl<T: Type, A: Clone + Debug> Prio3<T, A> {
+    /// The suite used for this instance of prio3.
+    pub fn suite(&self) -> Suite {
+        self.suite
+    }
+
+    /// The output lenngth of the underlying FLP.
```
tgeoghegan, line 201:
```suggestion
    /// The output length of the underlying FLP.
```

tgeoghegan (commented):
This looks OK, but can you expand on why you need these methods? This looks suspiciously like the values I added to `Prio3VerifyParam` for use in decoding `Prio3InputShare`.

cjpatton:
In my implementation of PPM:
* I use `veriifer_len()` to decode the `Prio3PrepareStep`. (We'll need serialization of this structure for our deployment.)
* I use `output_len()` for decoding a serialized aggregate share.

Could be that there are alternatives for either or both of these use cases. Any ideas?

tgeoghegan (approved):
Oh, interesting, I had not thought about encoding and decoding `PrepareStep`! My prototype keeps those in memory, but I'm guessing you do something like write them to an object store so that two different instances of your aggregator can handle two different rounds of prepare?

I think we could do something like require `Decode<Self::VerifyParam>` and `Encode` on the `PrepareStep` associated type on `trait Aggregator`... but then again neither the VDAF nor PPM specifications discuss transmitting or serializing the prepare step so perhaps we're better off providing the necessary tools for it as you do here and not being any more prescriptive than that.

cjpatton:
If we implemented serialization for `Prio3PrepareStep` here then we could avoid exposing `suite()` and `verifier_len()`. `output_len()` would still be needed for the aggregate share (at least as far as I can tell). I don't think there's a need to add a trait bound to `Aggreagtor::PrepareStep` to *require* this, however.

I'd be cool with providing serialization of `Prio3PrepareStep` here, but as you say, it's not strictly needed for an implementation. I'll leave it to you.

cjpatton:
As discussed offline, the plan is to implement serialiation for Prio3PrepareStep in this crate. (We'll also add Poplar1PrepareStep for parity.) However because the wire format isn't standardized (for now), we won't require these traits for generic Vdafs (for now).

cjpatton:
After discussing this more offline, it seems like the implementation could be a bit messy, because in practice only one variant of `Prio3PrepareStep` needs to be serialized (`Waiting`). @tgeoghegan are we good to merge as is? (does your approval still stand?)

tgeoghegan:
Sure, we can take this as-is since it unblocks your work.

PR #174, opened 2022-02-24
Remove helper methods on `vdaf::Aggregator`

tgeoghegan:
Remove the methods `Aggregator::{prepare_start, prepare_next,
prepare_finish}`. Using these methods correctly requires the caller to
know statically how many rounds a VDAF requires, because calling
`prepare_next` on a finished preparation or `prepare_finish` on one with
remaining rounds is an error. In order to write a PPM implementation
that is generic over trait `vdaf::Aggregator`, we would have to
introduce a generic means to discover how many rounds a VDAF needs or
whether the `Aggregator::PrepareStep` is on the last round.

The alternative is for a generic implementation to use
`Aggregator::prepare_step`, which returns an enumeration telling the
caller what to do next. Exposing two interfaces for driving the prepare
protocol is more complexity and risk than we need, so we will force
callers to only use one.

This commit also adopts `vdaf::run_vdaf{_prepare}` in `vdaf/poplar1`'s
tests, since the existing tests depended on the helper methods. This
does introduce a signficant performance regression in the unit tests,
because `vdaf::run_vdaf` makes a point of encoding and decoding
messages. With our `ToyIdpf`, this involves encoding and decoding
several hundred thousands of `Field128` values, so `test_poplar1` takes
about 20 seconds on my desktop. Exercising encode/decode of `poplar1`
messages also revealed a bug in `Poplar1InputShare::decode` that this
commit fixes.

### DIFF
--- a/src/vdaf.rs
+++ b/src/vdaf.rs
@@ -153,8 +153,9 @@ pub trait Aggregator: Vdaf {
     /// The type of messages exchanged among the Aggregators during the Prepare process.
     type PrepareMessage: Clone + Debug + Decode<Self::PrepareStep> + Encode;
 
-    /// Begins the Prepare process with the other Aggregators. The result of this process is
-    /// the Aggregator's output share.
+    /// Begins the Prepare process with the other Aggregators. The [`Self::PrepareStep`] returned
+    /// is passed to [`Aggregator::prepare_step`] to get this aggregator's first-round prepare
+    /// message.
     fn prepare_init(
         &self,
         verify_param: &Self::VerifyParam,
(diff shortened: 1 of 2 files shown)


src/vdaf.rs
```diff
@@ -163,64 +164,25 @@ pub trait Aggregator: Vdaf {
...
     ) -> Result<Self::PrepareMessage, VdafError>;
 
     /// Compute the next state transition from the current state and the previous round of input
-    /// messages.
+    /// messages. If this returns [`PrepareTransition::Continue`], then the returned
+    /// [`Self::PrepareMessage`] should be combined with the other aggregator's `PrepareMessage`s from
+    /// this round and passed into another call to this method. This continues until this method
+    /// returns [`PrepareTransition::Finish`], at which point the returned output share may be
+    /// aggregated. If the method returns [`PrepareTransition::Fail`], the aggregator should alert
+    /// its peers as specified in the PPM protocol.
```
cjpatton, line 179: It's best to not be too specific about the application.

```suggestion
    /// aggregated. If the method returns [`PrepareTransition::Fail`], the aggregator should consider its input share invalid and not attempt to process it any further.
```

src/vdaf/poplar1.rs
```diff
@@ -304,23 +305,28 @@ impl<I: Idpf<2, 2>> Decode<Poplar1VerifyParam> for Poplar1InputShare<I> {
...
+    input_length: usize,
     phantom: PhantomData<I>,
 }
 
 impl<I> Poplar1<I> {
     /// Create an instance of the poplar1 VDAF. The caller provides a cipher suite `suite` used for
-    /// deriving pseudorandom sequences of field elements.
-    pub fn new(suite: Suite) -> Self {
+    /// deriving pseudorandom sequences of field elements, and a input length in bits, corresponding
+    /// to `BITS` as defined in [the `poplar1` VDAF][1].
```
cjpatton, line 315:
```suggestion
    /// to `BITS` as defined in in the [VDAF spec][1].
```

src/vdaf/poplar1.rs
```diff
@@ -822,41 +833,23 @@ mod tests {
 
     #[test]
     fn test_poplar1() {
-        let vdaf: Poplar1<ToyIdpf<Field128>> = Poplar1::new(Suite::Blake3);
-        assert_eq!(vdaf.num_aggregators(), 2);
+        const INPUT_LEN: usize = 16;
```
cjpatton, line 836: to cut down on the runtime of the unit tests, we ought to cut this down to 8 bits. We should be able to modify the tests cases accordingly and still have the same coverage.

cjpatton (changes requested):
Looks great. 20 seconds for unit tests is definitely unacceptable, so let's see if we can't resolve this by reduing the input size for the tests (as discussed inline).

cjpatton (approved):
LGTM! I totally forgot about the `0b0000_0000` syntax. God, rust is great.

tgeoghegan:
With this change that trims down the input size in the `poplar1` test, the tests run in about half a second on my relatively beefy desktop.

Issue #176, opened 2022-02-25
Make VDAF structs opaque to clients


tgeoghegan:
The VDAF traits define a number of associated types (input shares, prepare messages, prepare step, aggregate shares, etc.) that are emitted and consumed by VDAF implementations. Clients of crate `prio` have no reason to examine the fields of these structs, much less mutate them. To minimize the risk of client error invalidating PPM/VDAF's security guarantees, we should make these as opaque as possible. At a minimum, we should audit the existing messages and make all their fields private or `pub(crate)` at most. It'd also be nice if we could somehow enlist the compiler or some other automation to make sure we don't accidentally make fields public in the future.

divergentdave:
I don't think there's any feature in the compiler to deny `pub` fields (short of writing our own attribute procedural macro), and a quick look at the Clippy lints didn't turn up anything relevant. I think it's just up to us to not make them `pub`.

PR #177, opened 2022-02-25
prio3: Align joint rand seed generation with spec

cjpatton:
If no joint randomness is required for the underlying FLP, then the spec
requires the client to not append the joint randomness seed hint or the
blind to the input share message. (Similarly, the spec requires the
aggregator to not send its joint randomness seed share in its prepare
message,)

### DIFF
--- a/src/vdaf/prio3.rs
+++ b/src/vdaf/prio3.rs
@@ -212,6 +212,7 @@ impl<T: Type, A: Clone + Debug + Sync + Send> Vdaf for Prio3<T, A> {
                     aggregator_id,
                     input_len: self.typ.input_len(),
                     proof_len: self.typ.proof_len(),
+                    joint_rand_len: self.typ.joint_rand_len(),
                 })
                 .collect(),
         ))
@@ -236,6 +237,9 @@ pub struct Prio3VerifyParam {
 
     /// Length in field elements of an uncompressed proof.
     proof_len: usize,
+
+    /// Length of the joint randomness.
+    joint_rand_len: usize,
 }
 
 /// The message sent by the client to each aggregator. This includes the client's input share and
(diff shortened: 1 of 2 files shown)


tgeoghegan:
@cjpatton could you please clarify which commit or PR on cjpatton/vdaf these changes should be checked against?

cjpatton:
https://github.com/cjpatton/vdaf/commit/5de2062ac1f1524a9ce8e23c08a24938c5f32fc1

src/vdaf/prio3.rs
```diff
@@ -249,13 +253,10 @@ pub struct Prio3InputShare<F> {
...
     /// The sum of the joint randomness seed shares sent to the other aggregators.
-    //
-    // TODO(cjpatton) If `input.joint_rand_len() == 0`, then we don't need to bother with the joint
-    // randomness seed at all. See https://github.com/cjpatton/vdaf/issues/15.
-    pub joint_rand_seed_hint: Key,
+    pub joint_rand_seed_hint: Option<Key>,
 
     /// The blinding factor, used to derive the aggregator's joint randomness seed share.
-    pub blind: Key,
+    pub blind: Option<Key>,
```
tgeoghegan, line 259: This allows the possibility of a `Prio3InputShare` with `joint_rand_seed_hint = None` and `blind = Some(...)`, or vice versa, which are both invalid and shouldn't be allowed. `Prio3InputShare::decode` should check for this and reject such values, or we could define a field `joint_rand_seed_hint_and_blind: Option<(Key, Key)>` so that they must either both be set or neither is set. Even better would be to define another struct that contains the `joint_rand_seed_hint` and `blind` but I'm not sure what to name it.

tgeoghegan (changes requested):

src/vdaf/prio3.rs
cjpatton: Done (sorry about the rebase + force-push, but you can see the changes in the last commit ("SQUASH Tim round 1")

src/vdaf/prio3.rs
```diff
@@ -247,16 +247,14 @@ pub struct Prio3VerifyParam {
 #[derive(Clone, Debug, Eq, PartialEq)]
 pub struct Prio3InputShare<F> {
     /// The input share.
-    pub input_share: Share<F>,
+    input_share: Share<F>,
```
cjpatton, line 250: Per #176

tgeoghegan (approved):

PR #181, opened 2022-03-03
vdaf: Align PRG with next VDAF draft

### LINKED ISSUE #152: `vdaf::suite` rework (was "Rename `vdaf::suite::Key` to `Seed`")
And while at it, rename the associated structs and methods to match. We only ever use this type as a seed share and never as a key for any other purpose. This is also unlikely to change in the future.

cjpatton:
Closes #152.

Replaces module `vdaf::suite` with a module `vdaf::prg` that defines
traits for PRGs as defined in the VDAF spec. It also implements the
AES128-based PRG specified there and integrates it into `prio3` and
`poplar1`.

Refactors Prio 2 to use the new `SeedStream` trait.


src/vdaf/prg.rs
```diff
@@ -0,0 +1,232 @@
...
+    Aes128, Aes128Ctr,
+};
+use cmac::{Cmac, Mac, NewMac};
+use std::{
+    fmt::Debug,
+    io::{Cursor, Read},
+};
+
+/// Input of [`Prg`].
+#[derive(Clone, Debug, Eq)]
```
tgeoghegan, line 20: nit: I think deriving `Copy` would also be appropriate.

src/vdaf/prg.rs
```diff
@@ -0,0 +1,232 @@
...
+impl<const L: usize> Decode<()> for Seed<L> {
+    fn decode(_decoding_parameter: &(), bytes: &mut Cursor<&[u8]>) -> Result<Self, CodecError> {
+        let mut seed = [0; L];
+        bytes.read_exact(&mut seed)?;
+        Ok(Seed(seed))
+    }
+}
+
+/// A stream of pseudorandom bytes derived from a seed.
+pub trait SeedStream: Sized {
```
tgeoghegan, line 74: Is this `Sized` requirement because of the methods on `trait Prg` that return `Self::SeedStream`? I think it could be argued that the `Sized` bound then belongs in `Prg` (i.e., it should say `type SeedStream: SeedStream + Sized;`) but if we don't expect `SeedStream` to ever be used independently from `Prg` then this is fine.

src/vdaf/prg.rs
```diff
@@ -0,0 +1,232 @@
...
+            GenericArray::from_slice(iv),
+        ))
+    }
+}
+
+impl SeedStream for SeedStreamAes128 {
+    fn fill(&mut self, buf: &mut [u8]) {
+        for b in buf.iter_mut() {
+            *b = 0;
+        }
```
tgeoghegan, line 150: nit: the stdlib provides [`slice::fill`](https://doc.rust-lang.org/std/primitive.slice.html#method.fill) which could be faster than this loop, though I suspect what you've got will compile to a single `memset` anyway.

src/vdaf/prg.rs
```diff
@@ -0,0 +1,232 @@
...
+}
+"#;
```
tgeoghegan, line 171: Is the intent to eventually add a file or files containing multiple JSON test vectors? Otherwise you could spare yourself the hassle of serde by directly instantiating a `struct PrgTestVector`. Having the JSON here does have the virtue of being easy to check against the reference implementation's test vectors.

src/vdaf/prio3.rs
```diff
@@ -28,17 +28,22 @@ use std::io::Cursor;
...
 
+// TODO Add domain separation tag to field element generation as required by spec.
+
+// TODO Add test vectors and make sure they pass.
+
 /// The count type. Each measurement is an integer in `[0,2)` and the aggregate is the sum.
-pub type Prio3Count64 = Prio3<Count<Field64>, Prio3Result<u64>>;
+// XXX Question for reviewer: What do you think about removing the "64" suffix from this type name?
+// (Here and below.)
+pub type Prio3Aes128Count64 = Prio3<Count<Field64>, Prio3Result<u64>, PrgAes128, 16>;
```
tgeoghegan, line 38: Having `64` here is indeed potentially confusing since it's not clear if it's referring to the size of the underlying field or the size of the result. What's important is to make it clear to the caller what the range of values this metric can record is, so the naming convention depends on whether or not we ever expect to have `Prio3Aes128Count128` or `Prio3Aes128Count32`. The precedent I look to here is Prometheus. They have a single [Counter](https://prometheus.io/docs/concepts/metric_types/#counter) type and Prometheus' docs don't specify what type it is. If you look at [client libraries](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#Counter) it's clearly a 64 bit float. IMO if one-size-fits-all counters and gauges are good enough for Prometheus, they'll be good enough for us (though of course u64 has a much smaller range than float64).

Yeah, let's rename these to `Prio3Aes128Count`. TBH I think we could even do `Prio3Count`. Will we ever let the caller choose which PRG to use?

tgeoghegan (approved):
This is great! I really like the use of const generics to simplify the decoding parameter on `Seed` and `vdaf::Share`. I wonder now if we could use const generics (perhaps once the feature matures a bit more) to further reduce our usage of decoding parameters.

Approving since I just left some nits that shouldn't block. Happy to take another look if you wind up renaming the `Prio3` VDAFs.

src/vdaf/prg.rs
cjpatton: Done

src/vdaf/prg.rs
cjpatton: I think forcing the user to clone is better for this particular type. Happy to elaborate if you're interested.

src/vdaf/prg.rs
cjpatton: It's probably a good idea to move this to a file eventually, thought I don't think this is strictly necessary. I did it this way mostly because it allows you to copy-paste the test vectors generated by the reference implementation.

src/vdaf/prio3.rs
cjpatton: Removed the "64" suffix from each of the names, but left the PRG name. I suspect users may one day have a preference for a different PRG.

src/vdaf/prg.rs
cjpatton: This turned out to not be necessary, so I removed it.

PR #182, opened 2022-03-03
prio3: Add domain separation for field element generation

cjpatton:
Based on #181 (merge that first).

See https://github.com/cjpatton/vdaf/commit/72da615bca3cfa9687d6221006ad2b60c5083757.

### DIFF
--- a/src/vdaf/prio3.rs
+++ b/src/vdaf/prio3.rs
@@ -28,9 +28,14 @@ use std::io::Cursor;
 use std::iter::IntoIterator;
 use std::marker::PhantomData;
 
-// TODO Add domain separation tag to field element generation as required by spec.
-
 // TODO Add test vectors and make sure they pass.
+
+// Domain-separation tag used to bind the VDAF operations to the document version. This will be
+// reved with each new draft.
+//
+// NOTE The CFRG has not yet adopted this spec. Version "vdaf-00" will match
+// draft-irtf-cfrg-vdaf-00.
+const VERS_PRIO3: &[u8] = b"vdaf-00 prio3";
 
 /// The count type. Each measurement is an integer in `[0,2)` and the aggregate is the sum.
 pub type Prio3Aes128Count = Prio3<Count<Field64>, Prio3Result<u64>, PrgAes128, 16>;
@@ -405,6 +410,9 @@ where
         _public_param: &(),
         measurement: &T::Measurement,
     ) -> Result<Vec<Prio3InputShare<T::Field, L>>, VdafError> {
+        let mut info = [0; VERS_PRIO3.len() + 1];
+        info[..VERS_PRIO3.len()].clone_from_slice(VERS_PRIO3);
+
         let num_aggregators = self.num_aggregators;
         let input = self.typ.encode(measurement)?;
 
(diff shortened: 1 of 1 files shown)


src/vdaf/prio3.rs
```diff
@@ -28,10 +28,15 @@ use std::io::Cursor;
...
-// TODO Add domain separation tag to field element generation as required by spec.
-
 // TODO Add test vectors and make sure they pass.
 
+// Domain-separation tag used to bind the VDAF operations to the document version. This will be
+// reved with each new draft.
+//
+// NOTE The CFRG has not yet adopted this spec. Version "vdaf-00" will match
+// draft-irtf-cfrg-vdaf-00.
+const VERS_PRIO3: &[u8] = b"vdaf-00 prio3";
```
tgeoghegan, line 38: I like the idea of domain separation based on VDAF spec version. Does PPM need to deal with migration from one VDAF version to the next? Or do we claim that the VDAF version is an essential part of what VDAF is being executed, and thus in order to go from running `Prio3Count@vdaf-00` to `Prio3Count@vdaf-01`, you should deploy a completely new PPM task? In any case this question isn't germane to this PR?

src/vdaf/prio3.rs
```diff
@@ -416,7 +421,7 @@ where
             let mut deriver = P::init(&helper.joint_rand_param.blind);
             deriver.update(&[aggregator_id]);
             let prng: Prng<T::Field, _> =
-                Prng::from_seed_stream(P::seed_stream(&helper.input_share, b"input share"));
+                Prng::from_seed_stream(P::seed_stream(&helper.input_share, VERS_PRIO3));
```
tgeoghegan, line 424: This seems like it regresses in that we no longer have domain separation between PRNGs used for joint rand vs. input sharing vs. prove randomness. Why not do `b"vdaf-00 prio3 input share"` and `b"vdaf-00 prio3 joint rand"` and so on?

tgeoghegan (approved):
I have a couple of meta-questions but this PR achieves what it intends to do, which is alignment with the current VDAF draft.

src/vdaf/prio3.rs
cjpatton: Whether this is a regresssion depends on how we define security for VDAFs. My intuition is that domain separation for vector type (input share, proof share, query rand, etc.) isn't going to be needed.

That said, a binding to the draft version is useful. My goal is to keep the info string under 16 bytes so that it fits into a single AES block. Note that I will add an additional byte for input share and proof share: see the commit message for https://github.com/cjpatton/vdaf/commit/29840fd017e6dc9103d1fa996475e0ccaa71d222 for an explanation. I will tack this on when I do my pass of your comments.

src/vdaf/prio3.rs
cjpatton: What will end up happening is that a particular draft of PPM will end up using a particular draft of VDAF.

cjpatton:
@tgeoghegan: CI has been running for a while, maybe we should restart it?

tgeoghegan:
> @tgeoghegan: CI has been running for a while, maybe we should restart it?

Seems like it got kicked when you force-pushed and is making progress now.

PR #183, opened 2022-03-03
Add encoding parameter, allow parameters to be omitted if they are ().

BranLwyd:
I do this by creating non-parameterized versions of the traits, then
creating blanket implementations allowing the non-parameterized trait
to act as the parameterized trait with a parameter of ().
Implementations should implement the non-parameterized version of the
trait if they don't need a parameter.

The encoding parameter will practically be used in some PPM messages to
allow the required authentication key to be used as part of the encoding
logic.


src/codec.rs
```diff
@@ -50,21 +68,51 @@ pub trait Decode<D>: Sized {
...
+        _decoding_parameter: &(),
+        bytes: &mut Cursor<&[u8]>,
+    ) -> Result<Self, CodecError> {
+        Self::decode(bytes)
+    }
+}
+
 /// Describes how to encode objects into a byte sequence.
-pub trait Encode {
+pub trait Encode: Sized {
```
tgeoghegan, line 82: Why is `Sized` needed? Neither method o n the trait takes `self` by value or deals in `Self`.

src/vdaf/suite.rs
```diff
@@ -6,7 +6,7 @@
 //
```
tgeoghegan, line 6: Changes in this file will collide with https://github.com/abetterinternet/libprio-rs/pull/181

tgeoghegan (approved):
This is great and will save us so much `&()` noise.

src/codec.rs
BranLwyd: Without it, compilation fails with the following error:
```
error[E0599]: the method `get_encoded_with_param` exists for reference `&Self`, but its trait bounds were not satisfied
  --> src/codec.rs:88:14
   |
88 |         self.get_encoded_with_param(&())
   |              ^^^^^^^^^^^^^^^^^^^^^^ method cannot be called on `&Self` due to unsatisfied trait bounds
   |
   = note: the following trait bounds were not satisfied:
           `Self: Sized`
           which is required by `Self: ParameterizedEncode<()>`
           `&Self: Encode`
           which is required by `&Self: ParameterizedEncode<()>`
```


I think this was because of the implicit `Sized` bound in the blanket implementation; I added an explicit `?Sized` and compilation works out without specifying that `Encode` requires `Sized`.

tgeoghegan: In any case I don't think requiring `Sized` is a problem here.

src/vdaf/suite.rs
BranLwyd: It looks like `suite.rs` is removed by that PR, but it looks like trivial updates will be require to the `prg.rs` that replaces it (a `Decode` implementation will need to drop its `<()>` parameterization & unused `_decoding_parameter` parameter).

Want me to wait for that PR to make it in to merge this PR? I'd like to get this merged sooner rather than later to unblock work on `janus` but I think I could do some local development by overriding the crate location in my local `cargo.toml`.

tgeoghegan: Your call. I think Chris is likely to merge his change later today.

BranLwyd: I waited, the merge wasn't too bad & probably easier for me to handle than to foist off on Chris.

PR #185, opened 2022-03-04
Test vectors for `prio3`

cjpatton:
Based on #182 (merge that first).

Adds test vectors for prio3 generated by the reference implementation and instruments prio3 to run them. The test vectors revealed a couple minor bugs, which this change fixes.

NOTE: This isn't quite ready to merge, as I can't quite get the code to compile. See "XXX" comments inline. I'm hoping reviewers have some ideas.


src/vdaf/prio3.rs
```diff
@@ -223,6 +223,165 @@ where
     pub fn verifier_len(&self) -> usize {
         self.typ.verifier_len()
     }
+
```
cjpatton, line 226: This should be identical to the previous `shard()` modulo replacing calls to `Seed::generate()` with calls to `Seed::from_rand_source(rand_source)`.

divergentdave:
I got the last test working with some trait bound billiards. Take a look at 7de1fc0f27f5c819e866772603de34ff0dfce452

divergentdave:
I think it would be good to move this from an integration test to a unit test inside the crate, so we could take `Prio::test_vec_setup()`, `Prio3::test_vec_shard()`, etc. out of the public API. (perhaps as a follow-up to this PR)

cjpatton:
> I got the last test working with some trait bound billiards. Take a look at [7de1fc0](https://github.com/abetterinternet/libprio-rs/commit/7de1fc0f27f5c819e866772603de34ff0dfce452)

Nice! Hope you don't mind, I cherry-picked your commit onto this branch and squashed.

I think it would be good to move this from an integration test to a unit test inside the crate, so we could take `Prio::test_vec_setup()`, `Prio3::test_vec_shard()`, etc. out of the public API. (perhaps as a follow-up to this PR)

Makes sense to me. The reason I put the test vectors here is because this is where other test vectors are :) I'll move them unless anyone objects. (cc/ @tgeoghegan who set up this directory initially.)

tgeoghegan:
I added the backward compatibility test as an integration test in #88 and I can't really remember why. I was hoping the PR would have the context on why,  but actually it shows that I migrated two other integration tests into unit tests in that change. The only argument I can think of for keeping `backward_compatibility.rs` as an integration test would be to prove that it can be implemented using no crate-private symbols but I'm not sure if that really matters. I'd support moving it to a unit test, though not in this PR.

src/vdaf/prio3.rs
```diff
@@ -223,6 +223,165 @@ where
...
+    fn setup_with_rand_source<R: Fn(&mut [u8]) -> Result<(), getrandom::Error>>(
```
tgeoghegan, line 227: nit: this trait bound occurs in a few places, and its signature doesn't make its purpose (which is to fill a buffer with random values, right?) obvious. A well-named type alias could make this more clear.

And if you did that, then you could possibly define the type alias to also require `Copy`, which you end up requiring everywhere anyway, I assume because you have to pass the `rand_source` function by value into various places.

cjpatton:
Moved tests to `vdaf/prio3_test.rs` and removed `test_vec_setup()` and `test_vec_shard()` from the public API.

tgeoghegan (approved):

src/vdaf/prio3.rs
cjpatton: I'm not sure this works the way you expect. Maybe I'm missing something?

```
$ cargo test
   Compiling prio v0.6.1 (/Users/chris/github.com/abetterinternet/libprio-rs)
error[E0404]: expected trait, found type alias `RandSource`
  --> src/vdaf/prg.rs:31:39
   |
31 |     pub(crate) fn from_rand_source<R: RandSource>(
   |                                       ^^^^^^^^^^ type aliases cannot be used as traits

warning: trait objects without an explicit `dyn` are deprecated
  --> src/vdaf/prg.rs:19:30
   |
19 | pub(crate) type RandSource = Fn(&mut [u8]) -> Result<(), getrandom::Error>;
   |                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: use `dyn`: `dyn Fn(&mut [u8]) -> Result<(), getrandom::Error>`
```

tgeoghegan: I was hoping this would Just Work. I think maybe with some trait definitions it could be made to work, but it's really not worth the trouble. Let's just move ahead with the `R: Fn(&mut [u8]) -> Result<(), getrandom::Error>` bound.

divergentdave: I think you could get rid of the type variable R, create a type alias for the function pointer type `fn(&mut [u8]) -> Result<(), getrandom::Error>`, and have the rand_source arguments be of that type. (you'll need to coerce `getrandom::getrandom as MyTypeAlias` to get a function pointer)

PR #192, opened 2022-03-08
vdaf: Serialization of verification parameters

cjpatton:


tgeoghegan (changes requested):
I had also been thinking that we will need this so that the two aggregators can share verify params in a deployment where they're not in the same process. However I think we need to express this in `trait Vdaf`, which is to say, some trait bound on the `Vdaf::VerifyParam`, so that PPM implementations can do this generically for any VDAF.

cjpatton:
This is also an encoding that's not specified in the VDAF draft.** How about following this suggestion https://github.com/abetterinternet/libprio-rs/pull/195#pullrequestreview-903742114 here as well? I.e., don't bound the trait, but provide the implementation where needed? In particular I could implement serialization for Poplar1 in this PR as well, if you like.

** I can see an argument for having the draft specify this encoding. If you want to advocate for this, then let's open an issue here: https://github.com/cjpatton/vdaf/issues

cjpatton:
Issue raised: https://github.com/cjpatton/vdaf/issues/38

The format we chose here is quite simple and could easily be adopted in the VDAF draft. In TLS-syntax:

```
struct {
   opaque k_query_init[Prg.SEED_SIZE];
   uint8_t aggregator_id;
} Prio3VerifyParam;
```

We'd have a similar format for Poplar1:

```
struct {
   opaque k_query_init[Prg.SEED_SIZE];
   uint8_t is_leader; // boolean indicating whether the aggregator is the leader
} Poplar1VerifyParam;
```

@tgeoghegan let me know if this PR needs changes. I'm happy to provide an implementation for Poplar1 as well.

tgeoghegan:
This looks good for `prio3`! But yes, let's wire it up for `poplar1`, too.

tgeoghegan (changes requested):
This looks good for `prio3`! But yes, let's wire it up for `poplar1`, too.

cjpatton:
Done.

src/vdaf/poplar1.rs
```diff
@@ -427,12 +427,12 @@ where
...
 
 /// The verification parameter used by the aggregators to evaluate the VDAF on a distributed input.
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Eq, PartialEq)]
 pub struct Poplar1VerifyParam<const L: usize> {
     /// Key used to derive the verification randomness from the nonce.
     rand_init: Seed<L>,
 
-    /// Length of
+    /// Length of the IPDF input.
```
divergentdave, line 435:
```suggestion
    /// Length of the IDPF input.
```

cjpatton: Ugh, my dyslexia ...

tgeoghegan (approved):

PR #195, opened 2022-03-09
prio3: Encode prepare step in "waiting" state

cjpatton:
This functionality is useful for implementations of PPM that need to
offload the helper state to persistent storage while waiting for a reply
from the leader.

### DIFF
--- a/src/vdaf/prio3.rs
+++ b/src/vdaf/prio3.rs
@@ -238,6 +238,7 @@ where
                     aggregator_id,
                     input_len: self.typ.input_len(),
                     proof_len: self.typ.proof_len(),
+                    verifier_len: self.typ.verifier_len(),
                     joint_rand_len: self.typ.joint_rand_len(),
                 })
                 .collect(),
@@ -420,6 +421,9 @@ pub struct Prio3VerifyParam<const L: usize> {
     /// Length in field elements of an uncompressed proof.
     proof_len: usize,
 
+    /// Length of the verifier message.
+    verifier_len: usize,
+
     /// Length of the joint randomness.
     joint_rand_len: usize,
 }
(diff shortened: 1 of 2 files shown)


tgeoghegan (changes requested):
You explained how you don't want to complicate trait `vdaf::Vdaf` or `vdaf::Aggregator` with trait bounds on `PrepareStep` that don't correspond to anything in the VDAF specification. But if we want this stuff to be usable for implementations of PPM servers that are generic over VDAF, it's not sufficient for `Prio3PrepareStep` to expose new methods, because my helper implementation will not be dealing with `Prio3PrepareStep` concretely but will rather be generic over `Aggregator::PrepareStep`.

I think a compromise here would be to refactor your new methods on `Prio3PrepareStep` into implementations of `codec::{Encode, Decode}`. Then, I can write something like:

```
struct Helper<A> { ... }

impl<A> Helper<A> 
where
  A: vdaf::Aggregator,
  A::PrepareStep: Encode + Decode,
{ ... }
```
I haven't checked that this compiles, but the upshot is that I can impose the trait bounds I want in my generic code, and then my `struct Helper` can use `Prio3` or any other `vdaf::Aggregator` whose `PrepareStep` happens to implement `Encode` and `Decode`. But we don't have to further complicate the `Vdaf` or `Aggregator` trait definitions or imply that any behavior is defined by the spec.

cjpatton:
This seems like a reasonable way to go. One issue I see: I don't want to have to implement serialization the initial state, mainly because I don't see how this would ever be needed; but if all we do is provide implementations of the `Encode + ParmaeterizedDecode`, then this behavior won't be documented anywhere. (I suppose we could put it in the docucomment for `Prio3PrepareStep`?)

Can we live with this? If not, then I see two options:
1. Implement serialization for every step, including the initial step. Decoding will require knowledge of what step the caller is suppose to be in.
2. Refine the traits by refactoring `PrepareStep` into `PrepareStepInit` and `PrepareStepNext` (or maybe just `PrepareStep`) and only implement serialization for the latter.

tgeoghegan:
I think returning an error or even panicking is appropriate if something tries to encode `PrepareStepInit`. The caller knows where in the share preparation state machine they are when they call `PrepareStep::encode` and so it seems fair to me to rely upon them to avoid doing so at the wrong time, and to fail noisily and obviously if they do the wrong thing. The doccomment on `Prio3PrepareStep` seems like a good place to document this behavior (though of course the panic string or error description can also serve as documentation).

cjpatton:
@tgeoghegan I've addressed your comments and am ready for another round!

src/vdaf/prio3.rs
```diff
@@ -564,6 +573,48 @@ pub struct Prio3PrepareStep<F, const L: usize> {
     state: PrepareStep<F, L>,
 }
 
+impl<F: FieldElement, const L: usize> Encode for Prio3PrepareStep<F, L> {
+    fn encode(&self, bytes: &mut Vec<u8>) {
+        if let PrepareStep::Ready(..) = self.state {
+            panic!("encooding of ready state is not implemented");
```
tgeoghegan, line 579: This wording leaves open the possibility that we will implement the encoding of the ready state in the future. Let's be more definite about it and provide some more context to help PPM implementers who will encounter this panic string while developing and may need some more help.

```suggestion
            panic!("encoding of ready state is not permitted. See Prio3PrepareStep documentation for details.");
```

tgeoghegan (approved):

src/vdaf/prio3.rs
cjpatton: What about just adding a comment above the `panic!()`?

cjpatton: Done, though worded slightly different

PR #197, opened 2022-03-09
Documentation cleanup

divergentdave:
The first commit here adds documentation of panic behavior directly on `<Prio3PrepareStep as Encode>::encode()`, as a follow-up to #195. The second commit fixes some rustdoc warnings due to broken references.

### DIFF
--- a/src/field.rs
+++ b/src/field.rs
@@ -129,7 +129,7 @@ pub trait FieldElement:
     /// # Warnings
     ///
     /// This function should only be used within [`prng::Prng`] to convert a random byte string into
-    /// a field element. Use [`Self::try_from_reader`] to deserialize field elements. Use
+    /// a field element. Use [`Self::decode`] to deserialize field elements. Use
     /// [`field::rand`] or [`prng::Prng`] to randomly generate field elements.
     #[doc(hidden)]
     fn try_from_random(bytes: &[u8]) -> Result<Self, FieldError>;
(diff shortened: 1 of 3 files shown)


src/field.rs
```diff
@@ -172,7 +172,7 @@ pub trait FieldElement:
     ///
     /// Returns an error if the length of the provided byte slice is not a multiple of the size of a
     /// field element, or if any of the values in the byte slice are invalid encodings of a field
-    /// element as documented in [`Self::try_from_reader`].
+    /// element, because the encoded integer is larger than the field modulus.
```
cjpatton, line 175:
```suggestion
    /// element, because the encoded integer is larger than or equal to the field modulus.
```

src/vdaf/prio3.rs
```diff
@@ -577,6 +577,11 @@ pub struct Prio3PrepareStep<F, const L: usize> {
 }
 
 impl<F: FieldElement, const L: usize> Encode for Prio3PrepareStep<F, L> {
+    /// Append the encoded form of this object to the end of `bytes`, growing the vector as needed.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the step is not in the "waiting" state. See [`Prio3PrepareStep`] for details.
```
cjpatton, line 584: Oh perfect, I didn't realize this would show up in the docs, but it does!

cjpatton (changes requested):

src/field.rs
divergentdave: Fixed this in one other place too.

cjpatton (approved):

PR #202, opened 2022-03-24
Implement Encode & Decode for OutputShare.

BranLwyd:
Users of a VDAF need to store output shares for some time, so it makes
sense to implement Encode/Decode.

I also added a requirement that Vdaf::OutputShare implement Encode & Decode, following the lead of a few other types related to Vdaf. But I could drop this at the cost of slightly complexifying the trait bounds in implementations--I'm not opinionated here.

I also arbitrarily guess that 64KiB is enough to encode an OutputShare. I think this is probably OK, but if folks are worried we could bump to `encode_u32_items` at the cost of a few bytes per encoded OutputShare.

### DIFF
--- a/src/vdaf.rs
+++ b/src/vdaf.rs
@@ -102,6 +102,9 @@ impl<F: FieldElement, const L: usize> Encode for Share<F, L> {
 /// The base trait for VDAF schemes. This trait is inherited by traits [`Client`], [`Aggregator`],
 /// and [`Collector`], which define the roles of the various parties involved in the execution of
 /// the VDAF.
+// TODO(brandon): once GATs are stabilized [https://github.com/rust-lang/rust/issues/44265],
+// state the "&AggregateShare must implement Into<Vec<u8>>" constraint in terms of a where clause
+// on the associated type instead of a where clause on the trait.
 pub trait Vdaf: Clone + Debug
 where
     for<'a> &'a Self::AggregateShare: Into<Vec<u8>>,
@@ -251,24 +254,30 @@ impl<F> AsRef<[F]> for OutputShare<F> {
     }
 }
 
+impl<F> From<Vec<F>> for OutputShare<F> {
+    fn from(other: Vec<F>) -> Self {
+        Self(other)
+    }
+}
+
 impl<F: FieldElement> TryFrom<&[u8]> for OutputShare<F> {
     type Error = FieldError;
 
     fn try_from(bytes: &[u8]) -> Result<Self, Self::Error> {
-        Ok(Self(F::byte_slice_into_vec(bytes)?))
+        fieldvec_try_from_bytes(bytes)
     }
 }
 
 impl<F: FieldElement> From<&OutputShare<F>> for Vec<u8> {
     fn from(output_share: &OutputShare<F>) -> Self {
-        F::slice_into_byte_vec(&output_share.0)
+        fieldvec_to_vec(&output_share.0)
     }
 }
 
 /// An aggregate share suitable for VDAFs whose output shares and aggregate
 /// shares are vectors of `F` elements, and an output share needs no special
 /// transformation to be merged into an aggregate share.
-#[derive(Clone, Debug, Serialize, Deserialize)]
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
 pub struct AggregateShare<F>(Vec<F>);
 
 impl<F> AsRef<[F]> for AggregateShare<F> {
@@ -325,14 +334,30 @@ impl<F: FieldElement> TryFrom<&[u8]> for AggregateShare<F> {
     type Error = FieldError;
 
     fn try_from(bytes: &[u8]) -> Result<Self, Self::Error> {
-        Ok(Self(F::byte_slice_into_vec(bytes)?))
+        fieldvec_try_from_bytes(bytes)
     }
 }
 
 impl<F: FieldElement> From<&AggregateShare<F>> for Vec<u8> {
     fn from(aggregate_share: &AggregateShare<F>) -> Self {
-        F::slice_into_byte_vec(&aggregate_share.0)
+        fieldvec_to_vec(&aggregate_share.0)
     }
+}
+
+/// fieldvec_try_from_bytes converts a slice of bytes to a type that is equivalent to a vector of
+/// field elements.
+#[inline(always)]
+fn fieldvec_try_from_bytes<F: FieldElement, T: From<Vec<F>>>(
+    bytes: &[u8],
+) -> Result<T, FieldError> {
+    F::byte_slice_into_vec(bytes).map(T::from)
+}
+
+/// fieldvec_to_vec converts a type that is equivalent to a vector of field elements into a vector
+/// of bytes.
+#[inline(always)]
+fn fieldvec_to_vec<'a, F: FieldElement, T: AsRef<[F]>>(val: T) -> Vec<u8> {
+    F::slice_into_byte_vec(val.as_ref())
 }
 
 #[cfg(test)]
@@ -435,6 +460,45 @@ where
     }
 
     Ok(out_shares)
+}
+
+#[cfg(test)]
+mod tests {
+    use super::{AggregateShare, OutputShare};
+    use crate::field::{Field128, Field64, FieldElement};
+    use itertools::iterate;
+    use std::convert::TryFrom;
+    use std::fmt::Debug;
+
+    fn fieldvec_roundtrip_test<F, T>()
+    where
+        F: FieldElement,
+        for<'a> T: Debug + PartialEq + From<Vec<F>> + TryFrom<&'a [u8]>,
+        for<'a> <T as TryFrom<&'a [u8]>>::Error: Debug,
+        for<'a> Vec<u8>: From<&'a T>,
+    {
+        // Generate a value based on an arbitrary vector of field elements.
+        let g = F::generator();
+        let want_value = T::from(iterate(F::one(), |&v| g * v).take(10).collect());
+
+        // Round-trip the value through a byte-vector.
+        let buf: Vec<u8> = (&want_value).into();
+        let got_value = T::try_from(&buf).unwrap();
+
+        assert_eq!(want_value, got_value);
+    }
+
+    #[test]
+    fn roundtrip_output_share() {
+        fieldvec_roundtrip_test::<Field64, OutputShare<Field64>>();
+        fieldvec_roundtrip_test::<Field128, OutputShare<Field128>>();
+    }
+
+    #[test]
+    fn roundtrip_aggregate_share() {
+        fieldvec_roundtrip_test::<Field64, AggregateShare<Field64>>();
+        fieldvec_roundtrip_test::<Field128, AggregateShare<Field128>>();
+    }
 }
 
 pub mod poplar1;
--- a/src/vdaf.rs
+++ b/src/vdaf.rs
@@ -6,7 +6,9 @@
 //!
 //! [draft-patton-cfrg-vdaf-01]: https://datatracker.ietf.org/doc/html/draft-patton-cfrg-vdaf-01
 
-use crate::codec::{CodecError, Decode, Encode, ParameterizedDecode};
+use crate::codec::{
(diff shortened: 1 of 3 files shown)


src/vdaf.rs
```diff
@@ -123,7 +125,7 @@ pub trait Vdaf: Clone + Debug {
     type InputShare: Clone + Debug + ParameterizedDecode<Self::VerifyParam> + Encode;
 
     /// An output share recovered from an input share by an Aggregator.
-    type OutputShare: Clone + Debug;
+    type OutputShare: Clone + Debug + Decode + Encode;
```
tgeoghegan, line 128: I think putting `Encode/Decode` on `struct OutputShare` is sound. However I'm not sure it makes sense to put this bound into `trait Vdaf`. Generally what's in the trait corresponds to requirements of the VDAF specification, and there's never a need for one VDAF implementation to transmit an output share to another, so draft-patton-cfrg-vdaf doesn't specify how to encode those. Then again, VDAF doesn't say anything about `Clone + Debug`, either, so maybe we don't have to treat this trait definition as equivalent to the VDAF spec. I'm curious to hear @cjpatton's opinion.

I'll note that an alternative to putting this bound on `trait Vdaf` is to declare it in your code that is generic over `Vdaf` (or over `Aggregator`). For instance:

```
fn do_thing<V>() 
where
    V: Vdaf,
    Vdaf::OutputShare: Decode + Encode,
{ ... }
```

The above trait bound holds for `struct Prio3` and `struct Poplar1` because they both have `OutputShare = vdaf::OutputShare`, which does implement `Encode/Decode`, but you wouldn't be able to use any implementation of `Vdaf` whose `OutputShare` associated type doesn't implement those traits.

BranLwyd:
I added `#[derive(PartialEq, Eq)]` to `OutputShare` as well. This is very helpful to test that OutputShares can be roundtripped through serialization properly -- outside the crate there is no access to the internal of OutputShare.

src/vdaf.rs
BranLwyd: Agreed, part of this PR's review is to determine if it's acceptable to add `Encode`/`Decode` here or not:

> I also added a requirement that Vdaf::OutputShare implement Encode & Decode, following the lead of a few other types related to Vdaf. But I could drop this at the cost of slightly complexifying the trait bounds in implementations--I'm not opinionated here.

We have to declare many trait bounds over what is guaranteed by `Vdaf` in our implementation (e.g. from WIP: https://github.com/divviup/janus/compare/aggregate#diff-89c14b8f557baa95520c0bf22f9aeaa0c885731a65acee9f629ce7937208663dR108-R113), so determining the correct "line" for what goes in `libprio-rs` vs what goes in implementations using the library would be helpful.

tgeoghegan (changes requested):
I think we should put the `Encode/Decode` impls on `struct OutputShare`, but not add the trait bound to the associated type `OutputShare` on `trait Vdaf`.

Implementations that are dealing concretely with `Prio3`, `Poplar1`, or some other `Vdaf` implementation whose `OutputShare` is `Encode + Decode` will Just Work, and implementations wishing to work generically with `trait Vdaf` can express a trait bound on `Vdaf::OutputShare` [as I described](https://github.com/divviup/libprio-rs/pull/202/files#r834769936) (note that you will probably have to do the same thing to require `Send + Sync` on various associated types).

src/vdaf.rs
BranLwyd: Removed Decode/Encode bound on OutputShare; will move to the implementation instead.

src/vdaf.rs
```diff
@@ -240,6 +242,18 @@ impl<F> AsRef<[F]> for OutputShare<F> {
     }
 }
 
+impl<F: Encode> Encode for OutputShare<F> {
+    fn encode(&self, bytes: &mut Vec<u8>) {
+        encode_u16_items(bytes, &(), &self.0)
```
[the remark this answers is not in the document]
BranLwyd: It makes sense to avoid the length prefix if we can, but I want to make sure I understand the suggestion.

AFAICT, the expected length of an output share/aggregate share will vary from VDAF to VDAF, and is not exposed by any existing VDAF methods. (e.g. prio3 output shares & aggregate shares will both have a length based on the `flp::Type` in use; for poplar1, output shares & aggregate shares will have a length based on the number of candidate prefixes in the aggregation param. FWIW prio3 does have an `output_len` method, but this is on the `prio3`-specific struct and not the Vdaf trait)

Is there a way for a user of this library to determine the expected length of an output share/aggregation share to pass to `decode_with_param`? Or is the suggestion that, if implementations want to deserialize these structures, they should serialize the additional information required for decoding themselves in a "wrapper" message around what is encoded by these `Encode` implementations?

(BTW, I found only one place that was calling `AggregateShare::decode_with_param`, and it is prio3-specific in that it assumes that the aggregate share length is equal to the output length & that those lengths are constant based on the VDAF [https://github.com/divviup/ppm-prototype/blob/40fcb1abe6b9480bfaa5beec56bf0d6fce474473/src/bin/collector.rs#L16]. So we may not be OK with `AggregateShare` as it stands!)

tgeoghegan (approved):

src/vdaf.rs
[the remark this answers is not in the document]
BranLwyd: Gentle ping @cjpatton. I'd like to resolve this.

After discussion with our group, two solutions come to mind:
1) Drop the `usize` decoding parameter entirely, logically decoding with no parameter. Have `AggregateShare` & `OutputShare` implement `TryFrom<&[u8]>` and `TryInto<Vec<u8>>`. (`Encode`/`Decode` are not appropriate because they operate on streams of values, while in this scenario we'd be decoding based on the total length of the vector, since we aren't encoding any kind of length information.)

2) Extend the `Vdaf` trait enough to have associated types for "Aggregate Share Decoding Param" & "Output Share Decoding Param", along with methods to retrieve the appropriate parameters. This would allow VDAFs more control over the decoding process, at the cost of complxifying the `Vdaf` trait even more.

Thoughts?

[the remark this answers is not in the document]
cjpatton: I think I see the problem: You can't decode an out or agg share without knowing the expected length, and right now there's nothing generic about `Vdaf` that tells you what length to expect. (1) works because `AggregateShare` and `OutputShare` are generic in the `FieldElement`, which specifies a a fixed length for each field element. (2) works because it requires the VDAF to specify this length.

My preference would be (1) because I think it's best to avoid restraining the VDAF API as much as possible, in particular for behaviors that are unspecified.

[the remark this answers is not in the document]
BranLwyd: Yep, that's exactly the problem.

Following your suggestion, I switched from Encode/Decode to From/Into<Vec<u8>>. The trait bounds required to make From/Into work on _references to_ the shares rather than the shares themselves are a little funky, but this is cleaner.

I left the bounds off of Vdaf::OutputShare, but updated the bounds on Vdaf::AggregateShare. This makes sense to me since an aggregate share _must_ be transmitted between aggregators (and so must be serialized), while an output share is not transmitted.

src/vdaf.rs
```diff
@@ -326,6 +343,7 @@ pub(crate) fn run_vdaf<V, M>(
 ) -> Result<V::AggregateResult, VdafError>
 where
     V: Client + Aggregator + Collector,
+    for<'a> &'a V::AggregateShare: Into<Vec<u8>>,
```
cjpatton, line 346: Why is this bound required here? `run_vdaf` exercises agg share serialization. (Same comment below.)

cjpatton (commented):
I'm a big fan of the minimal API changes here. A couple high-level questions: 
* It loos like the implementations of `From<&OutputShare<F>>` and `From<&AggregateShare<F>>` for `Vec<u8>` are identical. Does it make sense to refactor this so that they share the same implementation, then maybe `#[inline]` the implementation?
   * EDIT similar comment for `TryFrom<...>`.
* Do any of our tests exercise this serialization logic? If not, I think we might as well add some basic functionality tests.

src/vdaf.rs
BranLwyd: > I'm a big fan of the minimal API changes here. A couple high-level questions:
> 
> * It loos like the implementations of `From<&OutputShare<F>>` and `From<&AggregateShare<F>>` for `Vec<u8>` are identical. Does it make sense to refactor this so that they share the same implementation, then maybe `#[inline]` the implementation?
>   
>   * EDIT similar comment for `TryFrom<...>`.
> * Do any of our tests exercise this serialization logic? If not, I think we might as well add some basic functionality tests.

* I pulled the new `[Try]From` implementations to a common function; TBH, I was a little happier before (less indirection led to easier code reading IMO), but I definitely understand the desire to not repeat ourselves. Note I had to implement `From<Vec<F>>` for `OutputShare<F>` to make this work out.

* Not AFAICT, I added a couple of quick tests for round-tripping.

BranLwyd: > I'm a big fan of the minimal API changes here. A couple high-level questions:
> 
> * It loos like the implementations of `From<&OutputShare<F>>` and `From<&AggregateShare<F>>` for `Vec<u8>` are identical. Does it make sense to refactor this so that they share the same implementation, then maybe `#[inline]` the implementation?
>   
>   * EDIT similar comment for `TryFrom<...>`.
> * Do any of our tests exercise this serialization logic? If not, I think we might as well add some basic functionality tests.

* I pulled the new `[Try]From` implementations to a common function; TBH, I was a little happier before (less indirection led to easier code reading IMO), but I definitely understand the desire to not repeat ourselves. Note I had to implement `From<Vec<F>>` for `OutputShare<F>` to make this work out.

* Not AFAICT, I added a couple of quick tests for round-tripping.

BranLwyd:
> I'm a big fan of the minimal API changes here. A couple high-level questions:
> 
> * It loos like the implementations of `From<&OutputShare<F>>` and `From<&AggregateShare<F>>` for `Vec<u8>` are identical. Does it make sense to refactor this so that they share the same implementation, then maybe `#[inline]` the implementation?
>   
>   * EDIT similar comment for `TryFrom<...>`.
> * Do any of our tests exercise this serialization logic? If not, I think we might as well add some basic functionality tests.

* I pulled the new `[Try]From` implementations to a common function; TBH, I was a little happier before (less indirection led to easier code reading IMO), but I definitely understand the desire to not repeat ourselves. Note I had to implement `From<Vec<F>>` for `OutputShare<F>` to make this work out.

* Not AFAICT, I added a couple of quick tests for round-tripping.

cjpatton (approved):
LGTM, thanks @BranLwyd for entertaining my comments!

src/vdaf.rs
BranLwyd: It's required because of the bound that was on Vdaf:
```
where
    for<'a> &'a Self::AggregateShare: Into<Vec<u8>>,
```

For reasons I don't fully understand, this constraint becomes "sticky" & must be re-stated everywhere a generic `Vdaf` parameter is used. See https://stackoverflow.com/questions/50090578/how-to-write-a-trait-bound-for-a-reference-to-an-associated-type-on-the-trait-it for more information.

Doing better will require a feature called "generic associated types" (per the SO above). Unfortunately, GATs are not stabilized yet; it looks like they are set to be stabilized in a few months (https://github.com/rust-lang/rust/issues/44265#issuecomment-1066173042).

Other than GATs, we can also state the required constraints in terms of a new trait, which allows us to refer to references to the associated types without the need for trait bounds on references to the associated types. While the new trait would be simple enough & would be easy to refactor into equivalent From/Into implementations once GATs are stable, I avoided doing so because this is more complex than copying the trait bound everywhere until GATs land.

FTR, the new trait would look something like:
```
/// ByteConvertible allows for a type to be encoded into bytes, and attempt to be decoded from
/// bytes.
pub trait ByteConvertible: Sized {
    /// The type of the error returned by try_from_bytes.
    type Error;

    /// Converts a value to an equivalent byte sequence.
    fn to_bytes(&self) -> Vec<u8>;

    /// Attempts to convert a byte sequence into a value.
    fn try_from_bytes<B: AsRef<[u8]>>(bytes: B) -> Result<Self, Self::Error>;
}
```

PR #204, opened 2022-04-06
Extend blanket `Parameterized{En,De}code` impls

tgeoghegan:
Extends the implementation of `ParameterizedDecode<()> for D: Decode` to
be `ParameterizedDecode<T> for D: Decode, T`, and similarly extends the
blanket `ParameterizedEncode` impl. This way, VDAFs like the [fake one
in `janus`][1] that use `()` as their associated types can still satisfy
bounds like `Self::VerifyParam: ParameterizedDecode<Self>`. `()`
satisfies the trivial trait bound `T` so this is strictly more
permissive than what was there before.

[1]: https://github.com/divviup/janus/blob/main/janus_server/src/aggregator.rs#L789

### DIFF
--- a/src/codec.rs
+++ b/src/codec.rs
@@ -70,10 +70,11 @@ pub trait ParameterizedDecode<P>: Sized {
     }
 }
 
-// Provide a blanket implementation so that any Decode can be used as a ParameterizedDecode<()>.
-impl<D: Decode + ?Sized> ParameterizedDecode<()> for D {
+// Provide a blanket implementation so that any Decode can be used as a ParameterizedDecode<T> for
+// any T.
+impl<D: Decode + ?Sized, T> ParameterizedDecode<T> for D {
     fn decode_with_param(
-        _decoding_parameter: &(),
+        _decoding_parameter: &T,
         bytes: &mut Cursor<&[u8]>,
     ) -> Result<Self, CodecError> {
         Self::decode(bytes)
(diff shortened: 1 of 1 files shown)


tgeoghegan:
The motivation here: [`Prio3VerifyParam`](https://github.com/divviup/libprio-rs/blob/229cd9c45924c7dae3ba754a6eb46b2a05ca8451/src/vdaf/prio3.rs#L438) is `ParameterizedDecode<Prio3<...>>` and [`Poplar1VerifyParam`](https://github.com/divviup/libprio-rs/blob/229cd9c45924c7dae3ba754a6eb46b2a05ca8451/src/vdaf/poplar1.rs#L463) is `ParameterizedDecode<Poplar1<...>>`. The fake VDAF in Janus follows that pattern, and its `VerifyParam` is `()`, so we need this broader blanket impl so that [#55](https://github.com/divviup/janus/pull/55) can do:
```
impl<A, C> Aggregator<A, C>
where
    A: 'static + vdaf::Aggregator,
    A::AggregationParam: Send + Sync,
    A::VerifyParam: ParameterizedDecode<A>, // <-- this is the trait bound enabled by this change
    for<'a> &'a A::AggregateShare: Into<Vec<u8>>,
    A::PrepareStep: Send + Sync + Encode + ParameterizedDecode<A::VerifyParam>,
    A::OutputShare: Send + Sync + for<'a> TryFrom<&'a [u8]>,
    for<'a> &'a A::OutputShare: Into<Vec<u8>>,
    C: Clock,
{ ... }
```

I need that trait bound on `janus::aggregator::Aggregator` so that its constructor can decode a verify param from bytes.

divergentdave (approved):

BranLwyd (approved):

PR #206, opened 2022-04-18
Fix in docstring of FieldParameters::roots

divergentdave:
I noticed an error in a doc comment, I think what it was getting at is that `roots[0]` is the 2^0'th root of unity, i.e. the 1st root of unity, i.e. the identity.


cjpatton (approved):
Nice catch :)

divergentdave:
Rebasing this to fix CI

PR #207, opened 2022-04-18
clippy: rename Montgomery conversion methods

divergentdave:
Our CI is failing with [`clippy::wrong_self_convention`](https://rust-lang.github.io/rust-clippy/master/index.html#wrong_self_convention). This PR renames `FieldParameters::elem()` to `FieldParameters::map()`, and `FieldParameters::from_elem()` to `FieldParameters::inverse_map()` to resolve the issue.

### DIFF
--- a/src/field.rs
+++ b/src/field.rs
@@ -270,7 +270,7 @@ macro_rules! make_field {
                 if int >= $fp.p {
                     return Err(FieldError::ModulusOverflow);
                 }
-                Ok(Self($fp.elem(int)))
+                Ok(Self($fp.map(int)))
             }
         }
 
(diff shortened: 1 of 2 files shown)


BranLwyd (approved):

cjpatton (commented):
What about `into_mont() / from_mont()`?

cjpatton:
> ... (methods with the `into_` prefix are expected to take ownership of self) I'd prefer to avoid `from_` and `into_` prefixes, as those are conventionally for conversions of owned values, taking single arguments, whereas these need to take in two arguments, and `&self` is not itself being converted into anything.

I see, thanks for clarifying. It may still be useful to have the function names denote the fact that we're converting between integer and Montgomery rempresentation. What about `map_to_elem` and `map_to_elem_inv`?

divergentdave:
How about `map_to_montgomery` and `map_to_residue`? I fear "element" may be confusing, as both the domain and range are fields, and the doc comments talk about "field elements" on both sides. (runner up idea: `map_to_mont`/`map_to_mont_inv`)

cjpatton:
Those names sound good to me.

cjpatton (approved):
Looks good. One nit, in case clippy doesn't complain: Would be nice to remove the "map_to" prefix :)

Issue #216, opened 2022-05-13
Changes for vdaf-01


cjpatton:
Version 0.7.0 is locked to vdaf-00. Upgrading to vdaf-01 will require some changes, which this issue will track.

Because these changes will break compatibility with vdaf-00, it will be a good idea to cut a minor version before we start working on them (e.,g., move `x.y.z` yo `x.y.(z+1)`). Note that #210 should be solved before cutting the version.

Changes in the queue:
* 2022-05-13 https://github.com/cfrg/draft-irtf-cfrg-vdaf/pull/62 Drop public param and refactor verify parm
* 2022-05-17 https://github.com/cfrg/draft-irtf-cfrg-vdaf/pull/64 Breaking change to Prio3
* Update references to vdaf-01
* Update test vectors

tgeoghegan:
I'd like to release a prio 0.7.1 that implements vdaf-00, and then we can start implementing vdaf-01 and eventually release prio 0.8.x from that.

Janus currently pins an unreleased commit of `libprio` in its `Cargo.toml` that is a few commits beyond `prio 0.7.0`. I've been holding off on releasing `prio 0.7.1` until we can be confident that the whole DAP interop branch can be implemented (we might still need to add some trait bounds to this-and-that).

So I think we should:

- finish Janus implementation of DAP/PPM interop branch + vdaf-00, taking whatever libprio-rs changes are needed along the way into `main`
- release prio 0.7.1, noting in the README that it implements vdaf-00
- start taking changes into libprio-rs `main` to support vdaf-01
- eventually release prio 0.8.0, noting in its README that it implements vdaf-01, and if you want vdaf-00, you should use prio 0.7

If you want to get started on vdaf-01 support in libprio-rs right away, then we could cut a `release/0.7` branch from `main` and continue to take vdaf-00 changes there, and then take on vdaf-01 changes on main.

>minor version before we start working on them (e.,g., move x.y.z yo x.y.(z+1))

n.b. in [SemVer](https://semver.org/) `z` is the _patch_ version.

cjpatton:
Sounds like a plan to me, though it's not strictly necessary to make the change in 0.8.0. If an interm releasse is needed for some reason, we can do vdaf-01 in 0.9.0.

I don't think we're in a rush to get started on vdaf-01 support, especially since there are no breaking changes (yet).

PR #218, opened 2022-05-17
vdaf: Have prepare_init() output the first-round prep message

cjpatton:
Modifies the Aggregator API so that `prepare_init()` outputs the first
round Prepare message. This is less general, since it excludes VDAFs
that may have 0 rounds of communication during the Prepare phase. The
upside is that it eliminates a number of rough edges in both the API and
the implementation. For example, there is never a need to serialize the
initial state prior to producing the first-round message.

FWIW, I think this change is useful regardless of if/how we resolve https://github.com/cfrg/draft-irtf-cfrg-vdaf/issues/60.

### DIFF
--- a/src/vdaf.rs
+++ b/src/vdaf.rs
@@ -175,7 +175,7 @@ where
         agg_param: &Self::AggregationParam,
         nonce: &[u8],
         input_share: &Self::InputShare,
-    ) -> Result<Self::PrepareState, VdafError>;
+    ) -> Result<(Self::PrepareState, Self::PrepareMessage), VdafError>;
 
     /// Preprocess a round of prepare messages into a single input to [`Aggregator::prepare_step`].
     fn prepare_preprocess<M: IntoIterator<Item = Self::PrepareMessage>>(
@@ -193,7 +193,7 @@ where
     fn prepare_step(
         &self,
         state: Self::PrepareState,
-        input: Option<Self::PrepareMessage>,
+        input: Self::PrepareMessage,
     ) -> PrepareTransition<Self>;
 
     /// Aggregates a sequence of output shares into an aggregate share.
@@ -417,8 +417,9 @@ where
         .map(|input_share| input_share.get_encoded());
 
     let mut states = Vec::new();
+    let mut outbound = Vec::new();
     for (verify_param, input_share) in verify_params.iter().zip(input_shares) {
-        let state = vdaf.prepare_init(
+        let (state, msg) = vdaf.prepare_init(
             verify_param,
             agg_param,
             nonce,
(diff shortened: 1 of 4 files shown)


tgeoghegan:
I haven't reviewed these changes yet, but IIUC this is part of implementing vdaf-01, so we should do something like what is described in https://github.com/divviup/libprio-rs/issues/216 before we merge this to main.

cjpatton:
I don't think it's strictly related to changes in -01, since we're not actually changing the crypto. However I'm fine with holding off on merging.

BranLwyd (approved):
I like this change! It fits well with how Janus implements things currently, and smooths off (what I understand to be) a rough-edge of the API.

tgeoghegan (approved):
I agree that this change is good, I'm just worried that we will end up doing our interop tests with an implementation of an unclear mix of `vdaf-00` and `vdaf-01` which may make it harder to debug problems. So that's why I want to be explicit about whether we are taking this change into an implementation of `vdaf-00` or `-01`.

cjpatton:
Ack, @tgeoghegan -- I'm happy to hold off on merging until we're closer to -01.

src/vdaf/prio3.rs
```diff
@@ -611,9 +610,6 @@ impl<F: FieldElement, const L: usize> Encode for Prio3PrepareStep<F, L> {
     ///
     /// Panics if the step is not in the "waiting" state. See [`Prio3PrepareStep`] for details.
```
cjpatton, line 611: Remove this comment, as it's no longer relevant.

cjpatton: Done (similarly above).

src/vdaf.rs
```diff
@@ -415,18 +415,24 @@ where
...
                 .expect("failed to decode input share"),
         )?;
         states.push(state);
+        outbound.push(msg.get_encoded());
     }
 
-    let mut inbound = None;
+    let mut inbound = vdaf.prepare_preprocess(outbound.iter().map(|encoded| {
+        V::PrepareMessage::get_decoded_with_param(&states[0], encoded)
+            .expect("failed to decode papare message")
```
tgeoghegan, line 433:
```suggestion
            .expect("failed to decode prepare message")
```
But in any case, why not propagate the error? I think something like
```
let mut inbound = vdaf.prepare_preprocess(outbound.iter().map(|encoded| V::PrepareMessage::get_decoded_with_param(...)).transpose().collect::<Vec<Result<_, _>>>())
```
might work.

src/vdaf.rs
```diff
@@ -447,10 +453,10 @@ where
 
         if outbound.len() == vdaf.num_aggregators() {
             // Another round is required before output shares are computed.
-            inbound = Some(vdaf.prepare_preprocess(outbound.iter().map(|encoded| {
+            inbound = vdaf.prepare_preprocess(outbound.iter().map(|encoded| {
                 V::PrepareMessage::get_decoded_with_param(&states[0], encoded)
                     .expect("failed to decode papare message")
```
tgeoghegan, line 458:
```suggestion
                    .expect("failed to decode prepare message")
```
But as above, would prefer to propagate the error instead of crashing.

tgeoghegan (changes requested):

tgeoghegan (approved):

PR #219, opened 2022-05-17
Compare FieldElement via Montgomery representation

### LINKED ISSUE #208: Optimization: Compare field elements by their Montgomery representation
I noticed that the implementations of `PartialEq` and `Hash` for our fields convert from the internal representation, used for Montgomery multiplication, back to their p-residue form. We could save several multiplications per comparison if we used the Montgomery representation directly. Quoting from [the paper](https://www.ams.org/journals/mcom/1985-44-170/S0025-5718-1985-0777282-X/S0025-5718-1985-0777282-X.pdf), "Also unchanged are the algorithms for ... equality/inequality test". ~~It's possible we may need to perform the last conditional subtraction of p in `modp()`, as it appears that part of REDC has been broken out of `FieldParameters::mul()`.~~

divergentdave:
This changes field element comparison (and hashing) to use Montgomery representations directly, rather than converting to the natural representation first. The correctness of this change is dependent on an invariant that the integer representing a field element is never greater than the modulus, p, so comments are added throughout to document it. (Previously, the conversion function would have reduced any too-large values as part of its multiplication) The field element tests are expanded to exercise the Hash implementation.

This closes #208. cc @armfazh

### DIFF
--- a/src/field.rs
+++ b/src/field.rs
@@ -235,9 +235,10 @@ macro_rules! make_field {
         /// This structure represents a field element in a prime order field. The concrete
         /// representation of the element is via the Montgomery domain. For an element n in GF(p),
         /// we store n * R^-1 mod p (where R is a given power of two). This representation enables
-        /// using a more efficient multiplication algorithm, at the expense of having to convert
-        /// elements between their Montgomery domain representation and natural representation.
-        /// For calculations with many multiplications or exponentiations, this is worthwhile.
+        /// using a more efficient (and branchless) multiplication algorithm, at the expense of
+        /// having to convert elements between their Montgomery domain representation and natural
+        /// representation. For calculations with many multiplications or exponentiations, this is
+        /// worthwhile.
         ///
         /// As an invariant, this integer representing the field element in the Montgomery domain
         /// must be less than the prime p.
--- a/src/field.rs
+++ b/src/field.rs
@@ -231,15 +231,16 @@ macro_rules! make_field {
         $elem:ident, $int:ident, $fp:ident, $encoding_size:literal, $encoding_order:expr,
     ) => {
         $(#[$meta])*
-        // This structure represents a field element in a prime order field. The concrete
-        // representation of the element is via the Montgomery domain. For an element n in GF(p),
-        // we store n * R^-1 mod p (where R is a given power of two). This representation enables
-        // using a more efficient multiplication algorithm, at the expense of having to convert
-        // elements between their Montgomery domain representation and natural representation.
-        // For calculations with many multiplications or exponentiations, this is worthwhile.
-        //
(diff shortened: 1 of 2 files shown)


src/field.rs
```diff
@@ -279,7 +290,12 @@ macro_rules! make_field {
                 // The fields included in this comparison MUST match the fields
                 // used in Hash::hash
                 // https://doc.rust-lang.org/std/hash/trait.Hash.html#hash-and-eq
-                $fp.residue(self.0) == $fp.residue(rhs.0)
+
+                // Check the invariant that the integer representation is fully reduced.
+                assert!(self.0 < $fp.p);
+                assert!(rhs.0 < $fp.p);
```
cjpatton, line 296: Is it feasible/safe to only run these checks in tests? Would be nice to safe a few cycles here. (Similar question below.)

src/field.rs
```diff
@@ -230,6 +230,15 @@ macro_rules! make_field {
         $elem:ident, $int:ident, $fp:ident, $encoding_size:literal, $encoding_order:expr,
     ) => {
         $(#[$meta])*
+        // This structure represents a field element in a prime order field. The concrete
+        // representation of the element is via the Montgomery domain. For an element n in GF(p),
+        // we store n * R^-1 mod p (where R is a given power of two). This representation enables
+        // using a more efficient multiplication algorithm, at the expense of having to convert
+        // elements between their Montgomery domain representation and natural representation.
+        // For calculations with many multiplications or exponentiations, this is worthwhile.
```
cjpatton, line 238: Another reason for the choice of multiplication algorithm is to make it closer to constant time.

src/field.rs
divergentdave: Sure, I can swap it out for `debug_assert!`

src/field.rs
```diff
@@ -749,6 +790,46 @@ mod tests {
         let serialized_vec = F::slice_into_byte_vec(&test_inputs);
         let deserialized = F::byte_slice_into_vec(&serialized_vec).unwrap();
         assert_eq!(deserialized, test_inputs);
+
```
armfazh, line 793: it would be good to add a test that tries to create an element from a u128 value not in range [0,p).

cjpatton (approved):

src/field.rs
```diff
@@ -230,6 +231,15 @@ macro_rules! make_field {
         $elem:ident, $int:ident, $fp:ident, $encoding_size:literal, $encoding_order:expr,
     ) => {
         $(#[$meta])*
+        // This structure represents a field element in a prime order field. The concrete
```
tgeoghegan, line 234: Doccomment should start with `///`

armfazh (approved):

PR #222, opened 2022-05-19
Add a test of Prng rejection sampling

divergentdave:
This test uses a crafted test vector to confirm that rejection sampling is performed correctly when the AES-CTR output is larger than the modulus.

### DIFF
--- a/src/prng.rs
+++ b/src/prng.rs
@@ -107,7 +107,11 @@ where
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::field::FieldPriov2;
+    use crate::{
+        codec::Decode,
+        field::{Field96, FieldPriov2},
+        vdaf::prg::{Prg, PrgAes128, Seed},
+    };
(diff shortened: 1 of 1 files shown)


src/prng.rs
```diff
@@ -182,4 +186,18 @@ mod tests {
             .take(length)
             .collect()
     }
+
+    #[test]
+    fn rejection_sampling_test_vector() {
+        // These constants were found in a brute-force search, and they test that the PRG performs
+        // rejection sampling correctly when raw AES-CTR output exceeds the prime modulus.
+        let seed_stream = PrgAes128::seed_stream(
+            &Seed::get_decoded(&[0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0x5f]).unwrap(),
```
cjpatton, line 195: nit: Write the last byte of the seed in decimal for consistency.

tgeoghegan (approved):

PR #225, opened 2022-05-20
vdaf: Distinguish prep shares from prep messages

### LINKED ISSUE #220: vdaf: `Aggregator`: Handling prep-message shares and prep messages differently
A "prep-message share" is emitted by each Aggregator during each round of the prep phase. These are preprocessed into a "prep message" using [`Vdaf.prep_shares_to_prep()`](https://cfrg.github.io/draft-irtf-cfrg-vdaf/draft-irtf-cfrg-vdaf.html#section-5).

https://github.com/cfrg/draft-irtf-cfrg-vdaf/pull/64 introduces a change to Prio3 that results in prep shares and prep messages having different contents. (In particular, this PR avoids including the proof verifier in the prep message, thereby saving on communication overhead.) In general there's no reason to think prep shares and prep messages need to have the same structure, it just so happens to be true for (the current versions of) our VDAFs.

In the VDAF syntax of the spec, this is not an issue because the caller is responsible for deserializing these inputs. However, this does present a problem for the current `Aggregator` API. I've been working on a port of Prio2 to VDAF (#214) that illustrates this problem: https://github.com/divviup/libprio-rs/compare/cjpatton/vdaf-first-round...cjpatton/prio2. Setting aside the merits of this idea, the code illustrates the crucial problem: Because the same struct is used for both (`Aggregator::PrepareMessage`), we need some way to distinguish one variant from the other. The same is true for the new version of Prio3.

Here are some remedies I've thought of:
1. Spec change: Have the encoding describe its variant, e.g., using a type byte. This is not desirable because we would like to keep the encoding as compact as possible.
2. Implementation change: Use the length of the raw bytes to distinguish between the variants. This is currently impossible given the way we've defined the `Decode`, which takes in a byte stream rather than a byte slice. (And rightfully so, I think.)
3. API change: Move serialization of prep shares and messages into `Aggregator` so that the API just takes in raw byte strings for each. This is not desirable because it's inconsistent with the rest of the API. Plus, in DAP the leader would end up unnecessarily deserializing its own prep share.
4. API change: Define a new associated type for `Aggregator`, call it `PrepareShare`. This seems like the only viable option to me.
5. Any other possibilities?

cjpatton:
Based on #224 (merge that first).
Closes #220.

Modifies `Aggregator` by adding a new associated type, `PrepareShare`,
that is output by each Aggregator at each step of the Prepare
computation. The `prepare_preprocess()` method takes a sequence of
`PrepareShare` and outputs the `PrepareMessage` consumed by the
Aggregator in the next round.

### DIFF
--- a/src/vdaf.rs
+++ b/src/vdaf.rs
@@ -163,7 +163,10 @@ where
     /// State of the Aggregator during the Prepare process.
     type PrepareState: Clone + Debug;
 
-    /// The type of messages exchanged among the Aggregators during the Prepare process.
+    /// The type of messages broadcast by each aggregator at each round of the Prepare Process.
+    type PrepareShare: Clone + Debug + ParameterizedDecode<Self::PrepareState> + Encode;
+
+    /// Result of preprocessing a round of preparation shares.
     type PrepareMessage: Clone + Debug + ParameterizedDecode<Self::PrepareState> + Encode;
 
     /// Begins the Prepare process with the other Aggregators. The [`Self::PrepareState`] returned
@@ -175,21 +178,21 @@ where
         agg_param: &Self::AggregationParam,
         nonce: &[u8],
         input_share: &Self::InputShare,
-    ) -> Result<(Self::PrepareState, Self::PrepareMessage), VdafError>;
+    ) -> Result<(Self::PrepareState, Self::PrepareShare), VdafError>;
 
-    /// Preprocess a round of prepare messages into a single input to [`Aggregator::prepare_step`].
-    fn prepare_preprocess<M: IntoIterator<Item = Self::PrepareMessage>>(
+    /// Preprocess a round of preparation shares into a single input to [`Aggregator::prepare_step`].
+    fn prepare_preprocess<M: IntoIterator<Item = Self::PrepareShare>>(
         &self,
         inputs: M,
     ) -> Result<Self::PrepareMessage, VdafError>;
 
     /// Compute the next state transition from the current state and the previous round of input
     /// messages. If this returns [`PrepareTransition::Continue`], then the returned
-    /// [`Self::PrepareMessage`] should be combined with the other aggregator's `PrepareMessage`s
-    /// from this round and passed into another call to this method. This continues until this
-    /// method returns [`PrepareTransition::Finish`], at which point the returned output share may
-    /// be aggregated. If the method returns an error, the aggregator should consider its input
-    /// share invalid and not attempt to process it any further.
+    /// [`Self::PrepareShare`] should be combined with the other Aggregators' `PrepareShare`s from
+    /// this round and passed into another call to this method. This continues until this method
+    /// returns [`PrepareTransition::Finish`], at which point the returned output share may be
+    /// aggregated. If the method returns an error, the aggregator should consider its input share
+    /// invalid and not attempt to process it any further.
     fn prepare_step(
         &self,
         state: Self::PrepareState,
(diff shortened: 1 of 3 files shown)


src/vdaf.rs
```diff
@@ -164,7 +164,10 @@ where
     /// State of the Aggregator during the Prepare process.
     type PrepareStep: Clone + Debug;
 
-    /// The type of messages exchanged among the Aggregators during the Prepare process.
+    /// The type of message broadcast by each aggregator at each round.
+    type PrepareBroadcast: Clone + Debug + ParameterizedDecode<Self::PrepareStep> + Encode;
```
BranLwyd, line 168: Interestingly, DAP's model of communication would suggest naming these the other way around. In each DAP round, the leader sends to all helpers (i.e. broadcasts) a combined `PrepareMessage`. Each helper will then compute its next preparation state & `PrepareBroadcast` message by calling `prepare_step`, and send to the leader (i.e. unicast) that `PrepareBroadcast` message. The leader will then combine these messages into a `PrepareMessage` to prepare for the next round of communication.

I see the VDAF model though--I suppose the VDAF model is that every participant broadcasts their `PrepareBroadcast` message to all other participants, who then combine all of the broadcasted messages into a `PrepareMessage`.

I don't want to bikeshed too hard on naming, if for no other reason than DAP currently has exactly one helper so the distinction between broadcast & unicast is somewhat arbitrary. If other folks are weirded out by this too, my suggestion would be something like `PrepareMessage` and `CombinedPrepareMessage`? [also, have we considered if DAP's strategy of having only the leader produce combined messages is OK--could it give an evil leader some advantage if they sent a different `PrepareMessage` than would be honestly computed? probably this has already been thought about.]

BranLwyd (approved):
LGTM, modulo one piece of naming trivia.

src/vdaf.rs
BranLwyd: 👍🏻 to using `PrepareShare`/`PrepareMessage`, IMO this choice makes things understandable both in the VDAF's "broadcast" communication model & DAP's "star" communication model. (plus, matching the specification will likely make things easier to understand for folks attempting to translate between the two)

cjpatton: Done.

src/vdaf.rs
```diff
@@ -164,7 +164,10 @@ where
     /// State of the Aggregator during the Prepare process.
     type PrepareStep: Clone + Debug;
 
-    /// The type of messages exchanged among the Aggregators during the Prepare process.
+    /// The type of message broadcast by each aggregator at each round.
+    type PrepareShare: Clone + Debug + ParameterizedDecode<Self::PrepareStep> + Encode;
+
+    /// Result of preprocessing a round of broadcast messages.
```
BranLwyd, line 170:
```suggestion
    /// Result of preprocessing a round of preparation shares.
```

src/vdaf.rs
```diff
@@ -176,10 +179,10 @@ where
         agg_param: &Self::AggregationParam,
         nonce: &[u8],
         input_share: &Self::InputShare,
-    ) -> Result<(Self::PrepareStep, Self::PrepareMessage), VdafError>;
+    ) -> Result<(Self::PrepareStep, Self::PrepareShare), VdafError>;
 
     /// Preprocess a round of prepare messages into a single input to [`Aggregator::prepare_step`].
```
BranLwyd, line 184:
```suggestion
    /// Preprocess a round of preparation shares into a single input to [`Aggregator::prepare_step`].
```

BranLwyd (approved):
LGTM, w/ two comment nits.

tgeoghegan (approved):

src/vdaf.rs
```diff
@@ -191,12 +194,13 @@ where
     /// method returns [`PrepareTransition::Finish`], at which point the returned output share may
     /// be aggregated. If the method returns an error, the aggregator should consider its input
     /// share invalid and not attempt to process it any further.
+    #[allow(clippy::type_complexity)]
     fn prepare_step(
```
divergentdave, line 198: The docs for `prepare_init()` and `prepare_step()` should be revised to refer to `PrepareShare`

PR #228, opened 2022-05-21
vdaf: Drop public param and refactor verify param

cjpatton:
Based on #226 (merge that first).
Partially addresses #216.

Aligns the API with the syntax in vdaf-01 by removing the public parameter and replacing the verification parameter with an aggregator key and ID.

2022-05-25: Ready for review.
2022-05-20: The PR is currently a draft. Befor merging I need to fix up Poplar1 as well. Before I get to deep, I wanted to get folks' feedback on the API changes and think through any rough edges they might create for applications. (On the whole I think this is a huge improvement.)

### DIFF
--- a/src/vdaf/poplar1.rs
+++ b/src/vdaf/poplar1.rs
@@ -276,20 +276,23 @@ impl<I: Idpf<2, 2>, const L: usize> Encode for Poplar1InputShare<I, L> {
     }
 }
 
-impl<I: Idpf<2, 2>, const L: usize> ParameterizedDecode<Poplar1VerifyParam<L>>
+impl<'a, I, P, const L: usize> ParameterizedDecode<(&'a Poplar1<I, P, L>, usize)>
     for Poplar1InputShare<I, L>
+where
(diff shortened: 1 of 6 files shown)


src/vdaf/poplar1.rs
```diff
@@ -531,8 +518,9 @@ where
         let level = get_level(agg_param)?;
 
         // Derive the verification randomness.
-        let mut verify_rand_prng: Prng<I::Field, _> =
-            Prng::from_seed_stream(P::seed_stream(&verify_param.rand_init, nonce));
+        let mut p = P::init(agg_key);
+        p.update(nonce);
+        let mut verify_rand_prng: Prng<I::Field, _> = Prng::from_seed_stream(p.into_seed_stream());
```
cjpatton, line 463: Should we modify `P::seed_stream` to take a `[u8; L]` rather than a `Seed<L>`? I made a similar change to `P::init` to avoid a copy. (See below.)

divergentdave (approved):

PR #232, opened 2022-05-22
Refactor PrepareTransition<S, M, O> to PrepareTransition<V: Aggregator>.

### LINKED ISSUE #231: Refactor PrepareTransition<S, M, O> to PrepareTransition<V>
Where `V: Vdaf`; the previous `S`, `M`, `O` parameters would become `V::PrepareStep`, `V::PrepareMessage`, `V::OutputShare`, respectively.

(marking this bug as a `vdaf-01` change, as it is breaking if only due to the change in generic parameters)

BranLwyd:
This makes the relation between the internal types clearer: in
practices, S was a V::PrepareStep, M was a V::PrepareMessage, and O was
a V::OutputShare for some V: Aggregator. This is now explicit in the
type definition.

Please let me know if I should hold off on merging this, as it is a breaking change.

Closes #231.

### DIFF
--- a/src/vdaf.rs
+++ b/src/vdaf.rs
@@ -194,7 +194,7 @@ where
         &self,
         state: Self::PrepareStep,
         input: Option<Self::PrepareMessage>,
-    ) -> PrepareTransition<Self::PrepareStep, Self::PrepareMessage, Self::OutputShare>;
+    ) -> PrepareTransition<Self>;
 
     /// Aggregates a sequence of output shares into an aggregate share.
     fn aggregate<M: IntoIterator<Item = Self::OutputShare>>(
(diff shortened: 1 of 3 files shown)


divergentdave (approved):
This change looks good to me. It would make sense to batch this breakage with VDAF changes to prep_init, etc., but given that 0.7.1 has been cut, and we have plans to release 0.7.2 from another branch if needed, I think we could go ahead and merge it.

src/vdaf.rs
```diff
@@ -219,12 +219,15 @@ where
 
 /// A state transition of an Aggregator during the Prepare process.
 #[derive(Debug)]
-pub enum PrepareTransition<S, M, O> {
+pub enum PrepareTransition<V: Aggregator>
+where
+    for<'a> &'a V::AggregateShare: Into<Vec<u8>>,
```
tgeoghegan, line 224: If we're going to put this trait bound in `libprio`, could we put it right on `vdaf::Aggregator`? I think then we could delete corresponding trait bounds from quite a few places in Janus.

tgeoghegan (approved):

src/vdaf.rs
BranLwyd: It is on `vdaf::Aggregator` [L161 of this commit]; unfortunately, Rust makes this trait bound "infectious", rather than having the `V: Aggregator` bound itself imply the required bound on `&V::AggregateShare` in this case.

I thought GATs might simplify this, but I hopped into nightly temporarily to try them out and in some cases the bound is still infectious. It did allow eliding the bound in some other cases, though.

Unfortunately, I think this is a gap in Rust's generics currently. We could work around it by factoring the bound differently: instead of `Into<_>` on `&AggregateShare`, create a new bound named something like `Serializable`, specify it on `AggregateShare` per normal, and make `Serializable::serialize` take `&self` rather than `self`. (that is, the problem AFAICT is that Rust's support for expressing bounds on _references_ to associated types is not currently great)

PR #236, opened 2022-05-26
Clean up API and documentation

### LINKED ISSUE #216: Changes for vdaf-01

cjpatton:
Closes #216.

* Update references to vdaf-01
* Clean up documentation a bit
* Tweak the Prio3 API slightly to make it more self-documenting.

### DIFF
--- a/src/vdaf.rs
+++ b/src/vdaf.rs
@@ -1,9 +1,9 @@
 // SPDX-License-Identifier: MPL-2.0
 
 //! This module defines an API for Verifiable Distributed Aggregation Functions (VDAFs) and
-//! implements two constructions described in [[draft-irtf-cfrg-vdaf-00]].
+//! implements two constructions described in [[draft-irtf-cfrg-vdaf-01]].
 //!
-//! [draft-irtf-cfrg-vdaf-00]: https://datatracker.ietf.org/doc/draft-irtf-cfrg-vdaf/00/
+//! [draft-irtf-cfrg-vdaf-01]: https://datatracker.ietf.org/doc/draft-irtf-cfrg-vdaf/01/
 
 use crate::codec::{CodecError, Decode, Encode, ParameterizedDecode};
 use crate::field::{FieldElement, FieldError};
--- a/src/vdaf.rs
+++ b/src/vdaf.rs
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: MPL-2.0
 
-//! This module defines an API for Verifiable Distributed Aggregation Functions (VDAFs) and
-//! implements two constructions described in [[draft-irtf-cfrg-vdaf-01]].
+//! Verifiable Distributed Aggregation Functions (VDAFs) as described in
+//! [[draft-irtf-cfrg-vdaf-01]].
 //!
 //! [draft-irtf-cfrg-vdaf-01]: https://datatracker.ietf.org/doc/draft-irtf-cfrg-vdaf/01/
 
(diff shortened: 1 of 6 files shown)


src/vdaf.rs
```diff
@@ -220,6 +220,23 @@ where
...
+{
+    /// Return the output share.
+    ///
+    /// # Panics
+    ///
+    /// This method panics if processing is not complete.
+    pub fn unwrap_finish(self) -> V::OutputShare {
+        match self {
+            Self::Finish(out_share) => out_share,
+            _ => panic!("tried to unwrap continue as finishe"),
```
tgeoghegan, line 235:
```suggestion
            _ => panic!("tried to unwrap continue as finish"),
```

src/vdaf.rs
```diff
@@ -220,6 +220,23 @@ where
...
+impl<V: Aggregator<L>, const L: usize> PrepareTransition<V, L>
+where
+    for<'a> &'a V::AggregateShare: Into<Vec<u8>>,
+{
+    /// Return the output share.
+    ///
+    /// # Panics
+    ///
+    /// This method panics if processing is not complete.
+    pub fn unwrap_finish(self) -> V::OutputShare {
```
tgeoghegan, line 232: This appears to only be used in a doctest. Can you restrict it to `[cfg(test)]` or `[cfg(doc)]` or whichever is appropriate?

src/vdaf/prio3.rs
```diff
@@ -196,12 +208,73 @@ fn check_num_aggregators(num_aggregators: u8) -> Result<(), VdafError> {
...
+/// assert_eq!(agg_res, Prio3Result(3));
+/// ```
+///
+/// [draft-irtf-cfrg-vdaf-01]: https://datatracker.ietf.org/doc/draft-irtf-cfrg-vdaf/01/
 #[derive(Clone, Debug)]
 pub struct Prio3<T, A, P, const L: usize>
 where
     T: Type,
-    A: Clone + Debug,
+    A: TryFrom<AggregateShare<T::Field>, Error = VdafError>,
```
tgeoghegan, line 277: Why did we lose the `Clone + Debug` trait bound here?

tgeoghegan (changes requested):
Doc improvements are great, thank you!

src/vdaf/prio3.rs
cjpatton: No reason, added back.

src/vdaf.rs
cjpatton: It seems icky to not export something in the public API that's used in a doctest.

tgeoghegan: I see your point, but exporting this from the public API implies that it's intended for usage by clients of libprio which isn't really true. Since you only use it one place, could you delete the method and "inline" the `match` block into the doctest on `struct Prio3`,

cjpatton: Done!

tgeoghegan (approved):

PR #238, opened 2022-05-27
 prio3: Move agg result definition to `Type`

### LINKED ISSUE #235: `Prio3`: What is the right abstraction boundary for measurements/results?
As of #228, this our API for `Prio3`:
```rust
  impl<T, A, P, const L: usize> Vdaf for Prio3<T, A, P, L>
  where
      T: Type,
      A: Clone + Debug + Sync + Send,
      P: Prg<L>,
  {
      type Measurement = T::Measurement;
      type AggregateResult = A;
      type AggregationParam = ();
      type InputShare = Prio3InputShare<T::Field, L>;
      type OutputShare = OutputShare<T::Field>;
      type AggregateShare = AggregateShare<T::Field>;

      fn num_aggregators(&self) -> usize {
          self.num_aggregators as usize
      }
  }
```

There are two features of this API I want to discuss:
1. `Prio3::Measurement` is determined by the underlying `T::Measurement`. Typically this is either the integer representation of elements of the finite field (`T::Field`) or a vector of such integers. However, this design also envisions more complex measurement types, such as some fixed point numbers.
2. `Prio3::AggregateResult` is generic. Similar to measurement, typically the concrete type is either the integer representation of elements of the finite field or a vector of such integers. Again, this design envisions more complex aggregate results.

These design choices add quite a bit of API complexity to `Prio3`. Luckily this complexity doesn't bleed into concrete instantiations of `Prio3`, i.e., `Prio3Aes128Count`, `Prio3Aes128CountVec`, and so on. Still, I think it's worth asking ourselves if the design decisions of (1.) and (2.) drew the right abstraction boundary.

Suppose instead we defined both `Prio3::Mesurement` and `Prio3::AggregateResult` to be `Vec<T::Field::Integer>`. I claim that this is sufficiently general. However the downside is that this shifts to the user the burden of encoding measurements as `Vec<T::Field::Integer>` and decoding aggregate results from `Vec<T::Field::Integer>`. On the other hand I can see at least two upsides:
* Less complexity for `Prio3` API
* Supports code reuse. For example, suppose we wanted a variant of `Prio3Aes128Sum` that is defined for fixed-point numbers rather than integers. In theory the same code could be used for both, without having to define separate types for each.

cjpatton:
Based on #236 (merge that first).
Closes #235.

Prio3's aggregate result is generic, except that we require that it
implements `TryFrom<AggregateShare<T::Field>>` when unsharding a set of
agg shares. This is a bit awkward, since not every Prio3 instance will
decode the aggregate result the same way. Thus it is necessary for the
instance to define the type of the agg result and implement the above
`TryFrom<...>`.

This change requires this type to be defined by the underlying `Type`.
This is more natural, since `Type` also defines the measurement type.
This also allows us to do away with the `Prio3Result` and
`Prio3ResultVec` types, which aren't going to be applicable to every
Prio3 instance.

### DIFF
--- a/src/flp.rs
+++ b/src/flp.rs
@@ -23,7 +23,7 @@
 //!
 //! // The prover chooses a measurement.
 //! let count = Count::new();
-//! let input: Vec<Field64> = count.encode(&0).unwrap();
+//! let input: Vec<Field64> = count.encode_measurement(&0).unwrap();
 //!
 //! // The prover and verifier agree on "joint randomness" used to generate and
 //! // check the proof. The application needs to ensure that the prover
@@ -79,9 +79,13 @@ pub enum FlpError {
     #[error("validity circuit error: {0}")]
     Valid(String),
 
-    /// Calling [`Type::encode`] returned an error.
+    /// Calling [`Type::encode_measurement`] returned an error.
     #[error("value error: {0}")]
     Encode(String),
+
+    /// Calling [`Type::decode_result`] returned an error.
+    #[error("value error: {0}")]
+    Decode(String),
 
     /// Calling [`Type::truncate`] returned an error.
     #[error("truncate error: {0}")]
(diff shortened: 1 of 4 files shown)


divergentdave (approved):

src/flp.rs
```diff
@@ -108,11 +112,20 @@ pub trait Type: Sized + Eq + Clone + Debug {
...
 
     /// Encodes a measurement as a vector of [`Self::input_len`] field elements.
-    fn encode(&self, measurement: &Self::Measurement) -> Result<Vec<Self::Field>, FlpError>;
+    fn encode_measurement(
+        &self,
+        measurement: &Self::Measurement,
+    ) -> Result<Vec<Self::Field>, FlpError>;
+
+    /// Decode an aggregate result.
+    fn decode_result(&self, data: &[Self::Field]) -> Result<Self::AggregateResult, FlpError>;
```
tgeoghegan, line 128: nit: `data` could be `I: IntoIterator<Item = Self::Field>` for flexibility

tgeoghegan (approved):

PR #240, opened 2022-05-27
Port Prio2 to a VDAF

### LINKED ISSUE #214: Consider implementing Prio2 as a VDAF
Right now the v2 code is tightly integrated with ECIES encryption. However, it might be possible refactor the code so that Prio2 also implements the `Vdaf` API.  This may end up being useful for folks interesting in using a proof system that is currently more widely deployed.

The one wrinkle I see is that, in a VDAF, the query randomness would be generated by applying a PRF to a nonce. In the current v2 code, the query randomness is simply handed to each of the aggregators. What this means is that the v2 could couldn't be written in a way that uses the `Vdaf` API directly.

cjpatton:
Closes #214.

Defines a new `Vdaf` implementor that wraps the ENPA Prio system into a
VDAF. The implementation is backwards compatible with `Client` and
`Server`.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -9,7 +9,7 @@ use crate::{
     polynomial::{poly_fft, PolyAuxMemory},
     prng::{Prng, PrngError},
     util::{proof_length, unpack_proof_mut},
-    vdaf::prg::SeedStreamAes128,
+    vdaf::{prg::SeedStreamAes128, VdafError},
 };
 
 use std::convert::TryFrom;
@@ -19,13 +19,8 @@ use std::convert::TryFrom;
 /// Client is used to create Prio shares.
 #[derive(Debug)]
 pub struct Client<F: FieldElement> {
-    prng: Prng<F, SeedStreamAes128>,
     dimension: usize,
-    points_f: Vec<F>,
-    points_g: Vec<F>,
-    evals_f: Vec<F>,
-    evals_g: Vec<F>,
-    poly_mem: PolyAuxMemory<F>,
+    mem: ClientMemory<F>,
     public_key1: PublicKey,
     public_key2: PublicKey,
 }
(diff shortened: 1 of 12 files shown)


src/vdaf/prio2.rs
```diff
@@ -0,0 +1,372 @@
...
+};
+use ring::hmac;
+use std::{
+    convert::{TryFrom, TryInto},
+    io::Cursor,
+};
+
+/// The Prio2 VDAF. It supports a the same measurement type as
+/// [`Prio3Aes128CountVec`](crate::vdaf::prio3::Prio3Aes128CountVec) but uses the proof system
+/// deployed in ENPA for input validation. It also uses a different finite field.
```
cjpatton, line 26:
```suggestion
/// The Prio2 VDAF. It supports the same measurement type as
/// [`Prio3Aes128CountVec`](crate::vdaf::prio3::Prio3Aes128CountVec) but uses the proof system
/// and finite field deployed in ENPA.
```

src/client.rs
```diff
@@ -67,13 +62,8 @@ impl<F: FieldElement> Client<F> {
         }
```
tgeoghegan, line 62: Seems like this check should move down to `ClientMemory::new` since `n` isn't used in this function.

src/client.rs
```diff
@@ -112,19 +102,52 @@ impl<F: FieldElement> Client<F> {
         Ok((encrypted_share1, encrypted_share2))
     }
 
-    pub(crate) fn prove_with<G>(&mut self, init_function: G) -> Vec<F>
+    pub(crate) fn gen_proof(&mut self, input: &[F]) -> Vec<F> {
```
tgeoghegan, line 105: I think this would be an opportune moment to add some doccomment explaining just what `input` is and what is returned from this function.

src/vdaf/prio2.rs
```diff
@@ -0,0 +1,372 @@
...
+        let prio2 = Prio2::new(data.len()).unwrap();
+        let input_shares = vec![
+            Share::get_decoded_with_param(&(&prio2, 0), &input_share1).unwrap(),
+            Share::get_decoded_with_param(&(&prio2, 1), &input_share2).unwrap(),
+        ];
+
+        let verify_key = rng.gen();
+        let mut nonce = [0; 16];
+        rng.fill(&mut nonce);
+        run_vdaf_prepare(&prio2, &verify_key, &(), &nonce, input_shares).expect("test failed");
```
tgeoghegan, line 335: I believe the `.expect` will discard the error from `run_vdaf_prepare`:

```suggestion
        run_vdaf_prepare(&prio2, &verify_key, &(), &nonce, input_shares).unwrap();
```

tgeoghegan (changes requested):

tgeoghegan (approved):
This looks good! Last suggestion would be to gate the `Prio2` VDAF behind a feature but I don't mind including this stuff by default as I don't think it represents a ton of extra code to carry around for clients who want `Prio3`.

cjpatton:
> This looks good! Last suggestion would be to gate the `Prio2` VDAF behind a feature but I don't mind including this stuff by default as I don't think it represents a ton of extra code to carry around for clients who want `Prio3`.

Done. The new feature is called "prio2".

PR #242, opened 2022-06-07
Fix `Continue` variant of `PrepareTransition`

cjpatton:
A prep share should be returned, not a prep message.


cjpatton:
Anyone have a clue why the compiler doesn't catch this?

tgeoghegan (approved):

Issue #237, opened 2022-05-26
API issues


cjpatton:
While cleaning up the documentation for Prio3, I noticed each Prio3 instance uses `new()` as its constructor. The documentation renders an ugly list of these methods as acting on `Prio3`, rather than the type alias for each instance:
![image](https://user-images.githubusercontent.com/3453007/170585567-636dbdaf-2c1f-4d00-8187-ef477e1f543d.png)

Any way to clean this up?

BranLwyd:
A couple of suggestions for consideration, I've attempted to highlight the tradeoffs in each one -- none of these are really great:

1. Make these free functions (e.g. `vdaf::new_prio3_aes128_count(...)`). This would resolve the issue in a straightforward way but would also make these functions less discoverable to folks looking at the `Prio3` type. (I suppose the `Prio3` documentation could reference these functions.)

[I wanted to make a similar suggestion to rename these functions and place them on `Prio3` directly so that we could type e.g. `Prio3::new_aes128_count(...)`, but I couldn't figure out how to make Rust not require the type annotations even though they logically aren't needed by the constructor method.]

2. Switch from using type aliases to newtypes. This would resolve the issue but would probably require an infeasible amount of re-exporting of functions & types, unless you wrote a macro to produce the desired specializations (which itself may be too much effort).

cjpatton:
I think I would prefer (1.)

PR #246, opened 2022-06-14
Bump crate version to v0.8.0.

BranLwyd:
This version implements draft-irtf-cfrg-vdaf-01.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -720,7 +720,7 @@ checksum = "ac74c624d6b2d21f425f752262f42188365d7b8ff1aff74c82e45136510a4857"
 
 [[package]]
 name = "prio"
-version = "0.7.1"
+version = "0.8.0"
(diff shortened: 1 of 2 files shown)


cjpatton:
The draft is no longer draft-patton-*, since it's now under CFRG. The current draft is draft-irtf-cfrg-vdaf-01 (https://datatracker.ietf.org/doc/draft-irtf-cfrg-vdaf/).

Can we try to address some of the simpler issues before cutting this?
https://github.com/divviup/libprio-rs/issues/237

cjpatton:
One more thing to check here before merging: Does `cargo doc` issue any warnings?

BranLwyd:
> One more thing to check here before merging: Does `cargo doc` issue any warnings?

One more PR: #250

cjpatton (approved):

PR #248, opened 2022-06-14
Clippy

BranLwyd:
* Fix the existing Clippy complaints (both minor, in test code).
* Update Clippy invocation in CI build to catch Clippy complaints in test code.
* Invoke Clippy sooner during a CI build, to catch Clippy complaints before the relatively-expensive build & test steps.

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -17,7 +17,7 @@ jobs:
     - name: Lint
       run: cargo fmt --message-format human -- --check
     - name: Clippy
-      run: cargo clippy --all-targets -- -D warnings
+      run: cargo clippy --workspace --all-targets -- -D warnings
     - name: Build (default features)
       run: cargo build --verbose --workspace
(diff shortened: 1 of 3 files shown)


.github/workflows/ci-build.yml
```diff
@@ -16,6 +16,8 @@ jobs:
     - uses: actions/checkout@v2
     - name: Lint
       run: cargo fmt --message-format human -- --check
+    - name: Clippy
+      run: cargo clippy --all-targets -- -D warnings
```
divergentdave, line 20: We could use both `--workspace` and `--all-targets` for completeness.

divergentdave (approved):

.github/workflows/ci-build.yml
BranLwyd: Hm, that's a good point -- this does seem to check "more" than with either flag on its own.

We should probably do this in other repos as well, I'll but together a PR for Janus shortly.

PR #255, opened 2022-06-24
Resolve #254

### LINKED ISSUE #254: Problems with using 2 gadgets in a prio3 type
We tried to implement a prio3 type which uses 2 kinds of gadgets, and stumbled upon an `index_out_of_bounds` error.

This might be because we chose the wrong constants for one of `proof_len()`, `joint_rand_len()`, etc.. If so, we would be happy about some guidance with regards to this.

### Details
We created a minimal example type producing our error by modifying the `Sum` type to have a second gadget. It even occurs when the second gadget is not called at all.

Interestingly, the existence of a second gadget alone is not enough to trigger the error: In the test that triggers it, `self.bits` is 16, and the error occurs when the first gadget is called `2 * self.bits` times. It does not occur when there are e.g. `16 + 14` calls.

Note: Of course, when testing the different numbers of calls, we updated all constants in what seemed to be the correct way.

**EDIT**: The number of calls seems to [decide](https://github.com/divviup/libprio-rs/blob/69771d804019ffef8535bf0f005d3fa160a94d7c/src/flp/gadgets.rs#L200) whether the `call_poly_fft` or `call_poly_direct` implementation is called. The error occurs only when the `fft` version is called.

### How to reproduce
The modified `Sum` type that triggers this bug can be found in this [fork](https://github.com/MxmUrw/libprio-rs).
1. Clone that repo.
2. Run the relevant test `cargo test -- test_prio3_sum`.

The output is:
```
> cargo test -- test_prio3_sum
    Finished test [unoptimized + debuginfo] target(s) in 0.06s
     Running unittests (target/debug/deps/prio-53972546cb9089a3)

running 1 test
test vdaf::prio3::tests::test_prio3_sum ... FAILED

failures:

---- vdaf::prio3::tests::test_prio3_sum stdout ----
thread 'vdaf::prio3::tests::test_prio3_sum' panicked at 'index out of bounds: the len is 128 but the index is 128', src/flp/gadgets.rs:171:43
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    vdaf::prio3::tests::test_prio3_sum

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 77 filtered out; finished in 0.00s

error: test failed, to rerun pass '-p prio --lib'
```

cjpatton:


src/flp/gadgets.rs
```diff
@@ -167,7 +173,7 @@ impl<F: FieldElement> PolyEval<F> {
 
         outp[0] = self.poly[0];
         for i in 1..self.poly.len() {
-            for j in 0..outp.len() {
```
cjpatton, line 170: This was the actual bug, FWIW.

src/flp/gadgets.rs
```diff
@@ -30,14 +30,20 @@ pub struct Mul<F: FieldElement> {
     num_calls: usize,
 }
 
+macro_rules! gadget_poly_fft_mem_len {
```
tgeoghegan, line 33: Is a macro necessary here? I think a function marked `#[inline]` would be easier to read.

tgeoghegan (changes requested):

src/flp/gadgets.rs
cjpatton: Done.

tgeoghegan (approved):

Issue #256, opened 2022-06-26
Best practice for affine (not linear) data encodings


MxmUrw:
While writing our fixed point, bound norm vector type, we had to deal with a difficulty caused by our choice of data encoding. To solve it, we needed to know the number of clients in the implementations of `decode_result()` and the number of aggregators in the implementation of `valid()`. In the following I will explain in some more detail why we need that information.

### Our data encoding
We deal with fixed point numbers, mostly in the range `[-1,1]`.
To represent them in prio, we use the range `[0,2^n)`. To calculate the "semantic value" of a number $y$ encoded in the prio range, we do the obvious thing: rescale to `[0,2]` and shift by 1, namely: 
$$V(y) = \frac{y}{2^{n-1}} -1$$

### Problem 1: decoding
The (now relevant part of our) use-case is that each client submits a fixed point number. We are interested in the sum of those numbers. The problem is that the sum that is automatically computed by prio during accumulation is done in "the prio range" `[0,2^n)`: It turns out that we cannot simply use the decoding function `V`, because it is not linear. What we have to do, e.g. for the case of 2 clients, is:
$$V(y) + V(z) \quad = \quad \frac{y + z}{2^{n-1}} - 2$$
And in the general case we have to subtract the number of clients instead of the `2`.

### Problem 2: validation circuit
A similar effect is seen in the circuit: Our calculation has a non-linear part (a constant summand that is required because of the "shift by 1" in the prio range). This works, as long as the circuit is evaluated locally by the client. In a distributed evaluation by the aggregators, that constant summand appears too often in the calculation: once for every aggregator. So what we do here, is to first multiply the constant by `1/num_of_aggregators` in the circuit.

Alternatively, we could also make this work by including a "constant 1 field element" in the shares. As that 1 is distributed among the aggregators, we multiply the constant summand with a share of that 1.

### Conclusion
We currently provide this information manually when constructing our prio type, and store it in the struct. Is that how this situation should be handled?

If some more details are required (for example about the validation circuit, etc.), I will gladly provide more.

divergentdave:
I think the decoding problem suggests we should provide the number of reports to `Vdaf.agg_shares_to_result` (here, `Type::decode_result()` and `Vdaf::unshard()`.

[...]

PR #258, opened 2022-07-04
Add num_measurements argument to unshard() and decode_result().

MxmUrw:
I implemented here what @divergentdave suggested in #256.

Currently this is only the `num_measurements` parameter being passed to the `decode_result()` function of prio3 types via `Collector::unshard()`.

Is there something else to be done? For example add an average type like was done in cfrg/draft-irtf-cfrg-vdaf#95?

### DIFF
--- a/src/field.rs
+++ b/src/field.rs
@@ -45,6 +45,12 @@ pub enum FieldError {
     /// Error encoding or decoding a field.
     #[error("Codec error")]
     Codec(#[from] CodecError),
+    /// Error converting to `FieldElement::Integer`.
+    #[error("Integer TryFrom error")]
+    IntegerTryFrom,
+    /// Error converting `FieldElement::Integer` into something else.
+    #[error("Integer TryInto error")]
+    IntegerTryInto,
 }
 
 /// Byte order for encoding FieldElement values into byte sequences.
@@ -190,6 +196,71 @@ pub trait FieldElement:
             vec.push(Self::get_decoded(chunk)?);
         }
         Ok(vec)
+    }
+
+    /// Encode `input` as `bits`-bit vector of elements of `Self` if it's small enough
+    /// to be represented with that many bits.
+    ///
+    /// # Arguments
+    ///
+    /// * `input` - The field element to encode
+    /// * `bits` - The number of bits to use for the encoding
+    fn encode_into_bitvector_representation(
+        input: &Self::Integer,
+        bits: usize,
+    ) -> Result<Vec<Self>, FieldError> {
+        // Create a mutable copy of `input`. In each iteration of the following loop we take the
+        // least significant bit, and shift input to the right by one bit.
+        let mut i = *input;
+
+        let one = Self::Integer::from(Self::one());
+        let mut encoded = Vec::with_capacity(bits);
+        for _ in 0..bits {
+            let w = Self::from(i & one);
+            encoded.push(w);
+            i = i >> one;
+        }
+
+        // If `i` is still not zero, this means that it cannot be encoded by `bits` bits.
+        if i != Self::Integer::from(Self::zero()) {
+            return Err(FieldError::InputSizeMismatch);
+        }
+
+        Ok(encoded)
+    }
+
+    /// Decode the bitvector-represented value `input` into a simple representation as a single field element.
+    /// This function errors if `2^input.len() - 1` does not fit into the field `Self`.
+    fn decode_from_bitvector_representation(input: &[Self]) -> Result<Self, FieldError> {
+        if !Self::valid_integer_bitlength(input.len()) {
+            return Err(FieldError::ModulusOverflow);
+        }
+
+        let mut decoded = Self::zero();
+        for (l, bit) in input.iter().enumerate() {
+            let w = Self::Integer::try_from(1 << l).map_err(|_| {FieldError::IntegerTryFrom})?;
+            decoded += Self::from(w) * *bit;
+        }
+        Ok(decoded)
+    }
+
+    /// Interpret `i` as [`Self::Integer`] if it's representable in that type and smaller than the field modulus.
+    fn valid_integer_try_from<N>(i: N) -> Result<Self::Integer, FieldError> where Self::Integer: TryFrom<N>{
+        let i_int = Self::Integer::try_from(i).map_err(|_| {FieldError::IntegerTryFrom})?;
+        if Self::modulus() <= i_int {
+            return Err(FieldError::ModulusOverflow);
+        }
+        Ok(i_int)
+    }
+
+    /// Check if the largest number representable with `bits` bits (i.e. 2^bits - 1) is representable in this field.
+    fn valid_integer_bitlength(bits: usize) -> bool {
+        if let Ok(bits_int) = Self::Integer::try_from(bits){
+            if Self::modulus() >> bits_int != Self::Integer::from(Self::zero()) {
+                return true;
+            }
+        }
+        false
     }
 }
 
--- a/src/flp.rs
+++ b/src/flp.rs
@@ -125,7 +125,11 @@ pub trait Type: Sized + Eq + Clone + Debug {
     ) -> Result<Vec<Self::Field>, FlpError>;
 
     /// Decode an aggregate result.
-    fn decode_result(&self, data: &[Self::Field]) -> Result<Self::AggregateResult, FlpError>;
+    fn decode_result(
+        &self,
+        data: &[Self::Field],
+        num_measurements: usize,
+    ) -> Result<Self::AggregateResult, FlpError>;
 
     /// Returns the sequence of gadgets associated with the validity circuit.
     ///
@@ -490,6 +494,46 @@ pub trait Type: Sized + Eq + Clone + Debug {
 
         Ok(true)
     }
+
+    /// Check whether `input` and `joint_rand` have the length expected by `self`,
+    /// return [`FlpError::Valid`] otherwise.
+    fn valid_call_check(
+        &self,
+        input: &[Self::Field],
+        joint_rand: &[Self::Field],
+    ) -> Result<(), FlpError> {
+        if input.len() != self.input_len() {
+            return Err(FlpError::Valid(format!(
+                "unexpected input length: got {}; want {}",
+                input.len(),
+                self.input_len(),
+            )));
+        }
+
+        if joint_rand.len() != self.joint_rand_len() {
+            return Err(FlpError::Valid(format!(
+                "unexpected joint randomness length: got {}; want {}",
+                joint_rand.len(),
+                self.joint_rand_len()
+            )));
+        }
+
+        Ok(())
+    }
+
+    /// Check if the length of `input` matches `self`'s `input_len()`,
+    /// return [`FlpError::Truncate`] otherwise.
+    fn truncate_call_check(&self, input: &[Self::Field]) -> Result<(), FlpError> {
+        if input.len() != self.input_len() {
+            return Err(FlpError::Truncate(format!(
+                "Unexpected input length: got {}; want {}",
+                input.len(),
+                self.input_len()
+            )));
+        }
+
+        Ok(())
+    }
 }
 
 /// A gadget, a non-affine arithmetic circuit that is called when evaluating a validity circuit.
@@ -841,7 +885,11 @@ mod tests {
             Ok(input)
         }
 
-        fn decode_result(&self, _data: &[F]) -> Result<F::Integer, FlpError> {
+        fn decode_result(
+            &self,
+            _data: &[F],
+            _num_measurements: usize,
+        ) -> Result<F::Integer, FlpError> {
             panic!("not implemented");
         }
     }
@@ -970,7 +1018,11 @@ mod tests {
             Ok(input)
         }
 
-        fn decode_result(&self, _data: &[F]) -> Result<F::Integer, FlpError> {
+        fn decode_result(
+            &self,
+            _data: &[F],
+            _num_measurements: usize,
+        ) -> Result<F::Integer, FlpError> {
             panic!("not implemented");
         }
     }
--- a/src/flp/types.rs
+++ b/src/flp/types.rs
@@ -6,8 +6,7 @@ use crate::field::FieldElement;
 use crate::flp::gadgets::{BlindPolyEval, Mul, ParallelSumGadget, PolyEval};
 use crate::flp::{FlpError, Gadget, Type};
 use crate::polynomial::poly_range_check;
-
-use std::convert::TryFrom;
+use std::convert::TryInto;
 use std::marker::PhantomData;
 
 /// The counter data type. Each measurement is `0` or `1` and the aggregate result is the sum of
@@ -38,7 +37,7 @@ impl<F: FieldElement> Type for Count<F> {
     type Field = F;
 
     fn encode_measurement(&self, value: &F::Integer) -> Result<Vec<F>, FlpError> {
-        let max = F::Integer::try_from(1).unwrap();
+        let max = F::valid_integer_try_from(1)?;
         if *value > max {
             return Err(FlpError::Encode("Count value must be 0 or 1".to_string()));
         }
@@ -46,7 +45,7 @@ impl<F: FieldElement> Type for Count<F> {
         Ok(vec![F::from(*value)])
     }
 
-    fn decode_result(&self, data: &[F]) -> Result<F::Integer, FlpError> {
+    fn decode_result(&self, data: &[F], _num_measurements: usize) -> Result<F::Integer, FlpError> {
         decode_result(data)
     }
 
@@ -61,12 +60,12 @@ impl<F: FieldElement> Type for Count<F> {
         joint_rand: &[F],
         _num_shares: usize,
     ) -> Result<F, FlpError> {
-        valid_call_check(self, input, joint_rand)?;
+        self.valid_call_check(input, joint_rand)?;
         Ok(g[0].call(&[input[0], input[0]])? - input[0])
     }
 
     fn truncate(&self, input: Vec<F>) -> Result<Vec<F>, FlpError> {
-        truncate_call_check(self, &input)?;
+        self.truncate_call_check(&input)?;
         Ok(input)
     }
 
@@ -107,8 +106,6 @@ impl<F: FieldElement> Type for Count<F> {
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct Sum<F: FieldElement> {
     bits: usize,
-    one: F::Integer,
-    max_summand: F::Integer,
     range_checker: Vec<F>,
 }
 
@@ -116,27 +113,14 @@ impl<F: FieldElement> Sum<F> {
     /// Return a new [`Sum`] type parameter. Each value of this type is an integer in range `[0,
     /// 2^bits)`.
     pub fn new(bits: usize) -> Result<Self, FlpError> {
-        let bits_int = F::Integer::try_from(bits).map_err(|err| {
-            FlpError::Encode(format!(
-                "bit length ({}) cannot be represented as a field element: {:?}",
-                bits, err,
-            ))
-        })?;
-
-        if F::modulus() >> bits_int == F::Integer::from(F::zero()) {
-            return Err(FlpError::Encode(format!(
-                "bit length ({}) exceeds field modulus",
-                bits,
-            )));
+        if !F::valid_integer_bitlength(bits) {
+            return Err(FlpError::Encode(
+                "invalid bits: number of bits exceeds maximum number of bits in this field"
+                    .to_string(),
+            ));
         }
-
-        let one = F::Integer::from(F::one());
-        let max_summand = (one << bits_int) - one;
-
         Ok(Self {
             bits,
-            one,
-            max_summand,
             range_checker: poly_range_check(0, 2),
         })
     }
@@ -148,23 +132,11 @@ impl<F: FieldElement> Type for Sum<F> {
     type Field = F;
 
     fn encode_measurement(&self, summand: &F::Integer) -> Result<Vec<F>, FlpError> {
-        if *summand > self.max_summand {
-            return Err(FlpError::Encode(
-                "value of summand exceeds bit length".to_string(),
-            ));
-        }
-
-        let mut encoded: Vec<F> = Vec::with_capacity(self.bits);
-        for l in 0..self.bits {
-            let l = F::Integer::try_from(l).unwrap();
-            let w = F::from((*summand >> l) & self.one);
-            encoded.push(w);
-        }
-
-        Ok(encoded)
+        let v = F::encode_into_bitvector_representation(summand, self.bits)?;
+        Ok(v)
     }
 
-    fn decode_result(&self, data: &[F]) -> Result<F::Integer, FlpError> {
+    fn decode_result(&self, data: &[F], _num_measurements: usize) -> Result<F::Integer, FlpError> {
         decode_result(data)
     }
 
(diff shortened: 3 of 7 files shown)


divergentdave (changes requested):
Thanks! If you would like to add an average type parallel to that in the Sage implementation, that should go in a unit test, in one of the `tests` modules throughout the tree. These changes need a pass through `cargo fmt`, but other than that, the CI should be good.

src/vdaf.rs
```diff
@@ -210,6 +210,7 @@ where
         &self,
         agg_param: &Self::AggregationParam,
         agg_shares: M,
+        num_shares: usize,
```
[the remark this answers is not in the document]
MxmUrw: Yes, that's true. Thanks for catching.

MxmUrw:
Then I'm also going to add that average type.

MxmUrw:
I added the average type. Some notes:
1. To have less code duplication, I kept it as minimal as possible (for instance the validity circuit always accepts).
2. It seemed difficult to put all code into a single "self contained" file because of visibility restrictions (`run_vdaf` cannot be used in the `tests` directory though I do not quite understand why; the fields of `Prio3` are private to the module). So I mostly followed the structure of how the other prio3 types are defined and tested.

divergentdave (approved):

src/flp/test_types.rs
```diff
@@ -0,0 +1,112 @@
+use crate::field::FieldElement;
```
cjpatton, line 1: I believe this needs the copyright notice, as seen in other files.
```
```

src/vdaf/prio3.rs
```diff
@@ -159,6 +162,34 @@ impl Prio3Aes128Histogram {
     }
 }
 
+/// The average type. Each measurement is an integer in `[0,2^bits)` for some `0 < bits < 64` and the
+/// aggregate is the arithmetic average. This type is a test for demonstrating that the number of measurements
+/// may be used in the `decode_result()` function. No validity checks are done on the input data.
+#[cfg(test)]
+pub type TestPrio3Aes128Average = Prio3<Average<Field128>, PrgAes128, 16>;
```
cjpatton, line 169: If this type is indeed only for testing, then it need not be `pub`; `pub(crate)` would be sufficient. However I think it might be worth exposing in the public API. In this case, we could drop the `Test` prefix.

src/flp/test_types.rs
```diff
@@ -0,0 +1,112 @@
...
+        Ok(Self { bits, one })
+    }
+}
+
+impl<F: FieldElement> Type for Average<F> {
+    type Measurement = F::Integer;
+    type AggregateResult = f64;
+    type Field = F;
+
+    fn encode_measurement(&self, summand: &F::Integer) -> Result<Vec<F>, FlpError> {
```
cjpatton, line 42: It should be possible to refactor `Average` and `Sum` to reuse code. In particular:
* encode_measurement()
* decode_result()
* truncate()
* valid()

all have lots of code in common. (There might be others, too.)

src/flp/test_types.rs
```diff
@@ -0,0 +1,112 @@
...
+
+        let one = F::Integer::from(F::one());
+
+        Ok(Self { bits, one })
+    }
+}
+
+impl<F: FieldElement> Type for Average<F> {
+    type Measurement = F::Integer;
+    type AggregateResult = f64;
```
cjpatton, line 39: As I mentioned [here](https://github.com/cfrg/draft-irtf-cfrg-vdaf/pull/95#discussion_r916372075), I'm a bit wary of having the result type (`f64`) have a different precision than the measurements (`u64`, `u128`, ...). Maybe this concern is overblown?

src/flp/test_types.rs
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    fn valid(
+        &self,
+        _g: &mut Vec<Box<dyn Gadget<F>>>,
+        _input: &[F],
+        _joint_rand: &[F],
+        _num_shares: usize,
+    ) -> Result<F, FlpError> {
+        // Accept everything since this is only a test type.
```
cjpatton, line 72: This looks to me like a critical vulnerability waiting too happen. Imagine someone comes along and enablers `Prio3Aes128Average` without thoroughly checking the code, or uses `Average` for some other purpose (it's in the public API afterall).

In my view, this `Type` should be implemented properly, especially those in the public API. It shouldn't be too bad if we refactor the code to crib off of `Sum`.

cjpatton (changes requested):
Thanks, @MxmUrw, this is great. I have one blocking comment, which is to implement input validation for the `Average` type properly. See in-line comment for details.

cjpatton:
> Thanks, @MxmUrw, this is great. I have one blocking comment, which is to implement input validation for the `Average` type properly. See in-line comment for details.

Another way to resolve this: I don't think that adding code to test the new code changes is strictly necessary, given that you plan to send a PR for adding a VDAF that will exercise the code changes.

MxmUrw:
Thanks for the feedback! If it would be useful to have an average type in the public API, then of course I can implement it properly.

> Another way to resolve this: I don't think that adding code to test the new code changes is strictly necessary, given that you plan to send a PR for adding a VDAF that will exercise the code changes.

Actually, since code reuse is going to be a similar concern when adding our fixed point type, I think I might as well tackle that problem already here. I will try it out and push my proposal.

src/flp/test_types.rs
MxmUrw: I do not quite understand the direction that you would like to go with this. It doesn't seem to be about possible loss of information, as in your alternative information is also lost in the `floor()`. Is this about floats being more "lossy" when dealing with larger numbers?

I think that if this (or sth. similar) is the concern, then it is going to depend on the concrete situation where such an average type is going to be used, and the trade-offs cannot really be decided a priori. Alternatively, it could be that something else is going on here that I don't see?

This is also interesting for our bounded-norm fixed-point vector type because there we currently also use floats. In particular, in our use-case, we are going to have a `float -> fixed-point (prio) -> float` pipeline in some form, simply because ML frameworks use floats.

cjpatton: Good point about information loss. And yeah, basically my concern is: When one is given a float type, they expect some imprecision inherent to the IEEE standard for floating point arithmetic. But if the measurements are integers (`u64` or `u128` depending on the field being used to represent them), one would expect to get full precision from the result. Put another way: `Average` is a fairly straightforward extension of `Sum`, but is strictly less useful because it loses precision. 

> This is also interesting for our bounded-norm fixed-point vector type because there we currently also use floats in our aggregate-result-type. In particular, in our use-case, we are going to have a `float -> fixed-point (prio) -> float` pipeline in some form, simply because ML frameworks use floats.

This sounds totally reasonable! One way to solve this might be to have the measurement type for `Average` be `f64` so that the measurement type matches the aggregate result, precisely as you would for your new type.

cjpatton (changes requested):

MxmUrw:
Regarding the measurement type of `Average`: If we are going to allow non-integral numbers as inputs, we are going to have a lot more to think about: These inputs would need to be encoded as (un-)signed fixed point numbers, which requires us (or the user of the type) to choose a range and a number of bits for the fractional part. These choices again depend heavily on the concrete application.

But these are all problems that needn't necessarily be dealt with by prio. The actual functionality that is needed for computing the average is just the `Sum` type (and knowing the number of clients). Everything else is pre- and postprocessing and might best be done when there is a concrete use-case.

Continuing in this spirit, I think the result type of `Average` could be a tuple `(aggregate_sum: F::Integer, num_of_measurements: usize)`, and the user of that type can decide later how he wants to deal with loss of information.

cjpatton:
> Regarding the measurement type of `Average`: If we are going to allow non-integral numbers as inputs, we are going to have a lot more to think about: These inputs would need to be encoded as (un-)signed fixed point numbers, which requires us (or the user of the type) to choose a range and a number of bits for the fractional part. These choices again depend heavily on the concrete application.
> 
> But these are all problems that needn't necessarily be dealt with by prio. The actual functionality that is needed for computing the average is just the `Sum` type (and knowing the number of clients). Everything else is pre- and postprocessing and might best be done when there is a concrete use-case.
> 
> Continuing in this spirit, I think the result type of `Average` could be a tuple `(aggregate_sum: F::Integer, num_of_measurements: usize)`, and the user of that type can decide later how he wants to deal with loss of information.

Yeah good point. Let's not open up this can of beans with this change. Let's just stick with returning `f64`, as you currently do.

src/vdaf/prio3_test.rs
```diff
@@ -149,6 +149,19 @@ fn test_vec_prio3_sum() {
     }
 }
 
+#[test]
```
cjpatton, line 152: This test should be removed. It would only be appropriate if we had test vectors `Prio3Aes128Average`, but we don't (yet!).

cjpatton (changes requested):
Great work! Just a few comments. One high-level thing:  Can we add some basic tests to `flp/types.rs` that exercise the new `Average` type? We don't need as much test coverage as `test_sum()`, but we should at least have tests that validate any code that is different. In particular I'm thinking of `.decode_result()`.

src/flp/types.rs
```diff
@@ -116,26 +120,12 @@ impl<F: FieldElement> Sum<F> {
...
-        })?;
-
-        if F::modulus() >> bits_int == F::Integer::from(F::zero()) {
-            return Err(FlpError::Encode(format!(
-                "bit length ({}) exceeds field modulus",
-                bits,
-            )));
-        }
-
+        let bits_int = valid_int_from_usize::<F>(bits)?;
```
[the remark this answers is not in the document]
tgeoghegan: As a stylistic note, I agree with Chris that in this project we prefer type annotations over the turbofish.

src/vdaf/prio3.rs
```diff
@@ -218,7 +243,7 @@ impl Prio3Aes128Histogram {
 ///     .map(|o| vdaf.aggregate(&(), o).unwrap());
 ///
 /// // Unshard
-/// let agg_res = vdaf.unshard(&(), agg_shares).unwrap();
+/// let agg_res = vdaf.unshard(&(), agg_shares, 5).unwrap();
```
tgeoghegan, line 246: nit: to make this more clear, how about changing the for loop above to
```
let measurements = [0, 1, 1, 1, 0];
for measurement in measurements  { ... }
```
... and then ...
```suggestion
/// let agg_res = vdaf.unshard(&(), agg_shares, measurements.len()).unwrap();
```
I think that helps make it clear what this magic value `5` is, especially for people reading this example on docs.rs and not in an editor that can tell them what the third argument to `vdaf.unshard` is.

src/flp/types_utilities.rs
```diff
@@ -0,0 +1,142 @@
...
+
+/// Encode `input` as `bits`-bit vector of elements of `T::Field` if it's small enough
+/// to be represented with that many bits.
+///
+/// # Arguments
+///
+/// * `input` - The field element to encode
+/// * `bits` - The number of bits to use for the encoding
+/// * `max_input` - The maximal number representable with `bits` bits. Equal to `bits^2-1`
+pub(crate) fn encode_into_bitvector_representation<F: FieldElement>(
```
tgeoghegan, line 35: This function doesn't really operate on any `Type`, so I'd say maybe it should be put into `field.rs`. I also think you could refactor it into an associated function on trait `FieldElement` by inserting the following function definition into `pub trait FieldElement`:
```
    /// Encode `input` as `bits`-bit vector of elements of `T::Field` if it's small enough
    /// to be represented with that many bits.
    ///
    /// # Arguments
    ///
    /// * `input` - The field element to encode
    /// * `bits` - The number of bits to use for the encoding
    /// * `max_input` - The maximal number representable with `bits` bits. Equal to `bits^2-1`
    fn encode_into_bitvector_representation(
        input: &Self::Integer,
        bits: usize,
        max_input: Self::Integer,
    ) -> Result<Vec<Self>, FieldError> {
        if *input > max_input {
            return Err(FieldError::InputSizeMismatch);
        }

        let one = Self::Integer::from(Self::one());
        let mut encoded = Vec::with_capacity(bits);
        for l in 0..bits {
            let l = Self::Integer::try_from(l).unwrap();
            let w = Self::from((*input >> l) & one);
            encoded.push(w);
        }
        Ok(encoded)
    }
```

I think this makes it easier for users of the library to discover these functions, since they'll now show up when they read the documentation on [trait `FieldElement`](https://docs.rs/prio/latest/prio/field/trait.FieldElement.html), alongside the other default trait functions provided there.

I suggest the same approach for `valid_int_from_usize`, `decode_from_bitvector_representation`, `call_gadget_on_vec_entries`, `decode_result` and `decode_result_vec`.

src/flp/types_utilities.rs
```diff
@@ -0,0 +1,142 @@
...
+    let mut r = rnd;
+    for chunk in input.chunks(1) {
+        range_check += r * g.call(chunk)?;
+        r *= rnd;
+    }
+    Ok(range_check)
+}
+
+/// Check whether `input` and `joint_rand` have the length expected by `typ`.
+pub(crate) fn valid_call_check<T: Type>(
```
tgeoghegan, line 88: Similar to my suggestion of adding default implementations to `trait FieldElement`, I think this function and `truncate_call_check` could be refactored into methods on `trait Type` in `flp.rs`. In this case, I argue for _methods_ and not associated functions since these two new functions already take `typ: &T` and so that could become `&self`. e.g:

```
    /// Check whether `input` and `joint_rand` have the length expected by `self`.
    fn valid_call_check(
        &self,
        input: &[Self::Field],
        joint_rand: &[Self::Field],
    ) -> Result<(), FlpError> {
        if input.len() != self.input_len() {
            return Err(FlpError::Valid(format!(
                "unexpected input length: got {}; want {}",
                input.len(),
                self.input_len(),
            )));
        }

        if joint_rand.len() != self.joint_rand_len() {
            return Err(FlpError::Valid(format!(
                "unexpected joint randomness length: got {}; want {}",
                joint_rand.len(),
                self.joint_rand_len()
            )));
        }

        Ok(())
    }
```

src/flp/types.rs
```diff
@@ -183,27 +160,109 @@ impl<F: FieldElement> Type for Sum<F> {
...
+    type AggregateResult = f64;
+    type Field = F;
+
+    fn encode_measurement(&self, summand: &F::Integer) -> Result<Vec<F>, FlpError> {
+        encode_into_bitvector_representation(summand, self.bits, self.max_summand)
+    }
+
+    fn decode_result(&self, data: &[F], num_measurements: usize) -> Result<f64, FlpError> {
+        // Compute the average from the aggregated sum.
+        let data = decode_result(data).unwrap();
```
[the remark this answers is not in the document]
tgeoghegan: I agree with Chris that we should try our best not to panic, since (1) it's quite rude for libraries to crash and (2) this function is already wired up to return an error.

tgeoghegan (changes requested):

src/flp/types.rs
```diff
@@ -183,27 +160,109 @@ impl<F: FieldElement> Type for Sum<F> {
...
+        let data: u64 = data.try_into().unwrap();
+        let result = (data as f64) / (num_measurements as f64);
```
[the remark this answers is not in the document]
MxmUrw: Somehow I cannot get your suggestion to work.
 - I think that `data` needs to be a `f64` here, because division seems only to be implemented in the case where both numerator and denominator have the same type. And I assume that we want to use the floating point division here.
 - There is no `from` or `try_from` for `u64 -> f64`, only for `u32 -> f64`, so I think the only other option is to use the "cast with `as`".
 - Similarly there is no `from` for `usize -> f64`.

Could you elaborate on how your comment was meant? Am I missing imports?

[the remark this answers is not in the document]
cjpatton: Ok, this looks fine as is.

src/field.rs
```diff
@@ -190,6 +196,55 @@ pub trait FieldElement:
         }
         Ok(vec)
     }
+
+    /// Encode `input` as `bits`-bit vector of elements of `Self` if it's small enough
```
cjpatton, line 200: I'm a bit wary of leaking these methods in the public API, but since @tgeoghegan suggested it feel free to ignore me :)
* encode_into_bitvector_representation
* decode_from_bitvector_representation
* valid_integer_try_from

src/field.rs
```diff
@@ -190,6 +196,55 @@ pub trait FieldElement:
...
+    /// to be represented with that many bits.
+    ///
+    /// # Arguments
+    ///
+    /// * `input` - The field element to encode
+    /// * `bits` - The number of bits to use for the encoding
+    /// * `max_input` - The maximal number representable with `bits` bits. Equal to `bits^2-1`
```
cjpatton, line 207: `max_input` is used as a function parameter because precomputing it is convenient. If chosen incorrectly (i.e., if `max_input < 2^bits - 1`, then this method will panic in a way that the user may not expect. This was fine when this method was crate-private, but I'm a bit wary of putting this in the public API.

The fundamental goal here is to ensure that `input` can be encoded with `bits` bits. I wonder if we can rewrite the function to enforce this some other way? One straightforward approach is to compute `max_input` ourselves, but we can probably do even better in terms of computational efficiency.

MxmUrw:
Hi! Thanks again for the detailed comments.

We have dealt with all of them, except that we need input with regards to the following things:
1. Our "utility functions" now live in `flp.rs` and `field.rs`, as suggested. Which means that for example `encode_into_bitvector_representation` (which is defined in `field.rs`) throws a `FieldError`, and not an `FlpError::Encode`. To this end we also added the new constructors `FieldError::IntegerTryFrom` and `FieldError::IntegerTryInto`. This might, or might not be how this situation should be dealt with.
2. Regarding the `as f64` cast, I added a comment [above](https://github.com/divviup/libprio-rs/pull/258#discussion_r931213308).

src/field.rs
```diff
@@ -190,6 +196,55 @@ pub trait FieldElement:
...
+            let l = Self::Integer::try_from(l).map_err(|_| {FieldError::IntegerTryFrom})?;
+            let w = Self::from((*input >> l) & one);
+            encoded.push(w);
+        }
+        Ok(encoded)
+    }
+
+    /// Decode the bitvector-represented value `input` into a simple representation as a single field element.
+    /// This function panics if `2 ^ input.len()` does not fit into the field `Self`.
+    fn decode_from_bitvector_representation(input: &[Self]) -> Self {
```
cjpatton: Now that this method is in the public API, I think it's best to err rather than panic.

src/field.rs
```diff
@@ -190,6 +196,55 @@ pub trait FieldElement:
...
+            decoded += w * *bit;
+        }
+        decoded
+    }
+
+    /// Interpret `i` as [`Self::Integer`] if it's representable in that type and smaller than the field modulus.
+    fn valid_integer_try_from<N>(i: N) -> Result<Self::Integer, FieldError> where Self::Integer: TryFrom<N>{
+        let i_int = Self::Integer::try_from(i).map_err(|_| {FieldError::IntegerTryFrom})?;
+
+        if Self::modulus() >> i_int == Self::Integer::from(Self::zero()) {
```
cjpatton, line 242: This doesn't look right to me: `p >> i` means `p` shifted to the right by `i` bits. Thus this condition holds if `p <= 2^i`. What we want is `p >= i`, right? (Go look at where this function is used. If what we really want is `p <= 2^i`, then the documentation needs to be updated.

src/flp.rs
```diff
@@ -490,6 +494,46 @@ pub trait Type: Sized + Eq + Clone + Debug {
...
     }
+
+    /// Check whether `input` and `joint_rand` have the length expected by `self`,
+    /// return [`FlpError::Valid`] otherwise.
+    fn valid_call_check(
+        &self,
+        input: &[Self::Field],
+        joint_rand: &[Self::Field],
+    ) -> Result<(), FlpError> {
+        if input.len() != self.input_len() {
```
cjpatton, line 505: nit: You could call `truncate_call_check()` instead in order to de-duplicate code.

src/field.rs
```diff
@@ -190,6 +196,55 @@ pub trait FieldElement:
...
+        let mut decoded = Self::zero();
+        for (l, bit) in input.iter().enumerate() {
+            let w = Self::from(Self::Integer::try_from(1 << l).unwrap());
+            decoded += w * *bit;
+        }
+        decoded
+    }
+
+    /// Interpret `i` as [`Self::Integer`] if it's representable in that type and smaller than the field modulus.
+    fn valid_integer_try_from<N>(i: N) -> Result<Self::Integer, FieldError> where Self::Integer: TryFrom<N>{
```
cjpatton, line 239: Consider removing the `valid_` prefix here. It' doesn't make sense in this context.

cjpatton (commented):
Nearly there!

src/field.rs
MxmUrw: Yep, we just did this.

src/field.rs
MxmUrw: The idea was that this is not simply a value of type `Field::Integer` but one that is smaller than the modulus, thus a "representable integer" or "valid integer". But we can remove that prefix.

src/field.rs
```diff
@@ -190,6 +196,72 @@ pub trait FieldElement:
...
+    /// * `input` - The field element to encode
+    /// * `bits` - The number of bits to use for the encoding
+    fn encode_into_bitvector_representation(
+        input: &Self::Integer,
+        bits: usize,
+    ) -> Result<Vec<Self>, FieldError> {
+        // Create a mutable copy of `input`.
+        // In each iteration of the following loop
+        // we take the least significant bit,
+        // and shift input to the right by one bit.
```
cjpatton, line 214: nit: Block comments should break at 100 characters
```suggestion
        // Create a mutable copy of `input`. In each iteration of the following loop we take the
        // least significant bit, and shift input to the right by one bit.
```

src/field.rs
```diff
@@ -190,6 +196,72 @@ pub trait FieldElement:
...
+            return Err(FieldError::InputSizeMismatch);
+        }
+
+        Ok(encoded)
+    }
+
+    /// Decode the bitvector-represented value `input` into a simple representation as a single field element.
+    /// This function errors if `2^input.len() - 1` does not fit into the field `Self`.
+    fn decode_from_bitvector_representation(input: &[Self]) -> Result<Self, FieldError> {
+
```
cjpatton, line 237: nit: Remove spurios newline

src/field.rs
cjpatton: Ack, you can leave as is.

src/field.rs
```diff
@@ -190,6 +196,72 @@ pub trait FieldElement:
...
+    fn valid_integer_try_from<N>(i: N) -> Result<Self::Integer, FieldError> where Self::Integer: TryFrom<N>{
+        let i_int = Self::Integer::try_from(i).map_err(|_| {FieldError::IntegerTryFrom})?;
+        if Self::modulus() <= i_int {
+            return Err(FieldError::ModulusOverflow);
+        }
+        Ok(i_int)
+    }
+
+    /// Check if the largest number representable with `bits` bits (i.e. 2^bits - 1) is representable in this field.
+    fn valid_integer_bitlength(bits: usize) -> Result<(), FieldError> {
```
cjpatton, line 258: I think it would be better if this method just returned a `bool` indicating whether the `2^bits-1` is encodable and let the caller return an error if not.

src/field.rs
cjpatton: Great work!

src/flp.rs
MxmUrw: Yes, though the problem would be that `truncate_call_check()` returns an `FlpError::Truncate` while `valid_call_check()` returns an `FlpError::Valid`.

We could do either:
- Choose another error type which we throw in both functions.
- Call `truncate_call_check`, then do `map_err` to change the error type.

cjpatton: Ack, good catch. You can leave as is.

cjpatton (commented):
LGTM! A couple of asks before we can merge:
* You will need to rebase your change. The following steps should suffice:
    1. You'll want to start by "squashing" your changes into a single commit.
        a. git branch backup      # First make sure your branch is backed up in case something goes wrong.
        b. git rebase main -i    # This drops you into an edit or with instructions on how to proceed. Basically your goal objective is to "squash" every commit into the previous one, except for the first, which you will "pick".
        c. git diff backup     # Confirm that your squashed branch hasn't changed
        d. git push -f   # Update your origin with the new branch. (BE CAREFUL, this rewrites your history.)
    2. Now that you have one commit with all of your changes, you''ll want to rebase on remote.  It looks like your changes on your fork of `main`, so you'll need to do something a bit funky. (Tip: In the future, when making a PR you should always give your branch a distinctive name.)
        a. git remote add upstream https://github.com/divviup/libprio-rs      # you may need to change this, depending on your setup.
        b. git fetch upstream      # Get changes from upstream divviup/libprio-rs
        c. git rebase upstream/main -i      # Rebase on main.
* Tests are currently failing. (See the CI builds at the end of the PR for information.) You should be able to reproduce locally via `cargo test --all-features` once you've properly rebased.

MxmUrw:
Great! Thanks for the detailed instructions. I did two force-pushes now, once after squashing and once after rebase on main + fixing the failing test, I hope that is correct.

cjpatton:
Rebase looks good! I'm taking a last pass to make sure. (Sorry it took me so long to get back to this.) Meanwhile, you've got one more thing to fix before we can merge. You guessed it, it's clippy time:
```
$ cargo clippy
    Checking prio v0.8.2 (/Users/chris/github.com/divviup/libprio-rs)
warning: unneeded `return` statement
   --> src/field.rs:263:9
    |
263 |         return false;
    |         ^^^^^^^^^^^^^ help: remove `return`: `false`
    |
    = note: `#[warn(clippy::needless_return)]` on by default
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_return

warning: `prio` (lib) generated 1 warning
    Finished dev [unoptimized + debuginfo] target(s) in 2.58s
```

MxmUrw:
> You've got one more thing to fix before we can merge.

Done.

cjpatton (approved):

tgeoghegan:
I'm going to review and merge this today. Thanks for your patience!

src/field.rs
tgeoghegan: Not a bad point there, @cjpatton. The goal here was to put these new methods on all implementations of `FieldElement` without having them be floating functions in other modules. If we want to hide them from clients outside the crate, we could move them to some new trait `FieldElementExt` that is `pub(crate)`. I don't want to make Mxm do that though since this PR has been hanging out long enough as it is.

tgeoghegan (approved):

tgeoghegan:
Thanks for the contribution @MxmUrw!

PR #262, opened 2022-07-07
Don't unify "real" and dev dependencies

tgeoghegan:
This commit moves the library and `binaries` targets to `resolver = "2"`
so that `dev-dependencies` don't get pulled in when building non-test
configurations of the `prio` library. Up until now, the crate releases
were enabling feature `test-vector`, unnecessarily including the
`test_vector` module and its dependencies. See [1] for details on
dependency resolution with the new resolver.

[1]: https://github.com/rust-lang/cargo/issues/4866

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,6 +7,7 @@ description = "Implementation of the Prio aggregation system core: https://crypt
 license = "MPL-2.0"
 repository = "https://github.com/divviup/libprio-rs"
 rust-version = "1.58"
+resolver = "2"
 
 [dependencies]
 aes = "0.8.1"
@@ -46,8 +47,7 @@ test-vector = ["rand", "serde_json"]
 multithreaded = ["rayon"]
 
 [workspace]
-members = ["binaries"]
-default-members = ["binaries", "."]
+members = [".", "binaries"]
 
 [[bench]]
 name = "speed_tests"


tgeoghegan:
This one has an extra nuance that didn't turn up in https://github.com/divviup/janus/pull/294: here, `binaries/Cargo.toml` turns on feature `test-vector` in its regular `dependencies`, not exclusively in its `dev-dependencies`. As a consequence, if we do `cargo build --all`, then `prio` gets built with `test-vector` because (AFAICT) features do get unified across the `dependencies` of all workspace members. However, I did verify with `cargo build --verbose --lib` and `cargo publish --verbose` that when building or publishing just the library crate, the `test-vector` feature does not get enabled.

BranLwyd (approved):

binaries/Cargo.toml
```diff
@@ -5,6 +5,7 @@ edition = "2018"
 description = "Prio utilities"
 license = "MPL-2.0"
 repository = "https://github.com/divviup/libprio-rs"
+resolver = "2"
```
BranLwyd, line 8: I think this may be extraneous because it's implied by the workspace's `resolver = "2"`. From https://doc.rust-lang.org/edition-guide/rust-2021/default-cargo-resolver.html:

> The resolver is a global setting for a [workspace](https://doc.rust-lang.org/cargo/reference/workspaces.html), and the setting is ignored in dependencies. The setting is only honored for the top-level package of the workspace.

PR #265, opened 2022-07-11
Prio2: Implement serialization of the prep state

cjpatton:
Implements encoding/decoding traits for Prio2 prep state. The Leader
stores its output share; the Helper stores the seed used to generate the
output share.

### DIFF
--- a/src/vdaf/prio2.rs
+++ b/src/vdaf/prio2.rs
@@ -84,7 +84,7 @@ impl Prio2 {
         .map_err(|e| VdafError::Uncategorized(e.to_string()))?;
 
         Ok((
-            Prio2PrepareState(OutputShare::from(data[..self.input_len].to_vec())),
+            Prio2PrepareState(input_share.truncated(self.input_len)),
             Prio2PrepareShare(verifier_share),
         ))
     }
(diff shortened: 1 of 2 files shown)


src/vdaf/prio2.rs
```diff
@@ -22,7 +22,7 @@ use std::{
 };
 
 /// The Prio2 VDAF. It supports the same measurement type as
-/// [`Prio3Aes128CountVec`](crate::vdaf::prio3::Prio3Aes128CountVec) but uses the proof system
+/// [`Prio2Aes128CountVec`](crate::vdaf::prio3::Prio2Aes128CountVec) but uses the proof system
```
tgeoghegan, line 25: Was this intentional? There's no such thing as `Prio2Aes128CountVec` AFAIK.

tgeoghegan (changes requested):

src/vdaf/prio2.rs
tgeoghegan: Indeed, this is causing the docs build to fail.

cjpatton: Whoopsie daisy!

src/vdaf/prio2.rs
```diff
@@ -131,8 +131,29 @@ impl Client for Prio2 {
...
+
+impl Encode for Prio2PrepareState {
+    fn encode(&self, bytes: &mut Vec<u8>) {
+        self.0.encode(bytes);
+    }
+}
+
+impl<'a> ParameterizedDecode<(&'a Prio2, usize)> for Prio2PrepareState {
+    fn decode_with_param(
+        (prio2, agg_id): &(&'a Prio2, usize),
```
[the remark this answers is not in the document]
cjpatton: Nope :(
```
error[E0053]: method `decode_with_param` has an incompatible type for trait
   --> src/vdaf/prio2.rs:145:26
    |
145 |         (prio2, agg_id): (&'a Prio2, usize),
    |                          ^^^^^^^^^^^^^^^^^^
    |                          |
    |                          expected reference, found tuple
    |                          help: change the parameter type to match the trait: `&(&'a Prio2, usize)`
```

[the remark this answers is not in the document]
tgeoghegan: Ah, right, I forgot that the trait forces a reference here. OK, good enough!

tgeoghegan (approved):

PR #267, opened 2022-07-18
Remove crypto dependencies: ring, aes-gcm

simon-friedberger:


divergentdave (changes requested):
Currently, the "prio2" Cargo feature flag only exposes the VDAF version of Prio2. As such, I think this is the wrong flag to use here for the ECIES-related crypto dependencies. I'd suggest using a new feature flag, say, `"enpa"` instead. This way, our consumers can fall into these categories.

- PPM-DAP implementation, just using VDAFs that are already in the draft. These consumers will not set any feature flags, and they will import what they need from `prio::vdaf::prio3`.
- A hypothetical PPM-DAP implementation, using the Prio2 VDAF. They will set the "prio2" feature flag alone, and import items from `prio::vdaf::prio2`.
- ENPA's `prio-server`/`facilitator`, which only touches `prio::client`, `prio::server`, etc., and not `prio::vdaf`. This can set the new "enpa" feature flag, and make use of the code we're flagging off here via imports from `prio::server`. Note that while this uses the same Prio2 mathematical construction, it does not use the VDAF API at `prio::vdaf::prio2`.

We should also add a step or two to the CI build, to confirm that our conditional compilation is right when using just the ENPA feature flag. We could also use some documentation at top-level of what each Cargo feature does.

src/vdaf.rs
```diff
@@ -524,7 +524,7 @@ mod tests {
 
 pub mod poplar1;
 pub mod prg;
-#[cfg(feature = "prio2")]
+#[cfg(feature = "enpa")]
```
tgeoghegan, line 527: Per David's comment, I think you'll want to put these back, because there are clients that want the Prio2 VDAF but don't want ENPA.

binaries/Cargo.toml
```diff
@@ -16,3 +16,9 @@ structopt = { version = "0.3.26" }
 [[bin]]
 name = "generate-test-vector"
 path = "src/bin/generate_test_vector.rs"
+required-features = ["enpa"]
+
+[[bin]]
+name = "crypt"
+path = "src/bin/crypt.rs"
+required-features = ["enpa"]
```
tgeoghegan, line 24: nit: no newline at EOF

tgeoghegan (changes requested):
I agree with David that we should have clarity on what the different features are meant to achieve, since the distinction between the `prio2` and `enpa` features is pretty subtle. We ought to discuss these things in the crate README and make sure CI exercises each supported combination.

Also, it seems we build with a bunch of dead code warnings with default features, so I think there are some more items that should be hidden behind either `prio2` or `enpa` features. However some of those warnings may predate your change here so we can tackle those cleanups separately.

src/vdaf.rs
cjpatton: I think the intent is that the following modules are behind the "enpa" feature:
* `util`
* `encrypt`
* `client`
* `server`

And the `vdaf::prio2` module is behind the "prio2" feature. This module in turn needs to compile the following structs from the "enpa" feature:
* `client::ClientMemory`
* `server::ValidationMemory`
* `server::VerificationMessage`
* `serverr::is_valid_share()`
* `util::proof_length`

So the question is how to resolve this triangular dependency. I see two ways:
1. Somehow make the "prio2" feature a superset of the "enpa" feature. There's likely a rusty way to do this but I wouldn't know how to spell it. (cc/ @tgeoghegan).
2. (My preferred solution) Move the functionality required by both into a new module and compile that module when either feature is in use.

tgeoghegan: I think Chris' (2) is nicer in general, but (1) seems to be less work and achieves much the same thing. I took a stab at it in #270. In my branch, enabling feature `prio2` enables feature `enpa`, and then I also re-worked the GitHub actions so that we build, test, etc. under all supported feature combinations.

It'd be nice to use the `foo = ["dep:bar", "baz"]` formulation for defining features in `Cargo.toml`, so that you can disambiguiate between features and crate dependency names but I believe that requires Rust 1.60 or up, and I'm not sure it's worth bumping `prio`'s MSRV.

tgeoghegan (approved):
This looks right to me. I'll merge it later today unless @divergentdave or @cjpatton sees a problem with this. Per [discussion on the IETF Slack](https://ietf.slack.com/archives/C02P9620XT6/p1658754450474259), we'll take another change after this to fix `field::{random_vector, split_vector}` so that they don't force the use of `PrgAes128`.

.github/workflows/ci-build.yml
```diff
@@ -8,29 +8,37 @@ on:
...
+        features: [
+          "--all-features",
+          "--features=default",
+          "--features=test-vector",
+          "--features=multithreaded",
+          "--features=test-vector,multithreaded",
+          "--features=prio2",
+          "--features=prio2,test-vector",
+          "--features=prio2,multithreaded",
+          "--features=prio2,test-vector,multithreaded",
```
cjpatton, line 26: This looks like a combinatorial explosion. If we end up needing more feature flags, will we have to continue testing all possible combinations? Anything we can do to avoid this complexity?

cjpatton (commented):
One minor comment.

.github/workflows/ci-build.yml
cjpatton: cc/ @tgeoghegan

tgeoghegan: IMO we don't have a choice but to test all the feature flag combinations we intend to support. We can try our best to make sure that one feature doesn't break another, but we can't have confidence in that unless we actually build and test the feature flag combo. I think this means we have to be careful about adding features. The alternative would be to carefully document what feature combinations we do test and make it clear to clients that they are responsible for testing other feature combos.

As for the length of this list: there's probably a more clever way to write a [workflow matrix strategy](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix), but since there are currently only 8 configurations, I find it easier to reason about if we spell it out this way.

cjpatton: Ack, in that case it's *great news* that we've lumped "enpa" and "prio2" together!

cjpatton (approved):
cc/ @divergentdave: Merge is blocked on your approval, since you requested changes.

.github/workflows/ci-build.yml
divergentdave: There's a tool available that could help with this, [https://github.com/frewsxcv/cargo-all-features](cargo-all-features). It doesn't do clippy or benchmarks, though.

divergentdave (approved):

Issue #241, opened 2022-06-03
Crypto dependencies


simon-friedberger:
It might be difficult to pull ring into a project that already has a crypto library dependency, maybe even boringSSL, which ring seems to be based on but not using.
Especially since both the versioning and the licensing of ring are a bit problematic.

I don't have a good overview of the Rust crypto landscape but if we could keep the VDAF core free of ring that might simplify things in some cases.

In general it would be ideal if we could somehow support multiple cryptolibs.

I am also not sure how dependencies should be handled in general. Currently, `cargo tree -d` shows that we have the following duplicates:
1. aes as 0.8.1 and 0.7.5
2. cipher as 0.3.0 and 0.4.3
3. ctr as 0.8.0 and 0.9.1
4. itoa as 0.4.7 and 1.0.1

cjpatton:
It seems like most of our ring dependencies are for ENPA (i.e, Prio v2) stuff. Perhaps ENPA could be put behind a feature flag?

Note that https://github.com/divviup/libprio-rs/issues/52 is related.

simon-friedberger:
I had a look and afaict the dependencies are caused by:
1. Prio2 ECIES which needs: ring, aes-gcm (which in turn uses aead, aes, cipher, ctr, ....)
2. The PRG in vdaf/prg.rs which needs: aes, cipher, cmac, ctr
Where the latter is iiuc still up for discussion.

tgeoghegan:
+1 to hiding ENPA behind a feature flag, since nobody but the existing deployments should ever use that stuff.

tgeoghegan:
@simon-friedberger is there anything left to do here now that #272 is merged?

tgeoghegan:
We discussed offline -- closing this, and I will ship a patch release of `prio` to crates.io, just as soon as I can get #277 merged.

PR #272, opened 2022-07-25
Add feature flag to hide PRG.

simon-friedberger:
The motivation for this PR is to remove the dependencies on the crates aes, ctr and cmac so a user of the crate can rely on an existing crypto library. This should be dable by implementing analogues of `PrgAes128` and `SeedStreamAes128` and constructing, e.g. `Prio3Aes128Count = Prio3<Count<Field64>, PrgAes128Alt, 16>;`

This would fix the remainder of #241.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,9 +10,9 @@ rust-version = "1.58"
 resolver = "2"
 
 [dependencies]
-aes = "0.8.1"
-ctr = "0.9.1"
-cmac = "0.7.1"
+aes = { version = "0.8.1", optional = true }
+ctr = { version = "0.9.1", optional = true }
+cmac = { version = "0.7.1", optional = true }
 base64 = "0.13.0"
 byteorder = "1.4.3"
 getrandom = { version = "0.2.7", features = ["std"] }
@@ -20,7 +20,7 @@ serde = { version = "1.0", features = ["derive"] }
 static_assertions = "1.1.0"
 thiserror = "1.0"
 
-# dependencies required if feature "test-vector" is enabled
+# dependencies required if feature "test-util" is enabled
 rand = { version = "0.8", optional = true }
 serde_json = { version = "1.0", optional = true }
 
@@ -41,13 +41,14 @@ serde_json = "1.0"
 hex = { version = "0.4.3" , features = ["serde"] }
 # Enable test_vector module for test targets
 # https://github.com/rust-lang/cargo/issues/2911#issuecomment-749580481
-prio = { path = ".", features = ["test-vector"] }
+prio = { path = ".", features = ["test-util"] }
 
 [features]
-default = []
-test-vector = ["rand", "serde_json"]
+default = ["crypto-dependencies"]
+test-util = ["rand", "serde_json"]
 multithreaded = ["rayon"]
 prio2 = ["aes-gcm", "ring"]
+crypto-dependencies = ["aes", "ctr", "cmac"]
 
 [workspace]
 members = [".", "binaries"]
--- a/src/vdaf/prio3.rs
+++ b/src/vdaf/prio3.rs
@@ -27,15 +27,21 @@
 //! [CGB17]: https://crypto.stanford.edu/prio/
 //! [draft-irtf-cfrg-vdaf-01]: https://datatracker.ietf.org/doc/draft-irtf-cfrg-vdaf/01/
 
+#[cfg(feature = "crypto-dependencies")]
+use super::prg::PrgAes128;
 use crate::codec::{CodecError, Decode, Encode, ParameterizedDecode};
-use crate::field::{Field128, Field64, FieldElement};
+use crate::field::FieldElement;
+#[cfg(feature = "crypto-dependencies")]
+use crate::field::{Field128, Field64};
 #[cfg(feature = "multithreaded")]
 use crate::flp::gadgets::ParallelSumMultithreaded;
+#[cfg(feature = "crypto-dependencies")]
 use crate::flp::gadgets::{BlindPolyEval, ParallelSum};
+#[cfg(feature = "crypto-dependencies")]
 use crate::flp::types::{Count, CountVec, Histogram, Sum};
 use crate::flp::Type;
 use crate::prng::Prng;
-use crate::vdaf::prg::{Prg, PrgAes128, RandSource, Seed};
+use crate::vdaf::prg::{Prg, RandSource, Seed};
 use crate::vdaf::{
     Aggregatable, AggregateShare, Aggregator, Client, Collector, OutputShare, PrepareTransition,
     Share, ShareDecodingParameter, Vdaf, VdafError,
@@ -51,37 +57,29 @@ use std::marker::PhantomData;
 const VERS_PRIO3: &[u8] = b"vdaf-01 prio3";
 
 /// The count type. Each measurement is an integer in `[0,2)` and the aggregate result is the sum.
+#[cfg(feature = "crypto-dependencies")]
 pub type Prio3Aes128Count = Prio3<Count<Field64>, PrgAes128, 16>;
 
+#[cfg(feature = "crypto-dependencies")]
 impl Prio3Aes128Count {
     /// Construct an instance of Prio3Aes128Count with the given number of aggregators.
     pub fn new_aes128_count(num_aggregators: u8) -> Result<Self, VdafError> {
-        check_num_aggregators(num_aggregators)?;
-
-        Ok(Prio3 {
-            num_aggregators,
-            typ: Count::new(),
-            phantom: PhantomData,
-        })
+        Prio3::new(num_aggregators, Count::new())
     }
 }
 
 /// The count-vector type. Each measurement is a vector of integers in `[0,2)` and the aggregate is
 /// the element-wise sum.
+#[cfg(feature = "crypto-dependencies")]
 pub type Prio3Aes128CountVec =
     Prio3<CountVec<Field128, ParallelSum<Field128, BlindPolyEval<Field128>>>, PrgAes128, 16>;
 
+#[cfg(feature = "crypto-dependencies")]
 impl Prio3Aes128CountVec {
     /// Construct an instance of Prio3Aes1238CountVec with the given number of aggregators. `len`
     /// defines the length of each measurement.
     pub fn new_aes128_count_vec(num_aggregators: u8, len: usize) -> Result<Self, VdafError> {
-        check_num_aggregators(num_aggregators)?;
-
-        Ok(Prio3 {
-            num_aggregators,
-            typ: CountVec::new(len),
-            phantom: PhantomData,
-        })
+        Prio3::new(num_aggregators, CountVec::new(len))
     }
 }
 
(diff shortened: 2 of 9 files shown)


src/vdaf/prio3_instances.rs
```diff
@@ -0,0 +1,224 @@
+//! Implementation of the Prio3 instances from [[draft-irtf-cfrg-vdaf-01]].
```
cjpatton, line 1: I think it's best to minimize code movement with this change. Instead of creating a new module, I would prefer to see [cfg(internal_prg)] tags on everything that you want to be able to exclude.

Cargo.toml
```diff
@@ -44,10 +44,11 @@ hex = { version = "0.4.3" , features = ["serde"] }
 prio = { path = ".", features = ["test-vector"] }
 
 [features]
-default = []
+default = ["internal_prg"]
 test-vector = ["rand", "serde_json"]
 multithreaded = ["rayon"]
 prio2 = ["aes-gcm", "ring"]
+internal_prg = ["aes", "ctr", "cmac"]
```
cjpatton, line 51: Can we pick a more descriptive name than `internal_prg`?

src/vdaf/poplar1.rs
```diff
@@ -76,12 +83,14 @@ impl IdpfInput {
 
     /// Construct a new input that is a prefix of `self`. Bounds checking is performed by the
     /// caller.
+    #[cfg(feature = "internal_prg")]
```
cjpatton, line 86: Poplar1 needs a lot of work, so minimally invasive change here is probably best. Please move these fences up to the module.

simon-friedberger:
The motivation for this PR is to remove the dependencies on the crates aes, ctr and cmac so a user of the crate can rely on an existing crypto library. This should be dable by implementing analogues of `PrgAes128` and `SeedStreamAes128` and constructing, e.g. `Prio3Aes128Count = Prio3<Count<Field64>, PrgAes128Alt, 16>;`

src/vdaf/prio3.rs
```diff
@@ -12,13 +12,7 @@
 //!
 //! Prio3 is a transformation of a Fully Linear Proof (FLP) system [[draft-irtf-cfrg-vdaf-01]] into
 //! a VDAF. The base type, [`Prio3`], supports a wide variety of aggregation functions, some of
-//! which are instantiated here:
-//!
-//! - [`Prio3Aes128Count`] for aggregating a counter (*)
-//! - [`Prio3Aes128CountVec`] for aggregating a vector of counters
-//! - [`Prio3Aes128Sum`] for copmputing the sum of integers (*)
-//! - [`Prio3Aes128Histogram`] for estimating a distribution via a histogram (*)
-//!
```
cjpatton, line 21: Since PrgAes128 is in the default feature set, I don't think this comment should be reverted.

cjpatton (changes requested):

Cargo.toml
simon-friedberger: Do you have a suggestion?

src/vdaf/prio3_instances.rs
simon-friedberger: Okay, I will undo and resubmit.

Cargo.toml
cjpatton: "internal_prg_aes128"? A comment would also be helpful here.
```
# This feature controls whether VDAFs that use `PrgAes128` will be compiled.
```

src/vdaf/prio3.rs
```diff
@@ -239,6 +121,16 @@ where
     T: Type,
     P: Prg<L>,
 {
+    /// Constructor, because our fields are not public
```
tgeoghegan, line 124: This doccomment will end up on docs.rs so let's try for a nicer explanation, consistent with the other `new_*` functions in the module:
```suggestion
    /// Construct an instance of this Prio3 VDAF with the given number of aggregators and the
    /// underlying type.
```

src/vdaf/prio3_instances.rs
```diff
@@ -0,0 +1,224 @@
...
+
+/// The count type. Each measurement is an integer in `[0,2)` and the aggregate result is the sum.
+pub type Prio3Aes128Count = Prio3<Count<Field64>, PrgAes128, 16>;
+
+impl Prio3Aes128Count {
+    /// Construct an instance of Prio3Aes128Count with the given number of aggregators.
+    pub fn new_aes128_count(num_aggregators: u8) -> Result<Self, VdafError> {
+        check_num_aggregators(num_aggregators)?;
+
+        Prio3::new(num_aggregators, Count::new())
```
tgeoghegan, line 27: I like that these constructors now all call the common constructor! Since `Prio3::new` calls `check_num_aggregators`, can we remove that call from here and the other constructors for the `Prio3` type aliases?

Cargo.toml
tgeoghegan: In #267 we discussed the risk of combinatorial explosion of features. With that in mind, I don't want us to end up with one feature flag for every PRG implementation we end up with. It seems like what's at issue here is cryptography dependencies that Simon doesn't want. What if we name the feature `builtin_cryptography`? The idea being: if you turn on this feature (which is on by default), `prio` will make opinionated choices about crypto dependencies and use them. If you don't turn it on, you need to BYO crypto and adapt it to the cryptography traits that `prio` exports. Going forward, if and when we add new cryptography dependencies to `libprio-rs`, we can gate them behind this existing feature instead of having to add new ones.

src/vdaf/poplar1.rs
tgeoghegan: This sounds like a good idea to me, provided that `poplar1` remains available with the crate's default features.

tgeoghegan (changes requested):

Cargo.toml
simon-friedberger: We did however also disable some "builtin_cryptography" with the prio2 feature. This is specifically about the PRG part.

simon-friedberger:
I've created a little copy & paste PoC to see if this actually works: https://github.com/simon-friedberger/priouser

It did require a lot of copy pasting of things for testing. Does anybody have an idea how that could be improved?

Cargo.toml
cjpatton: @simon-friedberger that's a somewhat separate issue since "prio2" hides the legacy stuff. I think @tgeoghegan's suggestion makes sense: A "catch-all" term is perhaps best. Perhaps "rust_crypto_deps", or just "crypto_deps"?

src/vdaf/prio3.rs
```diff
@@ -105,57 +105,43 @@ impl Prio3Aes128CountVecMultithreaded {
...
-        check_num_aggregators(num_aggregators)?;
-
         let buckets = buckets.iter().map(|bucket| *bucket as u128).collect();
 
-        Ok(Prio3 {
-            num_aggregators,
-            typ: Histogram::<Field128>::new(buckets)?,
-            phantom: PhantomData,
-        })
+        Prio3::new(num_aggregators, Histogram::<Field128>::new(buckets)?)
```
cjpatton, line 144: If possible (i.e., rustc is happy):
```suggestion
        Prio3::new(num_aggregators, Histogram::new(buckets)?)
```

src/vdaf/prio3.rs
```diff
@@ -364,8 +361,8 @@ where
         Ok(out)
     }
 
-    #[cfg(test)]
-    pub(crate) fn test_vec_shard(
+    /// Shard measurement with constant randomness of repeated 0x01 bytes.
```
cjpatton, line 364: Why is this change needed? (why does test_vec_shard need to be pub?)

src/vdaf/prio3.rs
```diff
@@ -105,57 +105,43 @@ impl Prio3Aes128CountVecMultithreaded {
...
         len: usize,
     ) -> Result<Self, VdafError> {
-        check_num_aggregators(num_aggregators)?;
-
-        Ok(Prio3 {
-            num_aggregators,
-            typ: CountVec::new(len),
-            phantom: PhantomData,
-        })
+        Prio3::new(num_aggregators, CountVec::new(len))
```
cjpatton, line 108: Nice refactor!

cjpatton (commented):
Apart from deciding what to call the new feature, this is looking good. Just a few minor comments.

src/vdaf/prio3.rs
simon-friedberger: If I want to do tests similar to what we are doing internally for my external implementation it seems I need it. See here: https://github.com/simon-friedberger/priouser/blob/5ed81adeed2521ad43c3e09b438bd5e03fbac391/src/main.rs#L156
Is there a better way?

tgeoghegan: Ah, it's really too bad that turning on `#[cfg(test)]` in your crate doesn't also enable it in dependencies you pull in. I think the best way forward here would be to do [what Tokio does](https://docs.rs/tokio/latest/tokio/#feature-flags) and [what we did in Janus](https://github.com/divviup/janus/blob/4716fe32d77b9f4dc55503df9e49534e214cf5e8/janus_core/Cargo.toml#L14) and gate things like this behind a `test-util` feature. We already have a `test-vector` feature that we use to conditionally compile the backward compatibility tests, so we could extend that to cover this and any other items you need to enable exclusively in your test code.

cjpatton: Oh, Ok gotcha. Can you add ore context to the comment?
```suggestion
    /// Shard measurement with constant randomness of repeated bytres.
    ///
    /// This method is not secure. It is used for running test vectors for Prio3. 
```

cjpatton: (Or taking @tgeoghegan's suggestion would also be sufficient as far as I'm concerned.)

tgeoghegan:
> I've created a little copy & paste PoC to see if this actually works: https://github.com/simon-friedberger/priouser
> 
> It did require a lot of copy pasting of things for testing. Does anybody have an idea how that could be improved?

Are you referring to [`struct TPrio3<M>`](https://github.com/simon-friedberger/priouser/blob/5ed81adeed2521ad43c3e09b438bd5e03fbac391/src/main.rs#L105) and [`struct TPrio3Prep`](https://github.com/simon-friedberger/priouser/blob/5ed81adeed2521ad43c3e09b438bd5e03fbac391/src/main.rs#L111)? I think we could take a change to expose items like `vdaf::{run_vdaf, run_vdaf_prepare}` or parts of `mod prio3_test` under feature `test-vector` (or `test-util` if you rename it) to make it easier to develop VDAFs or specific bits of VDAFs outside of crate prio. I think that could land in a subsequent PR, though.

simon-friedberger:
Yes, I think exposing `prio3_test.rs` under `test-util` would probably be helpful.

cjpatton (commented):
LGTM, I'm ready to approve. First, I think it would be useful to add some context to the commit message. Plus, we should edit the top comment of this PR to point to the issue.

cjpatton (approved):

tgeoghegan (approved):

tgeoghegan:
Looks great, thank you for this change @simon-friedberger!

PR #280, opened 2022-08-10
Build with `--no-default-features` in CI

tgeoghegan:


.github/workflows/ci-build.yml
```diff
@@ -25,6 +26,10 @@ jobs:
           "--features=prio2,multithreaded",
           "--features=prio2,test-util,multithreaded",
         ]
+        packages: [
```
divergentdave, line 29: Only one of the commands below makes use of this package specifier. As a result, we will have two jobs for each feature, largely performing duplicative work. I think we should take this out of the matrix, and run `cargo build -p prio ...` and `cargo build -p prio-binaries ...` as separate steps.

divergentdave (changes requested):

divergentdave (approved):

.github/workflows/ci-build.yml
tgeoghegan: We have to do this anyway because `prio-binaries` has no features.

tgeoghegan:
I shuffled the steps around a bit more so that the steps that don't use any of the `matrix` are in their own jobs. This way we get more parallelism and less redundant work (we don't need to re-run `cargo fmt` and `cargo doc` for every set of features).

divergentdave (approved):

tgeoghegan:
@simon-friedberger I believe this PR should fix your problem. If you can confirm, then I'll cut a prio release for you.

Issue #257, opened 2022-06-29
How to integrate our custom type into prio


MxmUrw:
We were told that we could integrate our custom type for fixed-point vectors with bounded norm directly into the libprio-rs repository. See [this readme](https://github.com/dpsa-project/libprio-rs) for some details about our type. It is pretty specific to our concrete use-case, so it might be worth discussing whether it really makes sense to integrate it - or at least in what form.

Nevertheless, our prototype is working, and we are only waiting for a small issue (#256) to be resolved. So, let's assume that we want to get our type into prio. Here are the questions that we currently face:
 1. Where in the code base should it reside? I think it makes sense to have our own module, where should that be?
 2. We use the [fixed crate](https://docs.rs/fixed/latest/fixed/). Is it ok to add it as dependency?
 3. Should we think about more generic use-cases? We are working with fixed point numbers after all, something which should be useful much more broadly. But it seems difficult to write code generically:
    - Signed vs unsigned fixed numbers would use different encodings.
    - Different use-cases might require different fixed-point type implementations / libraries.
    - Even restricting to the `fixed` crate, our use-case dictates us a specific fixed-point format (1 bit encodes value before the point, rest of the bits encode fractional value). Other applications would probably need other formats.
 4. We might include some (less restrictive) variants of our prio type, for example fixed-point vectors without a bounded-norm-check. But it seems that this would require some amount of code duplication. As such, I would say that this wouldn't really be useful?
 5. Are there plans to make prio types more composable?

**EDIT**: Our next steps involve adding differential privacy. I am sure there must have been thoughts regarding this topic, so it might be relevant to this discussion.

tgeoghegan:
I agree with @cjpatton, with just two things to add. First, you could hide your new bits behind a [cargo feature](https://doc.rust-lang.org/cargo/reference/features.html) so that not every dependent of `prio` has to take on the `fixed` dependency or the new code you're adding.

Second, with regards to DP: I agree again with Chris in that VDAF is unlikely to say anything about applying DP before a measurement is sharded or compensating for it after aggregate shares are merged, so I wouldn't expect anything to do with DP to appear on trait `prio::vdaf::Vdaf` (or the `Client, Aggregator, Collector` traits). However, it might make sense for the concrete type you define for your new VDAF's `Vdaf::Measurement` associated type to expose support for DP.

[...]

PR #283, opened 2022-08-22
Add prio3 type for sum of bounded-norm fixed-point vectors.

ooovi:
As announced in #257, we added a feature for our custom type to compute the sum of fixed point vectors, with a circuit for verifying their L2 norm is bounded by 1. A couple of questions from our side:

- In the [unit tests in prio3.rs](https://github.com/dpsa-project/libprio-rs/blob/835fbb1c28c74b275bc0c133c4a3ef7b37b99151/src/vdaf/prio3.rs#L1090-L1123), we duplicate code from the `Sum` type testing all sorts of invalid inputs for our type. Does that make sense?
- We chose [`verifier_len`](https://github.com/dpsa-project/libprio-rs/blob/835fbb1c28c74b275bc0c133c4a3ef7b37b99151/src/flp/types/fixedpoint_l2.rs#L433-L435) of our type by doing what we were told by [an error message from the `query` function](https://github.com/dpsa-project/libprio-rs/blob/835fbb1c28c74b275bc0c133c4a3ef7b37b99151/src/flp.rs#L461) associated to the `Type` trait, but had difficulties to understand the reason by looking at the [documentation](https://github.com/dpsa-project/libprio-rs/blob/835fbb1c28c74b275bc0c133c4a3ef7b37b99151/src/flp.rs#L189) and the code that produces the error. Did we do it right?
- We refactored the tests in [`types.rs`](https://github.com/dpsa-project/libprio-rs/blob/835fbb1c28c74b275bc0c133c4a3ef7b37b99151/src/flp/types.rs#L1005-L1185) to enable using `ValidityTestCase` and `flp_validity_test` in our file. The two are now `pub(crate)`, but still have the `test` attribute. Is that satisfactory?

Looking forward to your comments! :)

Co-Authored-By: Olivia <ovi@posteo.de>
Co-Authored-By: Maxim Urschumzew <u.maxim@live.de>

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,7 +6,7 @@ edition = "2018"
 description = "Implementation of the Prio aggregation system core: https://crypto.stanford.edu/prio/"
 license = "MPL-2.0"
 repository = "https://github.com/divviup/libprio-rs"
-rust-version = "1.58"
+rust-version = "1.61"
 resolver = "2"
 
 [dependencies]
@@ -15,6 +15,7 @@ ctr = { version = "0.9.1", optional = true }
 cmac = { version = "0.7.1", optional = true }
 base64 = "0.13.0"
 byteorder = "1.4.3"
+fixed = "1.19"
 getrandom = { version = "0.2.7", features = ["std"] }
 serde = { version = "1.0", features = ["derive"] }
 static_assertions = "1.1.0"
@@ -34,6 +35,7 @@ ring = { version = "0.16.20", optional = true }
 [dev-dependencies]
 assert_matches = "1.5.0"
 criterion = "0.4"
+fixed-macro = "1.1.1"
 iai = "0.1"
 itertools = "0.10.3"
 modinverse = "0.1.0"
--- a/benches/speed_tests.rs
+++ b/benches/speed_tests.rs
@@ -2,6 +2,7 @@
 
 use criterion::{criterion_group, criterion_main, Criterion};
 
+use fixed_macro::fixed;
 use prio::benchmarked::*;
 #[cfg(feature = "prio2")]
 use prio::client::Client as Prio2Client;
@@ -260,6 +261,48 @@ pub fn prio3_client(c: &mut Criterion) {
                 prio3.shard(&measurement).unwrap();
             })
         });
+    }
+
+    let len = 1000;
+    let prio3 = Prio3::new_aes128_fixedpoint_boundedl2_vec_sum(num_shares, len).unwrap();
+    let fp_num = fixed!(0.0001: I1F15);
+    let measurement = vec![fp_num; len];
+    println!(
+        "prio3 fixedpoint16 boundedl2 vec ({} entries) size = {}",
+        len,
+        prio3_input_share_size(&prio3.shard(&measurement).unwrap().1)
+    );
+    c.bench_function(
+        &format!("prio3 fixedpoint16 boundedl2 vec ({} entries)", len),
+        |b| {
+            b.iter(|| {
+                prio3.shard(&measurement).unwrap();
+            })
+        },
+    );
+
+    #[cfg(feature = "multithreaded")]
+    {
+        let prio3 =
+            Prio3::new_aes128_fixedpoint_boundedl2_vec_sum_multithreaded(num_shares, len).unwrap();
+        let fp_num = fixed!(0.0001: I1F15);
+        let measurement = vec![fp_num; len];
+        println!(
+            "prio3 fixedpoint16 boundedl2 vec multithreaded ({} entries) size = {}",
+            len,
+            prio3_input_share_size(&prio3.shard(&measurement).unwrap().1)
+        );
+        c.bench_function(
+            &format!(
+                "prio3 fixedpoint16 boundedl2 vec multithreaded ({} entries)",
+                len
+            ),
+            |b| {
+                b.iter(|| {
+                    prio3.shard(&measurement).unwrap();
+                })
+            },
+        );
     }
 }
 
--- a/src/codec.rs
+++ b/src/codec.rs
@@ -210,7 +210,7 @@ pub fn encode_u8_items<P, E: ParameterizedEncode<P>>(
     }
 
     let len = bytes.len() - len_offset - 1;
-    assert!(len <= u8::MAX.into());
+    assert!(len <= usize::from(u8::MAX));
     bytes[len_offset] = len as u8;
 }
 
@@ -245,7 +245,7 @@ pub fn encode_u16_items<P, E: ParameterizedEncode<P>>(
     }
 
     let len = bytes.len() - len_offset - 2;
-    assert!(len <= u16::MAX.into());
+    assert!(len <= usize::from(u16::MAX));
     for (offset, byte) in u16::to_be_bytes(len as u16).iter().enumerate() {
         bytes[len_offset + offset] = *byte;
     }
--- a/src/field.rs
+++ b/src/field.rs
@@ -201,6 +201,37 @@ pub trait FieldElement:
 
 /// Methods common to all `FieldElement` implementations that are private to the crate.
 pub(crate) trait FieldElementExt: FieldElement {
+    /// Encode `input` as bitvector of elements of `Self`. Output is written into the `output` slice.
+    /// If `output.len()` is smaller than the number of bits required to respresent `input`,
+    /// an error is returned.
+    ///
+    /// # Arguments
+    ///
+    /// * `input` - The field element to encode
+    /// * `output` - The slice to write the encoded bits into. Least signicant bit comes first
+    fn fill_with_bitvector_representation(
+        input: &Self::Integer,
+        output: &mut [Self],
+    ) -> Result<(), FieldError> {
+        // Create a mutable copy of `input`. In each iteration of the following loop we take the
+        // least significant bit, and shift input to the right by one bit.
+        let mut i = *input;
+
+        let one = Self::Integer::from(Self::one());
+        for bit in output.iter_mut() {
+            let w = Self::from(i & one);
+            *bit = w;
+            i = i >> one;
+        }
+
+        // If `i` is still not zero, this means that it cannot be encoded by `bits` bits.
+        if i != Self::Integer::from(Self::zero()) {
+            return Err(FieldError::InputSizeMismatch);
+        }
+
+        Ok(())
+    }
+
     /// Encode `input` as `bits`-bit vector of elements of `Self` if it's small enough
     /// to be represented with that many bits.
     ///
@@ -212,24 +243,9 @@ pub(crate) trait FieldElementExt: FieldElement {
         input: &Self::Integer,
         bits: usize,
     ) -> Result<Vec<Self>, FieldError> {
-        // Create a mutable copy of `input`. In each iteration of the following loop we take the
-        // least significant bit, and shift input to the right by one bit.
-        let mut i = *input;
-
-        let one = Self::Integer::from(Self::one());
-        let mut encoded = Vec::with_capacity(bits);
-        for _ in 0..bits {
-            let w = Self::from(i & one);
-            encoded.push(w);
-            i = i >> one;
-        }
-
-        // If `i` is still not zero, this means that it cannot be encoded by `bits` bits.
-        if i != Self::Integer::from(Self::zero()) {
-            return Err(FieldError::InputSizeMismatch);
-        }
-
-        Ok(encoded)
+        let mut result = vec![Self::zero(); bits];
+        Self::fill_with_bitvector_representation(input, &mut result)?;
+        Ok(result)
     }
 
     /// Decode the bitvector-represented value `input` into a simple representation as a single
@@ -239,13 +255,16 @@ pub(crate) trait FieldElementExt: FieldElement {
     ///
     /// This function errors if `2^input.len() - 1` does not fit into the field `Self`.
     fn decode_from_bitvector_representation(input: &[Self]) -> Result<Self, FieldError> {
+        let fi_one = Self::Integer::from(Self::one());
+
         if !Self::valid_integer_bitlength(input.len()) {
             return Err(FieldError::ModulusOverflow);
         }
 
         let mut decoded = Self::zero();
         for (l, bit) in input.iter().enumerate() {
-            let w = Self::Integer::try_from(1 << l).map_err(|_| FieldError::IntegerTryFrom)?;
+            let fi_l = Self::Integer::try_from(l).map_err(|_| FieldError::IntegerTryFrom)?;
+            let w = fi_one << fi_l;
             decoded += Self::from(w) * *bit;
         }
         Ok(decoded)
@@ -267,6 +286,9 @@ pub(crate) trait FieldElementExt: FieldElement {
     /// Check if the largest number representable with `bits` bits (i.e. 2^bits - 1) is
     /// representable in this field.
     fn valid_integer_bitlength(bits: usize) -> bool {
+        if bits >= 8 * Self::ENCODED_SIZE {
+            return false;
+        }
         if let Ok(bits_int) = Self::Integer::try_from(bits) {
             if Self::modulus() >> bits_int != Self::Integer::from(Self::zero()) {
                 return true;
@@ -956,5 +978,28 @@ mod tests {
     #[test]
     fn test_field128() {
         field_element_test::<Field128>();
+    }
+
+    #[test]
+    fn test_encode_into_bitvector() {
+        let zero = Field128::zero();
+        let one = Field128::one();
+        let zero_enc = Field128::encode_into_bitvector_representation(&0, 4).unwrap();
+        let one_enc = Field128::encode_into_bitvector_representation(&1, 4).unwrap();
+        let fifteen_enc = Field128::encode_into_bitvector_representation(&15, 4).unwrap();
+        assert_eq!(zero_enc, [zero; 4]);
+        assert_eq!(one_enc, [one, zero, zero, zero]);
+        assert_eq!(fifteen_enc, [one; 4]);
+        Field128::encode_into_bitvector_representation(&16, 4).unwrap_err();
+    }
+
+    #[test]
+    fn test_fill_bitvector() {
+        let zero = Field128::zero();
+        let one = Field128::one();
+        let mut output: Vec<Field128> = vec![zero; 6];
+        Field128::fill_with_bitvector_representation(&9, &mut output[1..5]).unwrap();
+        assert_eq!(output, [zero, one, zero, zero, one, zero]);
+        Field128::fill_with_bitvector_representation(&16, &mut output[1..5]).unwrap_err();
     }
 }
--- a/src/flp.rs
+++ b/src/flp.rs
@@ -255,9 +255,9 @@ pub trait Type: Sized + Eq + Clone + Debug {
                 let inner_arity = inner.arity();
                 if prove_rand_len + inner_arity > prove_rand.len() {
                     return Err(FlpError::Prove(format!(
-                        "short prove randomness: got {}; want {}",
+                        "short prove randomness: got {}; want at least {}",
                         prove_rand.len(),
-                        self.prove_rand_len()
+                        prove_rand_len + inner_arity
                     )));
                 }
 
--- a/src/flp/types.rs
+++ b/src/flp/types.rs
@@ -623,18 +623,11 @@ pub(crate) fn decode_result_vec<F: FieldElement>(
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::field::{random_vector, split_vector, Field64 as TestField};
+    use crate::field::{random_vector, Field64 as TestField};
     use crate::flp::gadgets::ParallelSum;
     #[cfg(feature = "multithreaded")]
     use crate::flp::gadgets::ParallelSumMultithreaded;
-
-    // Number of shares to split input and proofs into in `flp_test`.
-    const NUM_SHARES: usize = 3;
-
-    struct ValidityTestCase<F> {
-        expect_valid: bool,
-        expected_output: Option<Vec<F>>,
-    }
+    use crate::flp::types::test_utils::{flp_validity_test, ValidityTestCase};
 
     #[test]
     fn test_count() {
@@ -662,6 +655,7 @@ mod tests {
             &ValidityTestCase::<TestField> {
                 expect_valid: true,
                 expected_output: Some(vec![one]),
+                num_shares: 3,
             },
         )
         .unwrap();
@@ -672,6 +666,7 @@ mod tests {
             &ValidityTestCase::<TestField> {
                 expect_valid: true,
                 expected_output: Some(vec![zero]),
+                num_shares: 3,
             },
         )
         .unwrap();
@@ -683,6 +678,7 @@ mod tests {
             &ValidityTestCase::<TestField> {
                 expect_valid: false,
                 expected_output: None,
+                num_shares: 3,
             },
         )
         .unwrap();
@@ -718,6 +714,7 @@ mod tests {
             &ValidityTestCase {
                 expect_valid: true,
                 expected_output: Some(vec![TestField::from(1337)]),
+                num_shares: 3,
             },
         )
         .unwrap();
@@ -728,6 +725,7 @@ mod tests {
             &ValidityTestCase::<TestField> {
                 expect_valid: true,
                 expected_output: Some(vec![zero]),
+                num_shares: 3,
             },
         )
         .unwrap();
@@ -738,6 +736,7 @@ mod tests {
             &ValidityTestCase {
                 expect_valid: true,
                 expected_output: Some(vec![one]),
+                num_shares: 3,
             },
         )
         .unwrap();
@@ -748,6 +747,7 @@ mod tests {
             &ValidityTestCase::<TestField> {
                 expect_valid: true,
                 expected_output: Some(vec![TestField::from(237)]),
+                num_shares: 3,
             },
         )
         .unwrap();
@@ -759,6 +759,7 @@ mod tests {
             &ValidityTestCase::<TestField> {
                 expect_valid: false,
                 expected_output: None,
+                num_shares: 3,
             },
         )
         .unwrap();
@@ -769,6 +770,7 @@ mod tests {
             &ValidityTestCase::<TestField> {
                 expect_valid: false,
                 expected_output: None,
+                num_shares: 3,
             },
         )
         .unwrap();
@@ -852,6 +854,7 @@ mod tests {
             &ValidityTestCase::<TestField> {
                 expect_valid: true,
                 expected_output: Some(vec![one, zero, zero]),
+                num_shares: 3,
             },
         )
         .unwrap();
@@ -862,6 +865,7 @@ mod tests {
             &ValidityTestCase::<TestField> {
                 expect_valid: true,
                 expected_output: Some(vec![zero, one, zero]),
+                num_shares: 3,
             },
         )
         .unwrap();
@@ -872,6 +876,7 @@ mod tests {
             &ValidityTestCase::<TestField> {
                 expect_valid: true,
                 expected_output: Some(vec![zero, zero, one]),
+                num_shares: 3,
             },
         )
         .unwrap();
@@ -883,6 +888,7 @@ mod tests {
             &ValidityTestCase::<TestField> {
                 expect_valid: false,
                 expected_output: None,
+                num_shares: 3,
             },
         )
         .unwrap();
@@ -893,6 +899,7 @@ mod tests {
             &ValidityTestCase::<TestField> {
                 expect_valid: false,
                 expected_output: None,
+                num_shares: 3,
             },
         )
         .unwrap();
@@ -903,6 +910,7 @@ mod tests {
             &ValidityTestCase::<TestField> {
                 expect_valid: false,
                 expected_output: None,
+                num_shares: 3,
             },
         )
         .unwrap();
@@ -913,6 +921,7 @@ mod tests {
             &ValidityTestCase::<TestField> {
                 expect_valid: false,
                 expected_output: None,
+                num_shares: 3,
             },
         )
         .unwrap();
@@ -935,6 +944,7 @@ mod tests {
                 &ValidityTestCase::<TestField> {
                     expect_valid: true,
                     expected_output: Some(vec![one; len]),
+                    num_shares: 3,
                 },
             )
             .unwrap();
@@ -948,6 +958,7 @@ mod tests {
             &ValidityTestCase::<TestField> {
                 expect_valid: true,
                 expected_output: Some(vec![one; len]),
+                num_shares: 3,
             },
         )
         .unwrap();
@@ -961,6 +972,7 @@ mod tests {
                 &ValidityTestCase::<TestField> {
                     expect_valid: false,
                     expected_output: None,
+                    num_shares: 3,
                 },
             )
             .unwrap();
@@ -1028,8 +1040,21 @@ mod tests {
         assert_eq!(verifier.len(), typ.verifier_len());
         assert!(typ.decide(&verifier).unwrap());
     }
+}
 
-    fn flp_validity_test<T: Type>(
+#[cfg(test)]
+mod test_utils {
+    use super::*;
+    use crate::field::{random_vector, split_vector};
+
+    pub(crate) struct ValidityTestCase<F> {
+        pub(crate) expect_valid: bool,
+        pub(crate) expected_output: Option<Vec<F>>,
+        // Number of shares to split input and proofs into in `flp_test`.
+        pub(crate) num_shares: usize,
+    }
+
+    pub(crate) fn flp_validity_test<T: Type>(
         typ: &T,
         input: &[T::Field],
         t: &ValidityTestCase<T::Field>,
@@ -1101,24 +1126,24 @@ mod tests {
         }
 
         // Run distributed FLP.
-        let input_shares: Vec<Vec<T::Field>> = split_vector(input, NUM_SHARES)
+        let input_shares: Vec<Vec<T::Field>> = split_vector(input, t.num_shares)
             .unwrap()
             .into_iter()
             .collect();
 
-        let proof_shares: Vec<Vec<T::Field>> = split_vector(&proof, NUM_SHARES)
+        let proof_shares: Vec<Vec<T::Field>> = split_vector(&proof, t.num_shares)
             .unwrap()
             .into_iter()
             .collect();
 
-        let verifier: Vec<T::Field> = (0..NUM_SHARES)
+        let verifier: Vec<T::Field> = (0..t.num_shares)
             .map(|i| {
                 typ.query(
                     &input_shares[i],
                     &proof_shares[i],
                     &query_rand,
                     &joint_rand,
-                    NUM_SHARES,
+                    t.num_shares,
                 )
                 .unwrap()
             })
@@ -1197,3 +1222,5 @@ mod tests {
         Ok(())
     }
 }
+
+pub mod fixedpoint_l2;
--- a/src/flp/types/fixedpoint_l2.rs
+++ b/src/flp/types/fixedpoint_l2.rs
@@ -1,0 +1,786 @@
+// SPDX-License-Identifier: MPL-2.0
+
+//! A [`Type`] for summing vectors of fixed point numbers where the
+//! [L2 norm](https://en.wikipedia.org/wiki/Norm_(mathematics)#Euclidean_norm)
+//! of each vector is bounded by `1`.
+//!
+//! In the following a high level overview over the inner workings of this type
+//! is given and implementation details are discussed. It is not necessary for
+//! using the type, but it should be helpful when trying to understand the
+//! implementation.
+//!
+//! ### Overview
+//!
+//! Clients submit a vector of numbers whose values semantically lie in `[-1,1)`,
+//! together with a norm in the range `[0,1)`. The validation circuit checks that
+//! the norm of the vector is equal to the submitted norm, while the encoding
+//! guarantees that the submitted norm lies in the correct range.
+//!
+//! ### Different number encodings
+//!
+//! Let `n` denote the number of bits of the chosen fixed-point type.
+//! Numbers occur in 5 different representations:
+//! 1. Clients have a vector whose entries are fixed point numbers. Only those
+//!    fixed point types are supported where the numbers lie in the range
+//!    `[-1,1)`.
+//! 2. Because norm computation happens in the validation circuit, it is done
+//!    on entries encoded as field elements. That is, the same vector entries
+//!    are now represented by integers in the range `[0,2^n)`, where `-1` is
+//!    represented by `0` and `+1` by `2^n`.
+//! 3. Because the field is not necessarily exactly of size `2^n`, but might be
+//!    larger, it is not enough to encode a vector entry as in (2.) and submit
+//!    it to the aggregator. Instead, in order to make sure that all submitted
+//!    values are in the correct range, they are bit-encoded. (This is the same
+//!    as what happens in the [`Sum`](crate::flp::types::Sum) type.)
+//!    This means that instead of sending a field element in the range `[0,2^n)`,
+//!    we send `n` field elements representing the bit encoding. The validation
+//!    circuit can verify that all submitted "bits" are indeed either `0` or `1`.
+//! 4. The computed and submitted norms are treated similar to the vector
+//!    entries, but they have a different number of bits, namely `2n-2`.
+//! 5. As the aggregation result is a pointwise sum of the client vectors,
+//!    the numbers no longer (semantically) lie in the range `[-1,1)`, and cannot
+//!    be represented by the same fixed point type as the input. Instead the
+//!    decoding happens directly into a vector of floats.
+//!
+//! ### Fixed point encoding
+//!
+//! Submissions consist of encoded fixed-point numbers in `[-1,1)` represented as
+//! field elements in `[0,2^n)`, where n is the number of bits the fixed-point
+//! representation has. Encoding and decoding is handled by the associated functions
+//! of the [`CompatibleFloat`] trait. Semantically, the following function describes
+//! how a fixed-point value `x` in range `[-1,1)` is converted to a field integer:
+//! ```text
+//! enc : [-1,1) -> [0,2^n)
+//! enc(x) = 2^(n-1) * x + 2^(n-1)
+//! ```
+//! The inverse is:
+//! ```text
+//! dec : [0,2^n) -> [-1,1)
+//! dec(y) = (y - 2^(n-1)) * 2^(1-n)
+//! ```
+//! Note that these functions only make sense when interpreting all occuring
+//! numbers as real numbers. Since our signed fixed-point numbers are encoded as
+//! two's complement integers, the computation that happens in
+//! [`CompatibleFloat::to_field_integer`] is actually simpler.
+//!
+//! ### Norm computation
+//!
+//! The L2 norm of a vector xs of numbers in `[-1,1)` is given by:
+//! ```text
+//! norm(xs) = sqrt(sum_{x in xs} x^2)
+//! ```
+//! Instead of computing the norm, we make two simplifications:
+//! 1. We ignore the square root, which means that we are actually computing
+//!    the square of the norm.
+//! 2. We want our norm computation result to be integral and in the range `[0, 2^(2n-2))`,
+//!    so we can represent it in our field integers. We achieve this by multiplying with `2^(2n-2)`.
+//! This means that what is actually computed in this type is the following:
+//! ```text
+//! our_norm(xs) = 2^(2n-2) * norm(xs)^2
+//! ```
+//!
+//! Explained more visually, `our_norm()` is a composition of three functions:
+//!
+//! ```text
+//!                    map of dec()                    norm()          "mult with 2^(2n-2)"
+//!   vector of [0,2^n)    ->       vector of [-1,1)    ->       [0,1)         ->           [0,2^(2n-2))
+//!                                         ^                      ^
+//!                                         |                      |
+//!                     fractions with denom of 2^(n-1)     fractions with denom of 2^(2n-2)
+//! ```
+//! (Note that the ranges on the LHS and RHS of `"mult with 2^(2n-2)"` are stated
+//! here for vectors with a norm less than `1`.)
+//!
+//! Given a vector `ys` of numbers in the field integer encoding (in `[0,2^n)`),
+//! this gives the following equation:
+//! ```text
+//! our_norm_on_encoded(ys) = our_norm([dec(y) for y in ys])
+//!                         = 2^(2n-2) * sum_{y in ys} ((y - 2^(n-1)) * 2^(1-n))^2
+//!                         = 2^(2n-2)
+//!                           * sum_{y in ys} y^2 - 2*y*2^(n-1) + (2^(n-1))^2
+//!                           * 2^(1-n)^2
+//!                         = sum_{y in ys} y^2 - (2^n)*y + 2^(2n-2)
+//! ```
+//!
+//! Let `d` denote the number of the vector entries. The maximal value the result
+//! of `our_norm_on_encoded()` can take occurs in the case where all entries are
+//! `2^n-1`, in which case `d * 2^(2n-2)` is an upper bound to the result. The
+//! finite field used for encoding must be at least as large as this.
+//! For validating that the norm of the submitted vector lies in the correct
+//! range, consider the following:
+//!  - The result of `norm(xs)` should be in `[0,1)`.
+//!  - Thus, the result of `our_norm(xs)` should be in `[0,2^(2n-2))`.
+//!  - The result of `our_norm_on_encoded(ys)` should be in `[0,2^(2n-2))`.
+//! This means that the valid norms are exactly those representable with `2n-2`
+//! bits.
+//!
+//! ### Differences in the computation because of distribution
+//!
+//! In `decode_result()`, what is decoded are not the submitted entries of a
+//! single client, but the sum of the the entries of all clients. We have to
+//! take this into account, and cannot directly use the `dec()` function from
+//! above. Instead we use:
+//! ```text
+//! dec'(y) = y * 2^(1-n) - c
+//! ```
+//! Here, `c` is the number of clients.
+//!
+//! ### Naming in the implementation
+//!
+//! The following names are used:
+//!  - `self.bits_per_entry` is `n`
+//!  - `self.entries`        is `d`
+//!  - `self.bits_for_norm`  is `2n-2`
+//!
+//! ### Submission layout
+//!
+//! The client submissions contain a share of their vector and the norm
+//! they claim it has.
+//! The submission is a vector of field elements laid out as follows:
+//! ```text
+//! |---- bits_per_entry * entries ----|---- bits_for_norm ----|
+//!  ^                                  ^
+//!  \- the input vector entries        |
+//!                                     \- the encoded norm
+//! ```
+//!
+//! ### Value `1`
+//!
+//! We actually do not allow the submitted norm or vector entries to be
+//! exactly `1`, but rather require them to be strictly less. Supporting `1` would
+//! entail a more fiddly encoding and is not necessary for our usecase.
+//! The largest representable vector entry can be computed by `dec(2^n-1)`.
+//! For example, it is `0.999969482421875` for `n = 16`.
+
+pub mod compatible_float;
+
+use crate::field::{FieldElement, FieldElementExt};
+use crate::flp::gadgets::{BlindPolyEval, ParallelSumGadget, PolyEval};
+use crate::flp::types::fixedpoint_l2::compatible_float::CompatibleFloat;
+use crate::flp::{FlpError, Gadget, Type};
+use crate::polynomial::poly_range_check;
+use fixed::traits::Fixed;
+
+use std::{convert::TryFrom, convert::TryInto, fmt::Debug, marker::PhantomData};
+
+/// The fixed point vector sum data type. Each measurement is a vector of fixed point numbers of
+/// type `T`, and the aggregate result is the float vector of the sum of the measurements.
+///
+/// The validity circuit verifies that the L2 norm of each measurement is bounded by 1.
+///
+/// The [*fixed* crate](https://crates.io/crates/fixed) is used for fixed point numbers, in
+/// particular, exactly the following types are supported:
+/// `FixedI16<U15>`, `FixedI32<U31>` and `FixedI64<U63>`.
+///
+/// Depending on the size of the vector that needs to be transmitted, a corresponding field type has
+/// to be chosen for `F`. For a `n`-bit fixed point type and a `d`-dimensional vector, the field
+/// modulus needs to be larger than `d * 2^(2n-2)` so there are no overflows during norm validity
+/// computation.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct FixedPointBoundedL2VecSum<
+    T: Fixed,
+    F: FieldElement,
+    SPoly: ParallelSumGadget<F, PolyEval<F>> + Clone,
+    SBlindPoly: ParallelSumGadget<F, BlindPolyEval<F>> + Clone,
+> {
+    bits_per_entry: usize,
+    fi_bits_per_entry: F::Integer,
+    entries: usize,
+    bits_for_norm: usize,
+    range_01_checker: Vec<F>,
+    norm_summand_poly: Vec<F>,
+    phantom: PhantomData<(T, SPoly, SBlindPoly)>,
+
+    // range/position constants
+    range_norm_begin: usize,
+    range_norm_end: usize,
+
+    // configuration of parallel sum gadgets
+    gadget0_calls: usize,
+    gadget0_chunk_len: usize,
+    gadget1_calls: usize,
+    gadget1_chunk_len: usize,
+}
+
+impl<T, F, SPoly, SBlindPoly> FixedPointBoundedL2VecSum<T, F, SPoly, SBlindPoly>
+where
+    T: Fixed,
+    F: FieldElement,
+    SPoly: ParallelSumGadget<F, PolyEval<F>> + Clone,
+    SBlindPoly: ParallelSumGadget<F, BlindPolyEval<F>> + Clone,
+    u128: TryFrom<F::Integer>,
+{
+    /// Return a new [`FixedPointBoundedL2VecSum`] type parameter. Each value of this type is a
+    /// fixed point vector with `entries` entries.
+    pub fn new(entries: usize) -> Result<Self, FlpError> {
+        // (0) initialize constants
+        let fi_one = F::Integer::from(F::one());
+        let fi_two = fi_one + fi_one;
+
+        // (I) Check that the fixed type `F` is compatible.
+        //
+        // We only support fixed types that encode values in [-1,1].
+        // These have a single integer bit.
+        if <T as Fixed>::INT_NBITS != 1 {
+            return Err(FlpError::Encode(format!(
+                "Expected fixed point type with one integer bit, but got {}.",
+                <T as Fixed>::INT_NBITS,
+            )));
+        }
+
+        // Compute number of bits of an entry, and check that an entry fits
+        // into the field.
+        let bits_per_entry: usize = (<T as Fixed>::INT_NBITS + <T as Fixed>::FRAC_NBITS)
+            .try_into()
+            .map_err(|_| FlpError::Encode("Could not convert u32 into usize.".to_string()))?;
+        if !F::valid_integer_bitlength(bits_per_entry) {
+            return Err(FlpError::Encode(format!(
+                "fixed point type bit length ({}) too large for field modulus",
+                bits_per_entry,
+            )));
+        }
+        let fi_bits_per_entry: F::Integer = F::valid_integer_try_from(bits_per_entry)
+            .map_err(|_| FlpError::Encode("Could not create FieldInteger from usize even though size check was successful".to_string()))?;
+
+        // (II) Check that the field is large enough for the norm.
+        //
+        // Valid norms encoded as field integers lie in [0,2^(2*bits - 2)).
+        let bits_for_norm = 2 * bits_per_entry - 2;
+        if !F::valid_integer_bitlength(bits_for_norm) {
+            return Err(FlpError::Encode(format!(
+                "maximal norm bit length ({}) too large for field modulus",
+                bits_for_norm,
+            )));
+        }
+
+        // In order to compare the actual norm of the vector with the claimed
+        // norm, the field needs to be able to represent all numbers that can
+        // occur during the computation of the norm of any submitted vector,
+        // even if its norm is not bounded by 1. Because of our encoding, an
+        // upper bound to that value is `entries * 2^(2*bits - 2)` (see docs of
+        // compute_norm_of_entries for details). It has to fit into the field.
+        let err = Err(FlpError::Encode(format!(
+            "number of entries ({}) not compatible with field size",
+            entries,
+        )));
+
+        if let Some(val) = (entries as u128).checked_mul(1 << bits_for_norm) {
+            if let Ok(modulus) = u128::try_from(F::modulus()) {
+                if val >= modulus {
+                    return err;
+                }
+            } else {
+                return err;
+            }
+        } else {
+            return err;
+        }
+
+        // Construct the polynomial that computes a part of the norm for a
+        // single vector entry.
+        //
+        // the linear part is 2^n,
+        // the constant part is 2^(2n-2),
+        // the polynomial is:
+        //   p(y) = 2^(2n-2) + -(2^n) * y + 1 * y^2
+        let linear_part = fi_one << fi_bits_per_entry;
+        let constant_part = fi_one << (fi_bits_per_entry + fi_bits_per_entry - fi_two);
+        let norm_summand_poly = vec![F::from(constant_part), -F::from(linear_part), F::one()];
+
+        // Compute chunk length and number of calls for parallel sum gadgets.
+        let len0 = bits_per_entry * entries + bits_for_norm;
+        let gadget0_chunk_len = std::cmp::max(1, (len0 as f64).sqrt() as usize);
+        let gadget0_calls = (len0 + gadget0_chunk_len - 1) / gadget0_chunk_len;
+
+        let len1 = entries;
+        let gadget1_chunk_len = std::cmp::max(1, (len1 as f64).sqrt() as usize);
+        let gadget1_calls = (len1 + gadget1_chunk_len - 1) / gadget1_chunk_len;
+
+        Ok(Self {
+            bits_per_entry,
+            fi_bits_per_entry,
+            entries,
+            bits_for_norm,
+            range_01_checker: poly_range_check(0, 2),
+            norm_summand_poly,
+            phantom: PhantomData,
+
+            // range constants
+            range_norm_begin: entries * bits_per_entry,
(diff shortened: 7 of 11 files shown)


src/flp/types/fixedpoint_l2.rs
```diff
@@ -0,0 +1,631 @@
...
+ * --- Validity ---
+ * In addition to checking that every submission entry is 0 or 1, the validation
+ * circuit of this type computes the norm and compares to what the client
+ * claimed.
+ *
+ *
+ * --- Value 1 ---
+ * We actually do not allow the submitted norm or vector entries to be
+ * exactly 1, but rather require them to be strictly less. Supporting 1 would
+ * entail a more fiddly encoding and is not necessary for our usecase.
```
cjpatton, line 176: This fact is definitely worth documenting in the public API.

src/flp/types/fixedpoint_l2.rs
```diff
@@ -0,0 +1,631 @@
...
+
+use std::{convert::TryInto, fmt::Debug, marker::PhantomData};
+
+/// The fixed point vector sum data type. Each measurement is a vector of fixed point numbers of
+/// type `T`, and the aggregate result is the float vector of the sum of the measurements.
+///
+/// The validity circuit verifies that the L2 norm of each measurement is bounded by 1.
+///
+/// The [*fixed* crate] is used for fixed point numbers, in particular, exactly the following types
+/// are supported: `FixedI8<U7>`, `FixedI16<U15>`, `FixedI32<U31>`, `FixedI64<U63>` and
```
cjpatton, line 24: It seems like only `FixedI16<U15>` is supported currently?

src/vdaf/prio3.rs
```diff
@@ -130,6 +136,33 @@ impl Prio3Aes128Sum {
     }
 }
 
+/// The fixed point vector sum type. Each measurement is a vector of 16-bit fixed point decimals
+/// with 15 fractional digits and the aggregate is the sum. The verification function ensures the
+/// L2 norm of the vector is <= 1.
+#[cfg(feature = "fixed")]
+#[cfg_attr(docsrs, doc(cfg(feature = "fixed")))]
+pub type Prio3Aes128FixedPointL2BoundedVecSum =
+    Prio3<FixedPointL2BoundedVecSum<FixedI16<U15>, Field64>, PrgAes128, 16>;
```
cjpatton, line 145: 1. The size of the field needs careful consideration, as the soundness error depends on it. Given that the number of gadget calls depends on the dimension of the input vector, I'd say Field128 is a safer choice. Note that we may end up going with smaller fields in the future, but this depends on security analysis; see https://github.com/cfrg/draft-irtf-cfrg-vdaf/issues/102.
2. What's the likelihood that we're going to want more precision than 16 bits in the future? If this is likely, then consider reflecting the precision in the name somehow.

cjpatton (commented):
This is gonna be neat! Some high level comments to start. Before going too deep, I think it would be useful to add some benchmarks to get a sense of how much this will cost for typical measurements. Would you mind extending `prio3_client()` in benches/speed_tests.rs to include benchmarks for input share generation (this includes generating the proof) and size of the input shares?


- In the [unit tests in prio3.rs](https://github.com/dpsa-project/libprio-rs/blob/835fbb1c28c74b275bc0c133c4a3ef7b37b99151/src/vdaf/prio3.rs#L1090-L1123), we duplicate code from the `Sum` type testing all sorts of invalid inputs for our type. Does that make sense?

I'll give a definitive answer when I review this part thoroughly, but for now: Yes, my expectation is that much of the code can be reused. The key trick is to take a logical-AND by taking a random linear combination of two sub-circuits.


- We chose [`verifier_len`](https://github.com/dpsa-project/libprio-rs/blob/835fbb1c28c74b275bc0c133c4a3ef7b37b99151/src/flp/types/fixedpoint_l2.rs#L433-L435) of our type by doing what we were told by [an error message from the `query` function](https://github.com/dpsa-project/libprio-rs/blob/835fbb1c28c74b275bc0c133c4a3ef7b37b99151/src/flp.rs#L461) associated to the `Type` trait, but had difficulties to understand the reason by looking at the [documentation](https://github.com/dpsa-project/libprio-rs/blob/835fbb1c28c74b275bc0c133c4a3ef7b37b99151/src/flp.rs#L189) and the code that produces the error. Did we do it right?

The correct computation of proof and verifier length is documented [here](https://www.ietf.org/archive/id/draft-irtf-cfrg-vdaf-02.html#name-validity-circuits). I will check this once I understand the circuit better.

- We refactored the tests in [`types.rs`](https://github.com/dpsa-project/libprio-rs/blob/835fbb1c28c74b275bc0c133c4a3ef7b37b99151/src/flp/types.rs#L1005-L1185) to enable using `ValidityTestCase` and `flp_validity_test` in our file. The two are now `pub(crate)`, but still have the `test` attribute. Is that satisfactory?

Yup, this looks great! I've added a couple suggestions for improvement.

Cargo.toml
```diff
@@ -49,6 +53,7 @@ test-util = ["rand", "serde_json"]
 multithreaded = ["rayon"]
 prio2 = ["aes-gcm", "ring"]
 crypto-dependencies = ["aes", "ctr", "cmac"]
+fixed = ["dep:fixed", "fixed-macro"]
```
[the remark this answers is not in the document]
ooovi: @tgeoghegan suggested a feature in https://github.com/divviup/libprio-rs/issues/257#issuecomment-1170407169

src/codec.rs
```diff
@@ -210,7 +210,8 @@ pub fn encode_u8_items<P, E: ParameterizedEncode<P>>(
     }
 
     let len = bytes.len() - len_offset - 1;
-    assert!(len <= u8::MAX.into());
+    let max: usize = u8::MAX.into();
+    assert!(len <= max);
```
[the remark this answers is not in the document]
ooovi: The `fixed` crate we depend on provides `<=` with a left-hand `usize` and a right-hand `Fixed`, so we need to annotate the `into()` call. Maybe there's a way around this?

src/field.rs
```diff
@@ -267,9 +283,12 @@ pub(crate) trait FieldElementExt: FieldElement {
     /// Check if the largest number representable with `bits` bits (i.e. 2^bits - 1) is
     /// representable in this field.
     fn valid_integer_bitlength(bits: usize) -> bool {
-        if let Ok(bits_int) = Self::Integer::try_from(bits) {
-            if Self::modulus() >> bits_int != Self::Integer::from(Self::zero()) {
-                return true;
+        if bits <= 8 * Self::ENCODED_SIZE {
```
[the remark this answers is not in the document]
ooovi: The right-shift will panic if `bits_int` is larger than the number of bits the `Self::Integer` type uses, even tho it might be small enough to be representable in that type (hence no `try_from()` error)

src/codec.rs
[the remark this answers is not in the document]
cjpatton: Oh, I see. I only tested w/o the "fixed" feature.

src/flp/types/fixedpoint_l2.rs
```diff
@@ -0,0 +1,632 @@
...
+    entries: usize,
+    bits_for_norm: usize,
+    range_01_checker: Vec<F>,
+    square_computer: Vec<F>,
+    phantom: PhantomData<T>,
+    // range/position constants
+    range_norm_begin: usize,
+    range_norm_end: usize,
+}
+
```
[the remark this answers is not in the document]
ooovi: Where is the right place to do something like this? Maybe in the docucomment for the `fixedpoint_l2` module? Are you thinking all of the huge block comment, or just some of the less detailed bits?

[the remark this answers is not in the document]
cjpatton: The package docucomment is the block above that looks like this:
```rust
//! Package `blah` implements ...
//!
//! A detailed description of the package.
```

I think this is where you should put this description. You can see it rendered by running `cargo doc --all-features --open`. Incidentally, you can use markdown to format things if you like.

src/flp/types/fixedpoint_l2.rs
```diff
@@ -0,0 +1,631 @@
...
+ *      are now represented by integers in the range [0,2^n), where -1 is
+ *      represented by 0 and +1 by 2^n.
+ *  (3) Because the field is not necessarily exactly of size 2^n, but might be
+ *      larger, it is not enough to encode a vector entry as in (2) and submit
+ *      it to the aggregator. Instead, in order to make sure that all submitted
+ *      values are in the correct range, they are bit-encoded. (This is the same
+ *      as what happens in the `Sum` type.)
+ *      This means that instead of sending a field element in the range [0,2^n),
+ *      we send n field elements representing the bit encoding. The validation
+ *      circuit can verify that all submitted "bits" are indeed either 0 or 1.
```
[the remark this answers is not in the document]
MxmUrw: For a d-sized vector of n-bit fixed point numbers the encoded input + proof is in the order of `d*n` (actually `(d+2)*n - 2`). And since `d` can become very high it seems likely that we will have to do everything we can to keep the communication cost as small as possible, so this idea is very relevant!

If instead of binary encoding, we switch to a b-ary encoding of our vector entries, we could reduce by a factor of `log_2(b)`, while the polynomial in the range-checker gets larger by a factor of `b/2`. I do not know currently how this will affect the circuit (or validation-) runtime, is a basis of b=256 realistic for example?

Anyways, (at least from our perspective) it would be convenient to begin thinking about such optimizations later, when we have a prototype for our full pipeline. Would it be ok to keep the encoding as is for now, and probably submit an update for it later?

src/vdaf/prio3.rs
MxmUrw: 1. What kind of soundness error do you refer to? We only checked that all number that might occur during validation fit into the field. Is there something else we need to be aware of when choosing the field size?
2. It is pretty likely, at least for 32 bits and possibly 64 bits. We will update the name.

tgeoghegan (commented):
Here's a whole bunch of nits, but I still need to spend time understanding `fixedpoint_l2.rs` properly before I can sign off on this.

src/flp/types/fixedpoint_l2.rs
```diff
@@ -0,0 +1,638 @@
...
+        //  - the submitted norm (encoded in `self.bits_for_norm` field
+        //    elements)
+        //
+        // Since all input vector entry (field-)bits, as well as the norm bits,
+        // are contiguous, we do the check directly for all bits from 0 to
+        // entries*bits_per_entry + bits_for_norm.
+        //
+        // Check that each element is a 0 or 1:
+        let range_check =
+            call_gadget_on_vec_entries(&mut g[0], &input[0..self.range_norm_end], joint_rand[0])?;
```
cjpatton, line 355: The gadget has degree 2 and the number of gadget calls is `entries * n + 2 * n  - 2`, Thus the length of the proof polynomial is 

```
2 * (next_power_of_2(1 + entries*n + 2*n - 2) - 1) + 1 
= 2 * (next_power_of_2(entries*n + 2*n - 1) - 1) + 1
<= 2 * (2 * (entries*n + 2*n - 1) - 1) + 1 // since next_power_of_2(x) <= 2*x for all x
= 2 * (entries*2*n + 4*n - 2 - 1) + 1
= entries*4*n + 8*n - 1
```

Does that look right?

It should be possible to shorten the proof to roughly `O(sqrt(entries*4*n + 8*n - 1))` using the `ParallelSum` gadget. The downside is that proof generation/verification is more expensive. Still, the trade-off might be worth it here. For a working example of this, see the `CountVec` type.

Before going much further, I suggest we benchmark this code as described here: https://github.com/divviup/libprio-rs/pull/283#pullrequestreview-1082914192. We need to get a sense of how long proof generation takes and how long the Prio3 input shares are. We could then measure the concrete cost, in terms of computation, of shortening the proof.

cjpatton (changes requested):
This is looking great! Before going much further, I think it's time to add benchmarks. See inline comments in the `valid()` function.

src/field.rs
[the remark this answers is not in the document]
cjpatton: Ah, good catch!

src/vdaf/prio3.rs
cjpatton: 1. The "soundness error" refers to the probability that an invalid input gets deemed valid by the Aggregators. This is non-zero for any FLP system (i.e., any circuit one specifies for use in Prio3). The exact error depends on the circuit. Generally, it increases as the number size of the proof increases and decreases as the field size grows. The bound  can be roughly inferred from [BBCGGI19, Theorem 4.3], but no one has yet worked out the precise bound for Prio3. This is work in progress, DM if you're interested.
2. You might want to also make the type generic in the number representation (i.e., some implementation of `Fixed`).

src/flp/types/fixedpoint_l2.rs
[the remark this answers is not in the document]
cjpatton: (Missed this in my second pass, sorry!) Yes, I'd be OK with that. Since we won't be standardizing this any time, it seems reasonable to keep this feature experimental for now. However, one fairly straight forward optimization is outlined here: https://github.com/divviup/libprio-rs/pull/283#discussion_r954458953. It shouldn't be too bad to evaluate quickly, if you have the time.

src/flp/types/fixedpoint_l2.rs
```diff
@@ -0,0 +1,638 @@
...
+//! of `our_norm_on_encoded()` can take occurs in the case where all entries are
+//! `2^n-1`, in which case `d * 2^(2n-2)` is an upper bound to the result. The
+//! finite field used for encoding must be at least as large as this.
+//! For validating that the norm of the submitted vector lies in the correct
+//! range, consider the following:
+//!  - The result of `norm(xs)` should be in `[0,1)`.
+//!  - Thus, the result of `our_norm(xs)` should be in `[0,2^(2n-2))`.
+//!  - The result of `our_norm_on_encoded(ys)` should be in `[0,2^(2n-2))`.
+//! This means that the valid norms are exactly those representable with `2n-2`
+//! bits.
```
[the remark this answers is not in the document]
ooovi: almost there, just two things:
- we don't use two's complement representation for the measurments. instead we de-two's-complement the fixd point number (happens in `ComplatibleFloat::to_field_integer`) and interpret as unsigned. this corresponds to moving the [-2^(n-1),2^(n-1)) interval to [0,2^n)
- we don't use the `Mul` gadget, but call `PolyEval` of the range-0-1 check on each measurment vector entry.

src/flp/types/fixedpoint_l2/compatible_float.rs
```diff
@@ -0,0 +1,43 @@
...
+
+/// Assign a `Float` type to this type and describe how to represent this type as an integer of the
+/// given field, and how to represent a field element as the assigned `Float` type.
+pub trait CompatibleFloat<F: FieldElement> {
+    /// The float type F can be converted into.
+    type Float: Debug + Clone;
+    /// Represent a field element as `Float`, given the number of clients `c`.
+    fn to_float(t: F, c: usize) -> Self::Float;
+    /// Represent a value of this type as an integer in the given field.
+    fn to_field_integer(t: Self) -> <F as FieldElement>::Integer;
```
tgeoghegan, line 19: This is much easier to use as:

```suggestion
    fn to_field_integer(&self) -> <F as FieldElement>::Integer;
```

src/flp/types/fixedpoint_l2.rs
```diff
@@ -0,0 +1,638 @@
...
+        let mut res = vec![];
+        for d in data {
+            let decoded = <T as CompatibleFloat<F>>::to_float(*d, num_measurements);
+            res.push(decoded);
+        }
+        Ok(res)
+    }
+
+    fn gadget(&self) -> Vec<Box<dyn Gadget<F>>> {
+        // This gadgets checks that a field element is zero or one.
```
tgeoghegan, line 317:
```suggestion
        // This gadget checks that a field element is zero or one.
```

src/flp/types/fixedpoint_l2.rs
```diff
@@ -0,0 +1,638 @@
...
+            }
+        };
+        F::valid_integer_try_from(usize_max_norm_value)?;
+
+        Ok(Self {
+            bits_per_entry,
+            entries,
+            bits_for_norm,
+            range_01_checker: poly_range_check(0, 2),
+            square_computer: vec![F::zero(), F::zero(), F::one()],
```
tgeoghegan, line 243: To help clarify that this is a polynomial, since we lack a decent type for that:
```suggestion
            // polynomial: 0 + 0 * x + 1 * x^2
            square_computer: vec![F::zero(), F::zero(), F::one()],
```

src/flp/types/fixedpoint_l2.rs
```diff
@@ -0,0 +1,638 @@
...
+//! our_norm(xs) = 2^(2n-2) * norm(xs)^2
+//! ```
+//! Given a vector ys of numbers in the field integer encoding (in `[0,2^n)`),
+//! this gives the following equation:
+//! ```text
+//! our_norm_on_encoded(ys) = our_norm([dec(y) for y in ys])
+//!                         = sum_{y in ys} y^2 - (2^n)*y + 2^(2n-2)
+//! ```
+//! The constant and linear terms in the sum appear because the decoding function
+//! is not linear but only affine.
```
tgeoghegan, line 86: Add a newline, because otherwise this and the next line are rendered as a single paragraph.

```suggestion
//! is not linear but only affine.

```

tgeoghegan, line 86: I think this is awkwardly phrased, because if the decode function _were_ linear and not affine, then there would still be the linear term in `our_norm_on_encoded`, right? As written, this implies that the function being affine is to blame for both terms.

src/flp/types/fixedpoint_l2.rs
```diff
@@ -0,0 +1,638 @@
...
+//! ```
+//! Instead of computing the norm, we make two simplifications:
+//! 1. We ignore the square root, which means that we are actually computing
+//!    the square of the norm.
+//! 2. Since we work with integers, fractional values should not occur in the
+//!    computation. This is done by working with a factor of `2^(2n-2)`.
+//! This means that what is actually computed in this type is the following:
+//! ```text
+//! our_norm(xs) = 2^(2n-2) * norm(xs)^2
+//! ```
```
tgeoghegan, line 78: I'm confused by this. The idea is that we are mapping fractional norms in the range `[0, 1)` to integers in [`0, 2^(2n-2))`, right?

tgeoghegan (changes requested):
I think I understand the math, and it all makes sense to me (though my head hurts a little now). This is really cool!

Besides the other comments I left, I agree with Chris that we should take some benchmarks and see if we can hit different tradeoffs of time vs. space using `ParallelSum`.

src/flp/types/fixedpoint_l2.rs
```diff
@@ -0,0 +1,638 @@
...
+        )?;
+
+        // The submitted norm is also decoded from its bit-encoding, and
+        // compared with the computed norm.
+        let submitted_norm_enc = &input[self.range_norm_begin..self.range_norm_end];
+        let submitted_norm = F::decode_from_bitvector_representation(submitted_norm_enc)?;
+
+        let norm_check = computed_norm - submitted_norm;
+
+        // Finally, we require both checks to be successfull by computing a
```
tgeoghegan, line 399:
```suggestion
        // Finally, we require both checks to be successful by computing a
```

src/flp/types/fixedpoint_l2.rs
```diff
@@ -0,0 +1,638 @@
...
+//! is actually simpler.
+//!
+//! ### Norm computation
+//! The L2 norm of a vector xs of numbers in `[-1,1)` is given by:
+//! ```text
+//! norm(xs) = sqrt(sum_{x in xs} x^2)
+//! ```
+//! Instead of computing the norm, we make two simplifications:
+//! 1. We ignore the square root, which means that we are actually computing
+//!    the square of the norm.
```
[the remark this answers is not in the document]
ooovi: or square the bound. but since we're only comparing to the client submission who also computes the square, it'd be fine. other bounds would require some extra work tho as the bound 1 is actually only enforced by the encoding size

src/flp/types/fixedpoint_l2.rs
```diff
@@ -0,0 +1,638 @@
...
+//! 2. Since we work with integers, fractional values should not occur in the
+//!    computation. This is done by working with a factor of `2^(2n-2)`.
+//! This means that what is actually computed in this type is the following:
+//! ```text
+//! our_norm(xs) = 2^(2n-2) * norm(xs)^2
+//! ```
+//! Given a vector ys of numbers in the field integer encoding (in `[0,2^n)`),
```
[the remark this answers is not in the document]
ooovi: we use `x` for things that live in the fixed-point domain and `y` for things that live in the field integer domain

src/flp/types/fixedpoint_l2/compatible_float.rs
```diff
@@ -0,0 +1,44 @@
...
+}
+
+impl<U> CompatibleFloat<Field64> for FixedI16<U> {
```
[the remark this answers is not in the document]
ooovi: yeah, this is still a bit confusing to me, as:
- if one choses any combination of field and fixed that do not fit into each other, one gets an error only upon attempting to use the functions provided by this trait, instead of being informed by the compiler that the combination is incompatible
- the associated `Float` type's precision follows from the size of the `FieldElement::Integer`, and i'm not sure i can manage to write that down in a generic fashion
- it seems the choice of field together with vector size affects the soundness error as [chris pointed out](https://github.com/divviup/libprio-rs/pull/283#discussion_r953209470), which will in turn in some way affect the differential privacy guarantee we can give IIUC. i felt like whitelisting field-fixed-float tuples might make analysis less involved and be less confusing for our end user...

src/flp/types/fixedpoint_l2.rs
```diff
@@ -0,0 +1,638 @@
...
+///
+/// * `entries` - Iterator over the vector entries.
+/// * `bits_per_entry` - Number of bits one entry has.
+/// * `constant_part_multiplier` - A share of 1.
+/// * `sq` - The function used to compute the square of an entry.
+fn compute_norm_of_entries<F, Fs, SquareFun>(
+    entries: Fs,
+    bits_per_entry: usize,
+    constant_part_multiplier: F,
+    sq: &mut SquareFun,
```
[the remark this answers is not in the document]
MxmUrw: Thanks. (But I had to write `mut sq: SquareFun` instead.)

src/flp/types/fixedpoint_l2.rs
cjpatton: Thanks @MxmUrw and @ooovi for adding a benchmark. Here's what I get on my laptop:
```
prio3 fixedpoint16 boundedl2 vec (1000 entries) size = 406864
Benchmarking prio3 fixedpoint16 boundedl2 vec (1000 entries): Collecting 100 samples in estimated 6.0199 s (300                                                                                                               prio3 fixedpoint16 boundedl2 vec (1000 entries)
                        time:   [19.945 ms 20.012 ms 20.080 ms]
                        change: [-29.655% -29.125% -28.655%] (p = 0.00 < 0.05)
                        Performance has improved.
```

The client runtime (20 ms) is definitely on the high end for Prio3 types, but I think reasonable for many applications. I'm more worried about the size of the input shares (406,864 bytes, or roughly 397KB). Thus I think it would be worth considering optimizations to the circuit that would drive down the size of the proof.

I would strongly recommend we use the `ParallelSum` gadget. I put together a change to demonstrate how this would be used for the range check part of the circuit: https://github.com/divviup/libprio-rs/commit/21491ac702ab06739f57dedea915b3b54ee76815.

```
prio3 fixedpoint16 boundedl2 vec (1000 entries) size = 152864
Benchmarking prio3 fixedpoint16 boundedl2 vec (1000 entries): 
[...]
prio3 fixedpoint16 boundedl2 vec (1000 entries)
                        time:   [64.192 ms 64.988 ms 65.953 ms]
                        change: [+220.42% +224.74% +230.01%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 14 outliers among 100 measurements (14.00%)
  4 (4.00%) high mild
  10 (10.00%) high severe
```

The headline here is that the input shares are now just 149KB. The downside is that the client runtime is tripled, due to the more expensive (i.e., higher arithmetic degree) circuit (see the BlindPolyEval gadget). However:
1. By multithreading the circuit evaluation, we can make this competitive with a lower degree circuit. See https://github.com/cfrg/draft-irtf-cfrg-vdaf/issues/83.
2. As the proof shrinks, proof generation time tends to improve. See https://github.com/cfrg/draft-irtf-cfrg-vdaf/issues/83.

We can use a similar trick for the second gadget as well, driving down the proof size even further. However, for this gadget we wouldn't even need the more expensive BlindPolyEval gadget, since all we want to compute is the sum of the squares.

src/vdaf/prio3.rs
```diff
@@ -130,6 +136,33 @@ impl Prio3Aes128Sum {
     }
 }
 
+/// The fixed point vector sum type. Each measurement is a vector of 16-bit fixed point decimals
+/// with 15 fractional digits and the aggregate is the sum. The verification function ensures the
+/// L2 norm of the vector is <= 1.
```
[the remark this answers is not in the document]
ooovi: i wrote a small description in the type docstring.

concerning the example, i'm not sure it's going to be very illustrative at this point. the norm bound is a requirement for differential privacy, but we have not yet worked out where and how to noise, so the example would just have this complicated-to-check bound for no reason. also we intend to use this in a crate that provides python bindings so the data and model handling can be done in a language familiar to ML users. what do you think?

src/flp/types/fixedpoint_l2.rs
MxmUrw: No, I would actually claim that the statement is correct as stated: Assume that the decoding function is linear, i.e.:

there is some `a : R`, such that:
```
dec : R -> R
dec(y) = a * y
```
then `our_norm_on_encoded`  would be (ignoring the `2^(2n-2)` factor which is irrelevant here, and which I am also going to explain in another comment)
```
our_norm_on_encoded = sum_{y in ys} dec(y)^2
                    = sum_{y in ys} (a*y)^2
```
where the summands have neither linear nor constant parts.

src/flp/types/fixedpoint_l2.rs
MxmUrw: The reason for the `2^(2n-2)` factor is the following: The decoding of values from the range `[0,2^n)` to `[-1,1)` involves fractions (as we go from the integer representation to the "real/float" representation). Thus, when plugging the decoding into the norm computation function (which is what we are doing here), we also get fractions in the composite. Looking at these fractions one notices that the both of them (in the square and in the linear term) disappear if we multiply the whole thing with `2^(2n-2)`. This way, in the computation now again only integral values appear (which is required if we are to execute it directly on field elements.

So our full `our_norm_on_encoded` function can be said to be a composition of three functions:
```
                    dec()                             norm()                   "mult with 2^(2n-2)"
vector of [0,2^n)    ->       vector of [-1,1)          ->   not negative reals         ->         [0,2^(2n-2))
                      (fractions with denom of 2^(n-1))      (denom of 2^(2n-2))
```

I hope that answers the question somewhat.

MxmUrw:
We tried to switch to `Field128` (as that is [apparently safer](https://github.com/divviup/libprio-rs/pull/283#discussion_r953209470)), and also to implement and test our type with larger fixedpoint-bitwidths. We ran into a problem with 64bit fixedpoints, when we check that the largest possibly occurring norm fits into the field: Converting numbers to field integers can currently only be done from `usize`, yet the number we want to convert is larger than that.

We see the following options:
 - Add a `TryFrom<u128>` for `FieldElement::Integer` bound. This has as disadvantage that every time one wants to call `try_from(n)` for a literal `n`, its type has to be annotated, as rust cannot infer which type to use.
 - Replace the `TryFrom<usize>` with `TryFrom<u128>`. This has as disadvantage that code needs to be updated in places where currently `usize` is provided.
 - Change the code on our side, to circumvent the restriction to `usize`: Instead of calculating `max_norm = entries * (2 * (1 << bits_per_entry))` and then checking that `max_norm` fits into the field, we could probably first put `entries` and `bits_per_entry` into the field and use some trick to ensure that the computation of `max_norm` inside the field does not overflow.

The last option would not require changes to other prio code, but the restriction to conversions from `usize` seems rather arbitrary, also because the bitwidth of `usize` is architecture dependent. Is there reason for that which we are not aware of?

cjpatton:
> We tried to switch to `Field128` (as that is [apparently safer](https://github.com/divviup/libprio-rs/pull/283#discussion_r953209470)), and also to implement and test our type with larger fixedpoint-bitwidths. We ran into a problem with 64bit fixedpoints, when we check that the largest possibly occurring norm fits into the field: Converting numbers to field integers can currently only be done from `usize`, yet the number we want to convert is larger than that.
> 
> We see the following options:
> 
>     * Add a `TryFrom<u128>` for `FieldElement::Integer` bound. This has as disadvantage that every time one wants to call `try_from(n)` for a literal `n`, its type has to be annotated, as rust cannot infer which type to use.
> 
>     * Replace the `TryFrom<usize>` with `TryFrom<u128>`. This has as disadvantage that code needs to be updated in places where currently `usize` is provided.
> 
>     * Change the code on our side, to circumvent the restriction to `usize`: Instead of calculating `max_norm = entries * (2 * (1 << bits_per_entry))` and then checking that `max_norm` fits into the field, we could probably first put `entries` and `bits_per_entry` into the field and use some trick to ensure that the computation of `max_norm` inside the field does not overflow.

I may be misunderstanding the problem, but why can't you convert to `F::Integer`, then to `F`? `Field128::Integer` is `u128`, so something like this should work (it's a bit overly verbose, I grant you, but it's generic!):
```rust
let max_norm = F::Integer::try_from(entries).unwrap()
                          * (F::Integer::try_from(2).unwrap()
                          * (F::Integer::try_from(1).unwrap() << F::integer::try_from(bits_per_entry).unwrap()));
```

FWIW, this is how we've solve this problem elsewhere in the repo.

> The last option would not require changes to other prio code, but the restriction to conversions from `usize` seems rather arbitrary, also because the bitwidth of `usize` is architecture dependent. Is there reason for that which we are not aware of?

You're right that's arbitrary. The history here is that we wanted one (and only) one way to convert a literal to the integer representation for the field. `usize` seemed like a reasonable choice because it's the type we have most often. If someone wants to make a convincing argument to replace `TryFrom<usize>` with `TryFrom<u128>`, I'm all ears. However I would not be in favor of adding it, as it sets a precedent for adding bounds for other primitive integers. This would incur tech debt, especially as we add additional fields and update implementations of existing ones.

@MxmUrw this is looking close to landing. I plan to do a complete pass next, but first, would you mind rebasing and squashing your branch? There are currently conflicts.

MxmUrw:
> If someone wants to make a convincing argument to replace `TryFrom<usize>` with `TryFrom<u128>`, I'm all ears. However I would not be in favor of adding it, as it sets a precedent for adding bounds for other primitive integers. This would incur tech debt, especially as we add additional fields and update implementations of existing ones.

Ok, that makes sense. Then we're not gonna suggest any changes here :)

> I may be misunderstanding the problem, but why can't you convert to F::Integer, then to F? Field128::Integer is u128, so something like this should work (it's a bit overly verbose, I grant you, but it's generic!):

Thanks, I did something like this now. But what was still additionally needed in our case is a bound `u128: TryFrom<F::Integer>` (which I simply added to *our* type), such that I could get the value of the modulus of the field, in order to ensure that all the numbers that occur during computations do not overflow/wrap around in the field.

ooovi:
> @MxmUrw this is looking close to landing. I plan to do a complete pass next, but first, would you mind rebasing and squashing your branch? There are currently conflicts.

i did, but CI is failing because our `fixed` dependency does things that made clippy unhappy back in version 1.58. clippy is happy with my local 1.63. i see you [support version 1.58 on purpose](https://github.com/divviup/libprio-rs/issues/284). want me to make a feature after all?

cjpatton:
> > @MxmUrw this is looking close to landing. I plan to do a complete pass next, but first, would you mind rebasing and squashing your branch? There are currently conflicts.
> 
> i did, but CI is failing because our `fixed` dependency does things that made clippy unhappy back in version 1.58. clippy is happy with my local 1.63. i see you [support version 1.58 on purpose](https://github.com/divviup/libprio-rs/issues/284). want me to make a feature after all?

Thanks @ooovi. I'm sort of hoping we can resolve this by bumping MSRV to 1.63, but I'll leave that up to @tgeoghegan. Are we using the latest version of `fixed`?

ooovi:
> Are we using the latest version of `fixed`?

we are now.

tgeoghegan:
I'm a little wary of bumping MSRV because we don't know how many people are out there using this crate and whether they are all in a position to update their toolchains. Is the only problem clippy diagnostics? Does the the code otherwise work as expected? If so I think we could get past this with a `#[allow(clippy::whichever_diagnostic_it_was)]`.

MxmUrw:
> Is the only problem clippy diagnostics? Does the the code otherwise work as expected?

No unfortunately we were a little bit confused. These were not clippy errors but compilation errors. After updating the `fixed` crate to the newest one it clearly says:
```
error: package `fixed v1.19.0` cannot be built because it requires rustc 1.61 or newer,
while the currently active rustc version is 1.58.1
```

cjpatton:
Alright, in my view putting `fixed` behind a feature is not a viable solution, since the MSRV would then depend on what feature flags you need. I see just two options:
1. Downgrade fixed to a version that builds with 1.58
2. Bump MSRV of prio to >=1.61

I think we should do (2.). It seems like Rust is still too immature as a language to count on other libraries to maintain backwards compatibility. In particular, I'm betting that we'll run into this situation in the future.

MxmUrw:
I would add that in our specific case there could be a third option:
 - Remove the `fixed` dependency altogether. We could switch our measurement type over to simply use `u16` (, `u32`, ...) and then document how a fixedpoint should be encoded in that. For users that (externally) use the `fixed` crate, converting a fixedpoint number to the bitrepresentation would be a simple call `number.to_bits()`, and, currently, this function is the only thing we actually use from `fixed` - mainly we work with our custom representations.

tgeoghegan:
I'm conflicted on this because I'm not sure what the etiquette is for bumping MSRV. Upgrading Rust toolchains is easy and the language's strong backward compatibility promises make it pretty low-risk, but I worry that there may be users of `prio` out there for whom using a newer Rust compiler is not just a matter of running `rustup update`. But Chris is also right that since this crate is pretty immature, users should expect lots of churn. I think we should go ahead and bump MSRV to 1.61, and if someone is trying to build Prio with a dusty old rustc that came with their distro, well, we'll deal with that.

cjpatton:
> I would add that in our specific case there could be a third option:
> 
>     * Remove the `fixed` dependency altogether. We could switch our measurement type over to simply use `u16` (, `u32`, ...) and then document how a fixedpoint should be encoded in that. For users that (externally) use the `fixed` crate, converting a fixedpoint number to the bitrepresentation would be a simple call `number.to_bits()`, and, currently, this function is the only thing we actually use from `fixed` - mainly we work with our custom representations.

This is an interesting point. If at some point we wanted to standardize your circuit in the VDAF spec, then we would need to spell out the fixed-length encoding anyway. But I think we should ONLY go this route IF we think the encoding implemented in `fixed` shouldn't be standardized, and we think a different encoding should be standardized instead. @MxmUrw, @ooovi, I think it's up to you to help make a call here:

* If you think we should standardize an encoding other than what's implemented in `fixed`, then we can try to work this out now and avoid the new dependency altogether.
* Regardless of whether we should standardize `fixed`, we can take the dependency now, bump MSRV to 1.61, and revisit later.

src/flp/types/fixedpoint_l2.rs
```diff
@@ -0,0 +1,816 @@
+// SPDX-License-Identifier: MPL-2.0
+
+//! A [`Type`] for summing vectors of fixed point numbers whose
+//! [L2 norm](https://en.wikipedia.org/wiki/Norm_(mathematics)#Euclidean_norm) is
+//! bounded by `1`.
```
cjpatton, line 5: Grammar: "where the L2 norm of each vector is bounded by 1".

benches/speed_tests.rs
```diff
@@ -188,63 +192,215 @@ pub fn count_vec(c: &mut Criterion) {
 /// Benchmark prio3 client performance.
 pub fn prio3_client(c: &mut Criterion) {
     let num_shares = 2;
+    let len = 1000;
```
cjpatton, line 195: There's a lot of code movement here, which makes it very difficult to review carefully. Because there is a functional change here, would you try writing the change in a way that minimizes size of the diff?

src/flp.rs
```diff
@@ -257,7 +257,7 @@ pub trait Type: Sized + Eq + Clone + Debug {
                     return Err(FlpError::Prove(format!(
                         "short prove randomness: got {}; want {}",
```
cjpatton, line 258: Hmm, I guess what we want this error to convey is that the proof rand is too long? Maybe we something like: "short prove randomness: got {}; want at least {}"

src/flp.rs
```diff
@@ -257,7 +257,7 @@ pub trait Type: Sized + Eq + Clone + Debug {
...
                         prove_rand.len(),
-                        self.prove_rand_len()
+                        prove_rand_len + inner_arity // self.prove_rand_len()
```
cjpatton, line 260: This comment is misleading, since prove_rand_len + inner_arity might not be equal to self.prove_rand_len(). Maybe just remove?

src/flp/types.rs
```diff
@@ -9,6 +9,8 @@ use crate::polynomial::poly_range_check;
 use std::convert::TryInto;
 use std::marker::PhantomData;
 
+pub mod fixedpoint_l2;
```
cjpatton, line 12: nit: We have been putting submodule declarations at the end of the file.

src/flp/types/fixedpoint_l2.rs
```diff
@@ -0,0 +1,816 @@
...
+//!
+//! In the following a high level overview over the inner workings of this type
+//! is given and implementation details are discussed. It is not necessary for
+//! using the type, but it should be very helpful when trying to understand the
```
cjpatton, line 9: nit: The writing style here should be formal(-ish). Avoid using words like "very" that don't add information.

src/flp/types/fixedpoint_l2.rs
```diff
@@ -0,0 +1,816 @@
...
+//!
+//! ```text
+//!                    map of dec()                    norm()          "mult with 2^(2n-2)"
+//!   vector of [0,2^n)    ->       vector of [-1,1)    ->       [0,1)         ->           [0,2^(2n-2))
+//!                                         ^                      ^
+//!                                         |                      |
+//!                     fractions with denom of 2^(n-1)     fractions with denom of 2^(2n-2)
+//! ```
+//! (note that the ranges on the LHS and RHS of `"mult with 2^(2n-2)"` are stated
+//!  here for vectors with a norm less than `1`)
```
cjpatton, line 92:
```suggestion
//! (Note that the ranges on the LHS and RHS of `"mult with 2^(2n-2)"` are stated
//! here for vectors with a norm less than `1`.)
```

src/flp/types/fixedpoint_l2.rs
```diff
@@ -0,0 +1,816 @@
...
+//! they claim it has.
+//! The submission is a vector of field elements laid out as follows:
+//! ```text
+//! |---- bits_per_entry * entries ----|---- bits_for_norm ----|
+//!  ^                                  ^
+//!  \- the input vector entries        |
+//!                                     \- the encoded norm
+//! ```
+//!
+//! ### Validity
```
cjpatton, line 147: I think this subsubsection can be deleted. (It's made pretty clear above what is validated.)

src/flp/types/fixedpoint_l2.rs
```diff
@@ -0,0 +1,816 @@
...
+//!
+//! In addition to checking that every submission entry is `0` or `1`, the validation
+//! circuit of this type computes the norm and compares to what the client
+//! claimed.
+//!
+//! ### Value `1`
+//!
+//! We actually do not allow the submitted norm or vector entries to be
+//! exactly `1`, but rather require them to be strictly less. Supporting `1` would
+//! entail a more fiddly encoding and is not necessary for our usecase.
```
cjpatton, line 157: It might be worth documenting the largest represented number, say, for the 16-bit fixed type.

src/flp/types/fixedpoint_l2.rs
```diff
@@ -0,0 +1,816 @@
...
+    SPoly: ParallelSumGadget<F, PolyEval<F>>,
+    SBlindPoly: ParallelSumGadget<F, BlindPolyEval<F>>,
+> {
+    bits_per_entry: usize,
+    fi_bits_per_entry: F::Integer,
+    entries: usize,
+    bits_for_norm: usize,
+    range_01_checker: Vec<F>,
+    norm_summand_poly: Vec<F>,
+    phantom: PhantomData<(T, SPoly, SBlindPoly)>,
```
cjpatton, line 196: nit: Add new line here

src/flp/types/fixedpoint_l2.rs
```diff
@@ -0,0 +1,816 @@
...
+}
+
+impl<
+        T: Fixed,
+        F: FieldElement,
+        SPoly: ParallelSumGadget<F, PolyEval<F>>,
+        SBlindPoly: ParallelSumGadget<F, BlindPolyEval<F>>,
+    > FixedPointBoundedL2VecSum<T, F, SPoly, SBlindPoly>
+where
+    u128: TryFrom<F::Integer>,
```
cjpatton, line 215: Nice one!

src/flp/types/fixedpoint_l2.rs
```diff
@@ -0,0 +1,816 @@
...
+    gadget1_calls: usize,
+    gadget1_chunk_len: usize,
+}
+
+impl<
+        T: Fixed,
+        F: FieldElement,
+        SPoly: ParallelSumGadget<F, PolyEval<F>>,
+        SBlindPoly: ParallelSumGadget<F, BlindPolyEval<F>>,
+    > FixedPointBoundedL2VecSum<T, F, SPoly, SBlindPoly>
```
cjpatton, line 213: Readability: Move trait bounds to the `where` block below
```suggestion
impl<T, F, SPoly, SBlindPoly>
```

src/flp/types/fixedpoint_l2.rs
```diff
@@ -0,0 +1,816 @@
...
+where
+    u128: TryFrom<F::Integer>,
+{
+    /// Return a new [`FixedPointBoundedL2VecSum`] type parameter. Each value of this type is a
+    /// fixed point vector with `entries` entries.
+    pub fn new(entries: usize) -> Result<Self, FlpError> {
+        // (0) initialize constants
+
```
cjpatton, line 221:
```suggestion
```

src/flp/types/fixedpoint_l2.rs
```diff
@@ -0,0 +1,816 @@
...
+        let fi_one = F::Integer::from(F::one());
+        let fi_two = fi_one + fi_one;
+
+        // (I) Check that the fixed type `F` is compatible.
+
```
cjpatton, line 226: Connect continuous block comments by adding a blank comment line. Or just remove the empty line, as you do below.
```suggestion
           //
```

src/flp/types/fixedpoint_l2.rs
```diff
@@ -0,0 +1,816 @@
...
+            return Err(FlpError::Encode(format!(
+                "fixed point type bit length ({}) too large for field modulus",
+                bits_per_entry,
+            )));
+        }
+        let fi_bits_per_entry: F::Integer = F::valid_integer_try_from(bits_per_entry)
+            .map_err(|_| FlpError::Encode("Could not create FieldInteger from usize even though size check was successful".to_string()))?;
+
+        // (II) Check that the field is large enough for the norm.
+
```
cjpatton, line 251: Connect continuous block comment

src/flp/types/fixedpoint_l2.rs
```diff
@@ -0,0 +1,816 @@
...
+        }
+
+        // Construct the polynomial that computes a part of the norm for a
+        // single vector entry.
+        //
+        // the linear part is 2^n,
+        // the constant part is 2^(2n-2),
+        // the polynomial is:
+        //   p(y) = 2^(2n-2) + -(2^n) * y + 1 * y^2
+        let linear_part = fi_one << (fi_bits_per_entry);
```
cjpatton, line 291: Curious that clippy doesn't wine here.
```suggestion
        let linear_part = fi_one << fi_bits_per_entry;
```

src/flp/types/fixedpoint_l2.rs
```diff
@@ -0,0 +1,816 @@
...
+            // configuration of parallel sum gadgets
+            gadget0_calls,
+            gadget0_chunk_len,
+            gadget1_calls,
+            gadget1_chunk_len,
+        })
+    }
+}
+
+impl<
```
cjpatton, line 332: Same comment (move trait bounds to `where` block)

src/flp/types/fixedpoint_l2.rs
```diff
@@ -0,0 +1,816 @@
...
+/// Depending on the size of the vector that needs to be transmitted, a corresponding field type has
+/// to be chosen for `F`. For a `n`-bit fixed point type and a `d`-dimensional vector, the field
+/// modulus needs to be larger than `d * 2^(2n-2)` so there are no overflows during norm validity
+/// computation.
+#[derive(Debug, PartialEq, Eq)]
+pub struct FixedPointBoundedL2VecSum<
+    T: Fixed,
+    F: FieldElement,
+    SPoly: ParallelSumGadget<F, PolyEval<F>>,
+    SBlindPoly: ParallelSumGadget<F, BlindPolyEval<F>>,
```
cjpatton, line 188: Move trait bounds to where block (see comment below)

src/flp/types/fixedpoint_l2.rs
```diff
@@ -0,0 +1,816 @@
...
+        })
+    }
+}
+
+impl<
+        T: Fixed,
+        F: FieldElement,
+        SPoly: ParallelSumGadget<F, PolyEval<F>>,
+        SBlindPoly: ParallelSumGadget<F, BlindPolyEval<F>>,
+    > Clone for FixedPointBoundedL2VecSum<T, F, SPoly, SBlindPoly>
```
cjpatton, line 337: Is it possible to derive this automatically? Try adding "Clone" to "#[derive(Debug, PartialEq, Eq)]" above the type declaration above.

src/flp/types/fixedpoint_l2.rs
```diff
@@ -0,0 +1,816 @@
...
+}
+
+impl<T, F, SPoly, SBlindPoly> Type for FixedPointBoundedL2VecSum<T, F, SPoly, SBlindPoly>
+where
+    T: Fixed + CompatibleFloat<F>,
+    F: FieldElement,
+    SPoly: ParallelSumGadget<F, PolyEval<F>> + Eq + 'static,
+    SBlindPoly: ParallelSumGadget<F, BlindPolyEval<F>> + Eq + 'static,
+{
+    const ID: u32 = 0xFFFF0000;
```
cjpatton, line 369: This codepoint is used for another type, so we'll need a different one. (cc/ @divergentdave, who might have an opinion about what point to allocate).

Relatedly, we should write unit tests to ensure that the codepoints for all of our "private" VDAFs are distinct.  This is separate from this PR, however.

src/flp/types/fixedpoint_l2.rs
```diff
@@ -0,0 +1,816 @@
...
+    type Measurement = Vec<T>;
+    type AggregateResult = Vec<f64>;
+    type Field = F;
+
+    fn encode_measurement(&self, fp_entries: &Vec<T>) -> Result<Vec<F>, FlpError> {
+        // Convert the fixed-point encoded input values to field integers. We do
+        // this once here because we need them for encoding but also for
+        // computing the norm.
+        let integer_entries: Vec<_> = fp_entries.iter().map(|x| x.to_field_integer()).collect();
```
cjpatton, line 378: You should be able to skip `collect()`, which would save a malloc. You might need a type hint elsewhere.

```suggestion
        let integer_entries: Vec<_> = fp_entries.iter().map(|x| x.to_field_integer());
```

src/flp/types/fixedpoint_l2.rs
```diff
@@ -0,0 +1,816 @@
...
+
+        // This gadget computes the square of a field element.
+        // It is called on each entry during norm computation.
+        let gadget1 = SPoly::new(
+            PolyEval::new(self.norm_summand_poly.clone(), self.gadget1_calls),
+            self.gadget1_chunk_len,
+        );
+
+        let res: Vec<Box<dyn Gadget<F>>> = vec![Box::new(gadget0), Box::new(gadget1)];
+        res
```
cjpatton, line 436: Readability: There shoiuldn't be a need to declare the type.
```suggestion
        vec![Box::new(gadget0), Box::new(gadget1)]
```

src/flp/types/fixedpoint_l2.rs
```diff
@@ -0,0 +1,816 @@
...
+
+        let f_num_shares = F::from(F::valid_integer_try_from(num_shares)?);
+        let constant_part_multiplier = F::one() / f_num_shares;
+
+        // Ensure that all submitted field elements are either 0 or 1.
+        // This is done for:
+        //  - all vector entries (each of them encoded in `self.bits_per_entry`
+        //    field elements)
+        //  - the submitted norm (encoded in `self.bits_for_norm` field
+        //    elements)
```
cjpatton, line 456: Links to the inline comments above
```suggestion
        //  (I) all vector entries (each of them encoded in `self.bits_per_entry`
        //     field elements)
        //  (II) the submitted norm (encoded in `self.bits_for_norm` field
        //    elements)
```

src/flp/types/fixedpoint_l2.rs
```diff
@@ -0,0 +1,816 @@
...
+        //
+        // Since all input vector entry (field-)bits, as well as the norm bits,
+        // are contiguous, we do the check directly for all bits from 0 to
+        // entries*bits_per_entry + bits_for_norm.
+        //
+        // In order to keep the proof size down, this is done using the
+        // `ParallelSum` gadget. For a similar application see the `CountVec`
+        // prio3 type.
```
cjpatton, line 464:
```suggestion
        // `ParallelSum` gadget. For a similar application see the `CountVec`
        // type.
```

src/flp/types/fixedpoint_l2.rs
```diff
@@ -0,0 +1,816 @@
...
+        //
+        // Check that each element is a 0 or 1:
```
cjpatton, line 466:
```suggestion
```

src/flp/types/fixedpoint_l2.rs
```diff
@@ -0,0 +1,816 @@
...
+            .map(F::decode_from_bitvector_representation)
+            .collect();
+
+        // run parallel sum gadget on the decoded entries
+        let computed_norm = {
+            let mut outp = F::zero();
+
+            for chunk in decoded_entries?.chunks(self.gadget1_chunk_len) {
+                let d = chunk.len();
+
```
cjpatton, line 514:
```suggestion
```

src/flp/types/fixedpoint_l2.rs
```diff
@@ -0,0 +1,816 @@
...
+        // inner polynomial on each (decoded) vector entry, and then sums the
+        // results. Note that the gadget is not called on the whole vector at
+        // once, but sequentially on chunks of size `self.gadget1_chunk_len` of
+        // it. The results of these calls are accumulated in the `outp` variable.
+        //
+        // decode the bit-encoded entries into elements in the range [0,2^n):
+        let decoded_entries: Result<Vec<_>, _> = input[0..self.entries * self.bits_per_entry]
+            .chunks(self.bits_per_entry)
+            .map(F::decode_from_bitvector_representation)
+            .collect();
```
cjpatton, line 506: You might be able to avoid this collect().

src/flp/types/fixedpoint_l2.rs
```diff
@@ -0,0 +1,816 @@
...
+
+        // run parallel sum gadget on the decoded entries
+        let computed_norm = {
```
cjpatton, line 509: Readability: I would expect this code block to look more or less like the computation of `range_check` above. This could we be a lot more readable if they had the same overall structure.

src/flp/types/fixedpoint_l2.rs
```diff
@@ -0,0 +1,816 @@
...
+        // Finally, we require both checks to be successful by computing a
+        // random linear combination of them.
+        let out = joint_rand[1] * range_check + (joint_rand[1] * joint_rand[1]) * norm_check;
+        Ok(out)
+    }
+
+    fn truncate(&self, input: Vec<F>) -> Result<Vec<Self::Field>, FlpError> {
+        self.truncate_call_check(&input)?;
+
+        let mut decoded_vector = vec![];
```
cjpatton, line 545: You can help the compiler save mallocs by predicting the size of buffer you'll need.
```suggestion
        let mut decoded_vector = Vec::with_capacity(self.entries);
```

src/flp/types/fixedpoint_l2.rs
```diff
@@ -0,0 +1,816 @@
...
+
+    fn input_len(&self) -> usize {
+        self.bits_per_entry * self.entries + self.bits_for_norm
+    }
+
+    fn proof_len(&self) -> usize {
+        // computed via
+        // `gadget.arity() + gadget.degree()
+        //   * ((1 + gadget.calls()).next_power_of_two() - 1) + 1;`
+
```
cjpatton, line 565: Spurious newline
```suggestion
```

src/flp/types/fixedpoint_l2.rs
```diff
@@ -0,0 +1,816 @@
...
+{
+    let fi_one = F::Integer::from(F::one());
+    let fi_two = fi_one + fi_one;
+
+    // The value that is computed here is:
+    //    sum_{y in entries} 2^(2n-2) + -(2^n) * y + 1 * y^2
+    //
+    // Check out the norm computation bit in the explanatory comment block for
+    // more information.
+
```
cjpatton, line 613:
```suggestion
       //
```

src/flp/types/fixedpoint_l2.rs
```diff
@@ -0,0 +1,816 @@
...
+    // Initialize `norm_accumulator`.
+    let mut norm_accumulator = F::zero();
+
+    // constants
+    let linear_part = fi_one << (fi_bits_per_entry); // = 2^(2n-2)
```
cjpatton, line 618:
```suggestion
    let linear_part = fi_one << fi_bits_per_entry; // = 2^(2n-2)
```

src/vdaf/prio3.rs
```diff
@@ -127,6 +130,77 @@ impl Prio3Aes128Sum {
     }
 }
 
+/// The fixed point vector sum type. Each measurement is a vector of fixed point numbers
+/// and the aggregate is the sum represented as 64-bit floats. The verification function
```
cjpatton, line 134:
```suggestion
/// and the aggregate is the sum represented as 64-bit floats. The preparation phase
```

src/vdaf/prio3.rs
```diff
@@ -127,6 +130,77 @@ impl Prio3Aes128Sum {
...
+/// ensures the L2 norm of the input vector is < 1.
+///
+/// This is useful for aggregating gradients in a federated version of
+/// [gradient descent](https://en.wikipedia.org/wiki/Gradient_descent) with
+/// [differential privacy](https://en.wikipedia.org/wiki/Differential_privacy),
+/// useful e.g. for [differentially private deep learning](https://arxiv.org/pdf/1607.00133.pdf).
```
cjpatton, line 140:
```suggestion
/// useful, e.g.. for [differentially private deep learning](https://arxiv.org/pdf/1607.00133.pdf).
```

src/vdaf/prio3.rs
```diff
@@ -127,6 +130,77 @@ impl Prio3Aes128Sum {
...
+#[cfg(feature = "crypto-dependencies")]
+impl<Fx: Fixed + CompatibleFloat<Field128>> Prio3Aes128FixedPointBoundedL2VecSum<Fx> {
+    /// Construct an instance of this VDAF with the given number of aggregators and number of
+    /// vector entries.
+    pub fn new_aes128_fixedpoint_boundedl2_vec_sum(
+        num_aggregators: u8,
+        entries: usize,
+    ) -> Result<Self, VdafError> {
+        check_num_aggregators(num_aggregators)?;
+
```
cjpatton, line 166:
```suggestion
```

src/vdaf/prio3.rs
```diff
@@ -127,6 +130,77 @@ impl Prio3Aes128Sum {
...
+#[cfg_attr(docsrs, doc(cfg(feature = "multithreaded")))]
+impl<Fx: Fixed + CompatibleFloat<Field128>> Prio3Aes128FixedPointBoundedL2VecSumMultithreaded<Fx> {
+    /// Construct an instance of this VDAF with the given number of aggregators and number of
+    /// vector entries.
+    pub fn new_aes128_fixedpoint_boundedl2_vec_sum_multithreaded(
+        num_aggregators: u8,
+        entries: usize,
+    ) -> Result<Self, VdafError> {
+        check_num_aggregators(num_aggregators)?;
+
```
cjpatton, line 199:
```suggestion
```

cjpatton (changes requested):
Very nice work. Looks functionally correct, just nit-picky comments left from me. 

Also, `cargo bench` fails:
```
chris:~/github.com/divviup/libprio-rs$ cargo bench
   Compiling prio v0.9.0 (/Users/chris/github.com/divviup/libprio-rs)
error[E0412]: cannot find type `ParallelSumMultithreaded` in this scope
   --> benches/speed_tests.rs:278:13
    |
278 |             ParallelSumMultithreaded<F, PolyEval<F>>,
    |             ^^^^^^^^^^^^^^^^^^^^^^^^ not found in this scope

error[E0412]: cannot find type `ParallelSumMultithreaded` in this scope
   --> benches/speed_tests.rs:279:13
    |
279 |             ParallelSumMultithreaded<F, BlindPolyEval<F>>,
    |             ^^^^^^^^^^^^^^^^^^^^^^^^ not found in this scope

For more information about this error, try `rustc --explain E0412`.
error: could not compile `prio` due to 2 previous errors
warning: build failed, waiting for other jobs to finish...
^C  Building [=======================> ] 153/155: cycle_counts(bench), prio(test)
```

src/flp/types/fixedpoint_l2.rs
ooovi: it's a bit different than the one above, as the norm computation does not use BlindPolyEval so the chunk structure is simpler. i added a comment to explain and adjusted things a bit.

benches/speed_tests.rs
MxmUrw: Removed our superfluous code changes, the diff should now be readable.

src/flp/types/fixedpoint_l2.rs
ooovi: apart from collecting the iterator it also collects the `Vec<Result<_,_>> into a `Result<Vec<_>,_>` which is nice! can it stay?

src/flp/types/fixedpoint_l2.rs
ooovi: `SPoly` and `SBlindPoly` would need to be `Clone`able for that :(

MxmUrw: No, unfortunately this does not work, because the `ParallelSumGadget`s are not clonable. In our type they only appear in as phantom data, but it seems that rust does not take that into account:

```
error[E0277]: the trait bound `SPoly: Clone` is not satisfied
   --> src/flp/types/fixedpoint_l2.rs:357:31
    |
357 | impl<T, F, SPoly, SBlindPoly> Type for FixedPointBoundedL2VecSum<T, F, SPoly, SBlindPoly>
    |                               ^^^^ the trait `Clone` is not implemented for `SPoly`
    |
note: required because of the requirements on the impl of `Clone` for `FixedPointBoundedL2VecSum<T, F, SPoly, SBlindPoly>`
   --> src/flp/types/fixedpoint_l2.rs:179:32
    |
179 | #[derive(Debug, PartialEq, Eq, Clone)]
    |                                ^^^^^
```

cjpatton: Indeed, the problem here is that the compiler doesn't know if the instantiation of ParallelSumGadget implements Clone. This fix here is tell the compiler that it must by refining the trait bounds. Here's one solution: https://github.com/divviup/libprio-rs/commit/b212dcf8a4744f3665492796cb81563dc425fcab

MxmUrw:
>  @MxmUrw, @ooovi, I think it's up to you to help make a call here

Our view on this matter is the following:
 - The encoding of the `fixed` crate seems sensible.
 - We use the macros from this crate to easily specify fixed point numbers. 
 - By using its types the external interface of our prio3 type is "typed". Users can make less accidental mistakes by using the wrong encoding.

Thus we think we should keep this dependency.

src/flp/types/fixedpoint_l2.rs
divergentdave: We currently use `0xFFFF0000` in six places, including Prio3Aes128Average, Prio3Aes128CountVec, Poplar with ToyIDPF, and Prio v2. I may have misunderstood you in https://github.com/divviup/libprio-rs/pull/289#discussion_r956228387, should we be using distinct codepoints for other non-standardized VDAFs as well? (If so, that'll call for another issue/PR)

cjpatton: You did not misunderstand me, I'm contradicting my past self! (Sorry about that.) Let's leave this here for now and we'll follow up in the issue you filed.

tgeoghegan:
We talked about this among the Divvi Up team. I think we're OK with bumping the MSRV to 1.61, and then releasing this as 0.10.x, on the premise that bumping the compiler version should not be considered semver-compatible. @branlwyd does that sound alright to you?

branlwyd:
> We talked about this among the Divvi Up team. I think we're OK with bumping the MSRV to 1.61, and then releasing this as 0.10.x, on the premise that bumping the compiler version should not be considered semver-compatible. @branlwyd does that sound alright to you?

Yes, this approach SGTM. IMO it is strictly true that an MSRV bump is a breaking change.

That said, it's pretty common for Rust libraries to have a less-strict policy on MSRV bumps, treating MSRV bumps as a non-breaking change. https://github.com/rust-lang/api-guidelines/discussions/123 has a pretty good discussion of the current state of things. Another approach that would work would be to decide on (& publish) a less-strict MSRV-bump policy. It would certainly be nice to be able to do an MSRV bump without having to publish a new major version of the library.

tgeoghegan:
All right, I think we have consensus. @MxmUrw, I think you should go ahead and bump the Rust version to 1.61 [in Cargo.toml](https://github.com/divviup/libprio-rs/blob/7e7971dfa6f1b1824ea00385a3cd950ba9ad343a/Cargo.toml#L9) and in [ci-build.yml](https://github.com/divviup/libprio-rs/blob/7e7971dfa6f1b1824ea00385a3cd950ba9ad343a/.github/workflows/ci-build.yml#L33). Then we (ISRG) will do a release of `prio` 0.10.0.

divergentdave (commented):
Very nice work! I have a few smaller suggestions here and there.

src/flp/types/fixedpoint_l2.rs
MxmUrw: Ah, thanks. I did that now.

src/flp/types/fixedpoint_l2/compatible_float.rs
```diff
@@ -0,0 +1,77 @@
...
+fn to_float_bits(d: Field128, c: usize, n: i32) -> f64 {
+    // get integer representation of field element
+    let i: u128 = <Field128 as FieldElement>::Integer::from(d);
+    // interpret integer as float
+    let f = i as f64;
+    // to decode a single integer, we'd use the function
+    // dec(y) = (y - 2^(n-1)) * 2^(1-n) = y * 2^(1-n) - 1
+    // as f is the sum of c encoded vector entries where c is the number of
+    // clients, we compute f * 2^(1-n) - c
+    f * f64::powi(2.0, -n) - (c as f64)
```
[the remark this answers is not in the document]
MxmUrw: This sounds good. Implementing it was slightly more involved though, because your new subtraction underflows with `u128`s in case the decoded number is going to be negative. I manually handled both alternatives.

src/flp/types/fixedpoint_l2.rs
```diff
@@ -0,0 +1,814 @@
...
+//! bits.
+//!
+//! ### Differences in the computation because of distribution
+//!
+//! In `decode_result()`, what is decoded are not the submitted entries of a
+//! single client, but the sum of the the entries of all clients. We have to
+//! take this into account, and cannot directly use the `dec()` function from
+//! above. Instead we use:
+//! ```text
+//! dec'(x) = d * 2^(1-n) - c
```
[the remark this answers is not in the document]
MxmUrw: Yes, thanks.

src/flp/types/fixedpoint_l2/compatible_float.rs
```diff
@@ -0,0 +1,94 @@
...
+    //   = (s - c*2^(n-1)) * 2^(1-n)
+    // where the subtraction of `c` is done on integers and only afterwards
+    // the conversion to floats is done.
+    //
+    // Since the RHS of the substraction may be larger than the LHS
+    // (when the number we are decoding is going to be negative),
+    // yet we are dealing with unsigned 128-bit integers, we manually
+    // check for the resulting sign while ensuring that the subtraction
+    // does not underflow.
+
```
cjpatton, line 87:
```suggestion
```

cjpatton (approved):
👏

cjpatton:
Here's the build issue. I think you just need to add `#[cfg(feature = "crypto-dependencies")]` to the unused imports.
```
chris:~/github.com/divviup/libprio-rs$ cargo build --no-default-features
   Compiling libc v0.2.132
   Compiling serde v1.0.144
   Compiling getrandom v0.2.7
   Compiling prio v0.9.0 (/Users/chris/github.com/divviup/libprio-rs)
warning: unused import: `crate::flp::types::fixedpoint_l2::compatible_float::CompatibleFloat`
  --> src/vdaf/prio3.rs:41:5
   |
41 | use crate::flp::types::fixedpoint_l2::compatible_float::CompatibleFloat;
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(unused_imports)]` on by default

warning: unused import: `crate::flp::types::fixedpoint_l2::FixedPointBoundedL2VecSum`
  --> src/vdaf/prio3.rs:42:5
   |
42 | use crate::flp::types::fixedpoint_l2::FixedPointBoundedL2VecSum;
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

warning: unused import: `fixed::traits::Fixed`
  --> src/vdaf/prio3.rs:52:5
   |
52 | use fixed::traits::Fixed;
   |     ^^^^^^^^^^^^^^^^^^^^

warning: `prio` (lib) generated 3 warnings
    Finished dev [unoptimized + debuginfo] target(s) in 5.00s
```

MxmUrw:
> Here's the build issue. I think you just need to add #[cfg(feature = "crypto-dependencies")] to the unused imports.

Thanks, did this. Also a revert of a small revert which mistakenly happened in our last force push.

tgeoghegan (approved):

tgeoghegan:
@MxmUrw @ooovi I've approve the PR; thank you so much for your patience and diligence! However it looks like it now has some merge conflicts. If you can tidy those up, I'll merge swiftly without merging more stuff that breaks you in.

PR #286, opened 2022-08-25
Preparations for VDAF draft 03 breaking changes

divergentdave:
This makes the following housekeeping changes, in preparation for supporting VDAF draft 03.

- Bump the crate's version from 0.8.3 to 0.9.0.
- Update the document version used in Prio3 PRG derivation.
- Ignore test vector tests, as they would otherwise be broken by the above version number change.
- Update comments and documentation with references to the VDAF document.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -726,7 +726,7 @@ checksum = "ac74c624d6b2d21f425f752262f42188365d7b8ff1aff74c82e45136510a4857"
 
 [[package]]
 name = "prio"
-version = "0.8.3"
+version = "0.9.0"
 dependencies = [
  "aes 0.8.1",
  "aes-gcm",
--- a/src/flp.rs
+++ b/src/flp.rs
@@ -1,12 +1,13 @@
 // SPDX-License-Identifier: MPL-2.0
 
 //! Implementation of the generic Fully Linear Proof (FLP) system specified in
-//! [[draft-irtf-cfrg-vdaf-01]]. This is the main building block of [`Prio3`](crate::vdaf::prio3).
+//! [[draft-irtf-cfrg-vdaf-03]]. This is the main building block of [`Prio3`](crate::vdaf::prio3).
 //!
 //! The FLP is derived for any implementation of the [`Type`] trait. Such an implementation
 //! specifies a validity circuit that defines the set of valid measurements, as well as the finite
 //! field in which the validity circuit is evaluated. It also determines how raw measurements are
-//! encoded as inputs to the validity circuit.
+//! encoded as inputs to the validity circuit, and how aggregates are decoded from sums of
+//! measurements.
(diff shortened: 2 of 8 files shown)


Cargo.lock
```diff
@@ -726,7 +726,7 @@ checksum = "ac74c624d6b2d21f425f752262f42188365d7b8ff1aff74c82e45136510a4857"
 
 [[package]]
 name = "prio"
-version = "0.8.3"
+version = "0.9.0"
```
cjpatton, line 729: Don't we need to cut the new version prior to changing the version? Does this change need to be coordinated with a  GitHub action? cc/ @tgeoghegan for guidance here.

cjpatton (changes requested):

Cargo.lock
divergentdave: prio 0.8.3 has already been published. The `publish-crate` action runs upon publishing a release, and invokes `cargo publish`, so we need to update the manifest prior to creating the git tag. I figured I may as well update the crate version well in advance, to avoid potential confusion when working with unreleased code.

cjpatton (approved):

src/flp.rs
```diff
@@ -1,12 +1,13 @@
...
 
 //! Implementation of the generic Fully Linear Proof (FLP) system specified in
-//! [[draft-irtf-cfrg-vdaf-01]]. This is the main building block of [`Prio3`](crate::vdaf::prio3).
+//! [[draft-irtf-cfrg-vdaf-03]]. This is the main building block of [`Prio3`](crate::vdaf::prio3).
 //!
 //! The FLP is derived for any implementation of the [`Type`] trait. Such an implementation
 //! specifies a validity circuit that defines the set of valid measurements, as well as the finite
 //! field in which the validity circuit is evaluated. It also determines how raw measurements are
-//! encoded as inputs to the validity circuit.
+//! encoded as inputs to the validity circuit, and how aggregates are decoded from sums of
```
tgeoghegan, line 9: nit: "sum" isn't the right term for all VDAFs. At the risk of making this less clear, could we use something more generic? "combinations of measurements"? Or 
```suggestion
//! encoded as inputs to the validity circuit, and how aggregation results are decoded from aggregations of measurements
```

tgeoghegan (approved):

src/flp.rs
divergentdave: This module is restricted to just Prio3, so there will be a sum of some kind. Maybe this would be more precise?
```suggestion
//! encoded as inputs to the validity circuit, and how aggregation results are decoded from sums of encoded
```

tgeoghegan: Good point, we do literally sum the aggregate shares together in the Prio case. Let's leave the text as is.

PR #287, opened 2022-08-25
Add public share to VDAF API

divergentdave:
This updates the VDAF API to include a public share along with per-aggregator input shares. A new associated type, `PublicShare`, is added to the `Vdaf` trait, and the signatures of `Client::shard()` and `Aggregator::prepare_init()` are updated to plumb the public share through. The public share will be required to implement Poplar1 as specified in drafts 02 and 03, but it is not used yet in the codebase. (The Poplar paper's IDPF requires a public share, but ToyIdpf does not)

This depends on #286.

### DIFF
--- a/src/vdaf/poplar1.rs
+++ b/src/vdaf/poplar1.rs
@@ -342,6 +342,7 @@ where
     type Measurement = IdpfInput;
     type AggregateResult = BTreeMap<IdpfInput, u64>;
     type AggregationParam = BTreeSet<IdpfInput>;
+    type PublicShare = (); // TODO: Replace this when the IDPF from [BBCGGI21] is implemented.
     type InputShare = Poplar1InputShare<I, L>;
     type OutputShare = OutputShare<I::Field>;
     type AggregateShare = AggregateShare<I::Field>;
@@ -357,7 +358,7 @@ where
     P: Prg<L>,
 {
     #[allow(clippy::many_single_char_names)]
-    fn shard(&self, input: &IdpfInput) -> Result<Vec<Poplar1InputShare<I, L>>, VdafError> {
+    fn shard(&self, input: &IdpfInput) -> Result<((), Vec<Poplar1InputShare<I, L>>), VdafError> {
         let idpf_values: Vec<[I::Field; 2]> = Prng::new()?
             .take(input.level + 1)
             .map(|k| [I::Field::one(), k])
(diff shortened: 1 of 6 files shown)


divergentdave:
I'm going to add another `#[allow(clippy::type_complexity)]` to `shard()`, as it's only triggering a warning on 1.58, not 1.63.

src/vdaf/poplar1.rs
```diff
@@ -342,6 +342,7 @@ where
     type Measurement = IdpfInput;
     type AggregateResult = BTreeMap<IdpfInput, u64>;
     type AggregationParam = BTreeSet<IdpfInput>;
+    type PublicShare = (); // TODO: Replace this when the IDPF from BBCG+21 is implemented.
```
cjpatton, line 345: For consistency
```suggestion
    type PublicShare = (); // TODO: Replace this when the IDPF from [BBCGGI21] is implemented.
```

cjpatton (approved):
LGTM, though given that we are once again complicating the API with something that Prio3 doesn't use, I wonder if it's worth revisiting our strategy here. An alternative to modifying `Vdaf` et al would be to create a new set of traits that consume/produce the public_share. 

The updside of this would be reducing deadcode. I can see a couple of downsides:
1. We would no longer be able to write code that is generic for both Prio3 and Poplar1.
2. We would have to write different testing frameworks for different VDAF variants, leading to lots of redundant code and probably increasing maintenance cost.

Before merging, I think it's worth spending a few minutes pondering if these downsides can be mitigated. Calling on @tgeoghegan and @branlwyd for thoughts they might have here.

tgeoghegan (approved):

tgeoghegan:
I think it's important to maintain a strong correspondence between this crate and the VDAF specification, because if we find that something is ugly, difficult to implement in `prio` or difficult for clients of `prio` to use, then that suggests a problem with the VDAF spec (or let's say an opportunity to improve the VDAF spec).

My hope is that in the future, new VDAFs will be defined and implemented in crates outside of `libprio-rs`, and then DAP implementations that are generic over `trait Vdaf` should be able to use those new VDAFs with no code changes (except those needed to map a configuration value to a particular `Vdaf` instantiation).

If DAP implementations wind up having to write different code to contend with `trait Vdaf` and `trait VdafWithPublicParam`, then I contend that the VDAF specification is not achieving its goal of abstracting over VDAFs, and ought to separately specify a VDAF and a VDAFWithPublicParam (much like how it contains distinct definitions of DAFs and VDAFs).

src/vdaf/poplar1.rs
```diff
@@ -357,7 +358,7 @@ where
     P: Prg<L>,
 {
     #[allow(clippy::many_single_char_names)]
-    fn shard(&self, input: &IdpfInput) -> Result<Vec<Poplar1InputShare<I, L>>, VdafError> {
+    fn shard(&self, input: &IdpfInput) -> Result<((), Vec<Poplar1InputShare<I, L>>), VdafError> {
```
glandium, line 361: The API would have been clearer if `()` had been `PublicShare`.

tgeoghegan: Good point! https://github.com/divviup/libprio-rs/pull/371

PR #288, opened 2022-08-25
Make integer-to-GF(p) element conversion fallible

divergentdave:
draft-irtf-cfrg-vdaf-02 says that construction of a field element from an integer should fail if the integer is greater than or equal to the field's modulus. Previously, `try_from_bytes()`, `byte_slice_into_vec()`, and related methods checked for this already. This change applies the same check to the conversion from individual integers to field elements, changing it from a `From` implementation to a `TryFrom` implementation.

Outside of tests, this conversion chiefly shows up when the number of aggregators appears inside a circuit.

This depends on #286 and #287.

PR #289, opened 2022-08-25
Add algorithm IDs and update PRG seed derivation

divergentdave:
draft-irtf-cfrg-vdaf-03 introduces algorithm identifiers for each VDAF, to provide domain separation between uses of Prio3 with different circuits. Various PRG derivations are updated to incorporate this algorithm identifier, in addition to the document version. I arbitrarily assigned some identifiers from the "reserved for private use" range for VDAF instances that are for test purposes only or are not standardized.

This depends on #286, #287, and #288.

### DIFF
--- a/src/flp.rs
+++ b/src/flp.rs
@@ -110,6 +110,9 @@ pub enum FlpError {
 /// as a vector of field elements and how validity of the encoded measurement is determined.
 /// Validity is determined via an arithmetic circuit evaluated over the encoded measurement.
 pub trait Type: Sized + Eq + Clone + Debug {
+    /// The Prio3 VDAF identifier corresponding to this type.
+    const ID: u32;
+
     /// The type of raw measurement to be encoded.
     type Measurement: Clone + Debug;
 
@@ -798,6 +801,7 @@ mod tests {
     }
 
     impl<F: FieldElement> Type for TestType<F> {
+        const ID: u32 = 0xFFFF3002;
         type Measurement = F::Integer;
         type AggregateResult = F::Integer;
         type Field = F;
@@ -932,6 +936,7 @@ mod tests {
     }
 
     impl<F: FieldElement> Type for Issue254Type<F> {
+        const ID: u32 = 0xFFFF3003;
         type Measurement = F::Integer;
         type AggregateResult = F::Integer;
         type Field = F;
--- a/src/flp.rs
+++ b/src/flp.rs
@@ -801,7 +801,7 @@ mod tests {
     }
 
     impl<F: FieldElement> Type for TestType<F> {
-        const ID: u32 = 0xFFFF3002;
+        const ID: u32 = 0xFFFF0000;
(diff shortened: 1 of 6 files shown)


src/flp.rs
```diff
@@ -936,6 +940,7 @@ mod tests {
     }
 
     impl<F: FieldElement> Type for Issue254Type<F> {
+        const ID: u32 = 0xFFFF3003;
```
cjpatton, line 943: No need to make these test codepoints unique. In fact, I think it's better if they're the same so that someone reading this code in the future doesn't think this is important.

src/vdaf.rs
```diff
@@ -15,6 +15,10 @@ use std::convert::TryFrom;
 use std::fmt::Debug;
 use std::io::Cursor;
 
+/// Domain-separation tag used to bind the VDAF operations to the document version. This will be
+/// revised with each draft with breaking changes.
+pub(crate) const VERSION: &[u8] = b"vdaf-03";
```
cjpatton, line 20: Would it be useful for the compiler to know the length?
```suggestion
pub(crate) const VERSION: [u8; 7] = b"vdaf-03";
```

src/vdaf/poplar1.rs
```diff
@@ -339,6 +339,9 @@ where
     I: Idpf<2, 2>,
     P: Prg<L>,
 {
+    // TODO: This currently uses a codepoint reserved for testing purposes. Replace it with
+    // 0x00001000 once the implementation is updated to match draft-irtf-cfrg-vdaf-03.
```
cjpatton, line 343: 👍  good call

src/vdaf/prio3.rs
```diff
@@ -281,8 +278,9 @@ where
         measurement: &T::Measurement,
         rand_source: RandSource,
     ) -> Result<Vec<Prio3InputShare<T::Field, L>>, VdafError> {
```
cjpatton, line 280: This would add some much needed clarity I think (maybe make declare it above for the whole `vdaf` module)
```suggestion
    ) -> Result<Vec<Prio3InputShare<T::Field, L>>, VdafError> {
    const DST_LEN: usize = VERSION.len() + 4;
```

cjpatton (approved):
Looks good! nits only.

src/vdaf.rs
divergentdave: I don't think so, the codegen should be practically identical either way. We'd just move some ampersands around, and then the reference to an array would be dereferenced into a slice.

src/vdaf/prio3.rs
tgeoghegan: Yeah, or write this in terms of `Self::ID.to_be_bytes().len()` to absolve my rotten decrepit brain from figuring out what 4 and 5 refer to.

tgeoghegan (approved):

PR #290, opened 2022-08-25
Rewrite joint rand derivation, CMAC instead of XOR

divergentdave:
This implements a change from draft-irtf-cfrg-vdaf-03, updating the joint randomness derivation to combine contributions from each share in a nonlinear manner.

This depends on #286, #287, #288, and #289.

PR #291, opened 2022-08-25
Remove intermediate seed step from query randomness

divergentdave:
This implements a change from draft-irtf-cfrg-vdaf-03, simplifying generation of query randomness. One trip through the PRG is eliminated, and domain separation information is only included once.

This depends on #286, #287, #288, #289, and #290.


cjpatton (approved):
💥

PR #292, opened 2022-08-25
Add VDAF draft 03 test vectors, re-enable tests

divergentdave:
This checks in the relevant test vectors from draft-irtf-cfrg-vdaf-03, and re-enables the associated tests.

This depends on #286, #287, #288, #289, #290, and #291.


cjpatton (approved):
I confirmed that the test vectors match those generated by the reference implementation.

tgeoghegan (approved):

PR #295, opened 2022-08-29
Add Valgrind-based cycle count benchmarks

divergentdave:
This adds a new set of benchmarks that use Valgrind's cachegrind tool, by way of the `iai` crate. This provides very low-noise results, without requiring any changes to the OS, SpeedStep settings, or other environmental influences. (Every benchmark except the multithreaded variant of CountVec reports the exact same results when run repeatedly) The tradeoff is that it requires Valgrind at runtime, and the CPU that cachegrind models will differ from any given actual CPU. While there is an emulation slowdown from Valgrind, I found that only having to run each benchmark function once more than made up for it. (I believe Criterion runs benchmarks repeatedly until a set time has elapsed)

I ported most of the benchmarks from `speed_tests` to the new suite, but left out some lower-level benchmarks (FLP proof and query, plus polynomial multiplication).

I also added some clarifications to the `speed_tests` benchmark output, to disambiguate between input sizes, proof sizes, and share sizes.


tgeoghegan (approved):

benches/cycle_counts.rs
```diff
@@ -0,0 +1,295 @@
...
+use prio::{field::FieldPrio2, server::VerificationMessage};
+
+fn fft_iterative(size: usize) -> Vec<Field128> {
+    let inp = random_vector(size).unwrap();
+    let mut outp = vec![Field128::zero(); size];
+    benchmarked_iterative_fft(&mut outp, &inp);
+    outp
+}
+
+fn fft_iterative_16() -> Vec<Field128> {
```
cjpatton, line 23: Unless you think they're useful, we might consider skipping the FFT benchmarks.

cjpatton (approved):

PR #296, opened 2022-08-30
Add binary to generate PRG test vector

simon-friedberger:
I generated some test vectors to make sure my implementation of the PRG works as expected. Might be useful for others.


cjpatton (commented):
I think these should be added to the reference implementation instead: https://github.com/cfrg/draft-irtf-cfrg-vdaf/blob/main/poc/prg.sage#L103

simon-friedberger:
Indeed probably belongs in VDAF.

cjpatton:
Feel free to send a PR if you need addiitional test vectors! We do have some for PrgAes128 already.

PR #298, opened 2022-09-01
codec: Export `decode_items()`

cjpatton:
I'm reworking message serialization a bit ahead of upgrading Daphne to DAP-02, and I found this method to be useful. Let me know if you'd also like me to upstream my code.

### DIFF
--- a/src/codec.rs
+++ b/src/codec.rs
@@ -304,7 +304,7 @@ pub fn decode_u24_items<P, D: ParameterizedDecode<P>>(
 }
 
 /// Decode the next `length` bytes from `bytes` into as many instances of `D` as possible.
-fn decode_items<P, D: ParameterizedDecode<P>>(
+pub fn decode_items<P, D: ParameterizedDecode<P>>(
     length: usize,
     decoding_parameter: &P,
     bytes: &mut Cursor<&[u8]>,


tgeoghegan:
Exporting this doesn't seem like a horrible burden for the library, but I'm curious whether the [`decode_u8_items`, `decode_u16_items` or `decode_u24_items`](https://docs.rs/prio/latest/prio/codec/index.html) functions we already export might solve your problem, too?

cjpatton:
Well, for DAP-02 we actually need 32-bit prefixed strings. Instead of re-implementing encode_u32_items, I decided to generalize: https://github.com/cloudflare/daphne/commit/c66b12f77e972a2beaff44c370bcb8826db86dd6#diff-d0d5542e8b7c00e585cd52340ebc87a80dfbac2b995743ba40939365cf8e2bfeR722

tgeoghegan:
I see. Well, I wouldn't want to expose `decode_items` without a corresponding `encode_items`. I think we should either:
- add `[encode,decode]_u32_items` to `prio::codec` or
- upstream your generic `[encode,decode]_prefixed_items` to `prio::codec`

cjpatton:
Probably best to upstream.

cjpatton:
I need to mull it over a bit :) To be honest, I might end up forking here, as I've noticed some rough edges with these traits. I'll get back to you next week.

PR #300, opened 2022-09-07
Optimization: use fold in ParallelSumMultithreaded

divergentdave:
This rewrites the parallel processing core of `ParallelSumMultithreaded` to use `ParallelIterator::fold()`, so that buffers can be reused between subsequent operations, and allocation can be taken out of the inner loop.

I got the following benchmark results:

```
prio3 parallel countvec (1000 len)                                                                             
                        time:   [1.8427 ms 1.9360 ms 2.0303 ms]
                        change: [-13.301% -7.7941% -2.0182%] (p = 0.01 < 0.05)
                        Performance has improved.
```

### DIFF
--- a/src/flp/gadgets.rs
+++ b/src/flp/gadgets.rs
@@ -640,47 +640,49 @@ mod tests {
     fn test_parallel_sum_multithreaded() {
         use std::iter;
 
-        let poly: Vec<TestField> = random_vector(10).unwrap();
-        let num_calls = 10;
-        let chunks = 23;
+        for num_calls in [1, 10, 100] {
+            let poly: Vec<TestField> = random_vector(10).unwrap();
+            let chunks = 23;
 
-        let mut g =
-            ParallelSumMultithreaded::new(BlindPolyEval::new(poly.clone(), num_calls), chunks);
-        gadget_test(&mut g, num_calls);
+            let mut g =
+                ParallelSumMultithreaded::new(BlindPolyEval::new(poly.clone(), num_calls), chunks);
+            gadget_test(&mut g, num_calls);
 
-        // Test that the multithreaded version has the same output as the normal version.
-        let mut g_serial = ParallelSum::new(BlindPolyEval::new(poly, num_calls), chunks);
-        assert_eq!(g.arity(), g_serial.arity());
-        assert_eq!(g.degree(), g_serial.degree());
-        assert_eq!(g.calls(), g_serial.calls());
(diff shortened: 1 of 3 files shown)


src/flp/gadgets.rs
```diff
@@ -438,25 +464,43 @@ where
...
+                || ParallelSumFoldState::new(&self.serial_sum.inner, outp.len()),
+                |mut state, chunk| {
+                    state
+                        .inner
+                        .call_poly(&mut state.partial_output, chunk)
+                        .unwrap();
+                    for (sum_elem, output_elem) in state
+                        .partial_sum
+                        .iter_mut()
+                        .zip(state.partial_output.iter())
```
tgeoghegan, line 485: nit: `Iterator::zip` takes `IntoIterator`, which ought to be implemented on `Vec<F>`, so I think you could do:
```suggestion
                        .zip(&state.partial_output)
```

src/flp/gadgets.rs
```diff
@@ -438,25 +464,43 @@ where
...
+                },
+            )
+            .map(|state| state.partial_sum)
             .reduce(
                 || vec![F::zero(); outp.len()],
                 |mut x, y| {
-                    for i in 0..x.len() {
-                        x[i] += y[i];
+                    for (xi, yi) in x.iter_mut().zip(y.iter()) {
+                        *xi += *yi;
```
tgeoghegan, line 497: There's so many places where we use `foo.iter_mut().zip(&bar)` where `foo` and `bar` are `Vec<F: FieldElement>` that I'm wondering if we should put an implementation of `std::ops::Add` for `Vec<F: FieldElement>` in `field.rs` and use it everywhere. That isn't needed in this PR, however.

tgeoghegan (approved):

src/flp/gadgets.rs
divergentdave: I need to retain ownership of `state` for reuse in the next fold operation, so that rules out partial move of its members.

src/flp/gadgets.rs
divergentdave: Sounds good to me, I'll open an issue.

cjpatton (approved):
Nice one! One thing to double check is that we have adequate regression tests, which I believe we do: https://github.com/divviup/libprio-rs/blob/main/src/flp/gadgets.rs#L532

divergentdave:
I beefed up the testing some with different iteration counts and an end-to-end test at the Prio3 level.

PR #303, opened 2022-09-09
0.8.x: Revert breaking change to Collector::unshard()

divergentdave:
This is a pull request against the branch for the 0.8.x release series, to roll back a breaking change to the Collector trait. As of version 0.8.2, `Collector::unshard()` took only two parameters, the aggregation parameter and the aggregate shares. Versions 0.8.3 and 0.8.4 added an additional argument, `num_measurements`. This undoes that change, and removes the `Prio3Aes128Average` VDAF that depended on it.

### DIFF
--- a/src/flp.rs
+++ b/src/flp.rs
@@ -125,11 +125,7 @@ pub trait Type: Sized + Eq + Clone + Debug {
     ) -> Result<Vec<Self::Field>, FlpError>;
 
     /// Decode an aggregate result.
-    fn decode_result(
-        &self,
-        data: &[Self::Field],
-        num_measurements: usize,
-    ) -> Result<Self::AggregateResult, FlpError>;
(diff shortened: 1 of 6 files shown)


branlwyd (approved):
The change LGTM from a technical standpoint.

I may have missed some comms -- are there folks using Average today, and if so what's their story moving forward? (they can continue to use the existing published version of 0.8.x they are currently using, but what if they need to upgrade in the future to pick up some new functionality?)

divergentdave:
My understanding is that Prio3Aes128Average was written up to test out the unsharding change. That change was in turn motivated by the fixed-point vector with bounded L2 norm type in #283. (There, the number of measurements is required because the AFE encoding/decoding is affine, so the decoding part of unshard needs to multiply the zero offset by the number of measurements to subtract it out) As such, I think it's unlikely there are any users of Prio3Aes128Average yet.

tgeoghegan (approved):

PR #335, opened 2022-10-14
AggregateShare: switch to TryFrom<Vec<u8>>, Into<Vec<u8>>.

branlwyd:
Previously, these were written in terms of transformations to and from references to sequences of bytes. While From & Into can be used in that way, this doesn't fit with their normal usage, which is to consume some type and transform it into another type (in this case, consume the aggregate share and turn it into an owned vector of its serialization, or vice versa). More practically, specifying things in terms of references required specifying a "where" clause which effectively became infectious, needing to be copied to every place the Vdaf trait was referenced.

**This is a breaking change, do not merge until a breaking change window.**

### DIFF
--- a/src/vdaf.rs
+++ b/src/vdaf.rs
@@ -122,10 +122,7 @@ impl<F: FieldElement, const L: usize> Encode for Share<F, L> {
 // TODO(brandon): once GATs are stabilized [https://github.com/rust-lang/rust/issues/44265],
 // state the "&AggregateShare must implement Into<Vec<u8>>" constraint in terms of a where clause
 // on the associated type instead of a where clause on the trait.
-pub trait Vdaf: Clone + Debug
-where
-    for<'a> &'a Self::AggregateShare: Into<Vec<u8>>,
-{
+pub trait Vdaf: Clone + Debug {
     /// Algorithm identifier for this VDAF.
     const ID: u32;
 
@@ -149,7 +146,9 @@ where
     type OutputShare: Clone + Debug;
 
     /// An Aggregator's share of the aggregate result.
-    type AggregateShare: Aggregatable<OutputShare = Self::OutputShare> + for<'a> TryFrom<&'a [u8]>;
+    type AggregateShare: Aggregatable<OutputShare = Self::OutputShare>
+        + TryFrom<Vec<u8>>
+        + Into<Vec<u8>>;
 
     /// The number of Aggregators. The Client generates as many input shares as there are
     /// Aggregators.
@@ -157,10 +156,7 @@ where
 }
 
 /// The Client's role in the execution of a VDAF.
-pub trait Client: Vdaf
-where
-    for<'a> &'a Self::AggregateShare: Into<Vec<u8>>,
-{
+pub trait Client: Vdaf {
     /// Shards a measurement into a public share and a sequence of input shares, one for each
     /// Aggregator.
     fn shard(
@@ -170,10 +166,7 @@ where
 }
 
 /// The Aggregator's role in the execution of a VDAF.
-pub trait Aggregator<const L: usize>: Vdaf
-where
-    for<'a> &'a Self::AggregateShare: Into<Vec<u8>>,
-{
+pub trait Aggregator<const L: usize>: Vdaf {
     /// State of the Aggregator during the Prepare process.
     type PrepareState: Clone + Debug;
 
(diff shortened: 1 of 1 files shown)


branlwyd:
Note that phrasing the trait requirements this way incurs an additional copy (which is why I didn't write them this way in the first place) -- but one which I think the optimizer can easily optimize out for release builds, since it's a move followed by a readonly usage & immediate drop.

I don't believe any of the existing benchmarks cover serialization/deserialization speed; I can write one if folks are dubious.

src/vdaf.rs
```diff
@@ -279,23 +263,23 @@ impl<F> AsRef<[F]> for OutputShare<F> {
     }
 }
 
-impl<F> From<Vec<F>> for OutputShare<F> {
+impl<F: FieldElement> From<Vec<F>> for OutputShare<F> {
```
tgeoghegan, line 266: AFAICT nothing in this `impl` depends on `F` being `FieldElement` so there's no need to constrain `F` at all.

src/vdaf.rs
```diff
@@ -149,18 +143,17 @@ where
     type OutputShare: Clone + Debug;
 
     /// An Aggregator's share of the aggregate result.
-    type AggregateShare: Aggregatable<OutputShare = Self::OutputShare> + for<'a> TryFrom<&'a [u8]>;
+    type AggregateShare: Aggregatable<OutputShare = Self::OutputShare>
+        + TryFrom<Vec<u8>>
```
tgeoghegan, line 147: I believe your claim that the copy would be optimized away, but that aside, I think the ergonomics of `TryFrom<Vec<u8>>` are worse than `for<'a> TryFrom<&'a [u8]>`, regardless of performance. I also get the argument that `TryFrom` should consume the input value when doing the conversion, but in the stdlib alone there are several `From` implementations that seem analogous to this one, like [`From<&str> for Vec<u8>`](https://doc.rust-lang.org/std/convert/trait.From.html#impl-From%3C%26str%3E-for-Vec%3Cu8%2C%20Global%3E).

I agree with you that there should be no or little impact from the copy forced by taking a `Vec<u8>` by value, so I can live with either trait bound.

tgeoghegan (approved):

src/vdaf.rs
branlwyd: Ah, I should have commented: this is required as otherwise it conflicts with the `TryFrom<Vec<u8>>` implementation -- specifically, I think what's going on is that, letting `F = u8`, we have a `From<Vec<u8>>` implementation here, which implies a `TryFrom<Vec<u8>>` implementation, which conflicts with the manual implementation below. By restricting `F` to `FieldElement` we resolve the conflict, since `u8` does not implement `FieldElement`.

I think this is an OK change as, despite being a general container type, `OutputShare`/`AggregateShare` are will always store some `FieldElement` in practice.

src/vdaf.rs
branlwyd: Interested in folks' thoughts on the ergonomics -- IMO, dropping the infectious where clause is so nice that almost any other changes in ergonomics are acceptable. (Could you clarify what parts of the ergonomics you dislike here, possibly I'm missing something bad.)

That said, using `From`/`Into` is not the only way to phrase this change. For example, we could introduce a custom trait that provides equivalent functionality without the need to worry about how smart the compiler is. Something like:

```rust
trait Serializable {
  fn serialize(&self) -> Vec<u8>;
  fn deserialize(buf: &[u8]) -> Result<Self, SomeError>;
}
```

...except that we'd probably want to pick different names to avoid confusion with `serde`'s serialization traits. A custom trait also has the downside that stdlib & other libraries are not going to know how to work with it.

divergentdave: I would prefer separate traits for this serialization/deserialization. We probably shouldn't reuse `Encode`/`Decode`, as here we rely on greedy "read to end" behavior, unlike everything in the codec module.

branlwyd: (another approach would be to have `AggregateShare` implement `Into<Vec<u8>>` and `for<'a> TryFrom<&'a [u8]>`. this would drop the infectious where clause while still allowing parsing of arbitrary slices of bytes.)

tgeoghegan: The bad ergonomics of `TryFrom<Vec<u8>>` are that callers who have a `&[u8]` are forced to do something like `to_vec`, but if we required `for<'a> TryFrom<&'a [u8]>`, then it's easy for callers who have a `Vec<u8>` to pass a reference to it. So I think we're on the same page: `AggregateShare` requiring `Into<Vec<u8>>` and `for<'a> TryFrom<&'a [u8]>` gets us all the wins we want.

branlwyd: OK -- I rephrased things in terms of `for<'a> TryFrom<&'a [u8]>`. This change also let me drop the `F: FieldElement` requirement on the `From<Vec<F>>` implementations, too.

While I was at it, I added trait requirements on `OutputShare` equivalent to `AggregateShare` -- they were already implemented (and tested), and are logically required to run a VDAF evaluation (assuming the different parties are separated by an IPC boundary), so I think `OutputShare` is equally justified in requiring these traits as `AggregateShare` is.

cjpatton (approved):

PR #345, opened 2022-10-26
Switch to dtolnay/rust-toolchain GitHub Action

tgeoghegan:
See https://github.com/divviup/janus/pull/685,
https://github.com/divviup/janus/issues/684 for context.


.github/workflows/ci-build.yml
```diff
@@ -37,7 +37,7 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - name: Install Rust toolchain
-      uses: actions-rs/toolchain@v1
+      uses: dtolnay/rust-toolchain@v1
```
divergentdave, line 40: The second and third invocations should use the `stable` branch, as the `v1` tag is stale.

tgeoghegan: I also forgot to strip off some obsolete arguments that aren't used by `dtolnay/rust-toolchain@v1`.

divergentdave (approved):

PR #371, opened 2022-12-08
Use more explicit types than `()` in vdaf impls

tgeoghegan:
A number of function signatures in implementations of the `Vdaf` family of traits were using `()` as the type for their public shares or aggregation parameters. We now use `Self::PublicShare` or `Self::AggregationParam` as appropriate to make signatures and docs more clear.

### DIFF
--- a/src/vdaf/poplar1.rs
+++ b/src/vdaf/poplar1.rs
@@ -361,7 +361,10 @@ where
     P: Prg<L>,
 {
     #[allow(clippy::many_single_char_names)]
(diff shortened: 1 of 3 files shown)


tgeoghegan:
Thanks to @glandium for pointing this out! I'm pretty sure this isn't a breaking change since the actual types used here don't change, so we wouldn't need to go to `prio 0.11` for this.

divergentdave (approved):

cjpatton (approved):

PR #381, opened 2022-12-13
Implement IdpfPoplar

divergentdave:
This is a working prototype of IdpfPoplar, as specified in VDAF-03. Key generation and evaluation work, and the codec traits are implemented for the public share.

This needs the following issues to be addressed before it's production-ready:

- Cut down on dependencies as needed. (I was liberal about adding dependencies to get things working, with the intention of replacing some or all of them)
  - `fiat-crypto` is used for $GF(2^{255}-19)$ field math. While the formal methods work behind this provides a great deal of assurance, we only need a tiny slice of the library. This dependency could be troublesome for consumers that are conservative about crypto libraries.
  - A crate from Parity's Ethereum ecosystem is used to implement a 256-bit unsigned integer type. This is needed for an associated type on `Field255`. This could be easily replaced with a handwritten type using four `u64` limbs.
  - `bitvec` is used for storage of IDPF inputs, to simplify looking up bits in their semantic order, without fiddling with manual shifts. This could be replaced with much simpler types, since we don't need the ability to mutate bit vectors, etc. `bitvec` is also used in serialization/deserialization of packed control bit correction words, for convenience.
- Implement a stateful API, so that keys and control bits from previous evaluations can be reused in later evaluations. This will be necessary for efficient evaluation over practical size inputs, as this memoization will bring runtime from $O(n^2)$ to $O(n)$.
- Clean up `ToyIdpf` and the existing, older `Poplar1` implementation. I renamed the existing `Idpf` trait and some existing types to avoid having to change them too much. These will need to be overhauled to work with the complete Poplar1 specification in VDAF-03. I think it might make sense to wait and merge `IdpfPoplar` and the new `Poplar1` together once both are ready.
- We should have benchmarks for Poplar1.

Other notes: I split the `FieldElement` trait in two, both because $p=2^{255}-19$ is not FFT-friendly, thus certain methods don't make sense on it, and so that I could avoid implementing field operations that IdpfPoplar and Poplar1 do not use.


cjpatton:
Heeerree we go!!! 🛵

A few high-level, somewhat opinionated suggestions (worth discussing). They mostly have to do with not over-indexing on generality:
1. Implementing finite field arithmetic is tricky, and I think we're better off using implementations that already exist. In particular, I think that a dependency on `fiat-crypto` for Field255 is perfectly fine. In fact, ideally we would upstream implementations of Field64 and Field128 to `fiat-crypto` :) 
2. The IDPF should be moved into its own module, similar to how FLP has its own module. The reason is that this thing is likely to be useful independent of Poplar1.
3. I don't think it's worth defining a trait for IDPF, as we will in all likelihood end up with a single instantiation of it. Accordingly, I think we should drop ToyIdpf altogether.

divergentdave:
Rebased to pick up a Clippy fix, and pushed another commit to get rid of the `primitive-types` dependency.

divergentdave:
Some notes before I stop thinking about this for a while:

On my machine, the IDPF key generation benchmark is consistently running at a throughput of ~75KiB/s.

Right now, the evaluation benchmark has a throughput of ~100 KiB/s on 1-byte inputs, ~80 KiB/s on 16-byte inputs, and ~20 KiB/s on 256-byte inputs. Profiling the largest case shows that CPU time is dominated by hashing bit slices for cache lookups and insertions.

With a better designed cache, we could keep the throughput from dropping so much. Since IDPF evaluations will be almost certainly part of a tree walking algorithm, I think a cache built on a very small ring buffer, with lookups done via a linear scan, would make sense. So long as the ring buffer capacity is more than 1.5x the number of candidate prefixes being considered, LRU eviction from the ring buffer would work great.

We can make sure our eventual replacement for `bitvec` is more sympathetic to comparison. (I see that the `bitvec` hashing code uses a stack of iterators, and calls `Hasher::write_u8()` many times)

On the other hand, production deployments will either bring their own cache implementation, backed by durable/distributed storage, or choose to trade off recomputing seeds instead of storing them. Perhaps the performance of the cache implementations `prio` provides only matters to the extent they get out of the way of our benchmarks.

Separately, I have an idea for an optimization for the evaluation side. Currently, `eval_next()` computes the IDPF's values, i.e. the pair of field elements, for every level. `eval_inner()` discards these from all but the last `eval_next()` call. This is approximately one wasted field addition and one wasted field negation per level, if no caching is used, though no field computations are wasted if a perfect cache is used. With a little bit of rearranging, we could pull application of the field element correction words out of `eval_next()`, and only run this at the end of `eval_inner()`, for the level where we know we will use the result.

src/field255.rs
```diff
@@ -0,0 +1,305 @@
+//! Finite field arithmetic for `GF(2^255 - 19)`.
```
cjpatton, line 1: I was hoping there would be a higher-level API for arithmetic in this field that we could take advantage of, i.e., that we wouldn't use fiat-crypto directly. However after doing some research, this appears not to be the case.

The best rust implementation of GF(2^255-19) is dalek, but they don't expose the field arithemtic directly. I also noticed that the code here is essentially a re-implementation of [dalek's old fiat backend](https://github.com/dalek-cryptography/curve25519-dalek/blob/main/src/backend/serial/fiat_u64/field.rs). As I understand it, maintenance has moved to the zkcrypto project (https://github.com/zkcrypto/curve25519-dalek-ng). Interestingly, there the maintainers have removed the fiat backend entirely.

I'm not sure what the best path forward here would be. If the zkcrypto folks are willing to consider exposing the field arithmetic (https://github.com/zkcrypto/curve25519-dalek-ng/issues/20), then I think that would be our best bet. In the meantime, re-implementing the fiat backend is probably our best option.

This code is delicate and will require careful review. Let's move splitting the FFT functionality of `FieldElement` and implementing `Field255` into a separate PR, then rebase this PR on top. I will see if I can get some feedback from folks on my team who have experience with fiat-crypto.

cjpatton (commented):
I've started digging in. I think the first step is to carve out a PR for the Field255 implementation; see inline comments.

src/field.rs
```diff
@@ -199,6 +181,26 @@ pub trait FieldElement:
     }
 }
```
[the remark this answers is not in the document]
divergentdave: True, I had just moved those two because Poplar wouldn't need them (it just squares), so "FFT friendly field element" is a slight misnomer because of that. I could implement those two for a cleaner API boundary.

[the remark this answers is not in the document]
cjpatton: For now, I would not have `Field255` implement them. Rather, I would just stick a `todo!("implement when needed")` in there.

src/vdaf/idpf.rs
```diff
@@ -0,0 +1,1620 @@
...
+        prg.update(b" idpf poplar extend");
+        let mut seed_stream = prg.into_seed_stream();
+
+        let mut seeds = [Seed([0u8; L]), Seed([0u8; L])];
+        seed_stream.fill(&mut seeds[0].0);
+        seed_stream.fill(&mut seeds[1].0);
+
+        let mut byte = [0u8];
+        seed_stream.fill(&mut byte);
+        let control_bits = [(byte[0] & 1) != 0, ((byte[0] >> 1) & 1) != 0];
```
cjpatton, line 201: I'm slightly worried about the comparisons here (i.e., `x != 0`) being a side channel. It should be sufficient to represent the control bits as `u8`'s right?

src/vdaf/idpf.rs
```diff
@@ -0,0 +1,1620 @@
...
+        ];
+
+        let previous_control_bits = *control_bits;
+        control_bits[0] =
+            control_bits_0[keep] ^ (previous_control_bits[0] & control_bit_correction_words[keep]);
+        control_bits[1] =
+            control_bits_1[keep] ^ (previous_control_bits[1] & control_bit_correction_words[keep]);
+
+        let seeds_corrected = [
+            if previous_control_bits[0] {
```
cjpatton, line 259: We need to rewrite this code to be constant time, i.e., not branch on the value of the secret control bits. (I'm making a note to adjust the spec accordingly.) One idea is to convert the control bits to bitmasks via wrapping subtraction:

```rust
let control_bit_mask_0 = 0_u8.wrapping_sub(previous_control_bits[0] as u8);
let control_bit_mask_1 = 0_u8.wrapping_sub(previous_control_bits[1] as u8);
for i in 0..L {
    seed_0[keep].0[i] ^= seed_correction_word.0[i] & control_bit_mask_0;
    seed_1[keep].0[i] ^= seed_correction_word.0[i] & control_bit_mask_1;
}
```

src/vdaf/idpf.rs
```diff
@@ -0,0 +1,1620 @@
...
+
+        let mut byte = [0u8];
+        seed_stream.fill(&mut byte);
+        let control_bits = [(byte[0] & 1) != 0, ((byte[0] >> 1) & 1) != 0];
+
+        (seeds, control_bits)
+    }
+
+    /// An auxiliary function that acts as a pseudorandom generator, returning field elements.
+    fn convert<F>(&self, seed: &Seed<L>) -> (Seed<L>, [F; OUT_LEN])
```
cjpatton, line 207: Suggestion: It might be easier to oparate on `[u8; L]` directly rather than `Seed<L>`.

divergentdave:
> 1. Implementing finite field arithmetic is tricky, and I think we're better off using implementations that already exist. In particular, I think that a dependency on `fiat-crypto` for Field255 is perfectly fine. In fact, ideally we would upstream implementations of Field64 and Field128 to `fiat-crypto` :)

As an intermediate step, I think we could just run the Coq code ourselves to stamp out functions implementing field math for our fields. One of the binaries produces implementations using Montgomery representations, much like our existing code. The prime modulus can be specified, along with limb size, and the list of functions we want. The invocation would look like:

```
src/ExtractionOCaml/word_by_word_montgomery --lang Rust --inline field64 64 '18446744069414584321' mul square add sub opp from_montgomery to_montgomery ...
```

cjpatton:
> > 1. Implementing finite field arithmetic is tricky, and I think we're better off using implementations that already exist. In particular, I think that a dependency on `fiat-crypto` for Field255 is perfectly fine. In fact, ideally we would upstream implementations of Field64 and Field128 to `fiat-crypto` :)
> 
> As an intermediate step, I think we could just run the Coq code ourselves to stamp out functions implementing field math for our fields. One of the binaries produces implementations using Montgomery representations, much like our existing code. The prime modulus can be specified, along with limb size, and the list of functions we want. The invocation would look like:
> 
> ```
> src/ExtractionOCaml/word_by_word_montgomery --lang Rust --inline field64 64 '18446744069414584321' mul square add sub opp from_montgomery to_montgomery ...
> ```

Yup! Let's hold off for now, however.

PR #383, opened 2022-12-16
Require codec traits on OutputShare/AggregateShare

### LINKED ISSUE #382: Poplar1: Need a new trait for greedy deserialization of output shares/aggregate shares
Currently, the `Vdaf` trait imposes the trait bound `for<'a> TryFrom<&'a [u8]>` on both the `OutputShare` and `AggregateShare` associated types. The pseudocode in VDAF-03 for `out_shares_to_agg_share()` and other procedures demonstrate that these types are directly serialized as a vector of field elements. Deserialization of these types requires a priori knowledge of the current tree level from the aggregation parameter. (to choose between the two fields) We will need to write a new trait that can capture this kind of deserialization, and make a breaking change to replace the `for<'a> TryFrom<&'a [u8]>` trait bounds.

divergentdave:
This implements the idea [here](https://github.com/divviup/libprio-rs/issues/382#issuecomment-1349900873), adding `Encode` and `ParameterizedDecode` trait bounds to the `OutputShare` and `AggregateShare` associated types, replacing `Into<Vec<u8>>` and `TryFrom<&'a [u8]>`.

This change is necessitated by Poplar1 as specified in VDAF-03, because encoded output shares for one prefix at the leaf level or four prefixes at an inner level are indistinguishable.

Note that this is a breaking change that affects all consumers of `prio::vdaf::Vdaf`.

Closes #382.

### DIFF
--- a/src/field.rs
+++ b/src/field.rs
@@ -167,9 +167,7 @@ pub trait FieldElement:
     /// impossible.
     fn slice_into_byte_vec(values: &[Self]) -> Vec<u8> {
         let mut vec = Vec::with_capacity(values.len() * Self::ENCODED_SIZE);
-        for elem in values {
-            vec.append(&mut (*elem).into());
-        }
+        encode_fieldvec(values, &mut vec);
         vec
     }
 
--- a/src/field.rs
+++ b/src/field.rs
@@ -784,10 +784,12 @@ pub(crate) fn decode_fieldvec<F: FieldElement>(
     input: &mut Cursor<&[u8]>,
 ) -> Result<Vec<F>, CodecError> {
     let mut vec = Vec::with_capacity(count);
-    let mut buffer = vec![0; count * F::ENCODED_SIZE];
-    input.read_exact(&mut buffer)?;
-    for chunk in buffer.chunks(F::ENCODED_SIZE) {
-        vec.push(F::try_from(chunk).map_err(|e| CodecError::Other(Box::new(e)))?);
+    let mut buffer = [0u8; 64];
+    for _ in 0..count {
+        input.read_exact(&mut buffer[..F::ENCODED_SIZE])?;
+        vec.push(
+            F::try_from(&buffer[..F::ENCODED_SIZE]).map_err(|e| CodecError::Other(Box::new(e)))?,
+        );
     }
     Ok(vec)
 }
--- a/src/field.rs
+++ b/src/field.rs
@@ -727,7 +727,6 @@ make_field!(
 /// # Errors
 ///
 /// Fails if the two vectors do not have the same length.
-#[cfg(any(test, feature = "prio2"))]
 pub(crate) fn merge_vector<F: FieldElement>(
     accumulator: &mut [F],
     other_vector: &[F],
@@ -793,20 +792,6 @@ pub(crate) fn decode_fieldvec<F: FieldElement>(
         vec.push(F::try_from(chunk).map_err(|e| CodecError::Other(Box::new(e)))?);
     }
     Ok(vec)
-}
-
-/// `add_fieldvec` adds one vector of field elements to another elementwise, in-place, and returns an
-/// error if the vectors have different lengths.
-pub(crate) fn add_fieldvec<F: FieldElement>(left: &mut [F], right: &[F]) -> Result<(), FieldError> {
-    if left.len() != right.len() {
-        return Err(FieldError::InputSizeMismatch);
-    }
-
-    for (x, y) in left.iter_mut().zip(right.iter()) {
-        *x += *y;
-    }
-
-    Ok(())
 }
 
 #[cfg(test)]
--- a/src/field.rs
+++ b/src/field.rs
@@ -772,6 +772,43 @@ pub fn random_vector<F: FieldElement>(len: usize) -> Result<Vec<F>, PrngError> {
     Ok(Prng::new()?.take(len).collect())
 }
 
+/// `encode_fieldvec` serializes a type that is equivalent to a vector of field elements.
+#[inline(always)]
+pub(crate) fn encode_fieldvec<F: FieldElement, T: AsRef<[F]>>(val: T, bytes: &mut Vec<u8>) {
+    for elem in val.as_ref() {
+        bytes.append(&mut (*elem).into());
+    }
+}
+
+/// `decode_fieldvec` deserializes some number of field elements from a cursor, and advances the
+/// cursor's position.
+pub(crate) fn decode_fieldvec<F: FieldElement>(
+    count: usize,
+    input: &mut Cursor<&[u8]>,
+) -> Result<Vec<F>, CodecError> {
+    let mut vec = Vec::with_capacity(count);
+    let mut buffer = vec![0; count * F::ENCODED_SIZE];
+    input.read_exact(&mut buffer)?;
+    for chunk in buffer.chunks(F::ENCODED_SIZE) {
+        vec.push(F::try_from(chunk).map_err(|e| CodecError::Other(Box::new(e)))?);
+    }
+    Ok(vec)
+}
+
+/// `add_fieldvec` adds one vector of field elements to another elementwise, in-place, and returns an
+/// error if the vectors have different lengths.
+pub(crate) fn add_fieldvec<F: FieldElement>(left: &mut [F], right: &[F]) -> Result<(), FieldError> {
+    if left.len() != right.len() {
+        return Err(FieldError::InputSizeMismatch);
+    }
+
+    for (x, y) in left.iter_mut().zip(right.iter()) {
+        *x += *y;
+    }
+
+    Ok(())
+}
+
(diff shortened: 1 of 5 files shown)


src/vdaf/prio2.rs
```diff
@@ -299,6 +301,95 @@ impl<'a> ParameterizedDecode<(&'a Prio2, usize)> for Share<FieldPrio2, 32> {
...
+
+impl<'a, F> ParameterizedDecode<(&'a Prio2, &'a ())> for Prio2AggregateShare<F>
+where
+    F: FieldElement,
+{
+    fn decode_with_param(
+        (prio2, _): &(&'a Prio2, &'a ()),
+        bytes: &mut Cursor<&[u8]>,
+    ) -> Result<Self, CodecError> {
+        decode_fieldvec(prio2.input_len, bytes).map(Prio2AggregateShare)
```
tgeoghegan, line 374:
```suggestion
        decode_fieldvec(prio2.input_len, bytes).map(Self)
```

src/vdaf/prio2.rs
```diff
@@ -299,6 +301,95 @@ impl<'a> ParameterizedDecode<(&'a Prio2, usize)> for Share<FieldPrio2, 32> {
...
+
+impl<F> AsRef<[F]> for Prio2AggregateShare<F> {
+    fn as_ref(&self) -> &[F] {
+        &self.0
+    }
+}
+
+impl<F> From<Prio2OutputShare<F>> for Prio2AggregateShare<F> {
+    fn from(value: Prio2OutputShare<F>) -> Self {
+        Prio2AggregateShare(value.0)
```
tgeoghegan, line 353:
```suggestion
        Self(value.0)
```

src/vdaf/prio2.rs
```diff
@@ -299,6 +301,95 @@ impl<'a> ParameterizedDecode<(&'a Prio2, usize)> for Share<FieldPrio2, 32> {
...
+
+impl<'a, F> ParameterizedDecode<(&'a Prio2, &'a ())> for Prio2OutputShare<F>
+where
+    F: FieldElement,
+{
+    fn decode_with_param(
+        (prio2, _): &(&'a Prio2, &'a ()),
+        bytes: &mut Cursor<&[u8]>,
+    ) -> Result<Self, CodecError> {
+        decode_fieldvec(prio2.input_len, bytes).map(Prio2OutputShare)
```
tgeoghegan, line 337:
```suggestion
        decode_fieldvec(prio2.input_len, bytes).map(Self)
```

src/vdaf/prio2.rs
```diff
@@ -299,6 +301,95 @@ impl<'a> ParameterizedDecode<(&'a Prio2, usize)> for Share<FieldPrio2, 32> {
...
+
+impl<F> AsRef<[F]> for Prio2OutputShare<F> {
+    fn as_ref(&self) -> &[F] {
+        &self.0
+    }
+}
+
+impl<F> From<Vec<F>> for Prio2OutputShare<F> {
+    fn from(vec: Vec<F>) -> Self {
+        Prio2OutputShare(vec)
```
tgeoghegan, line 316:
```suggestion
        Self(vec)
```

src/vdaf/prio3.rs
```diff
@@ -1057,11 +1058,122 @@ fn check_num_aggregators(num_aggregators: u8) -> Result<(), VdafError> {
...
+where
+    F: FieldElement,
+    T: Type,
+    P: Prg<L>,
+{
+    fn decode_with_param(
+        (vdaf, _): &(&'a Prio3<T, P, L>, &'a ()),
+        bytes: &mut Cursor<&[u8]>,
+    ) -> Result<Self, CodecError> {
+        decode_fieldvec(vdaf.output_len(), bytes).map(Prio3OutputShare)
```
tgeoghegan, line 1097:
```suggestion
        decode_fieldvec(vdaf.output_len(), bytes).map(Self)
```

src/vdaf/prio3.rs
```diff
@@ -1057,11 +1058,122 @@ fn check_num_aggregators(num_aggregators: u8) -> Result<(), VdafError> {
...
+where
+    F: FieldElement,
+    T: Type,
+    P: Prg<L>,
+{
+    fn decode_with_param(
+        (vdaf, _): &(&'a Prio3<T, P, L>, &'a ()),
+        bytes: &mut Cursor<&[u8]>,
+    ) -> Result<Self, CodecError> {
+        decode_fieldvec(vdaf.output_len(), bytes).map(Prio3AggregateShare)
```
tgeoghegan, line 1168:
```suggestion
        decode_fieldvec(vdaf.output_len(), bytes).map(Self)
```

src/vdaf/prio2.rs
```diff
@@ -299,6 +301,95 @@ impl<'a> ParameterizedDecode<(&'a Prio2, usize)> for Share<FieldPrio2, 32> {
     }
 }
 
+/// A Prio2 output share comprised of a vector of field elements.
+#[derive(Debug, Clone)]
+pub struct Prio2OutputShare<F>(Vec<F>);
```
tgeoghegan, line 306: `Prio2OutputShare` and `Prio2AggregateShare` seem to be identical, respectively, to `Prio3OutputShare` and `Prio3AggregateShare`. I think the existing `vdaf::OuputShare` and `vdaf::AggregateShare` would suffice for both VDAFs (with the change where the decoding parameter is `(vdaf, aggregation_param)`).

tgeoghegan (changes requested):

tgeoghegan (approved):

src/field.rs
```diff
@@ -772,6 +769,29 @@ pub fn random_vector<F: FieldElement>(len: usize) -> Result<Vec<F>, PrngError> {
...
+}
+
+/// `decode_fieldvec` deserializes some number of field elements from a cursor, and advances the
+/// cursor's position.
+pub(crate) fn decode_fieldvec<F: FieldElement>(
+    count: usize,
+    input: &mut Cursor<&[u8]>,
+) -> Result<Vec<F>, CodecError> {
+    let mut vec = Vec::with_capacity(count);
+    let mut buffer = vec![0; count * F::ENCODED_SIZE];
```
cjpatton, line 787: Could we avoid an alloc by reading a fixed number of elements at a time?
```suggestion
    let mut buffer = [0; F::ENCODED_SIZE];
```

src/vdaf.rs
```diff
@@ -315,60 +304,24 @@ impl<F: FieldElement> Aggregatable for AggregateShare<F> {
     }
 
     fn accumulate(&mut self, output_share: &Self::OutputShare) -> Result<(), VdafError> {
-        // For prio3 and poplar1, no conversion is needed between output shares and aggregation
-        // shares.
+        // For poplar1, prio2, and prio3, no conversion is needed between output shares and
```
cjpatton, line 307: nit
```suggestion
        // For Poplar1, Prio2, and Prio3, no conversion is needed between output shares and
```

cjpatton (approved):
I'm happy with this direction, just a couple minor comments.

src/field.rs
divergentdave: It appears we can't yet, due to const limitations. `#![feature(generic_const_exprs)]` would partly unblock this.

```
error: constant expression depends on a generic parameter
   --> src/field.rs:787:26
    |
787 |     let mut buffer = [0; F::ENCODED_SIZE];
    |                          ^^^^^^^^^^^^^^^
    |
    = note: this may fail depending on what value the parameter takes
```

cjpatton: Ooh right, I forgot :(

What about a fixed-length buffer, perhaps `[u8; 64]`? We're definitely never going to have a field that large.

Also, I might be over optimizing, so feel free to ignore.

src/field.rs
```diff
@@ -772,6 +769,31 @@ pub fn random_vector<F: FieldElement>(len: usize) -> Result<Vec<F>, PrngError> {
...
+/// `decode_fieldvec` deserializes some number of field elements from a cursor, and advances the
+/// cursor's position.
+pub(crate) fn decode_fieldvec<F: FieldElement>(
+    count: usize,
+    input: &mut Cursor<&[u8]>,
+) -> Result<Vec<F>, CodecError> {
+    let mut vec = Vec::with_capacity(count);
+    let mut buffer = [0u8; 64];
+    for _ in 0..count {
+        input.read_exact(&mut buffer[..F::ENCODED_SIZE])?;
```
tgeoghegan, line 789: I buy that 64 is going to be big enough for any `FieldElement` implementation we encounter, but maybe we should put in an assertion right before the `for` loop like `assert!(buffer.len() >= F::ENCODED_SIZE, "field is too big for buffer");` just to make the problem obvious should anyone ever create an excessively large field.

divergentdave: Sounds good, this will be clearer than a panic on slicing.

Issue #396, opened 2022-12-20
Provide `cargo vet` audit.


branlwyd:
`cargo vet` is an auditing tool for Rust crates: https://mozilla.github.io/cargo-vet

Mozilla has requested that a "safe to run" and "safe to deploy" audits for `libprio-rs`. "Safe to run" implies freedom from intentionally-malicious code (e.g. memory snoopers, exploits, bitcoin miners, ...). "Safe to deploy" implies safety even in the presence of untrusted inputs.

Mozilla is not currently requesting a "semantic correctness" audit. We might still do this, perhaps once `libprio-rs` is more mature.

divergentdave:
Here are some notes on the subject from a read-through of the manual and some brief experimentation:

Because we support multiple release branches simultaneously, we'll need to re-import our own audits. We could have all branches re-import audits from main, or we could bounce it through a `cargo vet aggregate` workflow.

I tried out importing Firefox and Zcash audit repositories, and I found that the overlap only covered ~10% of the audit backlog, by lines.

We should add a configuration stanza to set `policy.prio-binaries.criteria = "safe-to-run"`, since this provides development-only tools. (Then, I think we need to set `policy.prio.criteria = "safe-to-deploy"` to keep auditing its dependencies to that standard. This could possibly be a bug, or I may just not understand how default criteria work.)

I'd like to run `cargo vet` with the option `--no-all-features`, because Firefox uses this configuration, and auditing `ring` up front is a daunting proposition. However, this has no effect, presumably because `prio-binaries` enables features on `prio`, and they get unified with those that `cargo vet` sets on `prio` directly. We may have to remove `prio-binaries` from the workspace to get this to work as intended, since there's no `--package` option.

By default, `cargo vet init` sets `policy.prio.audit-as-crates-io = false, meaning that it won't expect audits on `prio` itself, since it's first party code. While we could override this, and record audits after every `prio` release, I'm not sure whether we'd want to. There's more on the rationale around reviewing first party code [here](https://mozilla.github.io/cargo-vet/curating-your-audit-set.html#self-certification) and [here](https://mozilla.github.io/cargo-vet/first-party-code.html).

bholley:
> Here are some notes on the subject from a read-through of the manual and some brief experimentation:

Thanks for working on this.

> Because we support multiple release branches simultaneously, we'll need to re-import our own audits. We could have all branches re-import audits from main, or we could bounce it through a `cargo vet aggregate` workflow.

[...]

PR #398, opened 2022-12-23
Initial cargo vet setup

divergentdave:
This PR checks in an initial set of files for `cargo vet`, with some changes made to the configuration, and exemptions for all dependencies for now.

Future work: (for when we backport this, or make a new release branch) We should import our own audits from the main branch for use on other branches, so we only have to enter new audits in one place. We'll also need to update the branches on which the workflow runs.

This is part of #396. cc @bholley

### DIFF
--- a/.github/workflows/supply-chain.yml
+++ b/.github/workflows/supply-chain.yml
@@ -1,0 +1,32 @@
+name: supply-chain
+
+on:
+  push:
+    branches:
+     - main
+  pull_request:
+    branches:
+     - main
+  workflow_dispatch:
+
+jobs:
+  vet:
+    name: Vet Dependencies
+    runs-on: ubuntu-latest
+    env:
+      CARGO_VET_VERSION: 0.3.0
+    steps:
+    - uses: actions/checkout@v3
+    - name: Install Rust toolchain
+      uses: dtolnay/rust-toolchain@stable
+    - name: Cache cargo vet
+      uses: actions/cache@v3
+      with:
+        path: ${{ runner.tool_cache }}/cargo-vet
+        key: cargo-vet-bin-${{ env.CARGO_VET_VERSION }}
+    - name: Add the tool cache directory to the search path
+      run: echo "${{ runner.tool_cache }}/cargo-vet/bin/" >> $GITHUB_PATH
+    - name: Ensure that the tool cache is populated with the cargo-vet binary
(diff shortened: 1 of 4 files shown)


.github/workflows/supply-chain.yml
```diff
@@ -0,0 +1,32 @@
+name: supply-chain
+
+on:
+  push:
+    branches:
+     - main
+  pull_request:
+    branches:
+     - main
```
tgeoghegan, line 9: Why restrict this to `main`? Shouldn't we be vetting our active release branches, too?

.github/workflows/supply-chain.yml
```diff
@@ -0,0 +1,32 @@
...
+  vet:
+    name: Vet Dependencies
+    runs-on: ubuntu-latest
+    env:
+      CARGO_VET_VERSION: 0.3.0
+    steps:
+    - uses: actions/checkout@v3
+    - name: Install Rust toolchain
+      uses: dtolnay/rust-toolchain@stable
+    - name: Cache cargo vet
```
tgeoghegan, line 22: Is installing `cargo vet` so expensive that it's worth caching the binary? We don't do this for installing other Rust toolchain components, do we?

tgeoghegan (approved):

.github/workflows/supply-chain.yml
bholley: It's not expensive in an absolute sense, but compiling the tool still takes a couple of minutes on GitHub infra, and it's a much better user experience to be notified of cargo-vet failures within a few seconds of submitting the PR.

.github/workflows/supply-chain.yml
divergentdave: I intend to enable this on release branches once they have their supply-chain directories set up. (see the top comment) Running `cargo vet` with no configuration would not work, plus I want the configuration on release branches to import audits from the main branch via a raw GitHub URL, so that will have to be done in a follow-up.

.github/workflows/supply-chain.yml
divergentdave: Toolchain components from rustup only require downloading and installing binaries, but `cargo install cargo-vet` does a release build locally, so it's worth it. I largely copied the cache action configuration from the `cargo vet` guide. The only other `cargo install` we do in CI is over in prio-server, for coverage tooling. We could save two and a half minutes there with the same approach.

branlwyd (approved):

PR #405, opened 2023-01-04
Set up cargo vet - 0.8

divergentdave:
This sets up `cargo vet` on the 0.8 release branch. This PR has the following differences from #398 on main:

* CI workflow targets (though this might not take effect until I do the same on main)
* Audits from the main branch are imported under the name `divviup`
* I didn't bother annotating exemptions on this branch.

### DIFF
--- a/.github/workflows/supply-chain.yml
+++ b/.github/workflows/supply-chain.yml
@@ -1,0 +1,34 @@
+name: supply-chain
+
+on:
+  push:
+    branches:
+     - main
+     - release/**
+  pull_request:
+    branches:
+     - main
+     - release/**
+  workflow_dispatch:
+
+jobs:
+  vet:
+    name: Vet Dependencies
+    runs-on: ubuntu-latest
+    env:
+      CARGO_VET_VERSION: 0.3.0
+    steps:
+    - uses: actions/checkout@v3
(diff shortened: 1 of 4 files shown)


.github/workflows/supply-chain.yml
```diff
@@ -0,0 +1,34 @@
+name: supply-chain
+
+on:
+  push:
+    branches:
+     - main
+     - release/**
```
tgeoghegan, line 7: If the branches listed here disagree with [this same file on `main`](https://github.com/divviup/libprio-rs/blob/6449bbab899d2025e218cc954fdaa6eafed12cad/.github/workflows/supply-chain.yml#L6), then which one wins? I guess we can take this change, and if we see that the `supply-chain` jobs don't get run on PRs into `release/**` branches, then we can update the `branches:` specification in `supply-chain.yaml` on `main`.

tgeoghegan (approved):

.github/workflows/supply-chain.yml
divergentdave: The new supply-chain workflow did run on this PR, so it appears that the target branch wins for PR triggers. I plan to update main in a follow-up PR anyway for the sake of the next time we create a release branch.

PR #408, opened 2023-01-06
Implement Field255

divergentdave:
This splits some methods from the `FieldElement` trait out into a new trait, `FftFriendlyFieldElement`, and implements `Field255` from VDAF-03, which is a prerequisite for Poplar1.


src/field255/integer.rs
```diff
@@ -0,0 +1,839 @@
+use std::{
```
cjpatton, line 1: Before reviewing this code too closely, I think we should first try to answer a basic question: How much of this functionality are we actually going to need?

To answer this question, I played around with the IDPF code (#381) and found that we only use the arithmetic in tests. (I replaced the arithmetic functions with `todo!()` to see what functions panicked; none of the IDPF tests panicked.)

Looking ahead, I don't think we're actually going to need 256-bit arithmetic. Certainly not for Poplar.

In fact, I think the only thing we need this code for is to satisfy the (rather stringent) requirements of the `FieldElement` trait, which demands that each field element has a corresponding integer representation, and that type has to implement a number of arithmetic operations. (We get these for free for rust's primitive types.)

This being the case, I wonder if it would be worth trying to relax the API requirements, rather than provide an implementation of an API we're not likely to use. Specifically:

(1) Remove the `Integer` associated type from `FieldElement` 

(2) Define a new trait for `FieldElement`s that have an integer representation. Something like:

```rust
trait IntoInteger: FieldElement {
    type Integer: Add + Sub + Mul + ...;

    // we would need to move this function here, as well any other trait method that operates on the integer representation
    fn modulus() -> Self::Integer;
}
```

Let me know what you think  ... I might be missing something important here.

cjpatton (commented):
Alrighty let's dooo this. First, I think we can prune the amount of code substantially, since we're already fiddling with traits. See inline comments.

src/field255/integer.rs
divergentdave: I implemented the change to add FromInteger, but as I was reaching for num-bigint to reimplement `Display for Field255`, I realized we could just use `num_bigint::BigInt` as `Field255::Integer`. Thoughts? I think it should be sufficient for the ancillary conversion uses, plus it would simplify the field element tests if we could keep the `Field::Integer` associated type.

divergentdave: Looking ahead to the IDPF, we will need some way to convert a control bit to a `Field64`/`Field255` element, so that we can use multiplication instead of flow control to apply field element corrections. From this decision point, that could either mean we add such a conversion method to the `FieldElement` trait (or another new trait) or bouncing through two `Into` conversions (u8 -> Field64::Integer=u64 -> Field64, or alternately u8 -> Field255::Integer=BigInt -> Field255).

cjpatton: Which tests do we need `Field::Integer` for?

divergentdave: The main things are test cases that use the field's modulus to construct interesting elements, (like p-1, or checking what happens when you pass the saturated modulus) and the exponent argument to `pow()`. I had to move some checks from `field_element_test_common()` to `field_element_test()`. The change to `pow()` keeps me from using the field modulus (or p - 1) in a field element exponentiation.

cjpatton: > The main things are test cases that use the field's modulus to construct interesting elements, (like p-1, or checking what happens when you pass the saturated modulus) and the exponent argument to pow().

Ah yes, I think that's still good to test. I think going with big_int::BigInt for Field255::Integer is the right way to go, but I would only compile this code in tests (i.e., use tag `#[cfg(tests)]`). The goal being to fence code we don't intend to be used by other applications.

> Looking ahead to the IDPF, we will need some way to convert a control bit to a Field64/Field255 element, ...

Can you remind where this happens? I can't seem to find the relevant bit in the spec ...

divergentdave: In `gen()`, there's the following snippet:

```
        if ctrl[1] == Field2(1):
            w_cw = vec_neg(w_cw)
```

And in `eval()`:

```
    if next_ctrl == Field2(1):
        y = vec_add(y, w_cw)
```

These branches could be eliminated with `w_cw ← w_cw * (1 - 2 * Field(ctrl[1]))` and `y ← y + w_cw * Field(next_ctrl)`.

cjpatton: Nice. Currently the conversion path is `usize -(TryInto)-> FieldElement::Integer -(Into)-> FieldElement`. WDYTA this simplification:

1. Require `TryFrom<usize>` on `FieldElement`
2. Remove `TryFrom<usize>` from `IntoInteger`.

I don't think (2) is not strictly necessary, but if we're making breaking changes, and this is better, than it's worth considering. I'm not sure how much effort it would be to refactor.

src/field255.rs
```diff
@@ -0,0 +1,308 @@
...
+impl From<Field255> for Vec<u8> {
+    fn from(elem: Field255) -> Vec<u8> {
+        <[u8; Field255::ENCODED_SIZE]>::from(elem).to_vec()
+    }
+}
+
+impl Display for Field255 {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        let bytes: [u8; 32] = (*self).into();
+        let bigint = BigInt::from_bytes_be(Sign::Plus, bytes.as_slice());
```
cjpatton, line 211: Instead of converting to decimal, WDYTA simply printing the big-endian byte string, or even converting to hex? It would be nice to avoid the dependency on `BigInt` (except in tests of course, which you suggested would be needed.)

Cargo.toml
```diff
@@ -15,8 +15,10 @@ ctr = { version = "0.9.2", optional = true }
 cmac = { version = "0.7.1", optional = true }
 base64 = "0.20.0"
 byteorder = "1.4.3"
+fiat-crypto = "0.1.17"
 fixed = { version = "1.21", optional = true }
 getrandom = { version = "0.2.8", features = ["std"] }
+num-bigint = "0.4.3"
```
cjpatton, line 21: Is it feasible to move this to `dev-dependencies`? See comment on `Display` for `Field255`.

src/field255.rs
```diff
@@ -0,0 +1,308 @@
...
+        let self_copy = self.0;
+        fiat_25519_carry_mul(&mut self.0, &self_copy, &rhs.0);
+    }
+}
+
+impl Div for Field255 {
+    type Output = Field255;
+
+    fn div(self, _rhs: Self) -> Self::Output {
+        unimplemented!()
```
cjpatton, line 163: I would add a comment here noting that this requires `inv()`.

src/field255.rs
```diff
@@ -0,0 +1,308 @@
...
+    fn deserialize<D: Deserializer<'de>>(deserializer: D) -> Result<Field255, D::Error> {
+        deserializer.deserialize_bytes(FieldElementVisitor {
+            phantom: PhantomData,
+        })
+    }
+}
+
+impl Encode for Field255 {
+    fn encode(&self, bytes: &mut Vec<u8>) {
+        // Note that VDAF uses big-endian serialization, while fiat-crypto uses little-endian
```
cjpatton, line 240:
```suggestion
        // Note that VDAF requires big-endian serialization, while fiat-crypto uses little-endian
```

src/field255.rs
```diff
@@ -0,0 +1,308 @@
...
+    fn try_from_random(bytes: &[u8]) -> Result<Self, FieldError> {
+        Field255::try_from_bytes(bytes, true)
+    }
+
+    fn zero() -> Self {
+        Field255([0, 0, 0, 0, 0])
+    }
+
+    fn one() -> Self {
+        Field255([1, 0, 0, 0, 0])
```
cjpatton, line 279: Let's add a test that asserts this is indeed a valid representation of one. (The byte slice rep should equal `[0, ..., 0, 1]`.) It might also be worth asserting that `Field255::one()` subtracted from itself is `Field255::zero()`.

Similarly for `zero()` above.

src/field255.rs
```diff
@@ -0,0 +1,308 @@
...
+}
+
+impl FieldElement for Field255 {
+    const ENCODED_SIZE: usize = 32;
+
+    fn pow(&self, _exp: u128) -> Self {
+        unimplemented!()
+    }
+
+    fn inv(&self) -> Self {
```
cjpatton, line 266: I think it's worth commenting here why we haven't implemented this yet. (The simple answer is we don't need it quite yet.)

cjpatton (commented):
Looking very good! Well scoped, and no major issues. I noticed some potential timing side channels that shouldn't be too bad to address, see inline comments.

High-level question:
* Should we fence `Field255` behind the experimental feature?

src/field.rs
```diff
@@ -331,6 +321,21 @@ impl<'de, F: FieldElement> Visitor<'de> for FieldElementVisitor<F> {
     }
 }
 
+/// Objects with this trait represent an element of `GF(p)`, where `p` is some prime and the
+/// field's multiplicative group has a subgroup with an order that is a power of 2, and at least
+/// 2<sup>20</sup>.
```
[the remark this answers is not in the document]
divergentdave: LaTeX in rustdoc is not supported, and it appears the current status is "waiting for someone to write an RFC".

src/field255.rs
```diff
@@ -0,0 +1,308 @@
...
+        let mut rhs_encoded = [0; 32];
+        fiat_25519_to_bytes(&mut rhs_encoded, &rhs.0);
+
+        self_encoded == rhs_encoded
+    }
+}
+
+impl Eq for Field255 {}
+
+impl Hash for Field255 {
```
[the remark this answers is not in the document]
divergentdave: It looks like we don't use this internally. (but it would allow using `HashSet<Field255>` or `HashMap<Field255, _>`) I had written this because the other field types do the same, but the trait does not actually require a `Hash` bound.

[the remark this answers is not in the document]
cjpatton: Maybe Janus needs it somewhere...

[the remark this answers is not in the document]
divergentdave: I tried patching in `prio` with the hash implementations in `field.rs` deleted, and both `prio-server` and Janus work fine without that. Dropping `impl Hash for Field255` should be OK.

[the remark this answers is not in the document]
cjpatton: In the interest of keeping PRs minimal, I suggest dropping it.

src/field255.rs
divergentdave: The subtraction check is covered by `field_element_test_common()`.

src/field/field255.rs
```diff
@@ -0,0 +1,483 @@
...
+use integer::Field255Integer;
+
+// `python3 -c "print(', '.join(hex(x) for x in (2**255-19).to_bytes(32, 'big')))"`
+const MODULUS_BIG_ENDIAN: [u8; 32] = [
+    0x7f, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff,
+    0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xed,
+];
+
+/// `GF(2^255 - 19)`, a 255-bit field.
+#[derive(Clone, Copy, Default)]
```
tgeoghegan, line 37: Is `Default` appropriate to derive? `fiat_25519_tight_field_element` is `[u64; 5]`, so I think the default value will be `[0u64; 5]`. Is that a valid `fiat_25519_tight_field_element` value?

src/field/field255.rs
```diff
@@ -0,0 +1,483 @@
...
+    fn neg(self) -> Field255 {
+        let mut loose_output = [0; 5];
+        fiat_25519_opp(&mut loose_output, &self.0);
+        let mut output = [0; 5];
+        fiat_25519_carry(&mut output, &loose_output);
+        Field255(output)
+    }
+}
+
+impl TryFrom<usize> for Field255 {
```
tgeoghegan, line 194: `TryFrom<usize>` feels sketchy given that `Field255` values are much, much bigger than any `usize` type. Is this some kind of scheme to "compress" sufficiently small values into 8 bytes? Why do we need this?

src/field/field255.rs
```diff
@@ -0,0 +1,483 @@
...
+    type Error = FieldError;
+
+    fn try_from(value: usize) -> Result<Self, FieldError> {
+        let usize_bytes = value.to_be_bytes();
+        let mut field_bytes = [0u8; 32];
+        let field_bytes_len = field_bytes.len();
+        field_bytes[field_bytes_len - usize_bytes.len()..].copy_from_slice(&usize_bytes);
+        Field255::try_from_bytes(&field_bytes, false)
```
tgeoghegan, line 202:
```suggestion
        Self::try_from_bytes(&field_bytes, false)
```

src/field/field255.rs
```diff
@@ -0,0 +1,483 @@
...
+        match (
+            bool::from(less_than_modulus),
+            bool::from(greater_than_modulus),
+        ) {
+            (true, false) => {}
+            (false, _) => return Err(FieldError::ModulusOverflow),
+            (true, true) => unreachable!(),
+        }
+
+        // Switch from big endian to little endian.
```
tgeoghegan, line 80: It would be helpful here to explain that VDAF and `fiat_crypto` expect opposite endianness, as you did in the `Encode` impl.

src/field/field255.rs
```diff
@@ -0,0 +1,483 @@
...
+impl<'de> Deserialize<'de> for Field255 {
+    fn deserialize<D: Deserializer<'de>>(deserializer: D) -> Result<Field255, D::Error> {
+        deserializer.deserialize_bytes(FieldElementVisitor {
+            phantom: PhantomData,
+        })
+    }
+}
+
+impl Encode for Field255 {
+    fn encode(&self, bytes: &mut Vec<u8>) {
```
tgeoghegan, line 294: This looks nearly identical to the `From<Field255> for [u8; Field255::ENCODED_SIZE]` impl above.

src/field/field255/integer.rs
```diff
@@ -0,0 +1,190 @@
+// Copyright (c) 2023 ISRG
```
tgeoghegan, line 1: I don't understand the purpose of this module. Is it so we can compare the performance of a bigint based `Field255` against the `fiat_crypto` one?

src/field/field255.rs
divergentdave: Yes, `Field255::zero()` produces the same thing. Though maybe I'll manually implement `Default` and dispatch to that, so it's clearer.

src/field/field255.rs
divergentdave: `TryFrom<usize>` is a (newly added) trait bound on `FieldElement`. We need some sort of bound like this so that we can construct field elements from integers when we don't have access to the `FieldElementWithInteger` trait. Poplar1 will use this by casting control bits to `usize`, then converting them to field elements, and using that resulting element to compute IDPF output values without branching on the control bit.

src/field/field255/integer.rs
divergentdave: No, this primarily exists to provide a big-enough integer type that we can run `field_element_test_common()` on `Field255`. The `Field255Integer` type and its impls here are needed to satisfy the `Integer` associated type on the relevant `TestFieldElementWithInteger` trait.

src/field/field255.rs
tgeoghegan: What concerns me here is that carrying this `TryFrom` impl on `Field255` implies that one can generally convert a `usize` to a `Field255`, which I don't think is true and I worry it could lead to bugs if misused by clients. It sounds like we have a specific use case of wanting to transform "control bits" into a `Field255`, rather than the general requirement of transforming `usize` into `Field255`. So assuming that some kind of `struct Poplar1ControlBits` appears, could we instead do `TryFrom<Poplar1ControlBits> for Field255`?

src/field/field255.rs
```diff
@@ -0,0 +1,487 @@
...
+        }
+
+        match (
+            bool::from(less_than_modulus),
+            bool::from(greater_than_modulus),
+        ) {
+            (true, false) => {}
+            (false, _) => return Err(FieldError::ModulusOverflow),
+            (true, true) => unreachable!(),
+        }
```
cjpatton, line 78: Readability
```suggestion
        if bool::from(greater_than_modulus | !less_than_modulus) {
            return Err(FieldError::ModulusOverflow);
        }
```

src/field/field255.rs
```diff
@@ -0,0 +1,487 @@
...
+        value.copy_from_slice(&bytes[..Self::ENCODED_SIZE]);
+
+        if mask_top_bit {
+            value[0] &= 0b0111_1111;
+        }
+
+        // Walk through the big-endian encoding of the provided value, and identify whether the
+        // first byte that differs from the field's modulus is less than the corresponding byte or
+        // greater than the corresponding byte, in constant time. (or whether the provided value is
+        // equal to the field modulus)
```
cjpatton, line 63:
```suggestion
        // greater than the corresponding byte, in constant time. (Or whether the provided value is
        // equal to the field modulus.)
```

src/field/field255.rs
```diff
@@ -0,0 +1,487 @@
...
+            return Err(FieldError::ShortRead);
+        }
+
+        let mut value = [0u8; Self::ENCODED_SIZE];
+        value.copy_from_slice(&bytes[..Self::ENCODED_SIZE]);
+
+        if mask_top_bit {
+            value[0] &= 0b0111_1111;
+        }
+
```
cjpatton, line 59: Nice work :) We should file an issue to do the same thing for the other fields. (We are less careful with checking for modulus overflow there.)

src/field/field255.rs
cjpatton: I would go with `u64` instead of `usize`, for consistency. (See comment above.)

src/field/field255.rs
```diff
@@ -0,0 +1,487 @@
...
+impl From<Field255> for Vec<u8> {
+    fn from(elem: Field255) -> Vec<u8> {
+        <[u8; Field255::ENCODED_SIZE]>::from(elem).to_vec()
+    }
+}
+
+impl Display for Field255 {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        let encoded: [u8; Self::ENCODED_SIZE] = (*self).into();
+        write!(f,
```
cjpatton, line 235: How about something like:
```rust
write!(f, "0x");
for byte in encoded {
     write!(f, "{:02x}", byte);
}
```

src/field/field255.rs
divergentdave: Any `usize` can be successfully converted to `Field255` if `usize` is 32 or 64 bits. For CHERI-like platforms where `usize` is 128 or 129 bits, the same holds. You'd need to have `usize` be 256 bits or larger to run into issues. (At 256 bits, `try_from_bytes` may return `FieldError::ModulusOverflow`, and beyond that, a slicing operation would panic)

Implementing `TryFrom<usize> for Field255` was motivated by [this comment thread](https://github.com/divviup/libprio-rs/pull/408#discussion_r1065237379). I'd like to avoid implementing conversions from more than one primitive integer type, as this would impede type inference of integer literals. Implementing `TryFrom<subtle::Choice> for Field255`/`TryFrom<sublte::Choice> for Field64` would be sufficient for the IDPF use case, but `TryFrom<usize> for Field...` is more generally applicable, and can simplify some Prio3 code that currently bounces throgh two conversions.

src/field/field255/integer.rs
```diff
@@ -0,0 +1,190 @@
...
+impl From<Field255> for Field255Integer {
+    fn from(value: Field255) -> Self {
+        let mut le_bytes = [0u8; 32];
+        fiat_25519_to_bytes(&mut le_bytes, &value.0);
+        Field255Integer::from_little_endian(&le_bytes)
+    }
+}
+
+impl TestFieldElementWithInteger for Field255 {
+    type Integer = Field255Integer;
```
cjpatton, line 146: It would be preferable to use the `BigUint` type directly so that we don't have to define a wrapper type. Playing with this a little bit, it looks like the only thing thing that `BigUint` doesn't have that we need is bit shifting by `BigUint`, i.e., `BigUint << BigUint`. Howver, `BigUint` does implement bit shifting for various primitive types.

My expectation is that we could just require bit shift by a primitive type, e.g., `u64`. I can implement this in a follow-up PR, if you prefer to not have to change things.

```suggestion
    type Integer = BigUint;
```

cjpatton (commented):
Looking good, just nit-picky things remaining.

src/field/field255.rs
cjpatton: Disregard, I'll send a follow-up PR

src/field.rs
```diff
@@ -79,7 +85,7 @@ pub trait FieldElement:
     + DivAssign
     + Neg<Output = Self>
     + Display
-    + From<<Self as FieldElement>::Integer>
+    + TryFrom<usize, Error = FieldError>
```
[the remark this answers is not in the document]
cjpatton: @tgeoghegan just wanted to bump this since you're wanting to be consistent about what primitive types we allow conversion to/from. I think a reasonable choice is `u64`.

src/field/field255.rs
tgeoghegan: All right, I'm willing to defer to your consensus since I think you both have a better understanding of how this will fit into a Poplar1 implementation.

src/field.rs
```diff
@@ -546,6 +565,18 @@ macro_rules! make_field {
             }
         }
 
+        impl TryFrom<usize> for $elem {
+            type Error = FieldError;
+
+            fn try_from(value: usize) -> Result<Self, FieldError> {
+                let int: $int = value.try_into().map_err(|_| FieldError::ModulusOverflow)?;
```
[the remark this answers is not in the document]
divergentdave: This conversion is to `$int` though, which is `u32` in the case of `Field32`/`FieldPrio2`. Thus, this error case would be reachable with those fields on a 64-bit machine.

src/field.rs
```diff
@@ -798,13 +833,226 @@ pub(crate) fn decode_fieldvec<F: FieldElement>(
...
+        hash::{Hash, Hasher},
+        io::Cursor,
+        ops::{Add, BitAnd, Div, Shl, Shr, Sub},
+    };
+
+    /// A test-only copy of `FieldElementWithInteger`.
+    ///
+    /// This trait is only used in tests, and it is implemented on some fields that do not have
+    /// `FieldElementWithInteger` implementations. This separate trait is used in order to avoid
+    /// affecting trait resolution with conditional compilation. Additionally, this trait only
```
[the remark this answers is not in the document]
divergentdave: My initial reasoning was to avoid interactions between conditional compilation and trait resolution. (i.e., code that compiles successfully in some settings, but is too ambiguous in others, for non-intuitive reasons) The second, and more important, reason is that the existing code includes a `Copy` bound on the `Integer` associated type. We can meet this with `u32`, `u64`, and `u128`, but not `num_bigint::BigUint`. This test-only trait only requires that the integer provide `Clone`.

src/field/field255/integer.rs
divergentdave: Yes, this sounds good, it looks like all the primitives support shifting by any other primitive. From a quick glance, it looks like most `F::Integers` constructed for use in shifting come from `usize` be they in the bitvector representation helper methods or `FieldParameters.num_roots`.

src/field.rs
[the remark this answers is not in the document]
divergentdave: `TryFrom<u64>` would require writing the implementations outside of the declarative macro, since `impl From<$int> for $elem` is a `u64` conversion already in some cases. However, we could use that `From<u64> for Field64` and write `From<u64> for Field255` to get what we need, without monkeying with the `FieldElement` trait any more.

src/field.rs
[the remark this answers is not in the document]
cjpatton: Ohh right, gotcha.

src/field.rs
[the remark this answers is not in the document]
cjpatton: Actually, is `TryFrom<usize>` even necessary? Without this bound, `cargo test --all-features` passes

```suggestion
```

src/field/field255/integer.rs
cjpatton: Suggestion: The amount of code in this module is now small enough that I think it would be reaonable to fold into the `tests` module in field255.rs

cjpatton (approved):
🥇

Before merging, I recommend squashing the branch.

src/field.rs
```diff
@@ -79,7 +85,7 @@ pub trait FieldElement:
     + DivAssign
```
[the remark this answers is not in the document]
cjpatton: Bump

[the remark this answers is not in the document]
divergentdave: I updated the trait name there (and dropped `IntegerTryInto` entirely, since it wasn't used)

tgeoghegan (approved):

divergentdave:
I've got one more comment typo fix to force-push

PR #409, opened 2023-01-09
Fix bug in fixedpoint type occuring for certain vector lengths.

MxmUrw:
For some vector lengths the validity circuit would not accept a valid encoded value.

This happened in the case where the length of the input chunk to the parallel sum gadget was not of the required length, and so had to be extended. What was wrong is that the extension happened with the field element zero, whereas it should have happened with (a share) of the encoded zero value, which is `2^(n-1)`. This is now fixed.

Also added a test which tests the `fixedpoint_l2` type on long vectors, as this is how this bug was discovered.

### DIFF
--- a/src/flp/types/fixedpoint_l2.rs
+++ b/src/flp/types/fixedpoint_l2.rs
@@ -481,6 +481,12 @@ where
         let computed_norm = {
             let mut outp = F::zero();
 
+            // Chunks which are too short need to be extended with a share of the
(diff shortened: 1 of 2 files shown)


tgeoghegan (approved):

tgeoghegan:
Thanks for the well-documented fix!

MxmUrw:
Thanks for the quick review and merge!

PR #414, opened 2023-01-13
Implement IdpfPoplar

divergentdave:
This implements IdpfPoplar, as specified in VDAF-03. Key generation, key evaluation, and all necessary codec traits are implemented. The test suite includes tests against a test vector generated by the Sage implementation. The existing `ToyIdpf` is removed, and parts of the existing incompatible `Poplar1` implementation is stubbed out for now.

See previous discussion in #381. (which GitHub auto-closed upon merging Field255)


src/idpf.rs
```diff
@@ -0,0 +1,1869 @@
...
+
+    let mut field_element_correction_words = [F::zero(); OUT_LEN];
+    for (((out, element_0), element_1), value) in field_element_correction_words
+        .iter_mut()
+        .zip(elements_0.iter())
+        .zip(elements_1.iter())
+        .zip(value.iter())
+    {
+        let bit_converted = F::from(control_bits[1].unwrap_u8() as u64);
+        let sign = F::one() - bit_converted - bit_converted;
```
cjpatton, line 251: Move these lines out of the loop so that we only compute this once.

cjpatton (commented):
Not a complete pass yet.

src/idpf.rs
```diff
@@ -0,0 +1,1869 @@
...
+    }
+
+    /// Get an iterator over the bits that make up this input.
+    pub fn iter(&self) -> impl Iterator<Item = bool> + '_ {
+        self.index.iter().by_vals()
+    }
+}
+
+impl From<&'_ BitSlice<u8, Lsb0>> for IdpfInput {
+    fn from(bit_slice: &'_ BitSlice<u8, Lsb0>) -> Self {
```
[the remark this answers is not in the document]
divergentdave: BitSlice is an unsized type, only usable as a reference. I could add a `to_owned()` at conversations to make the memory allocation more explicit, or change the argument type to BitVec or BitBox.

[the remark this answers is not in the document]
cjpatton: I would change the argument type to BitVec. That way we don't miss an opportunity to avoid a clone, if the opportunity arises!

src/idpf.rs
```diff
@@ -0,0 +1,1869 @@
...
+use subtle::{Choice, ConstantTimeEq};
+
+/// IDPF-related errors.
+#[derive(Debug, thiserror::Error)]
+pub enum IdpfError {
+    /// Error from incompatible shares at different levels.
+    #[error("tried to merge shares from incompatible levels")]
+    MismatchedLevel,
+
+    /// Invalid parameter.
```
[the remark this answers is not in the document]
divergentdave: Here are the different cases:

* `gen()`: The number of bits is zero (I realized that the bits argument is redundant with the input's length, so I'll get rid of this)
* `gen()`: The number of beta values and the input bit length don't match
* `gen()`: The input is empty
* `eval()`: The number of bits is zero
* `eval()`: The prefix is empty
* `eval()`: The length of the prefix is greater than the number of bits
* `eval()`: The number of correction words in the public share doesn't match the number of bits

src/idpf.rs
```diff
@@ -0,0 +1,1869 @@
...
+impl IdpfInput {
+    /// Convert a slice of bytes into an IDPF input, where the bits of each byte are processed in
+    /// LSB-to-MSB order. (and subsequent bytes are processed in their natural order)
+    pub fn from_bytes(bytes: &[u8]) -> IdpfInput {
+        let bit_box_u8_storage = BitBox::<u8, Lsb0>::from_boxed_slice(Box::from(bytes));
+        bit_box_u8_storage.as_bitslice().into()
+    }
+
+    /// Convert a slice of booleans into an IDPF input.
+    pub fn from_bools(bools: &[bool]) -> IdpfInput {
```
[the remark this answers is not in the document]
divergentdave: I mainly wanted to provide this as an entry point using `std` types only, so that there was a way to construct inputs without dealing with `bitvec` types. I added `clone_with_suffix()` as a helper for the candidate prefixes case.

src/idpf.rs
```diff
@@ -0,0 +1,1852 @@
...
+    mut seed: [u8; L],
+    mut control_bit: Choice,
+    prefix: &IdpfInput,
+    cache: &mut dyn IdpfCache<L>,
+) -> Result<IdpfOutputShare<OUT_LEN, Field64, Field255>, IdpfError>
+where
+    P: Prg<L>,
+{
+    let mut last_inner_output = None;
+    for ((correction_word, input_bit), level) in public_share.inner_correction_words[start_level..]
```
cjpatton, line 411: Hmm, something looks fishy here. Suppose we use `NoCache`. Do we end up evaluating a given prefix more than once?

```
eval():
   for start_level in 0..bits:
      eval_inner(start_level)

eval_inner(start_level):
   for level in start_level..bits:
       eval_next()
```

cjpatton (commented):
This looks pretty good. Two high-level things:
1. I'm a bit confused about the structure of `eval()`. It looks to me like we might be wasting some computation. It might help to make it look a bit more like the reference implementation. (See inline comments.)
2. This PR rips out some code form poplar1.rs that we're going to need later. I suggest leaving poplar1.rs alone in this change.

src/idpf.rs
divergentdave: eval() only calls eval_inner() once, and then breaks out of its loop, so we're only doing evaluation at each node once.

divergentdave:
For what it's worth, there is one area that I'm aware wastes computation, that I'm leaving as-is for the purposes of this PR. When we evaluate the IDPF at each level, we always compute the field vector representing each node's output, even though we only use this vector from the last computation out of each prefix. I dabbled in optimizations for this, but I wound up either computing something else twice or copying big AES structs around on the stack.

src/idpf.rs
```diff
@@ -0,0 +1,1852 @@
...
+        )));
+    }
+
+    let which_aggregator = agg_id != 0;
+
+    // Check for cached seeds first.
+    if prefix.len() > 1 {
+        // Skip checking for `prefix` in the cache, because we don't store field element
+        // values along with seeds and control bits. Instead, start looking one node higher
+        // up, so we can recompute everything for the last level of `prefix`.
```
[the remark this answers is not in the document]
divergentdave: Actual evaluation of the IDPF does start with level zero and `prefix[0]`, but this loop is intentionally operating in the reverse order because it's looking for the node at the greatest level where it can get a cache hit. I'll expand on the comments around this.

src/idpf.rs
```diff
@@ -0,0 +1,1852 @@
...
+        }
+        control_bits.extend(
+            self.leaf_correction_word
+                .control_bits
+                .iter()
+                .map(|x| bool::from(*x)),
+        );
+        control_bits.set_uninitialized(false);
+        let mut packed_control = control_bits.into_vec();
+        packed_control.reverse();
```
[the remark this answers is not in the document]
divergentdave: If I reversed the semantic order of the bit indexes and switched from `Lsb0` to `Msb0`, I think that would cancel out and give us the combination of bit endianness and byte endianness we want.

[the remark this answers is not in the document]
divergentdave: Wait, that wouldn't work for cases where we have unused bits, because the leading byte/most significant byte would have the wrong bits filled/empty. I think I'll have to ditch using `bitvec` here to avoid the `reverse()` call, and write the shifting and masking myself.

src/idpf.rs
```diff
@@ -0,0 +1,1831 @@
...
+    public_share: &IdpfPoplarPublicShare<Field64, Field255, L, OUT_LEN>,
+    key: &Seed<L>,
+    prefix: &IdpfInput,
+    cache: &mut dyn IdpfCache<L>,
+) -> Result<IdpfOutputShare<OUT_LEN, Field64, Field255>, IdpfError>
+where
+    P: Prg<L>,
+{
+    let bits = public_share.inner_correction_words.len() + 1;
+    let is_leader = agg_id == 0;
```
cjpatton, line 452: We should probably assert that agg_id is 0 or 1.

src/idpf.rs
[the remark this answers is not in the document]
cjpatton: Phew, the comments clear this up. Thanks for clarifying!

src/idpf.rs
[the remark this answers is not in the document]
cjpatton: (Sorry to do this.) Thanks for figuring out an alternative. The code is far more complex and therefore less clearly correct. In this instance, I would prefer (possibly) slower code that is more efficient. Would you mind reverting to using `reverse()`?

src/idpf.rs
cjpatton: Comments cleared up my confusion, thanks

cjpatton (commented):
Just a few minor comments left.

benches/cycle_counts.rs
```diff
@@ -152,67 +232,165 @@ cfg_if! {
...
-                    prio2_prove_and_verify_10,
-                    prio2_prove_and_verify_100,
-                    prio2_prove_and_verify_1000,
-                    prio3_client_count,
-                    prio3_client_histogram_11,
-                    prio3_client_sum_32,
-                    prio3_client_count_vec_1000,
-                    prio3_client_count_vec_multithreaded_1000,
-                );
+                cfg_if!{
```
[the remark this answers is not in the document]
cjpatton: Bump: @divergentdave addressed this in a way I think works, but more eyes on this would be helpful.

cjpatton (approved):
Looks good to me. Please be sure to squash before merging. Also, there is one outstanding comment that I'd appreciate another pair of eyes on (cc/ @tgeoghegan).

One final note: It occurs to me that we need to improve the coverage of the test vectors: For IDPF, if both aggregators have identical key shares (as they do in the current test), the correction words will cancel out. This prevents us from certain things, such as serialization of the info string input to the PRG.

I'll address this in the spec (https://github.com/cfrg/draft-irtf-cfrg-vdaf/pull/132#issuecomment-1386192389) and send a PR with new test vectors and any patches that are needed.

src/field.rs
```diff
@@ -420,6 +421,12 @@ macro_rules! make_field {
             }
         }
 
+        impl ConstantTimeEq for $elem {
```
tgeoghegan, line 424: This `impl` is available anywhere one is working with a concrete type that implements `FieldElement` via this macro. Should we add `ConstantTimeEq` to `trait FieldElement`'s bounds so that it's available when working generically with `impl FieldElement`?

src/lib.rs
```diff
@@ -22,6 +22,8 @@ mod fft;
 pub mod field;
 pub mod flp;
 mod fp;
+#[cfg(feature = "experimental")]
+pub mod idpf;
```
tgeoghegan, line 26: Clients of crate `prio` should be using the eventual `Poplar1` VDAF, not the underlying `idpf` module. Can this be `pub(crate)` or otherwise restricted?

src/vdaf.rs
```diff
@@ -47,6 +49,11 @@ pub enum VdafError {
     /// failure when calling getrandom().
     #[error("getrandom: {0}")]
     GetRandom(#[from] getrandom::Error),
+
+    /// IDPF error.
+    #[cfg(feature = "experimental")]
+    #[error("idpf error: {0}")]
+    Idpf(#[from] IdpfError),
```
tgeoghegan, line 56: I think we have a problem if we need to modify generic interfaces exported from `mod vdaf` to accommodate new VDAFs. Our goal is that someone should be able to implement new VDAFs outside of crate `prio` by implementing traits `Vdaf, Aggregator, Collector, Client`. If they don't get to extend this enum, what are they supposed to do? I think what's called for here is something like `codec::CodecError`'s `Other(#[source] Box<dyn Error + 'static + Send + Sync>)` variant.

That being said, this problem already existed (`VdafError::Flp` is arguably specific to the `Prio3` family), so I don't think it should be fixed in this PR: #424.

src/idpf.rs
```diff
@@ -0,0 +1,1831 @@
...
+        let bit_slice_u8_storage = bytes.view_bits::<Lsb0>();
+        let mut bit_vec_usize_storage = bitvec![0; bit_slice_u8_storage.len()];
+        bit_vec_usize_storage.clone_from_bitslice(bit_slice_u8_storage);
+        IdpfInput {
+            index: bit_vec_usize_storage.into_boxed_bitslice(),
+        }
+    }
+
+    /// Convert a slice of booleans into an IDPF input.
+    pub fn from_bools(bools: &[bool]) -> IdpfInput {
```
tgeoghegan, line 56: Seems odd that we have `From<BitVec>` and `From<BitBox>` below but these aren't `From<&[u8]>` and `From<&[bool]>`, but I'm OK with either.

src/idpf.rs
```diff
@@ -0,0 +1,1831 @@
...
+        IdpfInput {
+            index: bits.into_boxed_bitslice(),
+        }
+    }
+
+    /// Create a new IDPF input by appending one bit to this input.
+    pub fn clone_with_suffix(&self, suffix: &[bool]) -> IdpfInput {
+        let mut vec = BitVec::with_capacity(self.index.len() + 1);
+        vec.copy_from_bitslice(&self.index);
+        vec.extend(suffix);
```
tgeoghegan, line 67: AFAICT, [`Extend::extend`](https://doc.rust-lang.org/std/iter/trait.Extend.html#tymethod.extend) won't honor the `Vec`'s capacity. Won't this grow `vec` by more than one bit if `suffix.len() > 1`?

benches/cycle_counts.rs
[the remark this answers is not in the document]
tgeoghegan: I might not have done it with macros, but this seems fine to me.

src/lib.rs
cjpatton: I think it's useful to export. For example, I might end up implementing (a prototype of) Doplar elsewhere. However since it's behind `experimental` it won't show up with the normal feature set anyway.

Incidentally: What's our goal for API breaking changes for `experimental`? Do we care if we make breaking changes within the same version?

divergentdave: We will also need to either export or wrap `IdpfInput`, for use in the eventual Poplar1 API.

src/idpf.rs
divergentdave: This was done to avoid implicit copies in from/into conversions, see [this thread](https://github.com/divviup/libprio-rs/pull/414#discussion_r1072891268).

src/lib.rs
tgeoghegan: > Incidentally: What's our goal for API breaking changes for `experimental`? Do we care if we make breaking changes within the same version?

IMO anything experimental is unstable, i.e. the API can break without a major version change.

> We will also need to either export or wrap `IdpfInput`, for use in the eventual Poplar1 API.

Oh, OK, that's good enough for me. Just wanted to make sure we're deliberately committing to public API surface.

benches/cycle_counts.rs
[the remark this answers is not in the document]
tgeoghegan: To be clear: I'm not arguing against using macros here, it's just that I'm not as good at them as David is, so it's not the tool I would have reached for.

src/vdaf.rs
cjpatton: @divergentdave seems to be following a design pattern already in this crate: Each module has its own error type. That said I agree. And the error cases are generic enough that I think we could just return `VdafError` instead of creating our own error type.

tgeoghegan: I think it's correct for `mod Idpf` to define `IdpfError`, but it's not correct for that to appear as a variant in `VdafError`, which should be generic and not contain any VDAF's specifics.

cjpatton: OOoh sorry I misunderstood.

src/idpf.rs
tgeoghegan: The doccomment needs to be updated as clearly this can append more than "one bit".

cjpatton:
@divergentdave can you rebase?

tgeoghegan (approved):

PR #415, opened 2023-01-14
Speed up fpvec tests

divergentdave:
This makes some changes to `Prio3Aes128FixedPointBoundedL2VecSum`-related tests to speed them up. First, the regression test from #409 is changed to run on a 5-element vector, rather than 1000. I confirmed that this test still fails if the fix to the FLP is reverted. (I also renamed various items throughout the test to reflect the changes.) This took 3.2s on my machine before, now it takes 0.03s.

Secondly, I split up `flp::types::fixedpoint_l2::tests::test_bounded_fpvec_sum_parallel` into four tests, so that they can be run in parallel. Previously, this took 0.86s on my machine, now the longest takes 0.61s. This is still one of the long poles, but I don't immediately see any other easy wins to take here.

### DIFF
--- a/src/flp/types/fixedpoint_l2.rs
+++ b/src/flp/types/fixedpoint_l2.rs
@@ -606,6 +606,50 @@ mod tests {
     use fixed::{FixedI128, FixedI16, FixedI64};
     use fixed_macro::fixed;
 
+    #[test]
+    fn test_bounded_fpvec_sum_parallel_fp16() {
+        let fp16_4_inv = fixed!(0.25: I1F15);
+        let fp16_8_inv = fixed!(0.125: I1F15);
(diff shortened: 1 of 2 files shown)


src/flp/types/fixedpoint_l2.rs
```diff
@@ -606,8 +606,122 @@ mod tests {
     use fixed::{FixedI128, FixedI16, FixedI64};
     use fixed_macro::fixed;
 
+    fn test_fixed<F: Fixed>(fp_vec: Vec<F>, enc_vec: Vec<u128>)
```
cjpatton, line 609: Can we put this code back where it was so that we can review the changes line-by-line?

cjpatton (approved):

tgeoghegan (approved):

PR #420, opened 2023-01-20
Generalize `CountVec` into `SumVec`

cjpatton:
The new FLP type, `SumVec`, is used to compute the element-wise sum of a vector of integers in range `[0, 2^bits)`. Each element of the vector is encoded as a bit-vector then concatenated together into the FLP input. The prover then proves that each element of the flattened vector is a bit in the desired range.

Using `SumVec` is equivalent to using as many `Sum` proofs as there are elements of the vector. Combining them into one proof improves efficiency by reducing overall communication cost.

The valididty circuit is identical to `CountVec`; only the chunk size has changed. In fact, `SumVec` with `bits = 1` should be wire-compatible with `CountVec`.

cc/ @simon-friedberger

### DIFF
--- a/src/flp/types.rs
+++ b/src/flp/types.rs
@@ -8,9 +8,7 @@ use crate::flp::{FlpError, Gadget, Type};
 use crate::polynomial::poly_range_check;
 use std::convert::TryInto;
 use std::marker::PhantomData;
-
-/// The counter data type. Each measurement is `0` or `1` and the aggregate result is the sum of
-/// the measurements (i.e., the total number of `1s`).
+/// The counter data type. Each measurement is `0` or `1` and the aggregate result is the sum of the measurements (i.e., the total number of `1s`).
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct Count<F> {
     range_checker: Vec<F>,
@@ -419,47 +417,80 @@ impl<F: FftFriendlyFieldElement> Type for Histogram<F> {
     }
 }
 
-/// A sequence of counters. This type uses a neat trick from [[BBCG+19], Corollary 4.9] to reduce
-/// the proof size to roughly the square root of the input size.
+/// A sequence of integers in range `[0, 2^bits)`. This type uses a neat trick from [[BBCG+19],
+/// Corollary 4.9] to reduce the proof size to roughly the square root of the input size.
 ///
 /// [BBCG+19]: https://eprint.iacr.org/2019/188
 #[derive(Debug, PartialEq, Eq)]
-pub struct CountVec<F, S> {
+pub struct SumVec<F: FftFriendlyFieldElement, S> {
     range_checker: Vec<F>,
     len: usize,
+    bits: usize,
+    flattened_len: usize,
+    max: F::Integer,
     chunk_len: usize,
     gadget_calls: usize,
     phantom: PhantomData<S>,
 }
 
-impl<F: FftFriendlyFieldElement, S: ParallelSumGadget<F, BlindPolyEval<F>>> CountVec<F, S> {
-    /// Returns a new [`CountVec`] with the given length.
-    pub fn new(len: usize) -> Self {
+impl<F: FftFriendlyFieldElement, S: ParallelSumGadget<F, BlindPolyEval<F>>> SumVec<F, S> {
+    /// Returns a new [`SumVec`] with the desired bit width and vector length.
+    ///
+    /// # Errors
+    ///
+    /// * The length of the encoded measurement, i.e., `bits * len`, overflows addressable memory.
+    /// * The bit width cannot be encoded, i.e., `bits` is larger than or equal to the number of
+    ///   bits required to encode field elements.
+    pub fn new(bits: usize, len: usize) -> Result<Self, FlpError> {
+        let flattened_len = bits.checked_mul(len).ok_or_else(|| {
+            FlpError::InvalidParameter("`bits*len` overflows addressable memory".into())
+        })?;
+
+        // Check if the bit width is too large. This limit is defined to be one bit less than the
+        // number of bits required to encode `F::Integer`. (One less so that we can compute `1 <<
+        // bits` without overflowing.)
+        let limit = std::mem::size_of::<F::Integer>() * 8 - 1;
+        if bits > limit {
+            return Err(FlpError::InvalidParameter(format!(
+                "bit wdith exceeds limit of {limit}"
+            )));
+        }
+
+        // Compute the largest encodable measurement.
+        let one = F::Integer::from(F::one());
+        let max = (one << bits) - one;
+
         // The optimal chunk length is the square root of the input length. If the input length is
         // not a perfect square, then round down. If the result is 0, then let the chunk length be
         // 1 so that the underlying gadget can still be called.
-        let chunk_len = std::cmp::max(1, (len as f64).sqrt() as usize);
+        let chunk_len = std::cmp::max(1, (flattened_len as f64).sqrt() as usize);
 
-        let mut gadget_calls = len / chunk_len;
-        if len % chunk_len != 0 {
+        let mut gadget_calls = flattened_len / chunk_len;
+        if flattened_len % chunk_len != 0 {
             gadget_calls += 1;
         }
 
-        Self {
+        Ok(Self {
             range_checker: poly_range_check(0, 2),
             len,
+            bits,
+            flattened_len,
+            max,
             chunk_len,
             gadget_calls,
             phantom: PhantomData,
-        }
+        })
     }
 }
 
-impl<F: FftFriendlyFieldElement, S> Clone for CountVec<F, S> {
+impl<F: FftFriendlyFieldElement, S> Clone for SumVec<F, S> {
     fn clone(&self) -> Self {
         Self {
             range_checker: self.range_checker.clone(),
             len: self.len,
+            bits: self.bits,
+            flattened_len: self.flattened_len,
+            max: self.max,
             chunk_len: self.chunk_len,
             gadget_calls: self.gadget_calls,
             phantom: PhantomData,
@@ -467,7 +498,7 @@ impl<F: FftFriendlyFieldElement, S> Clone for CountVec<F, S> {
     }
 }
 
-impl<F, S> Type for CountVec<F, S>
+impl<F, S> Type for SumVec<F, S>
 where
     F: FftFriendlyFieldElement,
     S: ParallelSumGadget<F, BlindPolyEval<F>> + Eq + 'static,
@@ -486,14 +517,20 @@ where
             )));
         }
 
-        let max = F::Integer::from(F::one());
-        for value in measurement {
-            if *value > max {
-                return Err(FlpError::Encode("Count value must be 0 or 1".to_string()));
+        let mut flattened = Vec::with_capacity(self.flattened_len);
+        for summand in measurement {
+            if summand > &self.max {
+                return Err(FlpError::Encode(format!(
+                    "summand exceeds maximum of 2^{}-1",
+                    self.bits
+                )));
             }
+            flattened.append(&mut F::encode_into_bitvector_representation(
+                summand, self.bits,
+            )?);
         }
 
-        Ok(measurement.iter().map(|value| F::from(*value)).collect())
+        Ok(flattened)
     }
 
     fn decode_result(
@@ -546,11 +583,15 @@ where
 
     fn truncate(&self, input: Vec<F>) -> Result<Vec<F>, FlpError> {
         self.truncate_call_check(&input)?;
-        Ok(input)
+        let mut unflattened = Vec::with_capacity(self.len);
+        for chunk in input.chunks(self.bits) {
+            unflattened.push(F::decode_from_bitvector_representation(chunk)?);
+        }
+        Ok(unflattened)
     }
 
     fn input_len(&self) -> usize {
-        self.len
+        self.flattened_len
     }
 
     fn proof_len(&self) -> usize {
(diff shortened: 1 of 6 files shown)


src/flp/types.rs
```diff
@@ -419,55 +419,60 @@ impl<F: FftFriendlyFieldElement> Type for Histogram<F> {
...
 
-/// A sequence of counters. This type uses a neat trick from [[BBCG+19], Corollary 4.9] to reduce
-/// the proof size to roughly the square root of the input size.
+/// A sequence of integers in range `[0, 2^bits)`. This type uses a neat trick from [[BBCG+19],
+/// Corollary 4.9] to reduce the proof size to roughly the square root of the input size.
 ///
 /// [BBCG+19]: https://eprint.iacr.org/2019/188
 #[derive(Debug, PartialEq, Eq)]
-pub struct CountVec<F, S> {
+pub struct SumVec<F, S> {
```
tgeoghegan, line 427: Renaming this changes the crate's public API. This isn't a problem for VDAF compliance, since no draft specifies either `CountVec` or `SumVec`, but it does mean we'd want to bump the crate version before releasing a version that includes this. Unless, that is, we can put back in a `SumVec` type that's just an alias for `CountVec` with `bits = 1`, and then the API won't change. I'm comfortable with either.

src/flp/types.rs
```diff
@@ -486,14 +491,21 @@ where
             )));
         }
 
-        let max = F::Integer::from(F::one());
-        for value in measurement {
-            if *value > max {
-                return Err(FlpError::Encode("Count value must be 0 or 1".to_string()));
+        let max = F::Integer::try_from(1 << self.bits).unwrap();
+        let mut flattened = Vec::with_capacity(self.bits * self.len);
```
tgeoghegan, line 495: I think we should check whether `1 << self.bits` or `self.bits * self.len` overflows and return an error if so. `SumVec::new` is probably a good place to do that, since if you check it there, you can assume those values are sane everywhere else in the `impl`.

tgeoghegan (changes requested):

src/flp/types.rs
cjpatton: We could just define an alias, yeah. The only problem I foresee is that `CountVec::new()` would to `SumVec::new()`, which is ambiguous because this type doesn't take in a `bits` parameter. One option is to do

```rust
pub struct CountVec(SumVec);
```

instead of

```rust
pub type CountVec = SumVec
```

cjpatton: If we're bumping the version, it might be a good idea to merge this into a branch other than main so that we can accumulate more API-breaking changes before bumping.

divergentdave: For what it's worth, we do have some breaking changes floating on main currently -- the output share/aggregate share `Encode`/`ParameterizedDecode` trait bounds, and moving some items behind the experimental feature.

cjpatton: I'm good either way -- just let me know what y'all decide :)

simon-friedberger: Increase the major version!

src/flp/types.rs
```diff
@@ -419,55 +417,90 @@ impl<F: FftFriendlyFieldElement> Type for Histogram<F> {
...
 
-impl<F: FftFriendlyFieldElement, S: ParallelSumGadget<F, BlindPolyEval<F>>> CountVec<F, S> {
-    /// Returns a new [`CountVec`] with the given length.
-    pub fn new(len: usize) -> Self {
+impl<F: FftFriendlyFieldElement, S: ParallelSumGadget<F, BlindPolyEval<F>>> SumVec<F, S> {
+    /// Returns a new [`SumVec`] with the bit width and vector length.
+    ///
+    /// # Errors
+    ///
+    /// * The length of the encoded measurement, i.e., `bits * len`, overflows addressible memory
```
tgeoghegan, line 441:
```suggestion
    /// * The length of the encoded measurement, i.e., `bits * len`, overflows addressable memory
```

src/flp/types.rs
```diff
@@ -486,14 +519,20 @@ where
...
 
-        let max = F::Integer::from(F::one());
-        for value in measurement {
-            if *value > max {
-                return Err(FlpError::Encode("Count value must be 0 or 1".to_string()));
+        let mut flattened = Vec::with_capacity(self.flattened_len);
+        for summand in measurement {
+            if summand > &self.max {
+                return Err(FlpError::Encode(format!(
+                    "summands is out of range [0, {})",
```
tgeoghegan, line 526:
```suggestion
                    "summand is out of range [0, {})",
```

tgeoghegan (approved):
Couple more typos and then this is good to go.

src/flp/types.rs
tgeoghegan: Oh, thanks @divergentdave, I forgot we had such changes on `main` already. Plus we know there will be API breaking stuff for VDAF-04.

src/flp/types.rs
```diff
@@ -419,55 +417,90 @@ impl<F: FftFriendlyFieldElement> Type for Histogram<F> {
...
-    /// Returns a new [`CountVec`] with the given length.
-    pub fn new(len: usize) -> Self {
+impl<F: FftFriendlyFieldElement, S: ParallelSumGadget<F, BlindPolyEval<F>>> SumVec<F, S> {
+    /// Returns a new [`SumVec`] with the bit width and vector length.
+    ///
+    /// # Errors
+    ///
+    /// * The length of the encoded measurement, i.e., `bits * len`, overflows addressible memory
+    /// * The bit width cannot be encoded, i.e., `bits` is larger than the number of bits in
+    /// required to encode field elements.
```
divergentdave, line 443: nit: extra word
```suggestion
    /// * The bit width cannot be encoded, i.e., `bits` is larger than the number of bits
    /// required to encode field elements.
```

src/flp/types.rs
```diff
@@ -419,55 +417,90 @@ impl<F: FftFriendlyFieldElement> Type for Histogram<F> {
...
+    ///
+    /// Behavior is undefined if the size of `F::Integer` exceeds `F::ENCODED_SIZE`.
```
divergentdave, line 445: I think this is backwards, and we only get into trouble if `F::Integer` has a smaller range than `F`. (assuming this is about calculation of `max`)

src/flp/types.rs
```diff
@@ -419,55 +417,90 @@ impl<F: FftFriendlyFieldElement> Type for Histogram<F> {
...
+    pub fn new(bits: usize, len: usize) -> Result<Self, FlpError> {
+        let flattened_len = bits.checked_mul(len).ok_or_else(|| {
+            FlpError::InvalidParameter("`bits*len` overflows addressible memory".into())
+        })?;
+
+        // Check if the bit width might exceed its limit. This limit is defined to be the number of
+        // bits required to encode finite field elements. If the size of `F::Integer` is less than
+        // or equal to `F::ENCODED_SIZE`, then `max` below is guaranteed to not overflow.
```
divergentdave, line 453: Nit: technically, if `bits` is equal to `F::ENCODED_SIZE * 8`, then the `one << bits` shift will overflow and panic in debug builds. We'd need to use something like `1u128.overflowing_neg().0` with right shifts to safely compute `max` in this case, but that would require more trait jenga.

SumVec with `bits` equal to the field modulus's bit length is of limited use, since any one measurement could wrap around the sum. Thus, I think we should just say not to use it with `bits` equal to the bit width of the field/integer, and update the below error's boundary by one.

src/flp/types.rs
```diff
@@ -486,14 +519,20 @@ where
...
-        let max = F::Integer::from(F::one());
-        for value in measurement {
-            if *value > max {
-                return Err(FlpError::Encode("Count value must be 0 or 1".to_string()));
+        let mut flattened = Vec::with_capacity(self.flattened_len);
+        for summand in measurement {
+            if summand > &self.max {
+                return Err(FlpError::Encode(format!(
+                    "summands is out of range [0, {})",
+                    1 << self.bits
```
divergentdave, line 527: Just in case, we should make sure the compiler doesn't default this to i32, as that would overflow too early for some fields.
```suggestion
                    1u128 << self.bits
```

src/flp/types.rs
cjpatton: Sweet!

src/flp/types.rs
cjpatton: Good catch. Nit-picking a bit, this would overflow for fields larger than 128 bits. Instead of adding a weird edge case here, I think i'll just not try to print this number.

src/flp/types.rs
cjpatton: Good catch, done.

One minor note here: In theory, FLPs might have applications beyond aggregation, i.e., `SumVec` *could* be used for something besides summing. Of course we have constructed our API explicitly around aggregation, so this is maybe a moot point.

src/flp/types.rs
cjpatton: Yes, this is about computing `max`. My claim is that the computation of `max` is guaranteed to not overflow if the size of `F::Integer` is less than or equal to `F::ENCODED_SIZE`, hence the behavior being undefined if this is not he case.

However, it occurred to me that we can easily get around this undefined behavior since `F::Integer` is `Sized`:

```rust
let limit = std::mem::size_of::<F::Integer>() * 8 - 1;
```

src/vdaf/prio3.rs
```diff
@@ -75,41 +75,48 @@ impl Prio3Aes128Count {
...
-pub type Prio3Aes128CountVec =
-    Prio3<CountVec<Field128, ParallelSum<Field128, BlindPolyEval<Field128>>>, PrgAes128, 16>;
+pub type Prio3Aes128SumVec =
+    Prio3<SumVec<Field128, ParallelSum<Field128, BlindPolyEval<Field128>>>, PrgAes128, 16>;
 
 #[cfg(feature = "crypto-dependencies")]
-impl Prio3Aes128CountVec {
-    /// Construct an instance of Prio3Aes1238CountVec with the given number of aggregators. `len`
-    /// defines the length of each measurement.
-    pub fn new_aes128_count_vec(num_aggregators: u8, len: usize) -> Result<Self, VdafError> {
```
cjpatton, line 85: Reviewer note: It might be worth keeping this constructor around for convenience. (Similarly below.)

src/flp/types.rs
```diff
@@ -419,55 +417,89 @@ impl<F: FftFriendlyFieldElement> Type for Histogram<F> {
...
+    /// * The bit width cannot be encoded, i.e., `bits` is larger than or equal to the number of
+    ///   bits required to encode field elements.
+    pub fn new(bits: usize, len: usize) -> Result<Self, FlpError> {
+        let flattened_len = bits.checked_mul(len).ok_or_else(|| {
+            FlpError::InvalidParameter("`bits*len` overflows addressable memory".into())
+        })?;
+
+        // Check if the bit width is too large. This limit is defined to be one bit less than the
+        // number of bits required to encode `F::Integer`. (One less so that we can compute `1 <<
+        // bits` without overlfowing.)
```
divergentdave, line 451: nit: typo
```suggestion
        // bits` without overflowing.)
```

divergentdave (approved):

cjpatton:
Squashed!

src/flp/types.rs
cjpatton: Oh that's a good idea :)

cjpatton: Done.

src/flp/types.rs
```diff
@@ -980,13 +1035,26 @@ mod tests {
             .unwrap();
         }
 
+        let len = 23;
+        let sum_vec = f(2, len).unwrap();
+        flp_validity_test(
+            &sum_vec,
+            &vec![nine; 2 * len],
```
simon-friedberger, line 1042: 9 cannot be encoded with 2 bits. Maybe this should be fixed to be independent of the order of checks.

simon-friedberger (commented):
The changes look reasonable but I haven't extensively reviewed the old CountVec code.
I couldn't find any tests that actually go from measurement to aggregate. Shouldn't we have some?

cjpatton:
> The changes look reasonable but I haven't extensively reviewed the old CountVec code. I couldn't find any tests that actually go from measurement to aggregate. Shouldn't we have some?

I've tacked on a commit that expands the tests in `src/prio3.rs` to exercise aggregation over multiple measurements.

src/vdaf/prio3.rs
```diff
@@ -1176,35 +1183,39 @@ mod tests {
...
-                ]]
+                [
+                    vec![0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 1, 0, 0, 1, 0, 1, 1, 1, 0, 1],
+                    vec![0, 2, 0, 0, 1, 0, 0, 0, 1, 1, 1, 3, 0, 3, 0, 0, 0, 1, 0, 0],
+                    vec![1, 1, 1, 0, 1, 1, 0, 1, 1, 1, 1, 0, 1, 1, 1, 1, 0, 1, 0, 1],
+                ]
             )
             .unwrap(),
-            vec![0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 1, 0, 0, 1, 0, 1, 1, 1, 0, 1,]
+            vec![1, 3, 1, 0, 3, 1, 0, 1, 2, 2, 3, 3, 1, 5, 1, 2, 1, 3, 0, 2],
```
cjpatton, line 1199: @simon-friedberger here is the expanded test (Similarly for the multithreaded below.)

simon-friedberger (approved):

cjpatton:
Squashed.

PR #425, opened 2023-01-26
Clippy fixes, 1.67

divergentdave:
This updates the edition to 2021 (necessary for a change to `panic!()`) and fixes lints related to format strings.


branlwyd (approved):

tgeoghegan (approved):
Does changing the edition count as an API break?

divergentdave:
No, the edition only affects code inside the crate, and the public API is unchanged. (The relevant changes here are builtin macros and the prelude. Editions can also change the set of reserved keywords, off the top of my head.)

Issue #141, opened 2021-11-17
Complete `poplar1` implementation


cjpatton:
Our implementation of [`poplar1`](https://datatracker.ietf.org/doc/html/draft-patton-cfrg-vdaf-00#section-7) is here: https://github.com/abetterinternet/libprio-rs/blob/main/src/vdaf/poplar1.rs. It's functional, but lacks an efficient implementation of an IDPF. We will need an IPDF that is interoperable with https://github.com/google/distributed_point_functions.

cjpatton:
As of #414 we have the IDPF done. I will start working on plumbing it into the poplar1 module.

PR #429, opened 2023-01-27
prng: Implement conversion of `Prng` to new field type

cjpatton:
Partially addresses #141.

Poplar1 requires the same seed stream to be used for producing multiple field types. To facilitate this, add a new method on `Prng` for converting the state to the desired field.

### DIFF
--- a/src/prng.rs
+++ b/src/prng.rs
@@ -32,7 +32,6 @@ pub(crate) struct Prng<F, S> {
     seed_stream: S,
     buffer: Vec<u8>,
     buffer_index: usize,
-    output_written: usize,
 }
 
 #[cfg(feature = "crypto-dependencies")]
@@ -67,7 +66,6 @@ where
             seed_stream,
             buffer,
             buffer_index: 0,
-            output_written: 0,
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/prng.rs
```diff
@@ -76,23 +74,51 @@ where
...
+
+    /// Convert this object into a field element generator for a different field.
+    //
+    // TODO(cjpatton) spec: Consider using distinct seeds for distinct field types. Buffering the
+    // seed stream seems to be important for performance, at least according to the benchmarks for
+    // `PrgAes128`. But having to support multiple output types is delicate because the buffer size
+    // is computed from the field modulus.
+    //
+    // TODO(cjpatton) Remove "dead_code" marker after Poplar1 is implemented.
+    #[allow(dead_code)]
```
tgeoghegan, line 113: nit: `#[cfg(test)]` would work too, and it'd be harder to forget to remove it because the compiler would force you to do that when you want to use it from a non-test cfg. But this works fine and is going away in the PR stack you've already prepared so it's not worth rebasing for.

src/prng.rs
```diff
@@ -207,4 +233,39 @@ mod tests {
         let actual = prng.nth(145).unwrap();
         assert_eq!(actual, expected);
     }
+
+    // Test that the `Prng`'s internal buffer properly copies the end of the bufer to the front
```
tgeoghegan, line 237:
```suggestion
    // Test that the `Prng`'s internal buffer properly copies the end of the buffer to the front
```

src/prng.rs
```diff
@@ -76,23 +74,51 @@ where
...
-                    self.buffer_index = j;
-                    self.output_written += 1;
                     return x;
                 }
             }
 
-            // Refresh buffer with the next chunk of PRG output.
-            self.seed_stream.fill(&mut self.buffer);
+            // Refresh buffer with the next chunk of PRG output, filling the front of the buffer
+            // with the leftovers. This ensures continuity of the seeds stream after converting the
```
tgeoghegan, line 94:
```suggestion
            // with the leftovers. This ensures continuity of the seed's stream after converting the
```

tgeoghegan (approved):

src/prng.rs
```diff
@@ -76,23 +74,51 @@ where
...
+            // `Prng` to a new field type via `into_new_field()`.
+            let left_over = self.buffer.len() - self.buffer_index;
+            for i in 0..left_over {
+                self.buffer[i] = self.buffer[self.buffer_index + i];
+            }
```
divergentdave, line 99: nit: This could use the `copy_within()` method, which would probably hoist some bounds checks

src/prng.rs
```diff
@@ -76,23 +74,51 @@ where
...
+            self.seed_stream.fill(&mut self.buffer[left_over..]);
             self.buffer_index = 0;
         }
     }
+
+    /// Convert this object into a field element generator for a different field.
+    //
+    // TODO(cjpatton) spec: Consider using distinct seeds for distinct field types. Buffering the
```
divergentdave, line 107: If we don't make this spec change, we should have some additional PRG test vectors to cover generating elements in multiple fields from one hash input.

src/prng.rs
cjpatton: Should be "seed stream". Fixed.

src/prng.rs
cjpatton: Good idea, I'll make a note. The `left_over_buffer_back_fill` test is meant to replicate the conditions that lead to this, but it's sensible to just have test vectors as well.

I'm writing up an issue for the VDAF spec summarizing all of the changes I'd like to make.

src/prng.rs
cjpatton: Wow, that was the best thing that happened to me all day. What a find!

divergentdave (approved):

PR #432, opened 2023-01-27
field255: Implement `TryFrom<Field255>` for `u64`

cjpatton:
Based on #431 (merge that first).
Partially addresses #141.

Poplar1 unsharding requires conversion from the leaf field to `u64`.

### DIFF
--- a/src/field/field255.rs
+++ b/src/field/field255.rs
@@ -17,6 +17,7 @@ use std::{
     fmt::{self, Debug, Display, Formatter},
     io::{Cursor, Read},
     marker::PhantomData,
+    mem::size_of,
     ops::{Add, AddAssign, Div, DivAssign, Mul, MulAssign, Neg, Sub, SubAssign},
 };
 use subtle::{Choice, ConstantTimeEq, ConstantTimeGreater, ConstantTimeLess};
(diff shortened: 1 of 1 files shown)


src/field/field255.rs
```diff
@@ -289,6 +289,24 @@ impl Default for Field255 {
...
 
+impl TryFrom<Field255> for u64 {
+    type Error = FieldError;
+
+    fn try_from(elem: Field255) -> Result<u64, FieldError> {
+        let mut le_bytes = [0; 32];
+        fiat_25519_to_bytes(&mut le_bytes, &elem.0);
+        let mut is_zero = Choice::from(1);
+        for byte in &le_bytes[8..] {
+            is_zero &= !Choice::from(*byte);
```
divergentdave, line 300: This constructor should only be used with `0u8` or `1u8`, changing the "too_big" test to use `[2; 32]` hits a debug assertion inside subtle. Instead, we should use `<u8 as ConstantTimeEq>` methods.

divergentdave (changes requested):

src/field/field255.rs
```diff
@@ -289,6 +289,24 @@ impl Default for Field255 {
...
 }
 
+impl TryFrom<Field255> for u64 {
+    type Error = FieldError;
+
+    fn try_from(elem: Field255) -> Result<u64, FieldError> {
+        let mut le_bytes = [0; 32];
+        fiat_25519_to_bytes(&mut le_bytes, &elem.0);
+        let mut is_zero = Choice::from(1);
+        for byte in &le_bytes[8..] {
```
tgeoghegan, line 299: nit: to make it more clear why we're checking that the last 24 bytes are 0, `use std::mem::size_of` and:
```suggestion
        for byte in &le_bytes[size_of::<u64>()..] {
```

tgeoghegan (changes requested):

cjpatton:
Rebased.

divergentdave (approved):

tgeoghegan (approved):

cjpatton:
Rebased once more!

PR #433, opened 2023-01-28
poplar1: Stub out the code and revise types

cjpatton:
Based on #432 (merge that first).
Partially addresses #141.=

Prepare the `poplar1` module for the new-and-improved `Poplar1` by removing the old code. While at it, update the types to those we know we're going to need.

### DIFF
--- a/src/vdaf/poplar1.rs
+++ b/src/vdaf/poplar1.rs
@@ -1,646 +1,250 @@
 // SPDX-License-Identifier: MPL-2.0
 
-//! **(NOTE: This module is experimental. Applications should not use it yet.)** This module
-//! partially implements the core component of the Poplar protocol [[BBCG+21]]. Named for the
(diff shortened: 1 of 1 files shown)


src/vdaf/poplar1.rs
```diff
@@ -1,646 +1,250 @@
...
-            for j in 0..8 {
-                let bit = (byte >> j) & 1;
-                if i < level {
-                    index |= (bit as usize) << i;
-                }
-                i += 1;
-            }
+impl Poplar1Aes128 {
+    /// Create an instance of [`Poplar1Aes128`]. The caller provides the bit length of each
+    /// measurement (`BITS` as defined in the [[draft-irtf-cfrg-vdaf-03]].
```
tgeoghegan, line 23:
```suggestion
    /// measurement (`BITS` as defined in the [[draft-irtf-cfrg-vdaf-03]]).
```

tgeoghegan (approved):

PR #443, opened 2023-02-02
poplar1: Add speed tests for sharding and preparation

cjpatton:
Based on #434 (merge that first).
Partially addresses #141.

The computationally intensive parts of Poplar1 are sharding and initializing preparation (requires evaluating the IDPF shares). Add speed tests for both of these computations.

The speed of the preparation step is dominated by the length and distribution of the candidate prefixes. As such, it is necessary to pick a distribution that is realistic. Following [BBCG+21], we sample the measurements from a Zipf distribution and derive the candidate prefixes from the prefix tree for these measurements.

While at it, add an additional unit test for `poplar1` for exercising the end-to-end heavy hitters functionality.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,6 +50,7 @@ hex = { version = "0.4.3", features = ["serde"] }
 prio = { path = ".", features = ["test-util"] }
 cfg-if = "1.0.0"
 once_cell = "1.17.0"
+zipf = "7.0.0"
 
 [features]
 default = ["crypto-dependencies"]
--- a/benches/speed_tests.rs
+++ b/benches/speed_tests.rs
@@ -31,12 +31,18 @@ use prio::{
 use prio::{
     field::{Field255, Field64},
     idpf::{self, IdpfInput, RingBufferCache},
-    vdaf::prg::PrgAes128,
+    vdaf::{
+        poplar1::{Poplar1, Poplar1AggregationParam},
+        prg::PrgAes128,
+        Aggregator,
+    },
 };
 #[cfg(feature = "experimental")]
-use rand::random;
+use rand::{prelude::*, random};
 #[cfg(feature = "experimental")]
 use std::iter;
+#[cfg(feature = "experimental")]
+use zipf::ZipfDistribution;
 
 /// This benchmark compares the performance of recursive and iterative FFT.
 pub fn fft(c: &mut Criterion) {
@@ -383,7 +389,8 @@ pub fn idpf(c: &mut Criterion) {
             b.iter(|| {
                 // This is an aggressively small cache, to minimize its impact on the benchmark.
                 // In this synthetic benchmark, we are only checking one candidate prefix per level
-                // instead of the usual two, so the cache hit rate will be unaffected.
+                // (typically there are many candidate prefixes per level) so the cache hit rate
+                // will be unaffected.
                 let mut cache = RingBufferCache::new(1);
 
                 for prefix_length in 1..=size {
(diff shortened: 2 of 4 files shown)


Cargo.toml
```diff
@@ -23,6 +23,7 @@ serde = { version = "1.0", features = ["derive"] }
 static_assertions = "1.1.0"
 subtle = "2.4.1"
 thiserror = "1.0"
+zipf = "7.0.0" # XXX Figure out how to fence this dependency for benchmarking
```
cjpatton, line 26: Reviewer note: I'm not sure how to fence this dependency to benchmarks. We certainly don't want it as a main dependency.

benches/speed_tests.rs
```diff
@@ -396,10 +403,154 @@ pub fn idpf(c: &mut Criterion) {
...
+            let nonce: [u8; 16] = random();
+
+            // Parameters are chosen to match Chris Wood's experimental setup:
+            // https://github.com/chris-wood/heavy-hitter-comparison.
+            let (measurements, prefix_tree) = poplar1_generate_zipf_distributed_batch(
+                size, // bits
+                10,   // threshold
+                1000, // number of measurements
+                128,  // Zipf support
+                1.03, // Zipf exponent
```
cjpatton, line 441: @chris-wood can you take a look at this function to make sure I'm generating the outputs correctly? Also, what do you think would be reasonable worst-case parameters to use here?

Cargo.toml
divergentdave: dev-dependency should be good, as benchmarks get built like a different flavor of integration tests

benches/speed_tests.rs
```diff
@@ -396,10 +403,154 @@ pub fn idpf(c: &mut Criterion) {
...
+    let mut group = c.benchmark_group("poplar1_prepare_init");
+    for size in test_sizes.iter() {
+        group.throughput(Throughput::Bytes(*size as u64 / 8));
+        group.bench_with_input(BenchmarkId::from_parameter(size), size, |b, &size| {
+            let vdaf = Poplar1::new_aes128(size);
+            let verify_key = random();
+            let nonce: [u8; 16] = random();
+
+            // Parameters are chosen to match Chris Wood's experimental setup:
+            // https://github.com/chris-wood/heavy-hitter-comparison.
```
tgeoghegan, line 435:
```suggestion
            // https://github.com/chris-wood/heavy-hitter-comparison
```
Makes it easier to copy-paste the URL!

src/vdaf/poplar1.rs
```diff
@@ -234,6 +234,39 @@ pub struct Poplar1AggregationParam {
...
+impl Poplar1AggregationParam {
+    /// Construct an aggregation parameter from a set of candidate prefixes.
+    ///
+    /// # Errors
+    ///
+    /// * The list of prefixes is empty.
+    /// * The prefixes have different lengths (they must all be the same).
+    //
+    // TODO(cjpatton) spec: Ensure that prefixes don't repeat. To make this check easier, consider
+    // requiring them to appear in alphabetical order.
```
tgeoghegan, line 246:
```suggestion
    // TODO spec: Ensure that prefixes don't repeat. To make this check easier, consider
    // requiring them to appear in alphabetical order.
    // https://github.com/cfrg/draft-irtf-cfrg-vdaf/issues/134
```

benches/speed_tests.rs
```diff
@@ -396,10 +403,154 @@ pub fn idpf(c: &mut Criterion) {
...
+                    &public_share,
+                    &input_shares[0],
+                )
+                .unwrap();
+            });
+        });
+    }
+}
+
+/// Generate a set of Poplar1 measurements with the given bit length `bits`. Thehey are sapled
```
tgeoghegan, line 470:
```suggestion
/// Generate a set of Poplar1 measurements with the given bit length `bits`. They are sampled
```

tgeoghegan (approved):

divergentdave (approved):
FYI, when you retarget this PR to main, the supply-chain CI job will run, and fail because `zipf` is missing a safe-to-run audit. (Perhaps I shouldn't have specified branches on the pull_request event) It's just a few hundred lines, I'll look over it and submit an audit separately.

benches/speed_tests.rs
cjpatton: By correct I mean: Are we generating the measurements in the same way you did in your experiment?

Fair point about the threshold.

benches/speed_tests.rs
```diff
@@ -396,10 +403,154 @@ pub fn idpf(c: &mut Criterion) {
...
+        let bools: Vec<bool> = (0..bits).map(|_| rng.gen()).collect();
+        inputs.push(IdpfInput::from_bools(&bools));
+    }
+
+    // Sample a number of inputs according to the Zipf distribution.
+    let mut samples = Vec::with_capacity(measurement_count);
+    let zipf = ZipfDistribution::new(zipf_support, zipf_exponent).unwrap();
+    for _ in 0..measurement_count {
+        samples.push(inputs[zipf.sample(&mut rng) - 1].clone());
+    }
```
chris-wood, line 500: This all seems to match what I did 👍

cjpatton:
Squashed and rebased.

PR #445, opened 2023-02-03
Allow `non_snake_case` at the function level

tgeoghegan:
We suppress variable name convention lints in a couple of spots. While that works for `rustc`, unfortunately the variable names still trip a different lint in `rust-analyzer` (see [1]). Fortunately, there's a workaround: hoist the `#[allow()]` up to the enclosing item (in our case, functions) instead of the individual line.

[1]: https://github.com/rust-lang/rust-analyzer/issues/6541


cjpatton (approved):
👍 Let's just avoid putting "non_snake_case" on everything. I don't want us to get in the habbit of breaking this convention.

divergentdave (approved):

PR #446, opened 2023-02-03
Benchmark cleanup

divergentdave:
This cleans up the `speed_tests` benchmark by moving reporting of message sizes out to a separate binary, renaming benchmarks, organizing benchmarks hierarchically with groups, and moving sizes to the "parameter" part of `BenchmarkId`s.

### DIFF
--- a/benches/speed_tests.rs
+++ b/benches/speed_tests.rs
@@ -7,6 +7,8 @@ use criterion::{criterion_group, criterion_main, BenchmarkId, Criterion};
 use fixed_macro::fixed;
 #[cfg(feature = "multithreaded")]
 use prio::flp::gadgets::ParallelSumMultithreaded;
+#[cfg(feature = "prio2")]
+use prio::vdaf::prio2::Prio2;
 use prio::{
     benchmarked::*,
     field::{random_vector, Field128 as F, FieldElement},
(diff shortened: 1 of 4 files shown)


benches/speed_tests.rs
```diff
@@ -47,17 +43,18 @@ use zipf::ZipfDistribution;
 /// This benchmark compares the performance of recursive and iterative FFT.
 pub fn fft(c: &mut Criterion) {
```
cjpatton, line 44: FWIW, I don't think the FFT benchmarks are all that useful. The goal here was to compare the speed of the iterative FFT algorithm I wrote to the recursive algorithm from the original llbiprio repo.

binaries/src/bin/vdaf_message_sizes.rs
```diff
@@ -0,0 +1,120 @@
...
+        "prio3 fixedpoint16 boundedl2 vec multithreaded ({} entries) size = {}",
+        len,
+        prio3_input_share_size(&prio3.shard(&measurement).unwrap().1)
+    );
+
+    println!();
+
+    for size in [10, 100, 1_000] {
+        // Prio2
+        // Public keys used to instantiate the v2 client.
```
cjpatton, line 85: Consider using the Prio2 VDAF directly to simplify this code. That way we don't have to muck with encryption. (We have regression tests to make sure they are functionally equivalent.)

cjpatton (approved):
❤️

tgeoghegan (approved):

PR #451, opened 2023-02-06
Reduce variance of Poplar1 benchmark

divergentdave:
Previously, the Poplar1 benchmarks randomly generated one set of inputs, and measured the time it took to run the relevant operation over that input repeatedly. This leads to higher variability in benchmark results, primarily through the aggregation parameter. If the prefixes in the aggregation parameter themselves share enough common prefixes, memoization in the preparation algorithm's IDPF evaluation calls influences the total runtime. (We could also be over-training the CPU branch predictor, etc.)

This PR reduces the variability of these benchmarks by both using the same random inputs between runs and using many random inputs in a single run. Calls to `Bencher::iter()` are replaced with `Bencher::iter_batched()`, which takes an additional closure to generate inputs, and a memory consumption hint. Inputs are generated randomly as before, but using an RNG initialized with a fixed seed. The library will do some warmup runs, generate a large number of inputs stored in a `Vec`, and then run the function under test on each input in turn. In my experimentation, this made a noticeable difference in the run-to-run variability. (changes went from +/-7% to +/-2%)

I also made a couple other Criteron-related cleanups: calling `BenchmarkGroup::finish()` and dropping the throughput annotation on `poplar1_prepare_init`. (This benchmark evaluates only the last IDPF level on a somewhat-arbitrary number of prefixes, so I think there are too many confounding factors to turn the results into a bytes/sec throughput)

### DIFF
--- a/benches/speed_tests.rs
+++ b/benches/speed_tests.rs
@@ -1,8 +1,8 @@
 // SPDX-License-Identifier: MPL-2.0
 
-#[cfg(feature = "experimental")]
-use criterion::Throughput;
 use criterion::{criterion_group, criterion_main, BenchmarkId, Criterion};
+#[cfg(feature = "experimental")]
+use criterion::{BatchSize, Throughput};
 #[cfg(feature = "experimental")]
 use fixed_macro::fixed;
 #[cfg(feature = "multithreaded")]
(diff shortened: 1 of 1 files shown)


tgeoghegan (approved):

cjpatton (approved):

benches/speed_tests.rs
```diff
@@ -312,59 +325,84 @@ pub fn poplar1(c: &mut Criterion) {
...
         group.bench_with_input(BenchmarkId::from_parameter(size), size, |b, &size| {
-            let bits = iter::repeat_with(random).take(size).collect::<Vec<bool>>();
-            let measurement = IdpfInput::from_bools(&bits);
-            let vdaf = Poplar1::new_aes128(bits.len());
-
-            b.iter(|| {
-                vdaf.shard(&measurement).unwrap();
-            });
+            let vdaf = Poplar1::new_aes128(size);
+            let mut rng = StdRng::seed_from_u64(RNG_SEED);
```
cjpatton, line 329: * In the near future there will be a version of `shard()` that takes in a rand source for running test vectors, just like we have for Prio3.
* I don't understand why fixing the seed for `shard()` should improve variability. The only codepaths that *should* depend on this value in a way that impacts timing is rejection sampling, but this should be fairly unlikely. (Maybe not for 25519?)

divergentdave: Yeah, I think shard should be mostly constant-time, I mainly did this for consistency of the code.

Issue #454, opened 2023-02-10
VDAF-04 changes


cjpatton:
This is a tracking issue for changes in the next VDAF draft. (Not yet final, we should confirm interop with reference code before cutting.)

- [x] Add nonce to `Client::shard()`. (#455)
- [x] Update `Prg` API (https://github.com/divviup/libprio-rs/issues/423)
- [x] Add `PrgSha3` (mark `PrgAes128` as deprecated)
- [x] Update test vectors and re-enable tests
- [x] Bump `VERSION`

cjpatton:
We haven't yet decided how to spell the nonce size in the spec. (Please chime in on https://github.com/cfrg/draft-irtf-cfrg-vdaf/issues/127 if you have an opinion.) There are two options, basically: (1) have each VDAF define its own nonce size or (2) have fixed, global nonce size, i.e., 16 bytes.

My preference would be (1). My expectation is that this will most likely be tided to seed size, since both the seed size and the nonce size dictate the security level.

Another consideration: I would like to use a const generic for this so that we always have a fixed-size nonce in the API. I would suggest we just re-use `L` (the seed size). Thoughs? In particular @tgeoghegan

cjpatton:
Hmm, this logic actually doesn't work for Prio2, which has a seed size of 32 bytes (but doesn't have this high a security level.) I actually think I found a third option I like better: (3) Let the application determine the nonce size. This is more general than (1). #455 spells this out.

PR #455, opened 2023-02-10
Add nonce to `Client::shard()`

cjpatton:
Partially addresses #454.

This is required for VDAF-04.

### DIFF
--- a/src/vdaf.rs
+++ b/src/vdaf.rs
@@ -163,17 +163,18 @@ pub trait Vdaf: Clone + Debug {
 }
 
 /// The Client's role in the execution of a VDAF.
-pub trait Client: Vdaf {
+pub trait Client<const NONCE_SIZE: usize>: Vdaf {
     /// Shards a measurement into a public share and a sequence of input shares, one for each
     /// Aggregator.
     fn shard(
         &self,
         measurement: &Self::Measurement,
+        nonce: &[u8; NONCE_SIZE],
     ) -> Result<(Self::PublicShare, Vec<Self::InputShare>), VdafError>;
 }
 
 /// The Aggregator's role in the execution of a VDAF.
-pub trait Aggregator<const L: usize>: Vdaf {
+pub trait Aggregator<const VERIFY_KEY_SIZE: usize, const NONCE_SIZE: usize>: Vdaf {
     /// State of the Aggregator during the Prepare process.
     type PrepareState: Clone + Debug;
 
@@ -194,10 +195,10 @@ pub trait Aggregator<const L: usize>: Vdaf {
     /// message.
     fn prepare_init(
         &self,
-        verify_key: &[u8; L],
+        verify_key: &[u8; VERIFY_KEY_SIZE],
         agg_id: usize,
         agg_param: &Self::AggregationParam,
-        nonce: &[u8],
+        nonce: &[u8; NONCE_SIZE],
         public_share: &Self::PublicShare,
         input_share: &Self::InputShare,
     ) -> Result<(Self::PrepareState, Self::PrepareShare), VdafError>;
@@ -219,7 +220,7 @@ pub trait Aggregator<const L: usize>: Vdaf {
         &self,
         state: Self::PrepareState,
         input: Self::PrepareMessage,
-    ) -> Result<PrepareTransition<Self, L>, VdafError>;
+    ) -> Result<PrepareTransition<Self, VERIFY_KEY_SIZE, NONCE_SIZE>, VdafError>;
 
     /// Aggregates a sequence of output shares into an aggregate share.
     fn aggregate<M: IntoIterator<Item = Self::OutputShare>>(
@@ -242,7 +243,11 @@ pub trait Collector: Vdaf {
 
 /// A state transition of an Aggregator during the Prepare process.
 #[derive(Debug)]
-pub enum PrepareTransition<V: Aggregator<L>, const L: usize> {
+pub enum PrepareTransition<
+    V: Aggregator<VERIFY_KEY_SIZE, NONCE_SIZE>,
+    const VERIFY_KEY_SIZE: usize,
+    const NONCE_SIZE: usize,
+> {
     /// Continue processing.
     Continue(V::PrepareState, V::PrepareShare),
 
@@ -336,27 +341,25 @@ pub(crate) fn run_vdaf<V, M, const L: usize>(
     measurements: M,
 ) -> Result<V::AggregateResult, VdafError>
 where
-    V: Client + Aggregator<L> + Collector,
+    V: Client<16> + Aggregator<L, 16> + Collector,
     M: IntoIterator<Item = V::Measurement>,
 {
     use rand::prelude::*;
+    let mut rng = thread_rng();
     let mut verify_key = [0; L];
-    thread_rng().fill(&mut verify_key[..]);
-
-    // NOTE Here we use the same nonce for each measurement for testing purposes. However, this is
-    // not secure. In use, the Aggregators MUST ensure that nonces are unique for each measurement.
-    let nonce = b"this is a nonce";
+    rng.fill(&mut verify_key[..]);
 
     let mut agg_shares: Vec<Option<V::AggregateShare>> = vec![None; vdaf.num_aggregators()];
     let mut num_measurements: usize = 0;
     for measurement in measurements.into_iter() {
         num_measurements += 1;
-        let (public_share, input_shares) = vdaf.shard(&measurement)?;
+        let nonce = rng.gen();
+        let (public_share, input_shares) = vdaf.shard(&measurement, &nonce)?;
         let out_shares = run_vdaf_prepare(
             vdaf,
             &verify_key,
             agg_param,
-            nonce,
+            &nonce,
             public_share,
             input_shares,
         )?;
@@ -399,12 +402,12 @@ pub(crate) fn run_vdaf_prepare<V, M, const L: usize>(
     vdaf: &V,
     verify_key: &[u8; L],
     agg_param: &V::AggregationParam,
-    nonce: &[u8],
+    nonce: &[u8; 16],
     public_share: V::PublicShare,
     input_shares: M,
 ) -> Result<Vec<V::OutputShare>, VdafError>
 where
-    V: Client + Aggregator<L> + Collector,
+    V: Client<16> + Aggregator<L, 16> + Collector,
     M: IntoIterator<Item = V::InputShare>,
 {
     let input_shares = input_shares
--- a/src/vdaf/prio3.rs
+++ b/src/vdaf/prio3.rs
@@ -287,9 +287,8 @@ impl Prio3Aes128Average {
 /// let measurements = [0, 1, 1, 1, 0];
 /// for measurement in measurements {
 ///     // Shard
-///     let (public_share, input_shares) = vdaf.shard(&measurement).unwrap();
-///     let mut nonce = [0; 16];
-///     rng.fill(&mut nonce);
+///     let nonce = rng.gen::<[u8; 16]>();
+///     let (public_share, input_shares) = vdaf.shard(&measurement, &nonce).unwrap();
 ///
 ///     // Prepare
 ///     let mut prep_states = vec![];
@@ -377,9 +376,10 @@ where
         deriver.into_seed()
     }
 
-    fn shard_with_rand_source(
+    fn shard_with_rand_source<const N: usize>(
         &self,
         measurement: &T::Measurement,
+        _nonce: &[u8; N],
         rand_source: RandSource,
     ) -> Result<Vec<Prio3InputShare<T::Field, L>>, VdafError> {
         let mut info = [0; DST_LEN + 1];
@@ -513,11 +513,12 @@ where
     /// Shard measurement with constant randomness of repeated bytes.
     /// This method is not secure. It is used for running test vectors for Prio3.
     #[cfg(feature = "test-util")]
-    pub fn test_vec_shard(
+    pub fn test_vec_shard<const N: usize>(
         &self,
         measurement: &T::Measurement,
+        nonce: &[u8; N],
     ) -> Result<Vec<Prio3InputShare<T::Field, L>>, VdafError> {
-        self.shard_with_rand_source(measurement, |buf| {
+        self.shard_with_rand_source(measurement, nonce, |buf| {
             buf.fill(1);
             Ok(())
         })
@@ -710,7 +711,7 @@ impl<F: FftFriendlyFieldElement, const L: usize> ParameterizedDecode<Prio3Prepar
     }
 }
 
-impl<T, P, const L: usize> Client for Prio3<T, P, L>
+impl<T, P, const L: usize> Client<16> for Prio3<T, P, L>
 where
     T: Type,
     P: Prg<L>,
@@ -719,8 +720,9 @@ where
     fn shard(
         &self,
         measurement: &T::Measurement,
+        nonce: &[u8; 16],
     ) -> Result<(Self::PublicShare, Vec<Prio3InputShare<T::Field, L>>), VdafError> {
-        self.shard_with_rand_source(measurement, getrandom::getrandom)
+        self.shard_with_rand_source(measurement, nonce, getrandom::getrandom)
             .map(|input_shares| ((), input_shares))
     }
 }
@@ -780,7 +782,7 @@ where
     }
 }
 
-impl<T, P, const L: usize> Aggregator<L> for Prio3<T, P, L>
+impl<T, P, const L: usize> Aggregator<L, 16> for Prio3<T, P, L>
 where
     T: Type,
     P: Prg<L>,
@@ -797,7 +799,7 @@ where
         verify_key: &[u8; L],
         agg_id: usize,
         _agg_param: &Self::AggregationParam,
-        nonce: &[u8],
+        nonce: &[u8; 16],
         _public_share: &Self::PublicShare,
         msg: &Prio3InputShare<T::Field, L>,
     ) -> Result<
@@ -959,7 +961,7 @@ where
         &self,
         step: Prio3PrepareState<T::Field, L>,
         msg: Prio3PrepareMessage<L>,
-    ) -> Result<PrepareTransition<Self, L>, VdafError> {
+    ) -> Result<PrepareTransition<Self, L, 16>, VdafError> {
         if self.typ.joint_rand_len() > 0 {
             // Check that the joint randomness was correct.
             if step.joint_rand_seed.as_ref().unwrap() != msg.joint_rand_seed.as_ref().unwrap() {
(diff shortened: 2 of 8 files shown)


src/vdaf.rs
```diff
@@ -166,9 +166,10 @@ pub trait Vdaf: Clone + Debug {
 pub trait Client: Vdaf {
     /// Shards a measurement into a public share and a sequence of input shares, one for each
     /// Aggregator.
-    fn shard(
+    fn shard<const N: usize>(
```
tgeoghegan, line 169: If we put the generic separately on `Client::shard` and `Aggregator::prepare_init`, then there's no enforcement that the same `N` is used. Instead, I think we should have `pub trait Vdaf<const N: usize>` and then `pub trait Client<const N: usize>: Vdaf<N>` and `pub trait Aggregator<const L: usize, const N: usize>: Vdaf<N>` and then all the other places the compiler complains about.

tgeoghegan (changes requested):

src/vdaf.rs
cjpatton: Yeah I think you're right. That approach seems general enough for any solution for https://github.com/cfrg/draft-irtf-cfrg-vdaf/issues/127. I'm gonna send a PR to address that issue, then will update here.

branlwyd, line 169: Rather than using a const generic, could we instead use an [associated constant](https://doc.rust-lang.org/reference/items/associated-items.html#associated-constants)?

The issue with const generics is that they leak into code that is generic over the trait. For example, check out the many instances of `const L: usize` in [Janus](https://github.com/divviup/janus/blob/main/aggregator/src/aggregator.rs). I think using an associated const would avoid this issue.

An associated const would also make it easier to give a meaningful descriptive name to the constant -- right now we have `L` on `Aggregator` and now `N` on `Client`; we could rename these to (e.g.) `VERIFY_KEY_SIZE` and `NONCE_SIZE`, to aid understandability.

cjpatton: The main advantage of using an const generic instead of an associated constant is that the size of the buffer for the nonce can be specified by the API. (To my knowledge, you can't have `&[u8; Self::NONCE_SIZE]` be a type for trait method.) This avoids the edge case where the user supplies `shard()` with a `nonce: &[u8]` whose length is not `Self::NONCE_SIZE`. In fact, this is why we have `const L: usize` for the verification key.

I'd be cool with an associated constant, but I'd like to suggest that we be consistent.

cjpatton: Namely, we could add:

```
pub trait Vdaf:
   const VERIFY_KEY_SIZE: usize;
   const NONCE_SIZE: usize;
```

tgeoghegan: For one thing, nobody said we had to use `N` as the name of the parameter. We can just as easily make it `const NONCE_SIZE: usize` for clarity.

Further, I agree that the value of a const generic is that then the compiler can statically know what the size of `[u8; NONCE_SIZE]` is. Otherwise I think you end up with unnecessary allocations because you have to go through `Vec::new` to get a `&[u8; Self::NONCE_SIZE]`. But I'm happy to be proven wrong; I haven't tried it myself.

So I lean toward a const generic, given my current understanding of this.

cjpatton: Here is a minimal example:
```rust
pub trait Vdaf {
    const NONCE_SIZE: usize;
    fn shard(nonce: &[u8; Self::NONCE_SIZE]);
}
```

Neither stable nor nightly allows you to use an associated type this way:

```
0:20:18:~ $ rustc hella.rs
error: generic parameters may not be used in const operations
 --> hella.rs:3:27
  |
3 |     fn shard(nonce: &[u8; Self::NONCE_SIZE]);
  |                           ^^^^^^^^^^^^^^^^ cannot perform const operation using `Self`
  |
  = note: type parameters may not be used in const expressions

error[E0601]: `main` function not found in crate `hella`
 --> hella.rs:4:2
  |
4 | }
  |  ^ consider adding a `main` function to `hella.rs`

error: aborting due to 2 previous errors

For more information about this error, try `rustc --explain E0601`.
```

cjpatton: So if we wanted to use an associated type, then we'd pass in a `&[u8]` instead of a `&[u8; NONCE_SIZE]`.

branlwyd: That's unfortunate -- I hope they expand Rust to support this, for this reason:

> For one thing, nobody said we had to use N as the name of the parameter. We can just as easily make it const NONCE_SIZE: usize for clarity.

The problem here is that the name is redefined at every point the const generic is used -- so even if `libprio-rs` sensibly names the constant something like `NONCE_SIZE`, implementations have to either (a) switch to using a less-descriptive name like `N` or (b) repeat the longer `NONCE_SIZE` name at every point-of-use of the related trait. Neither is great, and perhaps even worse is that both permit application code to get the name _wrong_, which will start to matter more as we introduce more related constants.

branlwyd (approved):

src/vdaf.rs
cjpatton: Yeah there's not much we can do about that. We can at least nudge users in the right direction.

It sounds like we all might agree on two possibilities:
(1) Rename `L`  to `VERIFY_KEY_SIZE` and add `NONCE_SIZE` as const generic (downside: Users can rename these)
(2) Remove `L` and add `VERIFY_KEY_SIZE` and `NONCE_SIZE` as associated constants (downside: length checks are delegated to implementations of `Vdaf`)

cjpatton: I'm in favor of (1) I think.

branlwyd: Agreed, I am in favor of (1) unless/until we can use associated constants in method signatures.

divergentdave: For future reference, using a generic associated constant in array lengths is currently behind the `#![feature(generic_const_exprs)]` gate in nightly, to an extent.

cjpatton: As discussed in https://github.com/divviup/libprio-rs/pull/455#discussion_r1113368715, I've added a new const generic for the nonce size. I decided to put this on Client and Aggregator rather than VDAF. My thinking is that the Collector won't ever use the nonce length, so it doesn't make sense to add this as a trait bound.

On the other hand, it might be useful to put both `NONCE_SIZE` and `VERIFY_KEY_SIZE` on Vdaf so that the API looks a bit more like the spec. My only objection would be that it *seems to me* to be bad form to add bounds on traits that aren't strictly needed.

src/vdaf/prio3.rs
```diff
@@ -710,7 +711,7 @@ impl<F: FftFriendlyFieldElement, const L: usize> ParameterizedDecode<Prio3Prepar
     }
 }
 
-impl<T, P, const L: usize> Client for Prio3<T, P, L>
+impl<T, P, const L: usize> Client<16> for Prio3<T, P, L>
```
cjpatton, line 714: I think we should consider renaming `L` to `SEED_SIZE` in a follow-up PR.

src/vdaf.rs
tgeoghegan: The downside of putting the generic on `trait Client` and `Aggregator` is that now I can do this (very rough pseudocode):

```
let my_client: Client<NONCE_SIZE = 8> = whatever_mod::new_client();
let my_aggregator: Aggregator<NONCE_SIZE = 16> = whatever_mod::new_aggregator();

let (pub, shares) = my_client.shard(...)?;
let _ = my_aggregator.prep_init(...)?;
```

This won't work, because despite implementing the same `Vdaf`, the client and aggregator use different nonce sizes. I won't find out that it doesn't work until runtime. Now, if `NONCE_SIZE` were on `trait Vdaf`, then the compiler can statically verify that `my_client` and `my_aggregator` are using the same nonce size.

Now, this is a bit of a reach, because this scenario only comes up when writing tests, because otherwise VDAF clients and aggregators are going to be compiled into separate executables and the compiler can't enforce consistency in that scenario. So I'll defer to your preference as the author here.

tgeoghegan (approved):

src/vdaf.rs
cjpatton: I see how this could happen if the implementation of the `Client` and `Aggregator` were generic in the nonce size. However I imagine this will usually be hard-coded. (For example, in this PR, `Prio3` implements `Client<16>`, not `Client<NONCE_SIZE>`. We don't need this to be generic because the VDAF spec spells out the nonce size; any other choice of nonce size would not be compliant.) 

I tacked on a commit to see what this would look like. Overall the change isn't too bad. I would say the only major downside is that `fieldvec_roundtrip_test` now needs `VERIFY_KEY_SIZE` and `NONCE_SIZE` as parameters. To me this makes the code harder to read.

cjpatton: I've reverted the commit with this change. For the record, here is the patch:
```
diff --git a/src/vdaf.rs b/src/vdaf.rs
index 5c3beac..c95a39f 100644
--- a/src/vdaf.rs
+++ b/src/vdaf.rs
@@ -126,7 +126,7 @@ impl<F: FieldElement, const L: usize> Encode for Share<F, L> {
 /// The base trait for VDAF schemes. This trait is inherited by traits [`Client`], [`Aggregator`],
 /// and [`Collector`], which define the roles of the various parties involved in the execution of
 /// the VDAF.
-pub trait Vdaf: Clone + Debug {
+pub trait Vdaf<const VERIFY_KEY_SIZE: usize, const NONCE_SIZE: usize>: Clone + Debug {
     /// Algorithm identifier for this VDAF.
     const ID: u32;

@@ -163,7 +163,9 @@ pub trait Vdaf: Clone + Debug {
 }

 /// The Client's role in the execution of a VDAF.
-pub trait Client<const NONCE_SIZE: usize>: Vdaf {
+pub trait Client<const VERIFY_KEY_SIZE: usize, const NONCE_SIZE: usize>:
+    Vdaf<VERIFY_KEY_SIZE, NONCE_SIZE>
+{
     /// Shards a measurement into a public share and a sequence of input shares, one for each
     /// Aggregator.
     fn shard(
@@ -174,7 +176,9 @@ pub trait Client<const NONCE_SIZE: usize>: Vdaf {
 }

 /// The Aggregator's role in the execution of a VDAF.
-pub trait Aggregator<const VERIFY_KEY_SIZE: usize, const NONCE_SIZE: usize>: Vdaf {
+pub trait Aggregator<const VERIFY_KEY_SIZE: usize, const NONCE_SIZE: usize>:
+    Vdaf<VERIFY_KEY_SIZE, NONCE_SIZE>
+{
     /// State of the Aggregator during the Prepare process.
     type PrepareState: Clone + Debug;

@@ -231,7 +235,9 @@ pub trait Aggregator<const VERIFY_KEY_SIZE: usize, const NONCE_SIZE: usize>: Vda
 }

 /// The Collector's role in the execution of a VDAF.
-pub trait Collector: Vdaf {
+pub trait Collector<const VERIFY_KEY_SIZE: usize, const NONCE_SIZE: usize>:
+    Vdaf<VERIFY_KEY_SIZE, NONCE_SIZE>
+{
     /// Combines aggregate shares into the aggregate result.
     fn unshard<M: IntoIterator<Item = Self::AggregateShare>>(
         &self,
@@ -341,7 +347,7 @@ pub(crate) fn run_vdaf<V, M, const L: usize>(
     measurements: M,
 ) -> Result<V::AggregateResult, VdafError>
 where
-    V: Client<16> + Aggregator<L, 16> + Collector,
+    V: Client<L, 16> + Aggregator<L, 16> + Collector<L, 16>,
     M: IntoIterator<Item = V::Measurement>,
 {
     use rand::prelude::*;
@@ -407,7 +413,7 @@ pub(crate) fn run_vdaf_prepare<V, M, const L: usize>(
     input_shares: M,
 ) -> Result<Vec<V::OutputShare>, VdafError>
 where
-    V: Client<16> + Aggregator<L, 16> + Collector,
+    V: Client<L, 16> + Aggregator<L, 16> + Collector<L, 16>,
     M: IntoIterator<Item = V::InputShare>,
 {
     let input_shares = input_shares
@@ -476,10 +482,13 @@ where
 }

 #[cfg(test)]
-fn fieldvec_roundtrip_test<F, V, T>(vdaf: &V, agg_param: &V::AggregationParam, length: usize)
-where
+fn fieldvec_roundtrip_test<F, V, T, const VERIFY_KEY_SIZE: usize, const NONCE_SIZE: usize>(
+    vdaf: &V,
+    agg_param: &V::AggregationParam,
+    length: usize,
+) where
     F: FieldElement,
-    V: Vdaf,
+    V: Vdaf<VERIFY_KEY_SIZE, NONCE_SIZE>,
     T: Encode,
     for<'a> T: ParameterizedDecode<(&'a V, &'a V::AggregationParam)>,
 {
diff --git a/src/vdaf/poplar1.rs b/src/vdaf/poplar1.rs
index d631f76..9115dd7 100644
--- a/src/vdaf/poplar1.rs
+++ b/src/vdaf/poplar1.rs
@@ -298,7 +298,7 @@ impl Decode for Poplar1AggregationParam {
     }
 }

-impl<P: Prg<L>, const L: usize> Vdaf for Poplar1<P, L> {
+impl<P: Prg<L>, const L: usize> Vdaf<L, 16> for Poplar1<P, L> {
     const ID: u32 = 0x00001000;
     type Measurement = IdpfInput;
     type AggregateResult = Vec<u64>;
@@ -313,7 +313,7 @@ impl<P: Prg<L>, const L: usize> Vdaf for Poplar1<P, L> {
     }
 }

-impl<P: Prg<L>, const L: usize> Client<16> for Poplar1<P, L> {
+impl<P: Prg<L>, const L: usize> Client<L, 16> for Poplar1<P, L> {
     fn shard(
         &self,
         input: &IdpfInput,
@@ -600,7 +600,7 @@ impl<P: Prg<L>, const L: usize> Aggregator<L, 16> for Poplar1<P, L> {
     }
 }

-impl<P: Prg<L>, const L: usize> Collector for Poplar1<P, L> {
+impl<P: Prg<L>, const L: usize> Collector<L, 16> for Poplar1<P, L> {
     fn unshard<M: IntoIterator<Item = Poplar1FieldVec>>(
         &self,
         agg_param: &Poplar1AggregationParam,
diff --git a/src/vdaf/prio2.rs b/src/vdaf/prio2.rs
index ade2ed7..53e522d 100644
--- a/src/vdaf/prio2.rs
+++ b/src/vdaf/prio2.rs
@@ -91,7 +91,7 @@ impl Prio2 {
     }
 }

-impl Vdaf for Prio2 {
+impl Vdaf<32, 16> for Prio2 {
     const ID: u32 = 0xFFFF0000;
     type Measurement = Vec<u32>;
     type AggregateResult = Vec<u32>;
@@ -107,7 +107,7 @@ impl Vdaf for Prio2 {
     }
 }

-impl Client<16> for Prio2 {
+impl Client<32, 16> for Prio2 {
     fn shard(
         &self,
         measurement: &Vec<u32>,
@@ -269,7 +269,7 @@ impl Aggregator<32, 16> for Prio2 {
     }
 }

-impl Collector for Prio2 {
+impl Collector<32, 16> for Prio2 {
     fn unshard<M: IntoIterator<Item = AggregateShare<FieldPrio2>>>(
         &self,
         _agg_param: &Self::AggregationParam,
@@ -463,12 +463,20 @@ mod tests {
     #[test]
     fn roundtrip_output_share() {
         let vdaf = Prio2::new(31).unwrap();
-        fieldvec_roundtrip_test::<FieldPrio2, Prio2, OutputShare<FieldPrio2>>(&vdaf, &(), 31);
+        fieldvec_roundtrip_test::<FieldPrio2, Prio2, OutputShare<FieldPrio2>, 32, 16>(
+            &vdaf,
+            &(),
+            31,
+        );
     }

     #[test]
     fn roundtrip_aggregate_share() {
         let vdaf = Prio2::new(31).unwrap();
-        fieldvec_roundtrip_test::<FieldPrio2, Prio2, AggregateShare<FieldPrio2>>(&vdaf, &(), 31);
+        fieldvec_roundtrip_test::<FieldPrio2, Prio2, AggregateShare<FieldPrio2>, 32, 16>(
+            &vdaf,
+            &(),
+            31,
+        );
     }
 }
diff --git a/src/vdaf/prio3.rs b/src/vdaf/prio3.rs
index 42ebac7..eb109a0 100644
--- a/src/vdaf/prio3.rs
+++ b/src/vdaf/prio3.rs
@@ -532,7 +532,7 @@ where
     }
 }

-impl<T, P, const L: usize> Vdaf for Prio3<T, P, L>
+impl<T, P, const L: usize> Vdaf<L, 16> for Prio3<T, P, L>
 where
     T: Type,
     P: Prg<L>,
@@ -711,7 +711,7 @@ impl<F: FftFriendlyFieldElement, const L: usize> ParameterizedDecode<Prio3Prepar
     }
 }

-impl<T, P, const L: usize> Client<16> for Prio3<T, P, L>
+impl<T, P, const L: usize> Client<L, 16> for Prio3<T, P, L>
 where
     T: Type,
     P: Prg<L>,
@@ -1009,7 +1009,7 @@ where
     }
 }

-impl<T, P, const L: usize> Collector for Prio3<T, P, L>
+impl<T, P, const L: usize> Collector<L, 16> for Prio3<T, P, L>
 where
     T: Type,
     P: Prg<L>,
@@ -1509,13 +1509,21 @@ mod tests {
     #[test]
     fn roundtrip_output_share() {
         let vdaf = Prio3::new_aes128_count(2).unwrap();
-        fieldvec_roundtrip_test::<Field64, Prio3Aes128Count, OutputShare<Field64>>(&vdaf, &(), 1);
+        fieldvec_roundtrip_test::<Field64, Prio3Aes128Count, OutputShare<Field64>, 16, 16>(
+            &vdaf,
+            &(),
+            1,
+        );

         let vdaf = Prio3::new_aes128_sum(2, 17).unwrap();
-        fieldvec_roundtrip_test::<Field128, Prio3Aes128Sum, OutputShare<Field128>>(&vdaf, &(), 1);
+        fieldvec_roundtrip_test::<Field128, Prio3Aes128Sum, OutputShare<Field128>, 16, 16>(
+            &vdaf,
+            &(),
+            1,
+        );

         let vdaf = Prio3::new_aes128_histogram(2, &[1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11]).unwrap();
-        fieldvec_roundtrip_test::<Field128, Prio3Aes128Histogram, OutputShare<Field128>>(
+        fieldvec_roundtrip_test::<Field128, Prio3Aes128Histogram, OutputShare<Field128>, 16, 16>(
             &vdaf,
             &(),
             12,
@@ -1525,21 +1533,21 @@ mod tests {
     #[test]
     fn roundtrip_aggregate_share() {
         let vdaf = Prio3::new_aes128_count(2).unwrap();
-        fieldvec_roundtrip_test::<Field64, Prio3Aes128Count, AggregateShare<Field64>>(
+        fieldvec_roundtrip_test::<Field64, Prio3Aes128Count, AggregateShare<Field64>, 16, 16>(
             &vdaf,
             &(),
             1,
         );

         let vdaf = Prio3::new_aes128_sum(2, 17).unwrap();
-        fieldvec_roundtrip_test::<Field128, Prio3Aes128Sum, AggregateShare<Field128>>(
+        fieldvec_roundtrip_test::<Field128, Prio3Aes128Sum, AggregateShare<Field128>, 16, 16>(
             &vdaf,
             &(),
             1,
         );

         let vdaf = Prio3::new_aes128_histogram(2, &[1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11]).unwrap();
-        fieldvec_roundtrip_test::<Field128, Prio3Aes128Histogram, AggregateShare<Field128>>(
+        fieldvec_roundtrip_test::<Field128, Prio3Aes128Histogram, AggregateShare<Field128>, 16, 16>(
             &vdaf,
             &(),
             12,
```

PR #456, opened 2023-02-11
Align domain separation with draft

cjpatton:
Based on #455 (merge that first).
Partially addresses #454.

Update the `Prg` API by splitting the `info` string into `custom` and `binder`, as required by VDAF-04. Update `Prio3` and `Poplar1` to use the new API.

While at it, format each `custom` and `binder` string as specified in the draft. This is a breaking change, so ignore the test vector tests as well.


src/vdaf/prg.rs
```diff
@@ -124,8 +124,10 @@ pub struct PrgAes128(Cmac<Aes128>);
 impl Prg<16> for PrgAes128 {
     type SeedStream = SeedStreamAes128;
 
-    fn init(seed_bytes: &[u8; 16]) -> Self {
-        Self(Cmac::new_from_slice(seed_bytes).unwrap())
+    fn init(seed_bytes: &[u8; 16], custom: &[u8]) -> Self {
+        let mut mac = Cmac::new_from_slice(seed_bytes).unwrap();
+        mac.update(custom);
```
divergentdave, line 129: Should we length-delimit `custom` here? I suppose that PrgAes128 is no longer constrained by the draft. I think all of our customization strings in practice are eight bytes, but this could be a sharp edge if we change that before this updated PrgAes128 goes out of use.

divergentdave (approved):

src/vdaf/prg.rs
cjpatton: Yeah I think that's probably a good idea. I was trying to maintain backwards compatibility with the PrgAes128 unit tests in the reference implementation, but I think safety is more important.

cjpatton: https://github.com/cfrg/draft-irtf-cfrg-vdaf/pull/157

src/prng.rs
```diff
@@ -219,19 +219,20 @@ mod tests {
             .collect()
     }
 
+    #[ignore]
```
cjpatton, line 222: Reviewer note: We can either update this test or replace it with one for SHA-3. What would folks prefer?

divergentdave: I think taking a TODO to swap in the SHA-3 test once we implement its PRG would be best. I could also re-run the brute-force search in short order if need be, I have the script for it lying around still.

divergentdave (approved):

src/prng.rs
cjpatton: Filed #465 to track this.

PR #471, opened 2023-02-28
Implement PrgSha3

divergentdave:
This implements PrgSha3, but doesn't wire it up to VDAFs yet. It is tested against the test vector from cfrg/draft-irtf-cfrg-vdaf#164.

FWIW, I rolled the transitive dependency on `keccak` from 0.1.3 back to 0.1.2 in the lock file to get things unblocked, because version 0.1.3 introduces an ARM assembly backend that I didn't want to audit right away. The lock file choices won't affect downstream consumers of the library, and we can audit the 0.1.2-0.1.3 diff at a later date if need be.


src/vdaf/prg.rs
```diff
@@ -173,6 +178,50 @@ impl Debug for SeedStreamAes128 {
     }
 }
 
+/// The PRG based on SHA-3 as specified in [[draft-irtf-cfrg-vdaf-04]].
+///
+/// [draft-irtf-cfrg-vdaf-03]: https://datatracker.ietf.org/doc/draft-irtf-cfrg-vdaf/04/
```
tgeoghegan, line 183:
```suggestion
/// [draft-irtf-cfrg-vdaf-04]: https://datatracker.ietf.org/doc/draft-irtf-cfrg-vdaf/04/
```

src/vdaf/prg.rs
```diff
@@ -262,4 +311,29 @@ mod tests {
...
+            serde_json::from_str(include_str!("test_vec/04/PrgSha3.json")).unwrap();
+        let mut prg = PrgSha3::init(&t.seed.try_into().unwrap(), &t.custom);
+        prg.update(&t.binder);
+
+        assert_eq!(
+            prg.clone().into_seed(),
+            Seed(t.derived_seed.try_into().unwrap())
+        );
+
+        let mut bytes = std::io::Cursor::new(t.expanded_vec_field128.as_slice());
```
tgeoghegan, line 327: nit: could we `use std::io::Cursor` above?

tgeoghegan (approved):
VDAF-04 also removes `PrgAes128`. Is it being removed from `prio` in a later PR?

divergentdave:
I plan to cut the public APIs and tests over from PrgAes128 to PrgSha3 in a subsequent PR, I could remove the old PRG then.

cjpatton (approved):

PR #472, opened 2023-02-28
Prio3: Add public share, replacing joint rand hint

divergentdave:
This implements a change in Prio3 as of VDAF-04, removing joint randomness hints from input shares, and instead putting every aggregator's joint randomness part in the public share.

This PR also updates the parsing of Prio3 JSON test vectors to match the latest files, (including parsing the public share) though the related tests aren't ready to be re-enabled without other changes. Additionally, the "input share" inside `Prio3InputShare` is renamed to "measurement share", following a change that was made at the same time in VDAF.

### DIFF
--- a/src/vdaf/prio3.rs
+++ b/src/vdaf/prio3.rs
@@ -915,6 +915,13 @@ where
             }
             let own_joint_rand_part = joint_rand_part_prg.into_seed();
 
+            // Make an iterator over the joint randomness parts, but use this aggregator's
+            // contribution, computed from the input share, in lieu of the the corresponding part
+            // from the public share.
+            //
+            // The locally computed part should match the part from the public share for honestly
+            // generated reports. If they do not match, the joint randomness seed check during the
+            // next round of preparation should fail.
             let corrected_joint_rand_parts = public_share
                 .joint_rand_parts
                 .iter()
--- a/src/vdaf/prio3.rs
+++ b/src/vdaf/prio3.rs
@@ -396,7 +396,7 @@ where
         } else {
             None
         };
-        let mut leader_meas_share = encoded_measurement.clone();
+        let mut leader_measurement_share = encoded_measurement.clone();
         for agg_id in 1..num_aggregators {
             let helper = HelperShare::from_rand_source(rand_source)?;
             let mut joint_rand_part_prg = P::init(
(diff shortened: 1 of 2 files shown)


tgeoghegan (approved):

src/vdaf/prio3.rs
```diff
@@ -562,37 +575,70 @@ where
     }
 }
 
-/// Message sent by the [`Client`](crate::vdaf::Client) to each
-/// [`Aggregator`](crate::vdaf::Aggregator) during the Sharding phase.
+/// Message broadcast by the [`Client`] to every [`Aggregator`] during the Sharding phase.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct Prio3PublicShare<const L: usize> {
+    /// Contributions to the joint randomness from every aggregator's share.
+    joint_rand_parts: Vec<Seed<L>>,
```
cjpatton, line 582: nit: An `Option` would be slightly more aligned with semantics of optional fields elsewhere in this code. Unless you have reason to think has a hidden cost, I would suggest using `Option` her for consistency.
```suggestion
    joint_rand_parts: Option<Vec<Seed<L>>>,
```

src/vdaf/prio3.rs
```diff
@@ -862,23 +903,20 @@ where
...
-            let joint_rand_seed = Self::derive_joint_rand_seed(
-                hints[..agg_id as usize]
-                    .iter()
-                    .chain(std::iter::once(&joint_rand_part))
-                    .chain(hints[agg_id as usize..].iter()),
-            );
+            let own_joint_rand_part = joint_rand_part_prg.into_seed();
+
+            let mut corrected_joint_rand_parts = public_share.joint_rand_parts.clone();
+            corrected_joint_rand_parts[agg_id as usize] = own_joint_rand_part.clone();
```
cjpatton, line 917: We should be able to avoid these clones. Do you think doing so would be worth the effort? The downside would be less readable code I think.

cjpatton (commented):
Looks good overall, just a few comments.

src/vdaf/prio3.rs
```diff
@@ -913,10 +915,21 @@ where
             }
             let own_joint_rand_part = joint_rand_part_prg.into_seed();
 
-            let mut corrected_joint_rand_parts = public_share.joint_rand_parts.clone();
-            corrected_joint_rand_parts[agg_id as usize] = own_joint_rand_part.clone();
-
-            let joint_rand_seed = Self::derive_joint_rand_seed(corrected_joint_rand_parts.iter());
```
cjpatton: It's hard to tell (for me anyway) if this this code is correct. To make this readable to someone who doesn't know rust that well or doesn't know the spec, a comment would be useful.

The goal here is produce an iterator over `public_share.joint_rand_parts` in which the `agg_id`-th part is replaced by the properly derived value.

cjpatton (changes requested):

PR #473, opened 2023-02-28
Poplar1 nonce updates

divergentdave:
This makes two related changes to Poplar1, from VDAF-04. Since the nonce is now a fixed size, we drop a length prefix from an internal binder string. The nonce is additionally fed into the correlated randomness PRGs' binder strings as a defense in depth measure.


cjpatton (approved):
🤯

tgeoghegan (approved):

PR #475, opened 2023-02-28
Cut over VDAFs and tests to PrgSha3

divergentdave:
This cuts over the main public APIs to use the new PRG, and does the same the benchmarks and most (non-PRG specific) tests.

### DIFF
--- a/src/vdaf/poplar1.rs
+++ b/src/vdaf/poplar1.rs
@@ -10,7 +10,7 @@ use crate::{
     idpf::{self, IdpfInput, IdpfOutputShare, IdpfPublicShare, IdpfValue, RingBufferCache},
     prng::Prng,
     vdaf::{
-        prg::{CoinToss, Prg, PrgAes128, Seed, SeedStream},
+        prg::{CoinToss, Prg, PrgSha3, Seed, SeedStream},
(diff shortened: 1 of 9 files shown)


src/vdaf/poplar1.rs
```diff
@@ -28,22 +28,29 @@ const DST_CORR_INNER: u16 = 2;
...
         }
     }
 }
 
+impl Poplar1<PrgSha3, 16> {
+    /// Create an instance of [`Poplar1`] using [`PrgSha3`]. The caller provides the bit length of
+    /// each measurement (`BITS` as defined in the [[draft-irtf-cfrg-vdaf-03]]).
+    ///
+    /// [draft-irtf-cfrg-vdaf-03]: https://datatracker.ietf.org/doc/draft-irtf-cfrg-vdaf/03/
+    pub fn new_sha3(bits: usize) -> Self {
```
divergentdave, line 49: This is the most awkward naming choice I had to make -- I wanted to have one generic constructor, for BYO-PRG consumers, and one non-generic constructor, for general convenience, but they can't both be `new()`. Hopefully the parallelism break isn't too bad.

cjpatton: This is a little awkward, but not too bad IMO.

cjpatton (approved):

tgeoghegan (approved):

Issue #477, opened 2023-02-28
VDAF-05 changes


divergentdave:
- [ ] Little-endian field encoding (see cfrg/draft-irtf-cfrg-vdaf#175)
- [ ] Remaining codec impls for Poplar1 (note that the prepare message will change in VDAF-05, see cfrg/draft-irtf-cfrg-vdaf#174)
- [ ] Remove `RandSource` type alias, and instead pass byte slices of randomness to non-deterministic functions (see cfrg/draft-irtf-cfrg-vdaf#173)
- [ ] Tests for IDPF and Poplar1 using test vectors

PR #495, opened 2023-03-07
VDAF-05: Use little-endian byte-order for fields

cjpatton:
Partially addresses #477.

As required by the latest draft, switch the byte order of the fields to use little-endian. Accordingly, remove `Field32` (this now has the same byte order as `FieldPrio2`) and drop the big-endian byte order option from the `make_field()` macro.

### DIFF
--- a/src/field/field255.rs
+++ b/src/field/field255.rs
@@ -24,10 +24,10 @@ use subtle::{
     Choice, ConditionallySelectable, ConstantTimeEq, ConstantTimeGreater, ConstantTimeLess,
 };
 
-// `python3 -c "print(', '.join(hex(x) for x in (2**255-19).to_bytes(32, 'big')))"`
-const MODULUS_BIG_ENDIAN: [u8; 32] = [
-    0x7f, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff,
-    0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xed,
+// `python3 -c "print(', '.join(hex(x) for x in (2**255-19).to_bytes(32, 'little')))"`
+const MODULUS_LITTLE_ENDIAN: [u8; 32] = [
+    0xed, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff,
+    0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0x7f,
 ];
 
 /// `GF(2^255 - 19)`, a 255-bit field.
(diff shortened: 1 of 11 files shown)


src/field/field255.rs
```diff
@@ -436,20 +431,9 @@ mod tests {
         assert_matches!(
             Field255::try_from_bytes(
                 &[
-                    0x7f, 0xff, 0xff, 0xff, 0xff, 0xff, 0xfe, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff,
+                    0xec, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff,
                     0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff,
-                    0xff, 0xff, 0xff, 0xff, 0xff, 0xee,
```
cjpatton, line 441: To be honest I"m not sure why this was supposed to ever pass. As far as I can tell the value is larger than the modulus?

src/field/field255.rs
```diff
@@ -515,23 +499,19 @@ mod tests {
...
             format!("{:?}", Field255::zero()),
             "0x0000000000000000000000000000000000000000000000000000000000000000"
         );
         assert_eq!(
             format!("{:?}", Field255::one()),
-            "0x0000000000000000000000000000000000000000000000000000000000000001"
-        );
-        assert_eq!(
-            format!("{:?}", -Field255::one()),
-            "0x7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffec"
```
cjpatton, line 534: It looks like this was redundant (line 522 in the previous version of the file)

src/prng.rs
```diff
@@ -223,18 +223,20 @@ mod tests {
     fn rejection_sampling_test_vector() {
         // These constants were found in a brute-force search, and they test that the PRG performs
         // rejection sampling correctly when raw cSHAKE128 output exceeds the prime modulus.
-        let seed = Seed::get_decoded(&[0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0xd5]).unwrap();
-        let expected = Field96::from(74403028385650568506271441532);
+        let seed =
+            Seed::get_decoded(b"\x23\x1c\x40\x0d\xcb\xaf\xce\x34\x5e\xfd\x3c\xa7\x79\x65\xee\x06")
+                .unwrap();
+        let expected = Field64::from(13681157193520586550);
```
cjpatton, line 229: Now matches rejection test case from the reference code.

src/vdaf/prio3_test.rs
```diff
@@ -132,6 +132,7 @@ fn check_prep_test_vec<M, T, P, const SEED_SIZE: usize>(
     }
 }
 
+#[ignore] // TODO(issue #477)
```
cjpatton, line 135: @tgeoghegan Let me know if you want me to remove these TODOs.

src/field/field255.rs
divergentdave: The deleted test had an 0xfe n one of the higher words, so it wasn't more than the modulus.

cjpatton: Ohhh missed that :) Added the test back.

src/vdaf/prio3_test.rs
tgeoghegan: I'm happy enough with the issue references

tgeoghegan (approved):
SGTM, David's review is more pertinent than mine though.

src/field/field255.rs
```diff
@@ -51,16 +51,17 @@ impl Field255 {
...
         if mask_top_bit {
-            value[0] &= 0b0111_1111;
+            value[31] &= 0b0111_1111;
         }
 
-        // Walk through the big-endian encoding of the provided value, and identify whether the
-        // first byte that differs from the field's modulus is less than the corresponding byte or
-        // greater than the corresponding byte, in constant time. (Or whether the provided value is
-        // equal to the field modulus)
+        // Walk through the bytes of encoding of the provided value from most significant to least,
```
divergentdave, line 57:
```suggestion
        // Walk through the bytes of the provided value from most significant to least,
```

src/field/field255.rs
divergentdave: Yeah, probably a holdover from when the Debug and Display implementations differed. LGTM.

divergentdave (approved):

PR #496, opened 2023-03-08
Add length hints to encoding traits

divergentdave:
This adds new methods to the encoding traits that provide a size hint to speed up encoding, by right-sizing buffers. `encoded_len()` and `encoded_len_with_param()` are new provided methods that return `Option<usize>`, and their default implementations return `None`, so that the hints can be rolled out gracefully. All first-party types override this to provide accurate lengths. The `get_encoded_with_param()` method is changed to get a size hint first, and use `Vec::with_capacity()` if a size is available. This avoids subsequent reallocations and copies for larger messages. I added a benchmark for round-tripping a very small Poplar public share, and this shows a 3% improvement (albeit on a small number of total cycles). I expect that this should pay off more with bigger Poplar1 instances, and third-party DAP message types that reuse these traits.

### DIFF
--- a/src/codec.rs
+++ b/src/codec.rs
@@ -103,6 +103,11 @@ pub trait Encode {
     fn get_encoded(&self) -> Vec<u8> {
         self.get_encoded_with_param(&())
     }
+
+    /// Returns a hint indicating how many bytes will be required to encode this value.
+    fn encoded_len(&self) -> Option<usize> {
+        None
+    }
 }
 
 /// Describes how to encode objects into a byte sequence.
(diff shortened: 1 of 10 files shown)


src/codec.rs
```diff
@@ -103,6 +103,11 @@ pub trait Encode {
     fn get_encoded(&self) -> Vec<u8> {
         self.get_encoded_with_param(&())
     }
+
+    /// Returns a hint indicating how many bytes will be required to encode this value.
```
cjpatton, line 107: Justifying `Option<usize>` over `usize` as the return type will be useful IMO.

```suggestion
    /// Returns a hint indicating how many bytes will be required to encode this value.
    /// No hint is returned by default.
```

src/codec.rs
```diff
@@ -114,10 +119,19 @@ pub trait ParameterizedEncode<P> {
...
+        let mut ret = if let Some(length) = self.encoded_len_with_param(encoding_parameter) {
+            Vec::with_capacity(length)
+        } else {
+            Vec::new()
+        };
         self.encode_with_param(encoding_parameter, &mut ret);
         ret
     }
+
+    /// Returns a hint indicating how many bytes will be required to encode this value.
```
cjpatton, line 131: Same here.

cjpatton (approved):
Wow, this is a really nice idea! I particularly like the side-effect that `get_encoded()` now gets to predict the buffer size.

tgeoghegan (approved):

PR #498, opened 2023-03-08
Backport #497 and bump crate version to 0.11.1

tgeoghegan:
I'd like to release a `prio` with #497, because I'll need it soon for https://github.com/divviup/janus/issues/994.


divergentdave:
The main branch already has one VDAF-05 change on it, so we'd have to start a `release/0.11` before that, backport the `Clone` change onto it, and do a release from there.

divergentdave (approved):

PR #499, opened 2023-03-08
Explicltly pass random bytes to Prio3 and IDPF

### LINKED ISSUE #466: Optimization: skip unused joint randomness part computations
I noticed that sharding for Prio3Count will perform some unnecessary PRG derivations to compute joint randomness parts, even though they won't be used. We could speed this up by moving more behind `joint_rand_len() > 0` conditionals. `prepare_init()` does this well already.

divergentdave:
This adopts an internal API change in VDAF-05, explicitly passing random bytes to sharding and key generation functions. The `RandSource` type alias is removed. The Prio3 sharding method and the IDPF key generation function were updated; the Poplar1 sharding method doesn't support test vectors yet, so I left it for now. I had to rearrange Prio3 sharding to skip joint randomness computations when they aren't needed, since the new API doesn't let us get extra random bytes, so this closes #466 as well.

### DIFF
--- a/src/idpf.rs
+++ b/src/idpf.rs
@@ -312,7 +312,7 @@ fn gen_with_random<VI, VL, M: IntoIterator<Item = VI>, P, const SEED_SIZE: usize
     input: &IdpfInput,
     inner_values: M,
     leaf_value: VL,
-    random: &[u8],
+    random: &[[u8; SEED_SIZE]; 2],
 ) -> Result<(IdpfPublicShare<VI, VL, SEED_SIZE>, [Seed<SEED_SIZE>; 2]), VdafError>
 where
     VI: IdpfValue,
@@ -321,15 +321,8 @@ where
 {
     let bits = input.len();
 
-    if random.len() != SEED_SIZE * 2 {
-        return Err(VdafError::Uncategorized(
-            "incorrect random input length".to_string(),
-        ));
-    }
-    let initial_keys: [Seed<SEED_SIZE>; 2] = [
-        Seed::from_bytes(random[..SEED_SIZE].try_into().unwrap()),
-        Seed::from_bytes(random[SEED_SIZE..].try_into().unwrap()),
-    ];
+    let initial_keys: [Seed<SEED_SIZE>; 2] =
+        [Seed::from_bytes(random[0]), Seed::from_bytes(random[1])];
 
     let mut keys = [initial_keys[0].0, initial_keys[1].0];
     let mut control_bits = [Choice::from(0u8), Choice::from(1u8)];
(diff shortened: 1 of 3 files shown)


cjpatton (approved):
This looks OK to me, but it's kind of a shame that the algorithm is more complex. I wonder if there is a way to modify the spec such that the random seeds can just be read from an iterator?

```rust
let rand_seed_iter = random.chunks(SEED_SIZE);
```

divergentdave:
If the spec rearranged the order of the seeds in the coin input, so that they came in the order they would naturally be needed, that would let us simplify things. Interleaving helper measurement share seeds and proof share seeds would help. Interleaving the joint randomness blinds when present would help as well, though we only need the leader's joint randomness blind after we have processed every helper share (and thus know the contents of the leader's measurement share). Putting the leader's joint randomness blind after the helper-related coins would look weird, but it is also the second to last coin used in the proof-of-concept implementation, so I think it would be justified.

src/idpf.rs
```diff
@@ -308,11 +308,11 @@ where
     out
 }
 
-fn gen_with_rand_source<VI, VL, M: IntoIterator<Item = VI>, P, const SEED_SIZE: usize>(
+fn gen_with_random<VI, VL, M: IntoIterator<Item = VI>, P, const SEED_SIZE: usize>(
     input: &IdpfInput,
     inner_values: M,
     leaf_value: VL,
-    rand_source: RandSource,
+    random: &[u8],
```
tgeoghegan, line 315: Feel free to ignore this idea:
```suggestion
    random_1: [u8; SEED_SIZE],
    random_2: [u8; SEED_SIZE],
```

Then I think you could get rid of the check on the length as well as a couple awkward `.try_into().unwrap()`s below.

src/vdaf/prio3.rs
```diff
@@ -375,19 +375,37 @@ where
         prg.into_seed()
     }
 
+    fn random_size(&self) -> usize {
+        if self.typ.joint_rand_len() == 0 {
+            // Two seeds per helper for measurement and proof shares, plus one seed for proving
+            // randomness.
+            (usize::from(self.num_aggregators) * 2 - 1) * SEED_SIZE
```
tgeoghegan, line 382: Lines up better with the comment, IMO:
```suggestion
            ((usize::from(self.num_aggregators) - 1) * 2 + 1) * SEED_SIZE
```

src/vdaf/prio3.rs
```diff
@@ -400,67 +418,79 @@ where
...
-            let helper = HelperShare::from_rand_source(rand_source)?;
-            let mut joint_rand_part_prg = P::init(
-                helper.joint_rand_blind.as_ref(),
-                &Self::custom(DST_JOINT_RAND_PART),
-            );
-            joint_rand_part_prg.update(&[agg_id]); // Aggregator ID
-            joint_rand_part_prg.update(nonce);
-
+            let measurement_share_seed = random_seeds.next().unwrap().try_into().unwrap();
+            let proof_share_seed = random_seeds.next().unwrap().try_into().unwrap();
```
tgeoghegan, line 422: Some comments on unwrap safety would be nice.

tgeoghegan (approved):
Some nits, nothing blocking.

PR #505, opened 2023-03-10
Harmonize bit length across Sum, SumVec, Average

tgeoghegan:
`Prio3SumVec::new_sum_vec` and `Prio3Sum::new_sum` using different types to represent bit counts is awkward and surprising to API clients, and `Prio3Sum::new_sum` would immediately cast the value to `usize` anyway. While we're at it, make `Prio3Average` consistent, too, and fix a doccomment on `Prio3SumVec` left over from when it was `Prio3Count`.

### DIFF
--- a/src/vdaf/prio3.rs
+++ b/src/vdaf/prio3.rs
@@ -80,8 +80,8 @@ impl Prio3Count {
     }
 }
 
-/// The count-vector type. Each measurement is a vector of integers in `[0,2)` and the aggregate is
-/// the element-wise sum.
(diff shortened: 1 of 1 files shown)


tgeoghegan:
This constitutes a breaking API change (though I think `u32` can harmlessly be upgraded to `usize`, unless someone out there is on a 16 bit arch) so we should not backport it to `release/0.11`.

cjpatton (approved):
Good idea!

divergentdave (approved):

PR #509, opened 2023-03-10
Finish Poplar1 codec implementations

divergentdave:
This finishes up Encode/Decode/ParameterizedDecode implementations for Poplar1, particularly output shares, aggregate shares, public input shares, and aggregation parameters. The aggregation parameter serialization is somewhat complicated, (it reverses bits, then reverses bytes, to get the ordering and justification of packed prefixes right) so we may want to revisit the implementation and specification in the future to enable simplifications. I also changed the semantics of `IdpfInput::from_bytes()` and `IdpfInput::to_bytes()`, as I think I got this wrong the first time around. They now align with how candidate prefixes get encoded in the aggregation parameter (assuming they're byte-aligned to begin with).

### DIFF
--- a/src/vdaf/poplar1.rs
+++ b/src/vdaf/poplar1.rs
@@ -396,9 +396,6 @@ impl Aggregatable for Poplar1FieldVec {
 ///
 /// This includes an indication of what level of the IDPF tree is being evaluated and the set of
 /// prefixes to evaluate at that level.
-//
-// TODO(cjpatton) spec: Make sure repeated prefixes are disallowed. To make this check easier,
-// consider requring the prefixes to be in lexicographic order.
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct Poplar1AggregationParam {
     level: u16,
@@ -414,10 +411,6 @@ impl Poplar1AggregationParam {
     /// * The prefixes have different lengths (they must all be the same).
     /// * The prefixes are longer than 2^16 bits.
     /// * There are more than 2^32 - 1 prefixes.
-    //
-    // TODO spec: Ensure that prefixes don't repeat. To make this check easier, consider requiring
-    // them to appear in alphabetical order.
-    // https://github.com/cfrg/draft-irtf-cfrg-vdaf/issues/134
     pub fn try_from_prefixes(prefixes: Vec<IdpfInput>) -> Result<Self, VdafError> {
         if prefixes.is_empty() {
             return Err(VdafError::Uncategorized(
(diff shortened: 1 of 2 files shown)


src/vdaf/poplar1.rs
```diff
@@ -1343,5 +1421,122 @@ mod tests {
...
+        .unwrap();
+        assert_eq!(
+            agg_param.get_encoded().len(),
+            agg_param.encoded_len().unwrap()
+        );
+    }
+
+    #[test]
+    fn round_trip_agg_param() {
+        // Sage statements used to generate these test cases are given in comments.
```
cjpatton, line 1448:
```suggestion
        // Segments of the reference code (https://github.com/cfrg/draft-irtf-cfrg-vdaf/tree/main/poc) used to generate these test cases are given in comments.
```

cjpatton (approved):

src/vdaf/poplar1.rs
```diff
@@ -394,9 +399,9 @@ impl Aggregatable for Poplar1FieldVec {
 //
 // TODO(cjpatton) spec: Make sure repeated prefixes are disallowed. To make this check easier,
 // consider requring the prefixes to be in lexicographic order.
```
tgeoghegan, line 401: The checks on L439 suggest this spec change did happen. Let's delete this comment.

src/vdaf/poplar1.rs
```diff
@@ -407,6 +412,8 @@ impl Poplar1AggregationParam {
     ///
     /// * The list of prefixes is empty.
     /// * The prefixes have different lengths (they must all be the same).
+    /// * The prefixes are longer than 2^16 bits.
+    /// * There are more than 2^32 - 1 prefixes.
     //
     // TODO spec: Ensure that prefixes don't repeat. To make this check easier, consider requiring
     // them to appear in alphabetical order.
```
tgeoghegan, line 419: Based on the checks on L439-449, the spec must have been updated to do this. Can this comment be deleted?

src/vdaf/poplar1.rs
```diff
@@ -445,14 +470,70 @@ impl Poplar1AggregationParam {
...
+            .collect::<BitVec<u8, Lsb0>>();
+        packed.set_uninitialized(false);
+        let mut packed = packed.into_vec();
+        packed.reverse();
+        bytes.append(&mut packed);
+    }
+
+    fn encoded_len(&self) -> Option<usize> {
+        let packed_bit_count = (usize::from(self.level) + 1) * self.prefixes.len();
+        Some(6 + (packed_bit_count + 7) / 8)
```
tgeoghegan, line 507: I think a comment explaining at least `6` would be helpful.

tgeoghegan (approved):
nits about comments

PR #510, opened 2023-03-10
prg: Implement PrgFixedKeyAes128 as specified in VDAF-05

cjpatton:


src/vdaf/prg.rs
```diff
@@ -232,6 +237,116 @@ impl SeedStream for SeedStreamSha3 {
...
+impl SeedStream for SeedStreamFixedKeyAes128 {
+    fn fill(&mut self, buf: &mut [u8]) {
+        let next_length_consumed = self.length_consumed + buf.len();
+        let mut offset = self.length_consumed & 0b1111;
+        let mut index = 0;
+        let mut block = Block::from([0; 16]);
+
+        // NOTE(cjpatton) We might be able to speed this up by unrolling this loop and encrypting
+        // multiple blocks at the same time via `self.cipher.encrypt_blocks()`.
+        for block_counter in self.length_consumed >> 4..next_length_consumed >> 4 {
```
divergentdave, line 300: This loop incorrectly does nothing if the stream pointer doesn't advance to the next block. The IDPF's `extend()` calls this with a length of 1, which hits this issue.

It occurs to me that we could use RustCrypto's traits for buffer management to implement this, but that sounds like an entire project unto itself.

cjpatton: Fixed, and added a test for making sure a buffer gets filled even if it's less than a block.

divergentdave (approved):

src/vdaf/prg.rs
```diff
@@ -232,6 +237,116 @@ impl SeedStream for SeedStreamSha3 {
...
+        let mut index = 0;
+        let mut block = Block::from([0; 16]);
+
+        // NOTE(cjpatton) We might be able to speed this up by unrolling this loop and encrypting
+        // multiple blocks at the same time via `self.cipher.encrypt_blocks()`.
+        for block_counter in self.length_consumed >> 4..(next_length_consumed >> 4) + 1 {
+            block.clone_from(&self.base_block);
+            for (b, i) in block
+                .iter_mut()
+                .take(std::mem::size_of::<usize>())
```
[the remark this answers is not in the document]
cjpatton: Removed the .take(). and changed type of length_consumed to u64.

src/vdaf/prg.rs
```diff
@@ -232,6 +237,112 @@ impl SeedStream for SeedStreamSha3 {
...
+impl SeedStream for SeedStreamFixedKeyAes128 {
+    fn fill(&mut self, buf: &mut [u8]) {
+        let next_length_consumed = self.length_consumed + u64::try_from(buf.len()).unwrap();
+        let mut offset = usize::try_from(self.length_consumed % 16).unwrap();
+        let mut index = 0;
+        let mut block = Block::from([0; 16]);
+
+        // NOTE(cjpatton) We might be able to speed this up by unrolling this loop and encrypting
+        // multiple blocks at the same time via `self.cipher.encrypt_blocks()`.
+        for block_counter in self.length_consumed / 16..(next_length_consumed / 15) + 16 {
```
divergentdave, line 300: This has a typo. I presume the test vector test didn't catch it because we're not sampling much data out of the PRG, so all calls fit within 16 blocks. (as before, extra outer loop iterations wind up as no-ops because `read` is 0 once the buffer's full)
```suggestion
        for block_counter in self.length_consumed / 16..(next_length_consumed + 15) / 16 {
```

divergentdave (changes requested):

divergentdave (approved):

PR #520, opened 2023-03-16
Explicitly specify a trait to fix a Chalk issue

divergentdave:
Rust-analyzer was tripping over this line, presumably because Chalk incorrectly inferred this was `KeyInit::new()`, which takes one argument, instead of `KeyIvInit::new()`. We can explicitly specify the trait in this call to resolve it.

### DIFF
--- a/src/vdaf/prg.rs
+++ b/src/vdaf/prg.rs
@@ -166,7 +166,7 @@ pub struct SeedStreamAes128(Ctr64BE<Aes128>);
 #[cfg(feature = "crypto-dependencies")]
 impl SeedStreamAes128 {
     pub(crate) fn new(key: &[u8], iv: &[u8]) -> Self {
-        SeedStreamAes128(Ctr64BE::<Aes128>::new(key.into(), iv.into()))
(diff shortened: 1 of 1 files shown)


tgeoghegan (approved):
This change LGTM but I don't think I understand the relationship between Chalk and rust-analyzer. In particular, my rust-anaylzer doesn't complain about this line.

divergentdave:
Chalk is the experimental trait solver that rust-analyzer uses, it's analysis is more powerful in some respects, but it is still incomplete and in development. I have the following in `.vscode/settings.json`, which may explain the difference in what we're seeing. (because `KeyInit` is only in scope when `feature = "experimental"` is satisfied)

```json
{
    "rust-analyzer.cargo.features": [
        "experimental",
        "multithreaded",
        "prio2"
    ]
}
```

tgeoghegan:
Actually, I just needed to be patient and let rust-analyzer crunch through libprio-rs, and the error you describe eventually showed up. Thanks for the explanation!.

PR #523, opened 2023-03-16
Update documentation for VDAF-05

divergentdave:
Stacked on #522.

This updates document references for the new draft as appropriate, updates the version table in the README, and adds a reference to DPRS23 to the top-level docs.

### DIFF
--- a/src/vdaf/poplar1.rs
+++ b/src/vdaf/poplar1.rs
@@ -1,8 +1,8 @@
 // SPDX-License-Identifier: MPL-2.0
 
-//! Work-in-progress implementation of Poplar1 as specified in [[draft-irtf-cfrg-vdaf-04]].
+//! Implementation of Poplar1 as specified in [[draft-irtf-cfrg-vdaf-05]].
 //!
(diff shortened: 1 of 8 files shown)


src/vdaf/poplar1.rs
```diff
@@ -1,8 +1,8 @@
 // SPDX-License-Identifier: MPL-2.0
 
-//! Work-in-progress implementation of Poplar1 as specified in [[draft-irtf-cfrg-vdaf-04]].
+//! Implementation of Poplar1 as specified in [[draft-irtf-cfrg-vdaf-05]].
```
tgeoghegan, line 3: In this spirit, I think we could also remove `(EXPERIMENTAL)` from the `Verifiable Distributed Aggregation Function` header in `README.md`.

tgeoghegan (approved):

src/vdaf/poplar1.rs
divergentdave: Done, I also updated the following paragraph a bit.

PR #539, opened 2023-03-22
Deprecate Field96

### LINKED ISSUE #537: Deprecate Field96
We have five different fields implemented, three from the VDAF document, (Field64, Field128, and Field255) one for Prio v2 (FieldPrio2), and one that never wound up being used (Field96). We could mark this as deprecated and remove it in the future. Its size is such that it's overkill for plain secret sharing, but not big enough for Fiat-Shamir protocols or extractability.

divergentdave:
This deprecates `Field96`, which is not used in any constructions so far. Closes #537.

### DIFF
--- a/src/field.rs
+++ b/src/field.rs
@@ -712,6 +712,10 @@ make_field!(
     8,
 );
 
+/// This nested module is an implementation detail to limit the scope of a module-wide
+/// `allow(deprecated)` attribute. [`Field96`] is marked as deprecated, and deprecation warnings
+/// must be silenced on multiple implementation blocks, and the macro invocation itself that
+/// defines the struct and its implementation.
 mod field96 {
     #![allow(deprecated)]
 
--- a/src/field.rs
+++ b/src/field.rs
@@ -7,13 +7,11 @@
 //! [`FftFriendlyFieldElement`], and have an associated element called the "generator" that
 //! generates a multiplicative subgroup of order `2^n` for some `n`.
 
-#![allow(deprecated)] // for Field96
-
 #[cfg(feature = "crypto-dependencies")]
 use crate::prng::{Prng, PrngError};
 use crate::{
     codec::{CodecError, Decode, Encode},
-    fp::{FP128, FP32, FP64, FP96},
+    fp::{FP128, FP32, FP64},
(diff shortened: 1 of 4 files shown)


src/field.rs
```diff
@@ -7,6 +7,8 @@
 //! [`FftFriendlyFieldElement`], and have an associated element called the "generator" that
 //! generates a multiplicative subgroup of order `2^n` for some `n`.
 
+#![allow(deprecated)] // for Field96
```
cjpatton, line 10: What does the compiler say if we remove this?

src/field.rs
```diff
@@ -713,6 +715,7 @@ make_field!(
 );
 
 make_field!(
+    #[deprecated]
     /// `GF(79228148845226978974766202881)`, a 96-bit field.
```
cjpatton, line 719: Consider explaining why it's deprecated in the docucomment.

src/field.rs
divergentdave: Each use of Field96 or its field in the macro expansion flags a warning. Those can be addressed with method-level `#[allow(deprecated)]` attributes. The one I couldn't figure out how to silence without a module-level attribute was the `Field96` argument to the macro invocation itself. (The macro invocation does not accept an `#[allow]` attribute on it, and anything inside the parentheses becomes a macro argument)

tgeoghegan: I'm a little nervous about silencing deprecation warnings module-wide. If we move the declaration of `Field96` into a sub-module of `field.rs`, can we re-export `Field96` with `pub use crate::field::sub_module::Field96`, and then tag just the sub-module and the `pub use` as `#![allow(deprecated]`?

src/field.rs
```diff
@@ -712,13 +712,42 @@ make_field!(
...
 );
 
-make_field!(
-    /// `GF(79228148845226978974766202881)`, a 96-bit field.
-    Field96,
-    u128,
-    FP96,
-    12,
-);
+mod field96 {
```
tgeoghegan, line 715: Last suggestion: throw a doccomment on this module explaining its purpose so that future us won't inadvertently delete this otherwise useless module.

tgeoghegan (approved):

PR #554, opened 2023-04-05
Provide Encode/Decode implementation to Poplar1PrepareState.

### LINKED ISSUE #553: Serialization of Poplar1PrepareState
We should provide a way to serialize and deserialize a `Poplar1PrepareState` object, for aggregators that run as distributed systems. (even though the VDAF document does not define a serialization format for it) Currently there's no way to construct the struct other than the preparation methods.

I think the serialization should encode both whether it's for an inner level or a leaf level, and whether it's on round one or round two of the sketch, i.e. it should be deserializable without knowing either of those a priori. The prepare state is used as a decoding parameter for prepare messages/message shares, and in those implementations the prepare state is used to select between those four possible cases. If the serialization this library provides doesn't include discriminants for inner/leaf levels and first/second sketch rounds, then we'd have to take them in from decoding parameters, and a distributed application would have to store those alongside the serialized prepare state itself.

branlwyd:
The decoding parameter may seem strange at first, as the VDAF portion is unused; the choice of decoding parameter was made to match that used by Prio3's prepare state.

Closes #553.

### DIFF
--- a/src/vdaf/poplar1.rs
+++ b/src/vdaf/poplar1.rs
@@ -19,7 +19,9 @@ use std::{
     convert::TryFrom,
     fmt::Debug,
     io::{Cursor, Read},
+    iter,
     marker::PhantomData,
+    num::TryFromIntError,
     ops::{Add, AddAssign, Sub},
 };
 use subtle::{Choice, ConditionallyNegatable, ConditionallySelectable, ConstantTimeEq};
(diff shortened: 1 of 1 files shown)


src/vdaf/poplar1.rs
```diff
@@ -1477,6 +1636,115 @@ mod tests {
...
+                        Field255::from(31),
+                    ]),
+                })),
+            ),
+        ] {
+            let encoded_prep_state = prep_state.get_encoded();
+            let decoded_prep_state =
+                Poplar1PrepareState::get_decoded_with_param(&(&vdaf, agg_id), &encoded_prep_state)
+                    .unwrap();
+            assert_eq!(prep_state, decoded_prep_state);
```
divergentdave, line 1744: We should check that encoded_len() is correct as well here

tgeoghegan (approved):

PR #555, opened 2023-04-07
Poplar1: check underflow in Poplar1AggregationParam::try_from_prefixes.

branlwyd:
Previously, if an empty IdpfInput was provided, integer underflow would be encountered on the modified lines; this would cause debug builds to panic, and almost surely return the wrong error (except on platforms where usize is 16 bits, in which case very strange behavior was likely).

Also, update the comment about possible errors to include a few undocumented errors.

### DIFF
--- a/src/vdaf/poplar1.rs
+++ b/src/vdaf/poplar1.rs
@@ -407,8 +407,10 @@ impl Poplar1AggregationParam {
     ///
     /// * The list of prefixes is empty.
     /// * The prefixes have different lengths (they must all be the same).
-    /// * The prefixes are longer than 2^16 bits.
+    /// * The prefixes have length 0, or length longer than 2^16 bits.
     /// * There are more than 2^32 - 1 prefixes.
+    /// * The prefixes are not unique.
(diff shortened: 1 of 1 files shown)


branlwyd:
I notice a few places that compute `poplar1.bits - 1`, which can only underflow if the user constructs the `Poplar1` instance with a `bits` parameter of 0. I think that is considered bad input, so I think it should be checked for at time of construction -- but since the `Poplar1` constructors are infallible, fixing this case to return an error would be a breaking change.

Not sure if anyone cares enough to add guard rails to `bits = 0` -- if so, I don't mind sending a PR, but want to make sure it's not considered noise first.

tgeoghegan (approved):
I think this change is fine, but we might consider using things like [`std::num::NonZeroUsize`](https://doc.rust-lang.org/stable/std/num/struct.NonZeroUsize.html) to enforce and clearly express in the API that zero is not valid. However that wouldn't help much in cases where the value that can be zero comes from `Vec::len` and making the caller instantiate a `NonZeroWhicheverType` is a mild hassle.

PR #560, opened 2023-04-12
CI: Build benchmarks without optimization

divergentdave:
Building benchmarks is currently one of the slower steps in our CI. We can speed this up by building them with the dev profile, instead of release.


divergentdave:
This is cutting between one and three minutes off each build-crate matrix job. (which is the long pole)

tgeoghegan (approved):

PR #562, opened 2023-04-13
Remove SHA3 dependency from crypto-dependencies feature.

simon-friedberger:
Since the SHA3 dependency is pretty minor I want to take it.
If a separate feature, e.g. "sha3-dependency" is preferable just let me know and I will update!


divergentdave:
LGTM. I rebased this for you, since I broke it with a recent merge.

tgeoghegan (approved):

PR #586, opened 2023-05-23
Publish crate as `divviup-github-automation`

tgeoghegan:


inahga (approved):
I notice the secret is not an organization secret. Is there any particular opposition to this? It would make rotating the secret slightly easier--less repos to copy+paste into. Not a big deal in any case.

tgeoghegan:
Good point. I've promoted the secret to an org secret and blessed libprio-rs to access it. I'll do the same for other repos once this is shown to work.

branlwyd (approved):

PR #595, opened 2023-06-07
Clean up debug formatting of Prio3

divergentdave:
This customizes the Debug implementations of all the Type structs, to simplify and shorten their output. All fields used for pre-comupting values are excluded, and for Histogram, the list of bucket boundaries is replaced with just the number of buckets.

### DIFF
--- a/src/flp/types/fixedpoint_l2.rs
+++ b/src/flp/types/fixedpoint_l2.rs
@@ -201,12 +201,12 @@ pub struct FixedPointBoundedL2VecSum<
     gadget1_chunk_len: usize,
 }
 
-impl<
-        T: Fixed,
-        F: FftFriendlyFieldElement,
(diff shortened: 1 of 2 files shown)


src/flp/types/fixedpoint_l2.rs
```diff
@@ -201,6 +201,21 @@ pub struct FixedPointBoundedL2VecSum<
     gadget1_chunk_len: usize,
 }
 
+impl<
+        T: Fixed,
+        F: FftFriendlyFieldElement,
+        SPoly: ParallelSumGadget<F, PolyEval<F>> + Clone,
+        SBlindPoly: ParallelSumGadget<F, BlindPolyEval<F>> + Clone,
```
cjpatton, line 208: nit: elsewhere we're sticking long lists of trait bound in a `where:" block

PR #596, opened 2023-06-08
Bump cargo-vet to 0.7.0

bholley:
- Update cargo-vet to 0.7.0.
- Trust some crate publishers that Mozilla also trusts.
- Import fermyon.


bholley:
See [documentation](https://mozilla.github.io/cargo-vet/trusting-publishers.html) of the new trusted crates feature.

divergentdave (approved):
Thanks, these trusted publishers look like good ideas to me.

PR #602, opened 2023-06-14
Histogram changes for vdaf-06

divergentdave:
This updates Prio3Histogram for draft-06 of VDAF. It takes a number of buckets instead of a list of bucket boundaries, and the semantics of the measurement have similarly changed. I also bumped the crate version and updated the version table in the README. (and updated the Dependabot configuration)


tgeoghegan:
Please make sure to branch `release/0.12` from `main` before merging this.

PR #578, opened 2023-05-18
adding differential privacy to FixedPointBoundedL2VecSum

ooovi:
As announced in #440, we added a feature for our fixed point vector type to add discrete gaussian noise to the aggregation result.

Changes to prio code that is not part of our type, currently all hidden behind the `experimental` feature flag:

- The `Type` trait got an [`add_noise`](https://github.com/dpsa-project/libprio-rs/blob/0fbd24ca05f38cf035e2b47a2646d8efccce6c97/src/flp.rs#L552-L557) function that's the identity unless overridden.
- The `Aggregator` trait got a [`postprocess`](https://github.com/dpsa-project/libprio-rs/blob/0fbd24ca05f38cf035e2b47a2646d8efccce6c97/src/vdaf.rs#L252-L265) function that's the identity unless overridden. The `Aggregator` implementation for `Prio3` [overrides it calling the `add_noise` function](https://github.com/dpsa-project/libprio-rs/blob/0fbd24ca05f38cf035e2b47a2646d8efccce6c97/src/vdaf/prio3.rs#L1149-L1159)
- There's an [FlpError](https://github.com/dpsa-project/libprio-rs/blob/0fbd24ca05f38cf035e2b47a2646d8efccce6c97/src/flp.rs#L107-L110) for errors happening during noising.

Our type [overrides the default implementation of `add_noise`](https://github.com/dpsa-project/libprio-rs/blob/0fbd24ca05f38cf035e2b47a2646d8efccce6c97/src/flp/types/fixedpoint_l2.rs#L561), and adds noise obtained using our [sampler implementation](https://github.com/dpsa-project/libprio-rs/blob/feature-fixedvec-dp/src/flp/types/fixedpoint_l2/noise.rs).

Looking forward to your comments! :)

Co-Authored-By: Olivia <ovi@posteo.de>
Co-Authored-By: Maxim Urschumzew <u.maxim@live.de>

PR #607, opened 2023-06-16
API changes for Aggregator-side differential privacy

tholop:
As discussed yesterday, this PR makes some API changes required to add DP on aggregated shares (aka Server-DP, following the terminology from DPrio (https://petsymposium.org/popets/2023/popets-2023-0086.php). Once we settle on an API we can tackle DP implementations for specific VDAFs, such as https://github.com/divviup/libprio-rs/pull/578.

cc/ @cjpatton

### DIFF
--- a/src/dp.rs
+++ b/src/dp.rs
@@ -1,0 +1,68 @@
+// SPDX-License-Identifier: MPL-2.0
+
+//! Differential privacy (DP) primitives.
+use std::fmt::Debug;
+
+/// Positive rational number to represent DP and noise distribution parameters in protocol messages
+/// and manipulate them without rounding errors.
+#[derive(Clone, Debug)]
+pub struct Rational {
+    /// Numerator.
+    pub numerator: u32,
+    /// Denominator.
+    pub denominator: u32,
+}
+
+/// Marker trait for differential privacy budgets (regardless of the specific accounting method).
+pub trait DifferentialPrivacyBudget {}
+
+/// Marker trait for differential privacy scalar noise distributions.
+pub trait DifferentialPrivacyDistribution {}
+
+/// Zero-concentrated differential privacy (zCDP) budget as defined in [[BS16]].
+///
+/// [BS16]: https://arxiv.org/pdf/1605.02065.pdf
+pub struct ZeroConcentratedDifferentialPrivacyBudget {
+    /// Parameter `epsilon`, using the notation from [[CKS20]] where `rho = (epsilon**2)/2`
+    /// for a `rho`-zCDP budget.
+    ///
+    /// [CKS20]: https://arxiv.org/pdf/2004.00010.pdf
+    pub epsilon: Rational,
+}
+
+impl DifferentialPrivacyBudget for ZeroConcentratedDifferentialPrivacyBudget {}
+
+/// Zero-mean Discrete Gaussian noise distribution.
+///
+/// The distribution is defined over the integers, represented by arbitrary-precision integers.
+#[derive(Clone, Debug)]
+pub struct DiscreteGaussian {
+    /// Standard deviation of the distribution.
+    pub sigma: Rational,
+}
+
+impl DifferentialPrivacyDistribution for DiscreteGaussian {}
+
+/// Strategy to make aggregate shares differentially private, e.g. by adding noise from a specific
+/// type of distribution instantiated with a given DP budget
+pub trait DifferentialPrivacyStrategy {}
+
+/// A zCDP budget used to create a Discrete Gaussian distribution
+pub struct ZCdpDiscreteGaussian {
+    budget: ZeroConcentratedDifferentialPrivacyBudget,
+}
+
+impl DifferentialPrivacyStrategy for ZCdpDiscreteGaussian {}
+
+impl ZCdpDiscreteGaussian {
+    /// Creates a new Discrete Gaussian by following Theorem 4 from [[CKS20]]
+    ///
+    /// [CKS20]: https://arxiv.org/pdf/2004.00010.pdf
+    pub fn create_distribution(&self, sensitivity: Rational) -> DiscreteGaussian {
+        let sigma = Rational {
+            numerator: self.budget.epsilon.denominator * sensitivity.numerator,
+            denominator: self.budget.epsilon.numerator * sensitivity.denominator,
+        };
+        DiscreteGaussian { sigma }
+    }
+}
--- a/src/vdaf.rs
+++ b/src/vdaf.rs
@@ -5,6 +5,8 @@
 //!
 //! [draft-irtf-cfrg-vdaf-06]: https://datatracker.ietf.org/doc/draft-irtf-cfrg-vdaf/06/
 
+#[cfg(feature = "experimental")]
+use crate::dp::DifferentialPrivacyStrategy;
 #[cfg(all(feature = "crypto-dependencies", feature = "experimental"))]
 use crate::idpf::IdpfError;
 use crate::{
@@ -248,6 +250,25 @@ pub trait Aggregator<const VERIFY_KEY_SIZE: usize, const NONCE_SIZE: usize>: Vda
         agg_param: &Self::AggregationParam,
         output_shares: M,
     ) -> Result<Self::AggregateShare, VdafError>;
+}
+
+/// Aggregator that implements differential privacy with Aggregator-side noise addition.
+#[cfg(feature = "experimental")]
+pub trait AggregatorWithNoise<
+    const VERIFY_KEY_SIZE: usize,
+    const NONCE_SIZE: usize,
+    DPStrategy: DifferentialPrivacyStrategy,
+>: Aggregator<VERIFY_KEY_SIZE, NONCE_SIZE>
+{
+    /// Adds noise to an aggregate share such that the aggregate result is differentially private
+    /// as long as one Aggregator is honest.
+    fn add_noise_to_agg_share(
+        &self,
+        dp_strategy: &DPStrategy,
+        agg_param: &Self::AggregationParam,
+        agg_share: &mut Self::AggregateShare,
+        num_measurements: usize,
+    ) -> Result<(), VdafError>;
 }
 
 /// The Collector's role in the execution of a VDAF.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -51,3 +51,6 @@ pub mod test_vector;
 #[cfg_attr(docsrs, doc(cfg(feature = "prio2")))]
 pub mod util;
 pub mod vdaf;
+
+#[cfg(feature = "experimental")]
+pub mod dp;


src/dp.rs
```diff
@@ -0,0 +1,64 @@
+// SPDX-License-Identifier: MPL-2.0
+
+//! Differential privacy (DP) primitives.
+use crate::field::FieldElement;
+use crate::vdaf::prg::Seed;
+use std::fmt::Debug;
+
+// TODO(tholop): We could also use pairs of integers, or even
+// floats to store the DiscreteGaussian DP parameters?
+// Using BigInts might make it harder to communicate `DiscreteGaussian` to the Collector.
```
cjpatton, line 10: What do you mean by communicate here?

src/dp.rs
```diff
@@ -0,0 +1,64 @@
...
+/// and explain that `add_noise_to_agg_share` is in charge of scaling the noise properly.
+pub struct DiscreteGaussian {
+    epsilon: BigRational,
+    delta: BigRational,
+    sigma: BigRational,
+    sensitivity: BigRational,
+}
+
+impl DiscreteGaussian {
+    pub fn new(epsilon: BigRational, delta: BigRational, sensitivity: BigRational) -> Self {
```
cjpatton, line 64: I believe we'll need a docucomment here. This ought to be useful in any case.
```suggestion
    /// Blah blah blah
    pub fn new(epsilon: BigRational, delta: BigRational, sensitivity: BigRational) -> Self {
```

src/dp.rs
```diff
@@ -0,0 +1,64 @@
...
+use num_rational::BigRational;
+
+/// DP-related errors.
+#[derive(Debug, thiserror::Error)]
+pub enum DpError {
+    /// Error while sampling from a specific noise distribution
+    #[error("sampling error")]
+    SamplingError,
+}
+
```
cjpatton, line 20: It occurs to me that this API bears a striking resemblance to `prg::CoinToss`. Can we use that instead? The differences are:
* Here we use `seed`, but `CointToss::sample()` takes a `seed_stream`. The latter probably makes more sense.
* Here we have a `len` field, which is `CoinToss::sample()` doesn't have. However, I have been playing around with an IDPF application and noticed that I wanted to be able to provide `CoinToss` a "length hint" so that the length of the output shares can be determined at runtime instead of compile time.

What do you think, @divergentdave?

src/dp.rs
```diff
@@ -0,0 +1,67 @@
...
+    // TODO(tholop): do we need another method to send the distribution parameters
+    // to the Collector?
+}
+
+/// Discrete Gaussian noise distribution with mean 0 and standard deviation `sigma`.
+/// `sigma` can be computed as a function of DP parameters `epsilon`, `delta`
+/// and `sensitivity`. `sensitivity` depends on the VDAF and potentially the number of measurements.
+/// `sigma` is sufficient to characterize the distribution, but other parameters
+/// are kept for interpretability.
```
cjpatton, line 43: Style: In this paragraph you refer to variables that are not exposed in the public API. I'd suggest moving this discussion to the constructor (`::new()`) below.

src/dp.rs
```diff
@@ -0,0 +1,67 @@
...
+///
+/// The distribution is defined over Z, represented by BigInts
```
cjpatton, line 45: `Z` is not mathematical notation that all readers will be familiar with.
```suggestion
/// The distribution is defined over the integers, represented by BigInts
```

src/dp.rs
```diff
@@ -0,0 +1,67 @@
...
+    fn sample<F: FieldElement, const SEED_SIZE: usize>(
+        &self,
+        seed: Seed<SEED_SIZE>,
+        len: usize,
+    ) -> Result<Vec<F>, DpError>;
+
+    // TODO(tholop): do we need another method to send the distribution parameters
+    // to the Collector?
+}
+
```
cjpatton, line 38: Question about the distribution: In @ooovi's PR (#578) the noise applied to each element of the vector is i.i.d. Is this always going to be the case? If not, it might be sensible to specialize the name `DiscreteGaussian` a bit further.

src/dp.rs
```diff
@@ -0,0 +1,67 @@
...
+    /// The randomness comes from a Prg seeded by `seed`.
+    /// TODO(tholop): do we need a generic SEED_SIZE here, or 16 is fine?
+    fn sample<F: FieldElement, const SEED_SIZE: usize>(
+        &self,
+        seed: Seed<SEED_SIZE>,
+        len: usize,
+    ) -> Result<Vec<F>, DpError>;
+
+    // TODO(tholop): do we need another method to send the distribution parameters
+    // to the Collector?
```
cjpatton, line 36: What do you mean by "parameters" here?

src/dp.rs
```diff
@@ -0,0 +1,67 @@
...
+// floats to store the DiscreteGaussian DP parameters?
+// Using BigInts might make it harder to communicate `DiscreteGaussian` to the Collector.
+use num_rational::BigRational;
+
+/// DP-related errors.
+#[derive(Debug, thiserror::Error)]
+pub enum DpError {
+    /// Error while sampling from a specific noise distribution
+    #[error("sampling error")]
+    SamplingError,
```
cjpatton, line 18: What would cause this error?

src/dp.rs
```diff
@@ -0,0 +1,67 @@
+// SPDX-License-Identifier: MPL-2.0
+
+//! Differential privacy (DP) primitives.
+use crate::field::FieldElement;
+use crate::vdaf::prg::Seed;
+use std::fmt::Debug;
+
+// TODO(tholop): We could also use pairs of integers, or even
+// floats to store the DiscreteGaussian DP parameters?
```
cjpatton, line 9: Pairs of integers might be fine. I'm not sure we'd want to use floats: If we end up standardizing some algorithm we're going to need test vectors, and the precision of floating points might cause some issues.

One last thought here: It sounds like we're going to need `BigInt` in any case, based on our discussion about how machine representations of integers might invalidate existing DP performance analysis. Although the additional dependency on `BigRational` is pretty small potatoes, it might be prudent to avoid exposing it in the public API. We could alternatively define our own wrapper type:

```rust
pub struct Rational(BigRational);
```

Thoughts @ooovi? @tgeoghegan? @divergentdave? @branlwyd?

src/dp.rs
```diff
@@ -0,0 +1,67 @@
...
+/// are kept for interpretability.
+///
+/// The distribution is defined over Z, represented by BigInts
+/// To implement `Dist`, the `sample` method can project on a finite F if necessary,
+/// as long as the DP guarantees are the same as the mechanism over Z
+/// (the proof is implementation dependent, e.g. DP post-processing)
+///
+/// TODO(tholop): depending on how we handle `num_measurements`, rename `sensitivity`
+/// to something like `base_sensitivity` or `sensitivity_when_num_measurements_is_1`,
+/// and explain that `add_noise_to_agg_share` is in charge of scaling the noise properly.
```
cjpatton, line 52: Question: Is `base_sensitivity` always going to be `1`?

Another option would be to pass `num_measurements` here and remove the `len` parameter from `sample()`. That way it's up to the implementation of `Dist` to compute the length of the vector properly.

src/vdaf.rs
```diff
@@ -248,6 +255,38 @@ pub trait Aggregator<const VERIFY_KEY_SIZE: usize, const NONCE_SIZE: usize>: Vda
...
+
+    #[cfg(feature = "experimental")]
+    /// Adds noise to an aggregate share to achieve differential privacy.
+    /// The noise is drawn from the `dist` distribution, which contains
+    /// the relevant DP parameters.
+    ///
+    /// TODO(tholop): VDAFs that don't want DP can make it a no-op.
+    ///               How about VDAFs that sometimes want DP and sometimes not?
+    ///               Leave it up to them how they implement their `dist` parameter,
+    ///               potentially with a `None` variant?
```
cjpatton, line 267: Something we need to avoid is silently violating the user's expectations: If they call this method, then they expect the aggregate result to be DP. If an `Aggregator` doesn't actually implement DP (yet), then either the code should fail to compile or there should be an error at runtime.

Another option: Add yet another trait:

```rust
pub trait NoiseAdder: Aggregator {
    fn add_noise_to_agg_share();
}
```

What do folks think of this?

src/dp.rs
tholop: (solved IRL, protocol plumbing is out of scope here)

src/dp.rs
tholop: No, for instance I think that `base_sensitivity` is equal to `2^n_bits - 1` for the Prio3Sum VDAF. It also depends on the vector length in Prio3SumVec.

Also, `len` (e.g. number of buckets, gradient dimension) is not equal to `num_measurements` in general.

src/dp.rs
tholop: I thought these variables were exposed since they are fields of a `pub struct`. Is that not the case?

src/dp.rs
tholop: Nothing actually, if we use an infallible SeedStream! Will remove.

src/vdaf.rs
```diff
@@ -248,6 +255,38 @@ pub trait Aggregator<const VERIFY_KEY_SIZE: usize, const NONCE_SIZE: usize>: Vda
...
+        //    (i.e. `dist` is the true distribution we sample from)
+        // 2. else, we can add a "scaling" step in `add_noise_to_agg_share` to modify
+        //    the noise sampled from `dist` once we know `num_measurements`
+        //    (i.e. `dist` is a canonical disttribution, and we rescale it after sampling)
+        //    In that case, we need to communicate `num_measurements` to the Collector too.
+        //
+        // So far, all our Vdafs have sensitivity independent on `num_measurements`,
+        // so the scaling step would be a no-op. For a hypothetical `Prio3Average`, the
+        // scaling step would multiply the noise by something like `1/num_measurements`.
+        num_measurements: usize,
```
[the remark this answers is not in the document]
tholop: Thanks for the feedback, I am passing the aggregation parameter now.  And I kept `num_measurements`

src/dp.rs
tgeoghegan: No -- the structure itself is public so `DiscreteGaussian` values can be used outside this module, but its fields are private and not visible outside this module unless each is marked `pub`. See [docs](https://doc.rust-lang.org/rust-by-example/mod/struct_visibility.html). n.b. visibility of enum variants does not work the same way, which could be the source of your confusion?

src/dp.rs
cjpatton: Sorry, "DP performance analysis" is the term I learned from @tholop to describe theorems about privacy versus utility. By performance we don't mean CPU time or whatever.

cjpatton: I think `Rational::new(numerator: u32, denominator: u32)` would be a reasonable. (1) these numbers would be reasonably short and (2) we want them to be compatible with the field that's used. (`u32` is smaller than every `F::Integer` we have in this repo.)

src/vdaf.rs
cjpatton: Good question, but to refine it a bit more: should *server-DP* be optional, since this is what we're implementing here.

I think that it needs to be optional. There are a variety of DP mechanisms that may be compatible with a given VDAF, and I don't think we can or should try to encourage one over another. Ultimately the choice of DP mechanism comes down to the application and is not specific to the VDAF.

src/vdaf.rs
```diff
@@ -248,6 +250,24 @@ pub trait Aggregator<const VERIFY_KEY_SIZE: usize, const NONCE_SIZE: usize>: Vda
...
+    /// Adds noise to an aggregate share to achieve differential privacy.
+    /// The `dist` distribution contains the relevant DP parameters.
+    /// The noise is drawn from `dist` and then scaled by `num_measurements`
+    /// if necessary, in order to enforce the desired privacy level.
+    ///
+    /// TODO(tholop): VDAFs that don't want DP can make it a no-op.
+    ///               How about VDAFs that sometimes want DP and sometimes not?
+    ///               Leave it up to them how they implement their `dist` parameter,
+    ///               potentially with a `None` variant?
+    fn add_noise_to_agg_share<D: Dist>(
```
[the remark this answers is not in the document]
cjpatton: Ahh interesting point. One option would be to move `Dist` to the trait instead:
```rust
pub trait CentralDpAggregator<D: Distribution>: Aggregator {
    fn add_noise_to_agg_share( ... dist: &D) -> ....
}
```

that way you have to explicitly implement `CentralDpAggregator<DiscreteGaussian>` for your VDAF.

src/dp.rs
divergentdave: I think the big issue is that `CoinToss` implicitly assumes a uniform distribution. Extending it for DP-relevant distributions would require turning them into parallel universe version of `rand`'s traits.

Taking a `SeedStream` mutable reference sounds good. This opens the door to re-using a PRG within one `add_noise_to_agg_share` call, which could be a significant performance win. (`PrgSha3` setup/derivation is relatively expensive)

For the separate IDPF use case, we could write a helper function or provided method to handle calling `CoinToss::sample()` repeatedly and returning a `Vec<T>` of the desired length. But, since `CoinToss` exists to be a trait bound on `IdpfValue`, we may need to take a broader look at the whole IDPF API, as it sounds like `convert()` may need a way to pass that length when producing values.

src/dp.rs
```diff
@@ -0,0 +1,64 @@
...
+pub struct DiscreteGaussian {
+    epsilon: BigRational,
+    delta: BigRational,
+    sigma: BigRational,
+    sensitivity: BigRational,
```
divergentdave: I think we may need to explain what this is the sensitivity of. Alternately, if this is just the sensitivity of the VDAF's aggregate result with respect to any one valid measurement, then the sensitivity could be computed within `add_noise_to_agg_share()`. If the sensitivity is indeed a dependent variable, then we could split the other parameters apart from the implementation of discrete Gaussian sampling.

src/dp.rs
cjpatton: > I think the big issue is that `CoinToss` implicitly assumes a uniform distribution. Extending it for DP-relevant distributions would require turning them into parallel universe version of `rand`'s traits.

It assumes the output of `SeedStream` is uniform, but so would `Distribution`. But you're definitely on to something: we don't want to be duplicating work that has already been done.

My main goal here is to avoid defining two traits in libprio that are nearly identical. Are there traits in that crate we should be implementing instead? 

> For the separate IDPF use case, we could write a helper function or provided method to handle calling `CoinToss::sample()` repeatedly and returning a `Vec<T>` of the desired length. But, since `CoinToss` exists to be a trait bound on `IdpfValue`, we may need to take a broader look at the whole IDPF API, as it sounds like `convert()` may need a way to pass that length when producing values.

It's not super urgent, but it is something I'd like to investigate at some point. I do think it's going to be useful to have. I'll file an issue.

src/vdaf.rs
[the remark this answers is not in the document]
divergentdave: Here are a few scenarios for how we could handle this:

* Distributions are passed as a trait object, as described above. We could use `Any::downcast()` as a more idiomatic way to check the distribution's type, instead of checking codepoints. We could also consider providing a no-op distribution that doesn't read the seed stream and always returns zeros. This could make aggregator code paths a little simpler, as they could unconditionally call `add_noise_to_agg_share()`.
* The distribution and its parameters are passed as an enum, and each VDAF returns an error if an inappropriate distribution is chosen (i.e. a mechanism for categorical data is used with Prio3Sum).
* `add_noise_to_agg_share()` is moved to some new trait, called `AggregatorWithNoise`,  `NoiseAdder`, or similar, with `Aggregator` as a super trait. The noise distribution is a type parameter of this trait. Then, VDAFs that support DP implement it with the trait's type parameter equal to whichever noise distributions are appropriate. This rules out the need for a runtime check.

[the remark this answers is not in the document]
cjpatton: I like option 3 :)

src/vdaf.rs
cjpatton: (@divergentdave see this thread as well)

src/vdaf.rs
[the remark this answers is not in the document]
divergentdave: One other point in favor of a separate trait: with the current state of this PR, the shape of the Aggregator trait depends on the `experimental` feature. Thus, compilation would break if any foreign VDAF implementation, written assuming the feature is off, either turns it on for some other reason (Poplar1, Prio2, fixed point bounded L2 norm Prio3 instance etc.) or unintentionally turns it on through transitive dependencies that enable it. If we moved the new method to a separate trait, then the conditional compilation would only change the presence or absence of entire items, which is much easier to work with.

src/dp.rs
tholop: >  Alternately, if this is just the sensitivity of the VDAF's aggregate result with respect to any one valid measurement, then the sensitivity could be computed within add_noise_to_agg_share()

Yes, I think that the sensitivity of the aggregate result could be computed within `add_noise_to_agg_share`, by using both `num_measurements` and the particular VDAF. It seems cleaner than decomposing the sensitivity into two factors (`num_measurements`-dependent scaling factor in `add_noise_to_agg_share`, and `base_sensitivity` in `DiscreteGaussian`).

I'll remove `sensitivity` from `DiscreteGaussian`, which will then represent an unscaled distribution (sensitivity 1).

src/dp.rs
```diff
@@ -0,0 +1,67 @@
...
+/// that will enforce the VDAF-dependent DP guarantees.
+pub trait Dist: Clone + Debug {
+    /// Sample `len` elements from the distribution over `F`.
+    /// The randomness comes from a Prg seeded by `seed`.
+    /// TODO(tholop): do we need a generic SEED_SIZE here, or 16 is fine?
+    fn sample<F: FieldElement, const SEED_SIZE: usize>(
+        &self,
+        seed: Seed<SEED_SIZE>,
+        len: usize,
+    ) -> Result<Vec<F>, DpError>;
```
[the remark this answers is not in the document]
tholop: Yes, why not!

src/dp.rs
tholop: My bad, thanks!

src/dp.rs
tholop: I don't know how to address this one, so I kept the current implementation for now.

src/dp.rs
tholop: I used `pub struct Rational {numerator: u32, denominator: u32}` for now, to keep it simple. But this is not really part of the API anyway; we might have to change it when we do the actual discrete Gaussian implementation.

src/vdaf.rs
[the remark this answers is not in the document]
tholop: I moved `add_noise_to_agg_share()` to a new `AggregatorWithNoise` trait.

src/dp.rs
```diff
@@ -0,0 +1,64 @@
...
+/// To implement `Dist`, the `sample` method can project on a finite F if necessary,
+/// as long as the DP guarantees are the same as the mechanism over Z
+/// (the proof is implementation dependent, e.g. DP post-processing)
+///
+/// TODO(tholop): depending on how we handle `num_measurements`, rename `sensitivity`
+/// to something like `base_sensitivity` or `sensitivity_when_num_measurements_is_1`,
+/// and explain that `add_noise_to_agg_share` is in charge of scaling the noise properly.
+pub struct DiscreteGaussian {
+    epsilon: BigRational,
+    delta: BigRational,
```
ooovi: the DP flavour of the guarantee given in thm 4 of the [discrete gaussian paper](https://arxiv.org/pdf/2004.00010.pdf) is [zero-concentrated DP (zCDP)](https://programming-dp.com/ch8.html#zero-concentrated-differential-privacy), not epsilon-delta (ED). those can be converted into each other, but for composition (e.g. executing in a loop like for gradient descent) converting for every execution and then accumulating ED gives worse bounds than accumulating zCDP and converting to ED in the end. i hence suggest using the zCDP parameter here, confusingly also called epsilon in the paper. alternatively, i suppose an implementation of a [slightly involved](https://github.com/IBM/discrete-gaussian-differential-privacy/blob/master/cdp2adp.py) method to compute a tighter zCDP parameter given epsilon and delta could be provided, based on section 2.2 of the paper.

src/dp.rs
ooovi: the DP flavour of the guarantee given in thm 4 of the [discrete gaussian paper](https://arxiv.org/pdf/2004.00010.pdf) is [zero-concentrated DP (zCDP)](https://programming-dp.com/ch8.html#zero-concentrated-differential-privacy), not epsilon-delta (ED). those can be converted into each other, but for composition (e.g. executing in a loop like for gradient descent) converting for every execution and then accumulating ED gives worse bounds than accumulating zCDP and converting to ED in the end. i hence suggest using the zCDP parameter here, confusingly also called epsilon in the paper. alternatively, i suppose an implementation of a [slightly involved](https://github.com/IBM/discrete-gaussian-differential-privacy/blob/master/cdp2adp.py) method to compute a tighter zCDP parameter given epsilon and delta could be provided, based on section 2.2 of the paper.

tholop: Yes, as noted in the comment on the latest commit, I agree that a zCDP constructor would be good, potentially in addition to an Approximate DP/ED constructor (because some people might not need composition, or prefer a simpler definition of DP).

In that case, I think that the `DiscreteGaussian` struct should hold only `sigma`, and no other metadata? We can always provide some convenience functions to convert `sigma` from/to different DP definitions.

src/dp.rs
ooovi: the standard thing to do is using iid noise, and that's what we have privacy guarantee proofs for in the discrete gaussian paper. there are mechanisms that use non-iid noise for utility gain, so i can't promise it will always be the case.

src/dp.rs
```diff
@@ -0,0 +1,62 @@
...
+    }
+}
+/// Trait for a noise distribution over a (vector space over a) finite field.
+/// Useful for differential privacy.
+/// Inherited by specific mechanism instantiations such as `[DiscreteGaussian]`
+/// that will enforce the VDAF-dependent DP guarantees.
+pub trait Distribution: Clone + Debug {
+    /// Fills the `noise` buffer with elements from the distribution over `F`,
+    /// with randomness coming from `seed_stream`
+    fn sample<F: FieldElement, S: SeedStream>(&self, seed_stream: S, noise: &mut [F]) -> Vec<F>;
```
cjpatton, line 31: `seed_stream` should be a `&mut S`, otherwise we can't use the `seed_stream` again after calling `sample`

src/dp.rs
cjpatton: Fine to leave as is for now. We can decide later if we can merge these traits into one.

src/dp.rs
cjpatton: Yeah after asking a few folks it seems like i.i.d is going to be sufficient. Let's just make sure to note it in the documentation.

src/dp.rs
cjpatton: My only feedback here is that we want the API to be flexible enough to support multiple mechanisms. I'm not sure what's needed, so I'll just trust y'all's judgement here.

src/dp.rs
tholop: I realized that we can't always reduce to sensitivity-1 functions by scaling the result or the noise. Indeed, unlike the continuous case, when we work with integers `Delta * N_Z(0, sigma^2)` does not have the same distribution as `N_Z(0, Delta^2 * sigma^2)` for an integer Delta (the two distributions don't even have the same support!). 

Hence I think we should still give a `sensitivity` field to the constructor of the `Distribution` object, where the sensitivity definition and the privacy guarantees are given by the Discrete Gaussian paper.

divergentdave: Would it be simpler to make the `DiscreteGaussian` constructor to take only sigma, representing the distribution's standard deviation (or sigma squared, for variance) instead of the sensitivity and the sensitivity-1 value of sigma? Would the constructor otherwise just multiply the `Delta` and `sigma` arguments together? It appears to me that the "Sampling from a Discrete Gaussian" algorithm in section 5.3 only needs a single parameter to describe the probability distribution.

tholop: The distribution is completely described by a single parameter, but the DP guarantees depend on the sensitivity and the budget (e.g. Thm 4 or Thm 7 of https://arxiv.org/pdf/2004.00010.pdf). In the current draft the constructor takes `epsilon`, `delta` and `Delta`: https://github.com/divviup/libprio-rs/blob/d22d271c54f716507b3b5573bfd7da103a29d49e/src/dp.rs#L59 (and not `sigma` and `Delta`), and is supposed to compute the right value of `sigma`. 

Are you suggesting to construct `DiscreteGaussian` only from `sigma`, and leave it up to the user how to compute and interpret `sigma` (outside of libprio)? That would simplify some things (e.g. we wouldn't need to provide constructors for every possible accounting method like zCDP, RDP, Approximate DP; and the DP guarantees could be computed when `num_measurements` is known, maybe at the DAP level if we have a fixed batch size). However that would also leave more room for mistakes (e.g. if the user doesn't compute the sensitivity properly, although we could still provide some optional convenience functions to help computing sigma).

divergentdave: We could still handle this in libprio, but I think separating higher-level DP mechanisms from lower-level probability distributions while doing so would be valuable. Here's one way we could arrange it, with separate traits and types.

```rust
pub trait Distribution: Clone + Debug {
    fn sample<F: FieldElement, S: SeedStream>(&self, seed_stream: &mut S, noise: &mut [F]);
}

/// N_Z(0, sigma^2)
pub struct DiscreteGaussian {
    sigma: Rational,
}

impl DiscreteGaussian {
    pub fn new(sigma :Rational) -> Self {
        Self {
            sigma,
        }
    }
}

impl Distribution for DiscreteGaussian {
    fn sample<F: FieldElement, S: SeedStream>(&self, seed_stream: &mut S, noise: &mut [F]) {
        todo!()
    }
}

/// Marker trait for differential privacy mechanisms.
pub trait PrivacyMechanism {}

struct ZeroConcentratedDifferentialPrivacy {
    epsilon: Rational,
}

impl ZeroConcentratedDifferentialPrivacy {
    pub fn new(epsilon: Rational) -> Self {
        Self { epsilon }
    }
}

pub trait AggregatorWithNoise<const VERIFY_KEY_SIZE: usize, const NONCE_SIZE: usize, DP>:
    Aggregator<VERIFY_KEY_SIZE, NONCE_SIZE>
where DP: PrivacyMechanism,
{
    fn add_noise_to_agg_share(
        &self,
        mechanism: &DP,
        agg_share: &mut Self::AggregateShare,
        num_measurements: usize,
    ) -> Result<(), VdafError>;
}

impl AggregatorWithNoise<16, 16, ZeroConcentratedDifferentialPrivacy> for Prio3Count {
    fn add_noise_to_agg_share(
        &self,
        mechanism: &ZeroConcentratedDifferentialPrivacy,
        agg_share: &mut Self::AggregateShare,
        num_measurements: usize,
    ) -> Result<(), VdafError> {
        let sensitivity = Rational::new(1, 1); // differs per VDAF (and sensitivity definition)
        let sigma = sensitivity / mechanism.epsilon;
        let dg = DiscreteGaussian::new(sigma);

        let mut noise = [Field64::zero()];
        let seed = Seed::generate()?; // TODO: make seed an explicit parameter for testability.
        let mut seed_stream = PrgSha3::seed_stream(&seed, &[], &[]);
        dg.sample(&mut seed_stream, &mut noise);

        agg_share.sum(&noise)
    }
}
```

This arrangement gives us static dispatch on the VDAF (and thus the query function) and the DP mechanism at the same time. Do you think this would be a complete enough API to model what we need?

cjpatton: This looks good to me.

tholop: I think this is complete enough and I implemented it. Thanks for the idea! I just used  `DifferentialPrivacyBudget` instead of `PrivacyMechanism`, since this struct only tracks an epsilon (or epsilon-delta, or rho, etc.) and not a full mechanism (which would be a full noise-adding algorithm).

src/vdaf/prio2.rs
```diff
@@ -21,7 +21,6 @@ use std::{
     convert::{TryFrom, TryInto},
     io::Cursor,
 };
-
```
divergentdave, line 24: nit: please back out the unrelated whitespace changes in this directory

divergentdave (approved):

Cargo.toml
```diff
@@ -37,6 +37,9 @@ rayon = { version = "1.7.0", optional = true }
 aes-gcm = { version = "^0.10", optional = true }
 ring = { version = "0.16.20", optional = true }
 
+# dependencies for differential privacy, required if "experimental" is enabled
```
cjpatton: IIUC this dependency is no longer needed.

src/dp.rs
```diff
@@ -0,0 +1,81 @@
...
+/// Positive rational number to represent DP parameters and noise distributions.
+#[allow(dead_code)]
+#[derive(Clone, Debug)]
+pub struct Rational {
+    numerator: u32,
+    denominator: u32,
+}
+
+impl Rational {
+    #[allow(dead_code)]
```
cjpatton, line 17: Instead putting a dead_code here, I think we could just way to implement this constructor until we need it.

src/dp.rs
```diff
@@ -0,0 +1,81 @@
+// SPDX-License-Identifier: MPL-2.0
+
+//! Differential privacy (DP) primitives.
+use crate::field::FieldElement;
+use crate::vdaf::prg::SeedStream;
+use std::fmt::Debug;
+
+/// Positive rational number to represent DP parameters and noise distributions.
```
cjpatton, line 8: I think it's worth noting here why we think a floating point number shouldn't be used (eventually we're going to need to encode this value in a protocol message).

src/dp.rs
```diff
@@ -0,0 +1,81 @@
...
+            numerator,
+            denominator,
+        }
+    }
+}
+
+/// Marker trait for differential privacy budgets (regardless of the specific accounting method).
+pub trait DifferentialPrivacyBudget {}
+
+/// Zero-concentrated differential privacy (zCDP) budget.
```
cjpatton, line 29: Please add a reference to the paper with the standard definition. It can be formatted in markdown, e.g., https://github.com/divviup/libprio-rs/blob/main/src/flp.rs#LL47C1-L47C1

src/dp.rs
```diff
@@ -0,0 +1,81 @@
...
+#[allow(dead_code)]
+struct ZeroConcentratedDifferentialPrivacyBudget {
+    epsilon: Rational, // rho = (epsilon**2)/2
```
cjpatton, line 32: nit: `rho` is defined twice in comments, here and on line 40 below. When updating this code we'll have to keep track of the definition of `rho` in two places. I think it's better to just maintain it in one place, in order to reduce the chance of comments becoming inconsistent.
```suggestion
    epsilon: Rational,
```

src/dp.rs
```diff
@@ -0,0 +1,81 @@
...
+}
+
+impl DifferentialPrivacyBudget for ZeroConcentratedDifferentialPrivacyBudget {}
+
+#[allow(dead_code)]
+impl ZeroConcentratedDifferentialPrivacyBudget {
+    /// Creates a new rho-zCDP budget with parameter `epsilon`, using the
+    /// notation from https://arxiv.org/pdf/2004.00010.pdf where `rho = (epsilon**2)/2`.
```
cjpatton, line 40: Move this to a markdown reference (as suggested above).

src/dp.rs
cjpatton: +1 It would be great if we could reuse that. The problem is that their sample method can't be used to fill a buffer, as ours can. However since we're already asserting in the API that each element is i.i.d., then perhaps we don't actually need that feature?
```rust
    /// Fills the `noise` buffer with elements sampled independently from
    /// the distribution over `F` (i.i.d.) with randomness coming from `seed_stream`.
    fn sample<F: FieldElement, S: SeedStream>(&self, seed_stream: &mut S, noise: &mut [F]);
```

However would also need to implement the `RngCore`, which is slightly more sophisticated than `SeedStream`: https://rust-random.github.io/rand/rand/trait.RngCore.html

I think the trait is stable enough that we are safe to depend on it.

src/dp.rs
```diff
@@ -0,0 +1,81 @@
...
+/// that will enforce the VDAF-dependent DP guarantees.
+pub trait Distribution: Clone + Debug {
+    /// Fills the `noise` buffer with elements sampled independently from
+    /// the distribution over `F` (i.i.d.) with randomness coming from `seed_stream`.
+    fn sample<F: FieldElement, S: SeedStream>(&self, seed_stream: &mut S, noise: &mut [F]);
+}
+
+/// Zero-mean Discrete Gaussian noise distribution.
+///
+/// The distribution is defined over the integers, represented by BigInts.
```
cjpatton, line 58: The type is an implementation detail.
```suggestion
/// The distribution is defined over the integers, represented by arbitrary-precision integers.
```

src/dp.rs
```diff
@@ -0,0 +1,81 @@
...
+/// To implement `Distribution`, the `sample` method can project on a finite F if necessary,
+/// as long as the DP guarantees are the same as the mechanism over Z
+/// (the proof is implementation dependent, e.g. DP post-processing).
```
cjpatton, line 61: There's a few things in this comment that require unpacking in order to be understood by a non-DP expert. Stepping back, what is it you want to convey here?

src/vdaf.rs
```diff
@@ -250,6 +252,24 @@ pub trait Aggregator<const VERIFY_KEY_SIZE: usize, const NONCE_SIZE: usize>: Vda
     ) -> Result<Self::AggregateShare, VdafError>;
 }
 
+/// Aggregator that implements differential privacy with aggregator-side noise addition.
```
cjpatton, line 255:
```suggestion
/// Aggregator that implements differential privacy with Aggregator-side noise added.
```

src/vdaf.rs
```diff
@@ -250,6 +252,24 @@ pub trait Aggregator<const VERIFY_KEY_SIZE: usize, const NONCE_SIZE: usize>: Vda
...
+#[cfg(feature = "experimental")]
+pub trait AggregatorWithNoise<const VERIFY_KEY_SIZE: usize, const NONCE_SIZE: usize, DPBudget>:
+    Aggregator<VERIFY_KEY_SIZE, NONCE_SIZE>
+where
+    DPBudget: DifferentialPrivacyBudget,
+{
+    /// Adds noise to an aggregate share such that the aggregate result
+    /// is `DPBudget`-differentially private as long as one Aggregator is honest.
```
cjpatton, line 263: Ha, this is really cool thing to be able to express in the API :)

cjpatton (commented):
* nit: Block comments should break lines at 100 characters, for consistency.

src/dp.rs
tholop: Sounds good. Can I still keep the dead_code for ZeroConcentratedDifferentialPrivacyBudget and DiscreteGaussian? The constructor seems like a good place to keep some documentation about the struct fields.

cjpatton: Yup, I meant this one specifically.

src/dp.rs
tholop: Good point, actually that comment doesn't really belong here, so I removed it. It will be something to explain along the specific implementation.

src/dp.rs
cjpatton: Decision on 2023/6/21 call: Take Distribution and RngCore traits.

tholop: Thanks, I dropped the trait and rebased.

cjpatton (approved):

MxmUrw:
We have a remark regarding the name `DifferentialPrivacyBudget`: It seems that in the literature the term "privacy budget" is used in the context of iteration of privacy mechanisms (to track the privacy which was already spent), while simply "privacy parameter" is used to refer to the $(\epsilon,\delta)$ or $\rho$ parameters occurring in the definitions of the various notions of privacy, such as zCDP, etc. Compare with various papers such as [the definition of CDP](https://arxiv.org/pdf/1605.02065.pdf), [the analysis of the discrete gaussian](https://arxiv.org/pdf/2004.00010.pdf), [the "deep learning with dp" paper](https://arxiv.org/pdf/1607.00133.pdf).

Usually, the privacy budget is calculated by summing up the privacy parameters of each round, which makes the distinction somewhat vacuous - the privacy parameter is effectively just the privacy budget of a single differentially mechanism. We would be happy to leave the code as is, but wanted to give a heads-up regarding this. Of course, I assume there might be communities where using the term "budget" is more customary?

MxmUrw:
Thinking further about the discussed direction for this PR, we concluded the following:
 - It would be useful to separate the choice of noising mechanism (gaussian, laplace) from the choice of the parametrization, i.e., "notion of privacy" (e.g., epsilon-delta, zCDP)
 - It should be possible to implement dp-noising for VDAFs generically over a tuple of `(noising mechanism, parameter)`, this would require a function `new_with_param(param : Parameter, sensitivity : Rational) -> Distribution` for each supported (parameter,distribution) pair.

As a sketch, this might look as follows:
```rust
// We need a trait which inherits from `rand::Distribution`
trait ParametrizedDistribution<DPBudget : DifferentialPrivacyBudget> : Distribution<BigInt>
{
    fn new_with_budget(budget: DPBudget, sensitivity: Rational) -> Self;
}
```
And the `Aggregator` trait would take it as a further parameter: 
```rust
pub trait AggregatorWithNoise<DPBudget, PDistribution>: Aggregator
where
    DPBudget: DifferentialPrivacyBudget,
    PDistribution: ParametrizedDistribution<DPBudget>,
{
    /// Adds noise to an aggregate share such that the aggregate result
    /// is `DPBudget`-differentially private as long as one Aggregator is honest.
    fn add_noise_to_agg_share(
        &self,
        budget: &DPBudget,
        agg_share: &mut Self::AggregateShare,
        num_measurements: usize,
    ) -> Result<(), VdafError>;
}
```
Then implementations of `add_noise_to_agg_share` can use `new_with_budget` to create a distribution and sample from it.

tholop:
I like this proposal @MxmUrw. One concern I have is that your sketch hardcodes the type of the distribution to be `BigInt`. That can be problematic if some VDAFs prefer distributions that sample in different sets, e.g. a truncated discrete Gaussian sampling directly a `u128`. So I added a generic `T` in my implementation. Let me know if that looks good!

Another note: `DifferentialPrivacyDistribution` will sample individual elements (because it inherits from `Distribution`), so the constructor only specifies DP guarantees for scalar-valued queries. For vector-valued queries, it will be up to `add_noise_to_agg_share` to compute the right sensitivity for the scalar distribution, and sample each element i.i.d. I don't think it will be hard (e.g. with Theorem 14 of https://arxiv.org/pdf/2004.00010.pdf) but I wanted to keep track of this.

Regarding the name: I tend to use "privacy budget" more often than "privacy parameters" (as do some practitioners like [this survey](https://arxiv.org/pdf/2303.00654.pdf), the [Programming DP book](https://programming-dp.com/ch3.html) or even theory papers like [the RDP definition](https://arxiv.org/pdf/1702.07476.pdf)). Both terms seem pretty interchangeable. I just like the concrete feel of "budget" compared to generic "parameters" that could represent many things for the non-expert. I agree that "privacy parameters" might be more widespread though, so I don't have strong feelings.

tholop:
@MxmUrw @ooovi @divergentdave do you have more changes to suggest? If not, I can rebase and squash so you can merge.

MxmUrw:
If I understand correctly, this means that it would become unpractical to implement VDAFs generically over the choice of distribution. That is, the VDAF would hard-code the choice of distribution in `impl AggregatorWithNoise<..., MyDPBudget>` for a given `MyDPBudget`. (I think coupling choice of distribution to choice of budget type would be confusing, because usually these choices are quite orthogonal, as far as I know - @tholop @ooovi ?)

OTOH, it may not be necessary for the `AggregatorWithNoise` trait to be as generic as proposed. For instance, we only ever intend to use the discrete gaussian for our `fixedvec` type. Are there even cases where the same VDAF is intended to work with multiple distributions?

tholop:
>  Are there even cases where the same VDAF is intended to work with multiple distributions?

Yes. For instance, `Prio3Count` or `Prio3Sum` make sense with both Laplace and Gaussian noise, and the end user should be able to decide depending on their requirements (Pure DP, composition, type of error, etc). We could still hardcode the distribution, e.g. with `Prio3CountGaussian` and `Prio3CountLaplace`. That would leave room for a hypothetical future VDAF using non i.i.d noise, but I don't know if it's desirable according to @divergentdave:

> So long as our conception of the DifferentialPrivacyBudget trait encompasses a choice of noise distribution, I think we could eliminate DifferentialPrivacyDistribution, and leave it up to the impl AggregatorWithNoise<..., MyDPBudget> blocks to choose and parameterize the correct distributions.

Let me try to summarize some desired properties (they might not be all required, actually 1 and 3 seem incompatible):
1. Pass something that encapsulates a noise distribution and the corresponding budget to `AggregatorWithNoise`
2. Don't pollute the trait system with too many trait bounds
3. Allow some VDAFs to combine different distributions 
4. Keep some separation between the type of distribution (Gaussian/Laplace) and the budget (zCDP with epsilon=2, EpsilonDeltaDP with epsilon, delta = 1, 0.1)

And I think that we have two solutions:
-  @MxmUrw's proposal of `DPBudgetAndDistribution` satisfies 1, 2 and 4. I like that, it might actually make our life easier when we plug VDAF into DAP (e.g. we need codepoints to pass the type of distribution and the budget). Daphne has such a `DpConfig` type: https://github.com/cloudflare/daphne/blob/edda510353709496427e52f07ebbd29dd7302531/daphne/src/messages/taskprov.rs#L117.
-  I don't know any VDAF that would need 3 at the moment, but if we want to support that then I think that we should hardcode the distribution with something like `Prio3CountGaussian` (in which case we lose 1).

divergentdave:
The `DPBudgetAndDistribution`/`DpConfig` approach sounds good to me. (one other candidate name: `DifferentialPrivacyStrategy`?)

Regarding non-i.i.d. noise, I was only considering sampling noise from multiple distributions of the same kind, i.e. two different discrete Gaussian distributions for a VDAF that computes both a sum and a sum of squares. (not combining discrete Gaussian and Laplace) Thus, a `DPBudgetAndDistribution` argument should be expressive enough for such cases.

tholop:
>  I was only considering sampling noise from multiple distributions of the same kind, i.e. two different discrete Gaussian distributions for a VDAF that computes both a sum and a sum of squares. 

Ah indeed, this would be a realistic VDAF, and it's unclear if we can build it with `DifferentialPrivacyDistribution::from()` by playing with the `sensitivity` parameter (for a fixed budget). Instead, we can optionally provide a conversion function similar to `from` directly in the relevant `DPBudgetAndDistribution`/`DPConfig` object. I added such an example in `ZCdpDiscreteGaussian::create_distribution()`. 

The new implementation should be extensible enough to support non i.i.d VDAFs or other changes, and does static dispatch. The drawback is that we might have to implement multiple combinations of accounting methods & distributions, as @MxmUrw pointed out.  An alternative would be to pass a `(distribution_type, budget)` tuple and determine the compatibility at runtime, as suggested in https://github.com/divviup/libprio-rs/pull/607#discussion_r1232736343.

src/dp.rs
```diff
@@ -56,10 +49,40 @@ pub struct DiscreteGaussian {
...
+}
+
+impl DifferentialPrivacyStrategy for ZCdpDiscreteGaussian {}
+
+impl ZCdpDiscreteGaussian {
+    /// Creates a new Discrete Gaussian by following Theorem 4 from [[CKS20]]
+    ///
+    /// [CKS20]: https://arxiv.org/pdf/2004.00010.pdf
+    pub fn create_distribution(
+        budget: ZeroConcentratedDifferentialPrivacyBudget,
```
divergentdave, line 79: nit: this could take a reference instead
```suggestion
        budget: &ZeroConcentratedDifferentialPrivacyBudget,
```

divergentdave (approved):

src/dp.rs
tholop: Thanks!

tholop:
@divergentdave @tgeoghegan is there anything else I can do before merging? Thank you!

src/dp.rs
```diff
@@ -0,0 +1,88 @@
+// SPDX-License-Identifier: MPL-2.0
+
+//! Differential privacy (DP) primitives.
+use std::fmt::Debug;
+
+/// Positive rational number to represent DP and noise distribution parameters in protocol messages
+/// and manipulate them without rounding errors.
+#[allow(dead_code)]
```
tgeoghegan, line 8: Are these `allow`s needed? I thought any item marked `pub` was exempt from `dead_code` warnings.

src/dp.rs
```diff
@@ -0,0 +1,88 @@
...
+#[derive(Clone, Debug)]
+pub struct Rational {
+    numerator: u32,
+    denominator: u32,
```
tgeoghegan, line 12: If these aren't `pub` then there's no way to construct a `Rational` from outside the crate. Additionally, if there are any restrictions on what a valid `Rational` is (for instance, is it OK for `numerator > denominator`?) then we should provide a constructor that enforces that.

src/dp.rs
```diff
@@ -0,0 +1,88 @@
...
+}
+
+/// Marker trait for differential privacy budgets (regardless of the specific accounting method).
+pub trait DifferentialPrivacyBudget {}
+
+/// Marker trait for differential privacy scalar noise distributions
```
tgeoghegan, line 18:
```suggestion
/// Marker trait for differential privacy scalar noise distributions.
```

src/dp.rs
```diff
@@ -0,0 +1,88 @@
...
+    /// [CKS20]: https://arxiv.org/pdf/2004.00010.pdf
+    pub fn new(epsilon: Rational) -> Self {
+        Self { epsilon }
+    }
+}
+
+/// Zero-mean Discrete Gaussian noise distribution.
+///
+/// The distribution is defined over the integers, represented by arbitrary-precision integers.
+///
```
tgeoghegan, line 45:
```suggestion
```

src/dp.rs
```diff
@@ -0,0 +1,88 @@
...
+#[allow(dead_code)]
+#[derive(Clone, Debug)]
+pub struct DiscreteGaussian {
+    sigma: Rational,
+}
+
+impl DifferentialPrivacyDistribution for DiscreteGaussian {}
+
+#[allow(unused_variables)]
```
tgeoghegan, line 54: Is this needed? Seems like `sigma` is straightforwardly used in `new`.

src/vdaf.rs
```diff
@@ -250,6 +252,25 @@ pub trait Aggregator<const VERIFY_KEY_SIZE: usize, const NONCE_SIZE: usize>: Vda
     ) -> Result<Self::AggregateShare, VdafError>;
 }
 
+/// Aggregator that implements differential privacy with Aggregator-side noise addition.
+#[cfg(feature = "experimental")]
+pub trait AggregatorWithNoise<
+    const VERIFY_KEY_SIZE: usize,
+    const NONCE_SIZE: usize,
+    DPStrategy: DifferentialPrivacyStrategy,
```
tgeoghegan, line 260: How is a VDAF meant to indicate compatibility with an implementation of `DifferentialPrivacyStrategy`? Is the expectation that `add_noise_to_agg_share` returns an error if the DP strategy is not compatible?

src/dp.rs
tholop: Indeed. I actually had a constructor but we removed it since it wasn't used. We don't have any constraint on the fields, so I'll just stick to public fields.

src/vdaf.rs
tholop: With the current design we determine compatibility at compile time. I think we can indicate compatibility for a given VDAF by implementing something like this:
```rust
impl AggregatorWithNoise<16, 16, ZCdpDiscreteGaussian> for Prio3SumVec {
	fn add_noise_to_agg_share(
		&self,
		dp_strategy: &ZCdpDiscreteGaussian,
		agg_share: &mut Self::AggregateShare,
		num_measurements: usize,
		) -> Result<(), VdafError> {
			// Compute the L2 sensitivity (because we are using Gaussian noise)
			// Call `dp_strategy.create_distribution` (because the budget is in zCDP form)
			// Add i.i.d. noise
			todo!();
		}
}
```

For instance, we won't be able to compile a DAP program that passes a `dp_strategy` of type `EpsilonDeltaLaplace` to a `Prio3SumVec` VDAF, unless there is a corresponding implementation. Would that work?

tgeoghegan: Yes, I understand how this works now. Thanks!

tgeoghegan (approved):
Looks good to me!

PR #615, opened 2023-06-23
Inlining Fp arithmetic.

armfazh:
Saving some cycles by inlining field operations.

|     | `fft` (old) | `fft` (new)  | diff    |
|:----|:------------|:-------------|:--------|
| 1   | 259.07 ns   |  208.19 ns   |  19.6 % |
| 30  | 7.90 us     |  6.16 us     |  22.0 % |
| 60  | 17.51 us    |  13.82 us    |  21.0 % |
| 90  | 38.68 us    |  30.56 us    |  20.9 % |
| 120 | 38.72 us    |  30.55 us    |  21.1 % |
| 150 | 85.77 us    |  66.66 us    |  22.2 % |

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -66,6 +66,9 @@ crypto-dependencies = ["aes", "ctr", "cmac"]
 [workspace]
 members = [".", "binaries"]
 
+[lib]
+bench = false
+
 [[bench]]
 name = "speed_tests"
 harness = false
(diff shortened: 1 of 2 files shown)


divergentdave (approved):

tgeoghegan (approved):
Thanks!

cjpatton:
@armfazh what if we inlined this method as well? https://github.com/divviup/libprio-rs/blob/main/src/fft.rs#L31

armfazh:
> @armfazh what if we inlined this method as well? https://github.com/divviup/libprio-rs/blob/main/src/fft.rs#L31

After I ran some benchmarks, I didn't find significant changes.
Inlining small functions brought more benefits as they are called more often.

PR #616, opened 2023-06-23
Support IdpfValue types that need run-time parameters

### LINKED ISSUE #614: IDPF: Replace CoinToss trait
* Here we have a `len` field, which `CoinToss::sample()` doesn't have. However, I have been playing around with an IDPF application and noticed that I wanted to be able to provide `CoinToss` a "length hint" so that the length of the output shares can be determined at runtime instead of compile time.

_Originally posted by @cjpatton in https://github.com/divviup/libprio-rs/pull/607#discussion_r1232604120_

divergentdave:
This makes some changes to the `IdpfValue` trait, and the `idpf` module as a whole, in order to allow producing values that need run-time parameters, such as field vectors with lengths that are not known at compile time. This closes #614. The major changes are:

* Add a `Parameters` associated type to `IdpfValue`.
* Remove the `CoinToss` trait, and add an associated function to `IdpfValue` that randomly generates a value from PRG output and the above parameters.
* Add a `Parameters` argument to `IdpfValue::zero()`.
* Remove the `subtle::ConditionallySelectable` trait bound, since it requires `Copy`, and add an equivalent associated function to `IdpfValue`.
* Create an `Idpf` struct, which holds a pair of parameters for the inner node and leaf node value types, and move `gen()` and `eval()` to be methods of this struct.


src/idpf.rs
```diff
@@ -135,22 +136,70 @@ where
...
+    where
+        S: SeedStream;
+
     /// Returns the additive identity.
-    fn zero() -> Self;
+    fn zero(parameters: &Self::Parameters) -> Self;
+
+    /// Conditionally select between two values in constant time.
+    ///
+    /// This is the same as in [`subtle::ConditionallySelectable`], but without the [`Copy`] bound.
```
inahga, line 165: Why are we opposed to having a `Copy` bound?

divergentdave: The `Copy` bound is one of the two barriers keeping us from using `Vec<F>` in an IDPF value. As the `Vec` represents ownership of a memory allocation, it would be unsafe to duplicate a `Vec` by bytewise copying it. Since `Copy` is a special auto trait, this limitation is very hard to get around.

Looking at past discussions about `subtle::ConditionallySelectable`, part of the reason for requiring the `Copy` bound is that conditionally selecting between two heap-allocated objects with different sizes leads to unavoidable side channel leaks. In the context of an IDPF, we will be selecting between two identically-sized objects, and then immediately performing element-wise addition on the result. In the context of randomly generated values and secret sharing, this concern is ameliorated.

src/idpf.rs
```diff
@@ -135,22 +136,70 @@ where
...
     + Sub<Output = Self>
-    + ConditionallySelectable
     + ConditionallyNegatable
-    + CoinToss
     + Encode
     + Decode
     + Sized
 {
+    /// Any run-time parameters needed to produce a value.
+    type Parameters;
```
cjpatton, line 152: * nit: s/Parameters/Parameter for consistency with the codec traits
* Suggestion: A more specific sounding name might make the trait more self-documenting. Perhaps `OutputParameter` or `ValueParameter`?

src/idpf.rs
```diff
@@ -135,22 +136,70 @@ where
...
+
+    /// Generate a value from PRG output. This must have a uniform probability distribution over
+    /// the PRG output.
```
cjpatton, line 155: * We provide a seed stream here, not a `Prg`.
* A little nit-picky, but the value is not uniform random because it is not literally sampling from this distribution.
```suggestion
    /// Generate a psuedorandom value from the seed stream.
```

src/idpf.rs
```diff
@@ -135,22 +136,70 @@ where
...
+    fn generate<S>(seed_stream: &mut S, parameters: &Self::Parameters) -> Self
+    where
+        S: SeedStream;
+
     /// Returns the additive identity.
-    fn zero() -> Self;
+    fn zero(parameters: &Self::Parameters) -> Self;
+
+    /// Conditionally select between two values in constant time.
```
cjpatton, line 163: Make it clear that this is not a property of any implementation, but a design goal.
```suggestion
    /// Conditionally select between two values. Implementations must perform this operation in constant time.
```

src/idpf.rs
```diff
@@ -135,22 +136,70 @@ where
...
+            buffer.len() >= F::ENCODED_SIZE,
+            "field is too big for buffer"
+        );
+        loop {
+            seed_stream.fill(&mut buffer[..F::ENCODED_SIZE]);
+            match Self::try_from_random(&buffer[..F::ENCODED_SIZE]) {
+                Ok(x) => return x,
+                Err(FieldError::ModulusOverflow) => continue,
+                Err(err) => panic!("unexpected error: {err}"),
+            }
```
cjpatton, line 192: This is copied from the `prng` module, right? It might be useful to deduplicate it as it's a little hairy.

cjpatton (approved):
Nice work! Just a few minor comments around the trait definition.

src/idpf.rs
divergentdave: Finally, a reason to use `std::ops::ControlFlow` :stuck_out_tongue_closed_eyes:

branlwyd (approved):

src/idpf.rs
cjpatton: Nice :)

PR #628, opened 2023-06-26
CI: Replace features matrix with cargo-all-features

### LINKED ISSUE #617: Shrink CI matrix
I just saw a CI job fail to clone the repo from GitHub with a 429 Too Many Requests error. We should cut down the parallelism by reducing or eliminating the GHA matrix. We could instead use cargo-all-features to build and test with the same breadth of feature combinations, in series on the same host. This should be fast enough because the builds with different features will reuse almost all crates.

divergentdave:
This replaces our features matrix at the GitHub Actions level with `cargo-all-features`.

Positives:

* Automatically computes power set of Cargo features, so we won't miss combinations or forget to update the matrix.
* Reduces GHA-level parallelism, to play nice with rate limits. (Fixes #617)
* Compilations with different feature sets reuse almost all compiled artifacts.

Negatives:

* `cargo-all-features` does not yet support `clippy` or `bench`. This is mitigated by running each with the default feature set and `--all-features`.
* CI runtime is more sensitive to the test suite's runtime, since executions happen serially. Fortunately, it's pretty fast as-is.

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -46,26 +46,14 @@ jobs:
   build-crate:
     strategy:
       matrix:
-        features: [
-          "--no-default-features",
-          "--all-features",
-          "--features=default",
-          "--features=test-util",
-          "--features=multithreaded",
-          "--features=test-util,multithreaded",
-          "--features=prio2",
-          "--features=prio2,test-util",
-          "--features=prio2,multithreaded",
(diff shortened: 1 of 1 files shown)


tgeoghegan (approved):
Nice! I'm not too worried about not running `clippy` against the power set of feature combos, because I think that subtle feature combinations aren't likely to cause different lints to get tripped.

PR #659, opened 2023-07-26
Prio2 VDAF: Don't evaluate at a root of unity

### LINKED ISSUE #658: Prio2: VDAF preparation may evaluate polynomial at a root of unity
I was working on removing parts of the unused ENPA API, and I noticed that the VDAF version of Prio2 is missing a rejection sampling loop when choosing its evaluation point.

https://github.com/divviup/libprio-rs/blob/e51270cfd3a5ce65dd8e3d7ccec79cdd16e08515/src/server.rs#L148-L162

https://github.com/divviup/libprio-rs/blob/e51270cfd3a5ce65dd8e3d7ccec79cdd16e08515/src/vdaf/prio2.rs#L221-L223

This means it may choose a root of unity as an evaluation point, which leaks one digit of the input to the aggregators, breaking the zero knowledge property. The probability of this happening is approximately equal to the input's dimension over 2<sup>32</sup>. We should make a breaking change to fix this by adding the same rejection sampling loop to `prepare_init()`.

cc @cjpatton

divergentdave:
This fixes #658. Note that this is a wire-breaking change for communication between Prio2 aggregators. (sharding is unaffected)

### DIFF
--- a/src/vdaf/prio2.rs
+++ b/src/vdaf/prio2.rs
@@ -3,11 +3,13 @@
 //! Port of the ENPA Prio system to a VDAF. It is backwards compatible with
 //! [`Client`](crate::client::Client) and [`Server`](crate::server::Server).
 
-use super::{AggregateShare, OutputShare};
+use super::{prg::SeedStream, AggregateShare, OutputShare};
 use crate::{
     client as v2_client,
     codec::{CodecError, Decode, Encode, ParameterizedDecode},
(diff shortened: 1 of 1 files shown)


cjpatton (approved):
In Prio3 if we get a root of unity we raise an error instead of sampling the next point. We might consider harmonizing, but since the field for Prio2 is so tiny, I think it makes sense to sample the next point here.

tgeoghegan (approved):

Issue #32, opened 2021-04-19
Eliminate `mod util`


tgeoghegan:
Utility modules are a code smell (citation needed) so we should aim to get rid of `libprio-rs`' `mod util`. If anything in there is useful it should be refactored into more appropriate and meaningful modules, and there are likely opportunities to trim the crate's public API surface of interfaces that aren't quite useful enough to justify being exported.

cjpatton:
All that's left here is serializing/deserializing Prio2 proofs. Perhaps we could fold this code into `vdaf::prio2`? Would it get you all into trouble if the error types change?

tgeoghegan:
I don't think there's any better place for this stuff at the moment, except maybe `lib.rs`. At most I think we could make everything in there `pub(crate)` instead of `pub`, but let's wait and see what happens with ENPA before we make a decision here.

divergentdave:
Done in #660.

PR #660, opened 2023-07-26
Remove ENPA API from public interface, broad refactoring

divergentdave:
This removes the ENPA API from the crate's public interface, cuts out unneeded code and dependencies, and partly refactors the Prio2 VDAF. Support for encryption of Prio2 shares is removed, which lets us drop dependencies on `ring` and `aes-gcm`. And more:

* Two of the targets in the prio-binaries crate are removed, and the `test-util` feature on the `prio` crate (which supported them) is removed.
* I sorted the dependencies in Cargo.toml and got rid of some comments to make it more maintainable.
* There is still a Prio2 test vector, but I decrypted the input shares and changed its format. There is no longer any code to generate a test vector, as it's served its purpose.
* Prio2 benchmarks are rewritten to use the VDAF API.
* I changed the Prio2/Prio3CountVec comparison part of `vdaf_message_sizes` to output sizes in bytes (like the rest of the output) instead of number of field elements. This was also changed to use the VDAF API.
* The lone integration test was moved to a unit test, allowing us to make the Prio2 test vector struct private, instead of public and hidden from docs.
* Updated doc comments and the README.
* Folded the remnants of the `util` module into other Prio2-related modules, closing #32.
* Replaced ring's SHA256 and HMAC-SHA256 with RustCrypto equivalents in tests and the Prio2 VDAF.
* Removed unused dependencies in prio-binaries and pruned cargo-vet exemptions.

There's more that could be done to remove the `client` and `server` modules entirely, folding them into `src/vdaf/prio2.rs`, moving more methods onto the `Prio2` struct, and rewriting tests that still use `Server`. This PR gets all of that out of the public API, so it'll all be implementation details from here on out.


src/vdaf/prio2.rs
```diff
@@ -87,6 +85,27 @@ impl Prio2 {
...
+        S: SeedStream,
+    {
+        // Make sure the query randomness isn't a root of unity. Evaluating the proof at any of
+        // these points would be a privacy violation, since these points were used by the prover to
+        // construct the wire polynomials.
+        let n = (self.input_len + 1).next_power_of_two();
+        let proof_length = 2 * n;
+        loop {
+            let eval_at: FieldPrio2 = prng.get();
+            // Unwrap safety: the constructor checks that this conversion succeeds.
```
cjpatton, line 103: Since we aren't constrained by the old API anymore we might consider propagating this error instead of unwrapping.

cjpatton (approved):
LGTM. One high-level suggestion: I would probably keep around the test vectors for ENPA as they are, with the encryption bits. The reason is to keep a historical record: Someone in the future might want to know how encryption work, and this would give them something that can be easily reproduced for testing.

However that would entail moving the encryption logic into the test code, which has its own cost.

README.md
```diff
@@ -13,9 +13,10 @@ and Scalable Computation of Aggregate Statistics.
 
 ## Exposure Notifications Private Analytics
 
-This crate was used in the [Exposure Notifications Private Analytics][enpa] system. This is supported
-by the interfaces in modules `server` and `client` and is referred to in various places as Prio v2.
-See [`prio-server`][prio-server] or the [ENPA whitepaper][enpa-whitepaper] for more details.
+This crate was used in the [Exposure Notifications Private Analytics][enpa]
+system. This is referred to in various places as Prio v2. See
+[`prio-server`][prio-server] or the [ENPA whitepaper][enpa-whitepaper] for more
+details.
```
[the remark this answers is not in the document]
divergentdave: Already done :smile:

divergentdave (commented):
I think the historical purposes would be better served by the commit history and old versions of the crate. Even as test-only code, it would still add to maintenance burden and build times (mainly because of the ring dependency), and it's not worth it if we're not going to use it.

cjpatton:
@divergentdave can you resolve the conflicts?

src/client.rs
tgeoghegan, line 1: Let's move this whole file to `src/vdaf/prio2/client.rs`. I'd like for `prio2` to be just another VDAF and not get the special privilege of having its implementation in the root of the crate.

src/prng.rs
```diff
@@ -166,12 +166,16 @@ mod tests {
     /// takes a seed and hash as base64 encoded strings
     #[cfg(feature = "prio2")]
     fn random_data_interop(seed_base64: &str, hash_base64: &str, len: usize) {
+        use sha2::{Digest, Sha256};
```
tgeoghegan, line 169: nit: hoist these `use`s up to module level

src/server.rs
tgeoghegan, line 1: As with `src/client.rs`, I'd like to move this to `src/vdaf/prio2/server.rs`.

tgeoghegan (changes requested):

src/client.rs
cjpatton: Would this be a good time to move the sub-modules of `vdaf` into the base crate?

tgeoghegan: Yes and no: we could pile that API break into 0.14.0 with the others, but I'm not sure why we should move the VDAF modules. They're perfectly happy where they are.

cjpatton: Yeah I think I agree. We also have primitives, like `flp` and `idpf`, in the base crate, so it makes sense to keep `vdaf` at the same level. We *could* consider moving `prg` to the base crate for this reason? Not really sure about that one though.

tgeoghegan: I think we could, but not in this PR.

tgeoghegan (approved):

Issue #644, opened 2023-07-13
Migrate from SeedStream to RngCore


divergentdave:
We'd like to eventually replace the `SeedStream` trait with `rand_core::RngCore`. (see previous discussion on #607 and #578) We will still keep our own `Prg` trait, because we need to take in separate seed, context, and binder arguments to implement VDAF's interface, while `rand_core::SeedableRng` only takes in a (usually fixed-size) seed. For the time being, we can make use of both traits in parallel with a blanket implementation of `RngCore` for all `SeedStream` implementations. Proof-related operations will remain largely unchanged, and not use many provided methods from the `rand` crate, because of our hard interop requirements. Differential privacy code will make more use of the `rand::Distribution` trait, and may use provided methods like `gen_range()`, at least until we need to bring them in-tree and write test vectors for them. When we remove `SeedStream` in a future version, consumers such as Mozilla that maintain their own PRG implementations will need to make the same trait migration changes.

divergentdave:
Dropping a link to a [cautionary tale from Filippo Valsorda](https://words.filippo.io/dispatches/avoid-the-randomness-from-the-sky/) because it's such a good story.

> It’s tempting to say “just seed a stream cipher and use that as the CSPRNG” but that adopts the internals of the stream cipher, of how it’s instantiated, and how it’s used as a CSPRNG into your protocol, which might be a much worse outcome than a deliberate design. For example, the ZCash Powers of Tau MPC ceremony needed to generate a number of elliptic curve points from a seed, so they just called ChaChaRng::from_seed, got a Rust Rand trait implementation, and passed it down to Fq::Rand. Then they asked me to reimplement the protocol in Go. The result was… [fun](https://github.com/FiloSottile/powersoftau/blob/e2af113817477d26e6155f1aae478d3cb58d62c5/powersoftau/hash_to_g2.go?ref=words.filippo.io#L13-L55).

PR #661, opened 2023-07-28
Implement rand_core traits on SeedStreamSha3

divergentdave:
This implements `RngCore` and `SeedableRng` from `rand_core` on `SeedStreamSha3`. See previous discussion on #578; this partially addresses #644. I made the `Default for Seed<SEED_SIZE>` implementation return a zeroed array, as we expect users to call this, mutably dereference the new seed, and then fill it with random bytes.

### DIFF
--- a/src/vdaf/prg.rs
+++ b/src/vdaf/prg.rs
@@ -19,6 +19,10 @@ use aes::{
 use cmac::{Cmac, Mac};
 #[cfg(feature = "crypto-dependencies")]
 use ctr::Ctr64BE;
+use rand_core::{
+    impls::{next_u32_via_fill, next_u64_via_fill},
+    RngCore, SeedableRng,
+};
 use sha3::{
     digest::{ExtendableOutput, Update, XofReader},
     CShake128, CShake128Core, CShake128Reader,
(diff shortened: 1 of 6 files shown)


src/vdaf/prg.rs
```diff
@@ -48,12 +52,24 @@ impl<const SEED_SIZE: usize> Seed<SEED_SIZE> {
     }
 }
 
+impl<const SEED_SIZE: usize> Default for Seed<SEED_SIZE> {
```
tgeoghegan, line 55: I'm worried that this is too easy to use wrong. You mention in the PR description that we expect clients to take `Seed::default` and then fill the inner buf with random bytes, but nothing stops them from working with the default seed.

We already have `Seed::from_bytes` that lets callers create a seed with random bytes they provide. Is that not good enough? Is the concern that it forces them to somehow instantiate a `[u8; SEED_SIZE]` and move it in? If so, we could provide `pub from_byte_iter<I = Iterator<u8>>(seed_bytes: I) -> Self` which would read `SEED_SIZE` bytes from the iterator (maybe we could refactor the existing `from_bytes` into something that takes `IntoIterator<Item = u8>`). Is that flexible enough?

src/vdaf/prg.rs
```diff
@@ -48,12 +52,24 @@ impl<const SEED_SIZE: usize> Seed<SEED_SIZE> {
...
+    }
+}
+
 impl<const SEED_SIZE: usize> AsRef<[u8; SEED_SIZE]> for Seed<SEED_SIZE> {
     fn as_ref(&self) -> &[u8; SEED_SIZE] {
         &self.0
     }
 }
 
+impl<const SEED_SIZE: usize> AsMut<[u8]> for Seed<SEED_SIZE> {
```
tgeoghegan, line 67: I understand that this exists to be used in conjunction with `Seed::default`. Are there risks to allowing mutation of the seed after it's initialized, or more to the point after some randomness has already been generated from it?

As an alternative to the `Seed::from_byte_iter` idea I sketched above, maybe what this calls for is a `SeedBuilder` struct, such that only `SeedBuilder` allows mutating the inner buffer and then after you `.build()` it into a `Seed`, all you can do is read the bytes.

tgeoghegan (changes requested):

src/vdaf/prg.rs
divergentdave: Another way around this would be to make the SeedableRng::Seed associated type just `[u8; 16]`. (more in line with other implementations) That'll let us avoid any changes to our `Seed` struct.

tgeoghegan (approved):
Looks good, and we can add new (i.e. API-compat-preserving) interfaces to make this easier to use as we learn about use cases.

PR #668, opened 2023-08-01
Performance optimizations for FLP

cjpatton:
There are a few features of the implementation I have felt for a while could be improved:
1. When constructing the "wire polynomials", we perform an identical FFT operation on each of the wires. Some of the logic in this operation is identical across all invocations.
2. We construct `Vec<Vec<F>>` to store the wire polynomials, where the length of the outer `Vec<_>` is equal to the arity of the gadget. This results in a lot of allocations for high-arity gadgets, like `ParallelSum`.
3. We allocate storage for the wire polynomials separately for each gadget. Since the underlying type is the same (`F`), it would be possible to reuse the same buffer.

This PR implements the following changes:
1. Add a SIMD-style FFT algorithm and deduplicate some of the arithmetic in proof/query generation
2. Modify the `prove()` / `query()` to operate on a flat buffer that is allocated once
3. **API-breaking**: Have `Gadget::call_poly()` operate on a flat buffer rather than a slice of `Vec<F>`'s

These changes result in marked performance improvement for the verifier (up to 20% for long inputs). However the prover did not improve very much, and in some cases regressed slightly (no more than 3%). Before merging this, I think we need to spend more time understanding how this changes the code that's generated. Specifically, are we *actually* reducing the number of allocations? Can valgrind tell us?

### DIFF
--- a/src/fft.rs
+++ b/src/fft.rs
@@ -33,11 +33,28 @@ pub fn discrete_fourier_transform<F: FftFriendlyFieldElement>(
     inp: &[F],
     size: usize,
 ) -> Result<(), FftError> {
-    let d = usize::try_from(log2(size as u128)).map_err(|_| FftError::SizeTooLarge)?;
+    multi_discrete_fourier_transform(&mut [outp], &[inp], size)
+}
 
-    if size > outp.len() {
-        return Err(FftError::OutputTooSmall);
-    }
+/// For each `i in 0..inp.len()`, set `outp[i]` to the DFT of `inp[i]`.
+///
+/// Interpreting the input as the coefficients of a polynomial, the output is equal to the input
+/// evaluated at points `p^0, p^1, ... p^(size-1)`, where `p` is the `2^size`-th principal root of
+/// unity.
+///
+/// Assumpes that `outp.len() == inp.len()` and, for each `i in 0..inp.len()`, `outp[i].len() ==
+/// inp[i].len()`.
+#[allow(clippy::many_single_char_names)]
+pub fn multi_discrete_fourier_transform<
+    F: FftFriendlyFieldElement,
+    In: AsRef<[F]>,
+    Out: AsMut<[F]>,
+>(
+    outp: &mut [Out],
+    inp: &[In],
+    size: usize,
+) -> Result<(), FftError> {
+    let d = usize::try_from(log2(size as u128)).map_err(|_| FftError::SizeTooLarge)?;
 
     if size > 1 << MAX_ROOTS {
         return Err(FftError::SizeTooLarge);
@@ -50,7 +67,13 @@ pub fn discrete_fourier_transform<F: FftFriendlyFieldElement>(
     #[allow(clippy::needless_range_loop)]
     for i in 0..size {
         let j = bitrev(d, i);
-        outp[i] = if j < inp.len() { inp[j] } else { F::zero() }
+        for (inp, outp) in inp
+            .iter()
+            .map(AsRef::as_ref)
+            .zip(outp.iter_mut().map(AsMut::as_mut))
+        {
+            outp[i] = if j < inp.len() { inp[j] } else { F::zero() }
+        }
     }
 
     let mut w: F;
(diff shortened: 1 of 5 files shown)


divergentdave:
Here's a tactic for getting Valgrind to count allocations for us.

1. Run `cargo test --bench cycle_counts --no-run` to do a debug build. (alternately run `cargo bench --bench cycle_counts` for a release build)
2. Copy the path of the executable that is output, and substitute it into this: `valgrind --tool=dhat target/release/deps/cycle_counts-6b580cf861a9c836 --iai-run 7`. The argument reaches into `iai`'s internals, and just runs the `prio3_client_count_vec_1000` benchmark, with very little setup.
3. Load the resulting output file into `dh_view.html`. (I had to copy the webapp's directory from `/usr/libexec/valgrind` to get around either snap or apparmor policies on Firefox)

Comparing before and after, I see a slight increase in the total number of instructions (comparing release builds), and a slight decrease in the number of allocations.

When sharding a single Prio3SumVec report with a length of 1000 bits, there are three different call sites that allocate 1000 blocks, which is highly suspicious. Looking at them in a run with a debug build (for its better stack traces, due to a lack of inlining) these allocations come from the following:

- Hashing the helper measurement share during sharding, to produce joint randomness parts, uses `<impl From<Field128> for Vec<u8>>::from()`, which creates a lot of allocation chaff.
- The same happens again when hashing the leader measurement share.
- `SumVec::encode_measurement()` calls `encode_into_bitvector_representation` on each vector element, which results in allocation of 1000 vectors, each of which gets appended into one pre-allocated vector.

We could eliminate these first two cases easily by pre-allocating a vector for each measurement share and encoding into it. If we had a Field method or conversion that encoded an element and returned an array with a generic length on the stack, we could use that to eliminate allocations altogether. Encoding could be improved if `encode_into_bitvector_representation` returned an `impl Iterator` type instead.

To do the same with verification, we'll just need to write a new `cycle_counts` microbenchmark, and update the benchmark indexes we run. We could grab some report shares from a test vector and put them in `hex_literal::hex()` to avoid sharding and preparing in the same benchmark.

src/fft.rs
divergentdave, line 1: The only parts of `multi_discrete_fourier_transform()` that get reused are index math and taking powers of the first root of unity. The vast majority of the field element arithmetic still has to be done separately for each input polynomial. I ran cachegrind on the prio3_client_count_vec_1000 benchmark, before and after this change, and processed the results with `cg_diff` and `cg_annotate`. The total number of instructions in `multi_discrete_fourier_transform()` exceeds the number of instructions in `discrete_fourier_transform()`, by 9.6e6 to 7.0e6. Data reads and writes are up by similar ratios, and data first-layer cache misses jumped by an order of magnitude, 1,555 to 24,005. Based on this, I think the changes to support interleaving multiple input polynomials must have negatively affected compiler optimizations enough to outweigh the computations that it was able to reuse. Plus, interleaving processing of multiple polynomials increases the working set size, and hurts cache locality.

There is an alternate optimization we could make to `discrete_fourier_transform()`: currently we load a root of unity in an outermost loop, and exponentiate it in a nested loop. We have already precomputed all the roots of unity, so we could instead do some power-of-two modular arithmetic and look up all powers of the root from the precomputed table. (we would want `F::root((l * i) % num_roots)`) If we were to make that change, then all field arithmetic in the function would be input-dependent.

Based on the above, I don't think the `multi_discrete_fourier_transform()` change will pay off.

cjpatton: Thanks @divergentdave for the thorough evaluation.

Prio2 previously had a lookup table that was computed when `Prio2` was constructed. The iterative FFT algorithm (without lookup table) beats the recursive FFT algorithm (with lookup table) in our benchmarks, but I don't know to what extend the lookup table contributed to the longer runtime. But my hunch is that we'd be better off having the lookup table in the program memory than on the heap. I.e., we should evaluate whether adding

```rust
const ROOTS_OF_UNITY: &[Field128; ROOTS_OF_UNITY_LEN] = [ ... ];
```

for each field would improve performance. Note that there are lots of situations in cryptographic software where end up wanting to store precomputed things. The only downside to this approach is that `ROOTS_OF_UNITY_LEN` may be pretty dang long?

divergentdave: We actually already have this lookup table, as `Field128::root()` looks up a pre-calculated Montgomery representation of the root from an array inside a `FieldParameters` const.

cjpatton: Right, but I understand your proposal to be to exapnd out the powers of those values as well?

divergentdave: Ah nevermind, I misunderstood that function and array. They're all principal roots of unity of different powers, not successive roots of unity of the largest power. That could indeed get large when dealing with big polynomials.

cjpatton: I think it would definitely worth being evaluating this approach, but I think we'd want to limit the size of the table, say to 10K or so. We can fall back to computing on the fly for larger inputs.

Filed https://github.com/divviup/libprio-rs/issues/707 and closing this PR.

PR #683, opened 2023-08-17
VDAF-06+ ping-pong topology

tgeoghegan:
Implements the ping-pong topology introduced in VDAF-06, though the
implementation here is based on the revised ping-pong interface in a
yet-unpublished VDAF version.

We add a `topology` module, on the premise that we might someday add new
topologies and their implementations there, and `topology::ping_pong`.

This also adds an implementation of a dummy VDAF, brought over from
Janus ([1]). `vdaf::dummy` is only compiled if the `test-util` Cargo
feature is enabled. The dummy VDAF implements the `vdaf::{Vdaf,
Aggregator, Client, Collector}` traits and provides associated types for output
shares, prepare shares, etc., but it doesn't do anything except return
success, making it useful for testing higher-level constructions like
ping-pong.

Finally, we replace the derived `std::fmt::Debug` implementations on a
few `prio3` and `poplar1` associated types so that they redact fields
that are either sensitive secrets or just too big to be worth printing
when debugging. This is so that we can provide `Debug` impls on new
types in `topology::ping_pong` without pulling in crate `derivative`,
which would require us to do 9,000+ lines of audits.

[1]: https://github.com/divviup/janus

### DIFF
--- a/src/topology/mod.rs
+++ b/src/topology/mod.rs
@@ -1,0 +1,7 @@
+// SPDX-License-Identifier: MPL-2.0
+
+//! Implementations of the aggregator communication topologies specified in [VDAF].
+//!
+//! [VDAF]: https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-vdaf-06#section-5.7
+
+pub mod ping_pong;
--- a/src/topology/mod.rs
+++ b/src/topology/mod.rs
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: MPL-2.0
 
-//! Implementations of the aggregator communication topologies specified in [VDAF].
+//! Implementations of some aggregator communication topologies specified in [VDAF].
 //!
 //! [VDAF]: https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-vdaf-06#section-5.7
 
--- a/src/topology/ping_pong.rs
+++ b/src/topology/ping_pong.rs
@@ -50,11 +50,11 @@ pub enum PingPongError {
     },
 
     /// Message from peer indicates it is in an unexpected state
-    #[error("peer message state mismatch: message is {found} expected {expected}")]
-    PeerMessageStateMismatch {
+    #[error("peer message mismatch: message is {found} expected {expected}")]
+    PeerMessageMismatch {
         /// The state in the message from the peer.
         found: &'static str,
-        /// The state the message from the peer was expected to be in.
+        /// The message expected from the peer.
         expected: &'static str,
     },
 
@@ -549,7 +549,7 @@ where
             Self::PrepareShare::get_decoded_with_param(&prep_state, prep_share)
                 .map_err(PingPongError::CodecPrepShare)?
         } else {
-            return Err(PingPongError::PeerMessageStateMismatch {
+            return Err(PingPongError::PeerMessageMismatch {
                 found: inbound.variant(),
                 expected: "initialize",
             });
@@ -604,7 +604,7 @@ where
 
         let (prep_msg, next_peer_prep_share) = match inbound {
             PingPongMessage::Initialize { .. } => {
-                return Err(PingPongError::PeerMessageStateMismatch {
+                return Err(PingPongError::PeerMessageMismatch {
                     found: inbound.variant(),
                     expected: "continued",
                 });
@@ -650,13 +650,16 @@ where
             (PrepareTransition::Finish(output_share), None) => {
                 Ok(PingPongContinuedValue::FinishedNoMessage { output_share })
             }
-            (transition, _) => {
-                return Err(PingPongError::HostStateMismatch {
-                    found: match transition {
-                        PrepareTransition::Continue(_, _) => "continue",
-                        PrepareTransition::Finish(_) => "finished",
-                    },
-                    expected: inbound.variant(),
+            (PrepareTransition::Continue(_, _), None) => {
+                return Err(PingPongError::PeerMessageMismatch {
+                    found: inbound.variant(),
+                    expected: "continue",
+                })
+            }
+            (PrepareTransition::Finish(_), Some(_)) => {
+                return Err(PingPongError::PeerMessageMismatch {
+                    found: inbound.variant(),
+                    expected: "finish",
                 })
             }
         }
--- a/src/topology/ping_pong.rs
+++ b/src/topology/ping_pong.rs
@@ -237,23 +237,6 @@ impl<
     }
 }
 
-impl<
-        const VERIFY_KEY_SIZE: usize,
-        const NONCE_SIZE: usize,
-        A: Aggregator<VERIFY_KEY_SIZE, NONCE_SIZE>,
-    > Default for PingPongTransition<VERIFY_KEY_SIZE, NONCE_SIZE, A>
-where
-    A::PrepareState: Default,
-    A::PrepareMessage: Default,
-{
-    fn default() -> Self {
-        Self {
-            previous_prepare_state: A::PrepareState::default(),
-            current_prepare_message: A::PrepareMessage::default(),
-        }
-    }
-}
-
 impl<const VERIFY_KEY_SIZE: usize, const NONCE_SIZE: usize, A> Encode
     for PingPongTransition<VERIFY_KEY_SIZE, NONCE_SIZE, A>
 where
--- a/src/topology/ping_pong.rs
+++ b/src/topology/ping_pong.rs
@@ -573,11 +573,11 @@ where
 
 #[cfg(test)]
 mod tests {
-    use assert_matches::assert_matches;
+    use std::io::Cursor;
 
     use super::*;
-
     use crate::vdaf::dummy;
+    use assert_matches::assert_matches;
 
     #[test]
     fn ping_pong_one_round() {
@@ -754,5 +754,109 @@ mod tests {
             .unwrap();
         // 3 round VDAF, round 2: leader should finish and emit no message.
         assert_matches!(leader_state, ContinuedValue::FinishedNoMessage { .. });
+    }
+
+    #[test]
+    fn roundtrip_message() {
+        let messages = [
+            (
+                Message::Initialize {
+                    prep_share: Vec::from("prepare share"),
+                },
+                concat!(
+                    "00", // enum discriminant
+                    concat!(
+                        // prep_share
+                        "0000000d",                   // length
+                        "70726570617265207368617265", // contents
+                    ),
+                ),
+            ),
+            (
+                Message::Continue {
+                    prep_msg: Vec::from("prepare message"),
+                    prep_share: Vec::from("prepare share"),
+                },
+                concat!(
+                    "01", // enum discriminant
+                    concat!(
+                        // prep_msg
+                        "0000000f",                       // length
+                        "70726570617265206d657373616765", // contents
+                    ),
+                    concat!(
+                        // prep_share
+                        "0000000d",                   // length
+                        "70726570617265207368617265", // contents
+                    ),
+                ),
+            ),
+            (
+                Message::Finish {
+                    prep_msg: Vec::from("prepare message"),
+                },
+                concat!(
+                    "02", // enum discriminant
+                    concat!(
+                        // prep_msg
+                        "0000000f",                       // length
+                        "70726570617265206d657373616765", // contents
+                    ),
+                ),
+            ),
+        ];
+
+        for (message, expected_hex) in messages {
+            let mut encoded_val = Vec::new();
+            message.encode(&mut encoded_val);
+            let got_hex = hex::encode(&encoded_val);
+            assert_eq!(
+                &got_hex, expected_hex,
+                "Couldn't roundtrip (encoded value differs): {message:?}",
+            );
+            let decoded_val = Message::decode(&mut Cursor::new(&encoded_val)).unwrap();
+            assert_eq!(
+                decoded_val, message,
+                "Couldn't roundtrip (decoded value differs): {message:?}"
+            );
+            assert_eq!(
+                encoded_val.len(),
+                message.encoded_len().expect("No encoded length hint"),
+                "Encoded length hint is incorrect: {message:?}"
+            )
+        }
+    }
+
+    #[test]
+    fn roundtrip_transition() {
+        // VDAF implementations have tests for encoding/decoding their respective PrepareShare and
+        // PrepareMessage types, so we test here using the dummy VDAF.
+        let transition = Transition::<0, 16, dummy::Vdaf> {
+            previous_prepare_state: dummy::PrepareState::default(),
+            current_prepare_message: (),
+        };
+
+        let encoded = transition.get_encoded();
+        let hex_encoded = hex::encode(&encoded);
+
+        assert_eq!(
+            hex_encoded,
+            concat!(
+                concat!(
+                    // previous_prepare_state
+                    "00",       // input_share
+                    "00000000", // current_round
+                ),
+                // current_prepare_message (0 length encoding)
+            )
+        );
+
+        let decoded = Transition::get_decoded_with_param(&(), &encoded).unwrap();
+        assert_eq!(transition, decoded);
+
+        assert_eq!(
+            encoded.len(),
+            transition.encoded_len().expect("No encoded length hint"),
+        );
     }
 }
--- a/src/topology/ping_pong.rs
+++ b/src/topology/ping_pong.rs
@@ -598,7 +598,7 @@ where
         } else {
             return Err(PingPongError::HostStateMismatch {
                 found: "finished",
-                expected: "continued",
+                expected: "continue",
             });
         };
 
@@ -606,7 +606,7 @@ where
             PingPongMessage::Initialize { .. } => {
                 return Err(PingPongError::PeerMessageMismatch {
                     found: inbound.variant(),
-                    expected: "continued",
+                    expected: "continue",
                 });
             }
             PingPongMessage::Continue {
--- a/src/topology/ping_pong.rs
+++ b/src/topology/ping_pong.rs
@@ -92,7 +92,7 @@ impl Message {
 
 impl Debug for Message {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "Message::{}", self.state_name())
+        f.debug_tuple(self.state_name()).finish()
     }
 }
 
@@ -163,7 +163,7 @@ impl Decode for Message {
 /// # Discussion
 ///
 /// The obvious implementation would of `ping_pong_transition` would be a method on trait
-/// [`PingPongTopology`] that returns `(State, Message)`, and then `StateAndMessage::WithMessage`
+/// [`PingPongTopology`] that returns `(State, Message)`, and then `ContinuedValue::WithMessage`
 /// would contain those values. But then DAP implementations would have to store relatively large
 /// VDAF prepare shares between rounds of input preparation.
 ///
@@ -195,6 +195,8 @@ impl<
         &self,
         vdaf: &A,
     ) -> Result<(State<VERIFY_KEY_SIZE, NONCE_SIZE, A>, Message), PingPongError> {
+        let prep_msg = self.current_prepare_message.get_encoded();
+
         vdaf.prepare_step(
             self.previous_prepare_state.clone(),
             self.current_prepare_message.clone(),
@@ -203,16 +205,13 @@ impl<
             PrepareTransition::Continue(prep_state, prep_share) => (
                 State::Continued(prep_state),
                 Message::Continue {
-                    prep_msg: self.current_prepare_message.get_encoded(),
+                    prep_msg,
                     prep_share: prep_share.get_encoded(),
                 },
             ),
-            PrepareTransition::Finish(output_share) => (
-                State::Finished(output_share),
-                Message::Finish {
-                    prep_msg: self.current_prepare_message.get_encoded(),
-                },
-            ),
+            PrepareTransition::Finish(output_share) => {
+                (State::Finished(output_share), Message::Finish { prep_msg })
+            }
         })
         .map_err(PingPongError::VdafPrepareStep)
     }
@@ -314,12 +313,12 @@ pub enum State<
 /// type `Finished`.
 ///
 /// VDAF describes `Start` and `Rejected` states, but the `Start` state is never instantiated in
-/// code, and the `Rejected` state is represented as [`std::result::Result::Error`], so this enum does
-/// not include those variants.
+/// code, and the `Rejected` state is represented as [`std::result::Result::Error`], so this enum
+/// does not include those variants.
 ///
 /// [VDAF]: https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-vdaf-06#section-5.8
 #[derive(Clone, Debug)]
-pub enum StateAndMessage<
+pub enum ContinuedValue<
     const VERIFY_KEY_SIZE: usize,
     const NONCE_SIZE: usize,
     A: Aggregator<VERIFY_KEY_SIZE, NONCE_SIZE>,
@@ -346,8 +345,8 @@ pub trait PingPongTopology<const VERIFY_KEY_SIZE: usize, const NONCE_SIZE: usize
 {
     /// Specialization of [`State`] for this VDAF.
     type State;
-    /// Specialization of [`StateAndMessage`] for this VDAF.
-    type StateAndMessage;
+    /// Specialization of [`ContinuedValue`] for this VDAF.
+    type ContinuedValue;
 
     /// Initialize leader state using the leader's input share. Corresponds to
     /// `ping_pong_leader_init` in the forthcoming `draft-irtf-cfrg-vdaf-07`.
@@ -368,8 +367,8 @@ pub trait PingPongTopology<const VERIFY_KEY_SIZE: usize, const NONCE_SIZE: usize
     /// Initialize helper state using the helper's input share and the leader's first prepare share.
     /// Corresponds to `ping_pong_helper_init` in the forthcoming `draft-irtf-cfrg-vdaf-07`.
     ///
-    /// If successful, the returned [`Transition`] should be evaluated, yielding a [`Message`], which
-    /// should be transmitted to the leader, and a [`State`].
+    /// If successful, the returned [`Transition`] should be evaluated, yielding a [`Message`],
+    /// which should be transmitted to the leader, and a [`State`].
     ///
     /// If the state is `State::Continued`, then it should be used by the helper along with the next
     /// `Message` received from the leader as input to [`Self::continued`] to advance to the next
@@ -396,20 +395,20 @@ pub trait PingPongTopology<const VERIFY_KEY_SIZE: usize, const NONCE_SIZE: usize
     /// peer. `role` is the host's [`Role`]. Corresponds to `ping_pong_contnued` in the forthcoming
     /// `draft-irtf-cfrg-vdaf-07`.
     ///
-    /// If successful, the returned [`StateAndMessage`] will either be:
+    /// If successful, the returned [`ContinuedValue`] will either be:
     ///
-    /// - `StateAndMessage::WithMessage { transition }`: `transition` should be evaluated, yielding
-    ///   a [`Message`], which should be transmitted to the peer, and a [`State`].
+    /// - `ContinuedValue::WithMessage { transition }`: `transition` should be evaluated, yielding a
+    ///   [`Message`], which should be transmitted to the peer, and a [`State`].
     ///
     ///   If the state is `State::Continued`, then it should be used by this aggregator along with
     ///   the next `Message` received from the peer as input to [`Self::continued`] to advance to
     ///   the next round. The aggregator may store the `Transition` between rounds of preparation
     ///   instead of the `State` and `Message`.
     ///
-    ///   If the state is `State::Finished`, then preparation is finished and the output share may be
-    ///   accumulated.
+    ///   If the state is `State::Finished`, then preparation is finished and the output share may
+    ///   be accumulated.
     ///
-    /// - `StateAndMessage::FinishedNoMessage`: preparation is finished and the output share may be
+    /// - `ContinuedValue::FinishedNoMessage`: preparation is finished and the output share may be
     ///   accumulated. No message needs to be sent to the peer.
     ///
     /// # Errors
@@ -430,7 +429,7 @@ pub trait PingPongTopology<const VERIFY_KEY_SIZE: usize, const NONCE_SIZE: usize
         role: Role,
         host_state: Self::State,
         inbound: &Message,
-    ) -> Result<StateAndMessage<VERIFY_KEY_SIZE, NONCE_SIZE, Self>, PingPongError>;
+    ) -> Result<ContinuedValue<VERIFY_KEY_SIZE, NONCE_SIZE, Self>, PingPongError>;
 }
 
 impl<const VERIFY_KEY_SIZE: usize, const NONCE_SIZE: usize, A>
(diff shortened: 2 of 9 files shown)


tgeoghegan:
I might still need to tweak this as the corresponding Janus change evolves, but I think this is ready to be reviewed. Note there's a corresponding PR to `draft-irtf-cfrg-vdaf` coming to clean up some spec issues encountered while implementing.

src/vdaf/dummy.rs
```diff
@@ -0,0 +1,327 @@
...
+pub struct Vdaf {
+    rounds: u32,
+    prep_init_fn: ArcPrepInitFn,
+    prep_step_fn: ArcPrepStepFn,
+}
+
+impl Debug for Vdaf {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("Vdaf")
+            .field("prep_init_result", &"[redacted]")
```
branlwyd, line 35: nit: I think these field names need to be updated (or dropped). IMO, also include `rounds`.

src/vdaf/dummy.rs
```diff
@@ -0,0 +1,327 @@
...
+        _nonce: &[u8; 16],
+        _: &Self::PublicShare,
+        input_share: &Self::InputShare,
+    ) -> Result<(Self::PrepareState, Self::PrepareShare), VdafError> {
+        (self.prep_init_fn)(aggregation_param)?;
+        Ok((
+            PrepareState {
+                input_share: input_share.0,
+                current_round: 0,
+                max_rounds: self.rounds,
```
branlwyd, line 139: Q: why do we need to track `max_rounds` in `prepare_state` -- couldn't we just have the default `prep_step_fn` use `rounds`? (i.e. replace `state.max_rounds` on L53 with `rounds`)

branlwyd (changes requested):
Other than the in-line comments, one higher-level comment:

Currently, in Janus' experimental implementation of this change (as well as in the pre-ping-pong aggregation implementation), the Leader does not store the next `prep_share` to send between steps of an aggregation job; we store only the `prep_state` & `prep_msg`, and compute the `prep_share` in-memory when the job is picked up to continue onto the next step. Effectively, the Janus implementation stops its aggregation job steps after the `prep_shares_to_prep` call inside `ping_pong_transition` inside `ping_pong_continued`, before the following `prep_next`.

I don't see a way to use these functions to implement these semantics: `PingPongTopologyPrivate::transition` will always compute the prep share, and it will be included in the `Message` that is returned. So if the Leader calls `ping_pong_continued` in the same aggregation job step, the Leader will end up storing an extra (Leader) `prep_share` to durable storage; if the Leader calls `ping_pong_continued` in the next aggregation job step, the Leader will end up storing an extra (Helper) `prep_share` to durable storage.

I'm not sure the best way to address this; I'd fairly-strongly prefer not to store unnecessary data to durable storage. OTOH, splitting `ping_pong_continued` up in a way that just-so-happens to be what Janus does might be too inflexible for other implementations. Janus could always choose not to use these functions at all, of course, but that would be somewhat wasteful of this generic ping-pong implementation.

edit: see David's https://github.com/divviup/janus/blob/6dbab170ca9176b55661a1607b0c5cf0593ea527/core/src/task/message_sizes.rs#L173-L236 for formulas on how much extra space this would use, per in-progress report aggregation -- N.B. these are only for Prio3, which will never end up storing an in-progress report aggregation since it will terminate in a single network request/aggregation job step, but may be useful to estimate how many bytes per report aggregation we waste by storing an unnecessary `prep_share`.

tgeoghegan:
Good points about the size of objects we store between prep steps. I'm going to do a comparison between what we store in a `ReportAggregationState` without ping-pong and in the draft Janus change I've got in https://github.com/divviup/janus/pull/1741, and then we'll see what the difference is, and whether we can add affordances to the `libprio` interfaces that let us thread the needle on storage requirements.

tgeoghegan:
Using [David's branch](https://github.com/divviup/janus/blob/6dbab170ca9176b55661a1607b0c5cf0593ea527/core/src/task/message_sizes.rs#L173-L236), I worked up a table of sizes for prepare messages, prepare message shares, prepare states and output shares for various instantiations of Prio3. 

|VDAF | Prepare message size (bytes) | Prepare message share size (bytes) | Prepare state size (bytes) | Output share size (bytes) |
| ---- | --------------- | --------------------- | ------------- | ------------ |
Prio3Count | 0 | 32 | 16 | 8
Prio3CountVec len 4 | 16 | 112 | 32 | 64
Prio3CountVec len 8 | 16 | 112 | 32 | 128
Prio3CountVec len 16 | 16 | 176 | 32 | 256
Prio3Sum 8 bits | 16 | 64 | 32 | 16
Prio3Histogram 2 | 16 | 64 | 32 | 32
Prio3Histogram 4 | 16 | 64 | 32 | 64
Prio3Histogram 8 | 16 | 64 | 32 | 128
Poplar1 4 round 0 | 24 | 24 | 30 | 8
Poplar1 4 round 1 | 0 | 8 | 14 | 8

Crucially we observe that message shares and output shares are much bigger than the other objects.

Then, by code inspection in Janus, we work out the worst-case storage usage for the leader and helper, depending on the number of rounds in the VDAF. As a reminder, the reason that finished helpers must store anything is usualy to record the most recent response to the leader to handle replays. In the case of a 1-round VDAF, that response consists of the prepare message.

x | Helper waiting | Helper finished | Leader waiting | Leader finished
-- | --- | --- | --- | ---
Before pong-pong | prepare state + prepare share | nothing (computed output share is accumulated) | prepare state + prepare message | nothing (computed output share is accumulated)
After ping-pong (1 round VDAF) | never waits | prepare message | never waits | nothing (computed output share is accumulated)
After ping-pong (2 round VDAF) | prepare state + prepare message + prepare share | nothing (computed output share is accumulated) | output share* + prepare message | nothing (computed output share is accumulated)

\* The waiting 2-round leader stores the output share because it does not accumulate until the helper has also finished

Now let's plug in sizes from an 8 bit `Prio3Histogram` to make this more tangible.


Prio3Histogram | Helper waiting storage (bytes) | Helper finished storage (bytes) | Leader waiting storage (bytes) | Leader finished storage (bytes)
-- | --- | --- | --- | ---
Before pong-pong | 96 | 0 | 48 | 0
After ping-pong (1 round VDAF) | n/a | 16 | n/a | 0
After ping-pong (2 round VDAF) | 112 (_or 48, see note 1_) | 0 | 208 (_or 48, see note 2_) | 0

For 1-round VDAFs, the leader and helper both come out ahead. However both aggregators lose in the 2-round VDAF case, because they have to store a prepare message from round _n-1_ and a prepare share from round _n_. Since many of these objects can be computed from each other, I think we can do better. 

_Note 1_: In this state, the helper stores:

- round 1 prepare state
- round 1 prepare message (which it just used to advance to round 1)
- round 2 helper prepare share (big)

It could instead store:

- round 0 prepare state
- round 1 prepare message

Then, it can recompute round 1 prepare state and round 2 helper prepare state.

_Note 2_: In this state, the leader stores:

- output share (big)
- round 2 prepare message

It could instead store:

- round 1 prepare state
- round 2 prepare message

I need to tinker with this, but I think that I can make the `StateAndMessage` enum store a transcript of recent rounds of the VDAF preparation such that it can either give you the results of the current round or give you a minimal but equivalent representation that occupies the least possible storage. I'll follow up here after I've played with this a bit.

tgeoghegan:
Put another way: I think I can refactor these `helper_initialize` and `continue` definitions so that they return an object that the caller can then use to either serialize state into a database or continue the VDAF preparation. That will provide the break inside of `ping_pong_transition` that Brandon describes that enables Janus to avoid storing unnecessary objects.

src/topology/ping_pong.rs
```diff
@@ -0,0 +1,682 @@
...
+        let prep_message = self
+            .prepare_preprocess(prep_shares)
+            .map_err(PingPongError::VdafPreparePreprocess)?;
+
+        self.prepare_step(host_prep_state, prep_message.clone())
+            .map(|transition| match transition {
+                PrepareTransition::Continue(prep_state, prep_share) => (
+                    State::Continued(prep_state),
+                    Message::Continue {
+                        prep_msg: prep_message.get_encoded(),
```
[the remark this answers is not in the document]
tgeoghegan: Unfortunately not possible given that I refactored this function into the `Transition` type.

src/topology/ping_pong.rs
```diff
@@ -0,0 +1,682 @@
...
+/// For example, it's impossible to be in the `Continued` state but to send an outgoing `Message` of
+/// type `Finished`.
+///
+/// VDAF describes `Start` and `Rejected` states, but the `Start` state is never instantiated in
+/// code, and the `Rejected` state is represented as `std::result::Result::Error`, so this enum does
+/// not include those variants.
+///
+/// [VDAF]: https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-vdaf-06#section-5.8
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum StateAndMessage<PrepareState, OutputShare> {
```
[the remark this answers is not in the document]
tgeoghegan: I tried `ContinuedValue` which does make the type's purpose more clear.

src/topology/ping_pong.rs
```diff
@@ -0,0 +1,682 @@
...
+    fn state_name(&self) -> &'static str {
+        match self {
+            Self::Initialize { .. } => "Initialize",
+            Self::Continue { .. } => "Continue",
+            Self::Finish { .. } => "Finish",
+        }
+    }
+}
+
+impl Debug for Message {
```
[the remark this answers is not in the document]
tgeoghegan: `debug_tuple()` with no fields seems to give us identical behavior to a derived `Debug` impl on an enum (https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=cd6865f58f73a9580f0f7a170e144ab5)

tgeoghegan (commented):
OK, I think I have it. I've refactored `PingPongTopology::transition` into a new `struct Transition` which contains the necessary values to compute `transition()`, namely a `PrepareState` and a `PrepareMessage`. Now, implementations that want to store state to a DB between preparation rounds can store the encoded `Transition`, while those that don't care can do `vdaf.continued()?.evaluate()?` and get the `(State, Message)` tuple as above.

Besides avoiding the the storage hit @branlwyd discussed, I think this approach will also allow Janus to stop dealing with the `report_aggregations.prep_msg` column (and maybe `last_prep_step`?), because the stores `Transition` now contains enough information to reconstruct it.

src/topology/ping_pong.rs
```diff
@@ -0,0 +1,682 @@
...
+                        prep_msg: prep_message.get_encoded(),
+                    },
+                ),
+            })
+            .map_err(PingPongError::VdafPrepareStep)
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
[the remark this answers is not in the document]
tgeoghegan: Acknowledging that I still need to write these tests (now it's `Message` and `Transition`).

src/topology/ping_pong.rs
```diff
@@ -0,0 +1,862 @@
+// SPDX-License-Identifier: MPL-2.0
+
+//! Implements the Ping-Pong Topology described in [VDAF]. This topology assumes there are exactly
+//! two aggregators, designated "Leader" and "Helper". Note that while this implementation is
+//! compatible with VDAF-06, it actually implements the ping-pong wrappers specified in VDAF-07
+//! (forthcoming) since those line up better with the VDAF implementation in this crate.
```
tgeoghegan, line 6: These references will be fixed up once VDAF-07 is published as part of https://github.com/divviup/libprio-rs/issues/701

src/vdaf.rs
```diff
@@ -210,7 +210,12 @@ pub trait Aggregator<const VERIFY_KEY_SIZE: usize, const NONCE_SIZE: usize>: Vda
     ///
     /// Decoding takes a [`Self::PrepareState`] as a parameter; this [`Self::PrepareState`] may be
     /// associated with any aggregator involved in the execution of the VDAF.
-    type PrepareMessage: Clone + Debug + ParameterizedDecode<Self::PrepareState> + Encode;
+    type PrepareMessage: Clone
+        + Debug
+        + PartialEq
+        + Eq
```
tgeoghegan, line 216: I added `ParialEq + Eq` here and on `type PrepareState` so that `topology::ping_pong::Transition` could in turn be `PartialEq + Eq`. The relevant types for `prio3` and `poplar1` are already `PartialEq + Eq` so I think this change is harmless, and may let us delete some trait bound noise from DAP impementations.

src/topology/ping_pong.rs
```diff
@@ -0,0 +1,862 @@
...
+}
+
+/// Extension trait on [`crate::vdaf::Aggregator`] which adds the [VDAF Ping-Pong Topology][VDAF].
+///
+/// [VDAF]: https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-vdaf-06#section-5.8
+pub trait PingPongTopology<const VERIFY_KEY_SIZE: usize, const NONCE_SIZE: usize>:
+    Aggregator<VERIFY_KEY_SIZE, NONCE_SIZE>
+{
+    /// Specialization of [`State`] for this VDAF.
+    type State;
```
branlwyd, line 336: nit: can we do `type State = State<VERIFY_KEY_SIZE, NONCE_SIZE, Self>`?

src/topology/ping_pong.rs
```diff
@@ -0,0 +1,862 @@
...
+    },
+}
+
+/// Extension trait on [`crate::vdaf::Aggregator`] which adds the [VDAF Ping-Pong Topology][VDAF].
+///
+/// [VDAF]: https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-vdaf-06#section-5.8
+pub trait PingPongTopology<const VERIFY_KEY_SIZE: usize, const NONCE_SIZE: usize>:
+    Aggregator<VERIFY_KEY_SIZE, NONCE_SIZE>
+{
+    /// Specialization of [`State`] for this VDAF.
```
branlwyd, line 335: If we're defining specializations here, we might also want to specialize `Transition` to save on some typing.

src/topology/ping_pong.rs
```diff
@@ -0,0 +1,862 @@
...
+    ///
+    ///
+    /// [VDAF]: https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-vdaf-06#section-5.8
+    /// [issue]: https://github.com/divviup/libprio-rs/issues/670
+    fn continued(
+        &self,
+        role: Role,
+        host_state: Self::State,
+        inbound: &Message,
+    ) -> Result<ContinuedValue<VERIFY_KEY_SIZE, NONCE_SIZE, Self>, PingPongError>;
```
branlwyd, line 423: nit: `Self::ContinuedValue` instead of `ContinuedValue<...>`?

branlwyd (commented):
> OK, I think I have it. I've refactored `PingPongTopology::transition` into a new `struct Transition` which contains the necessary values to compute `transition()`, namely a `PrepareState` and a `PrepareMessage`. Now, implementations that want to store state to a DB between preparation rounds can store the encoded `Transition`, while those that don't care can do `vdaf.continued()?.evaluate()?` and get the `(State, Message)` tuple as above.

I like this abstraction. The only note I have is that a DAP implementation's stored state type will now be called `Transition`; this transition can fallibly compute a `State`. This naming scheme might be a bit confusing to folks new to this API, but since we're stuck with `State` to match the VDAF spec, I think we can't do much better here. (raising this in case someone else has a better idea)


> Besides avoiding the the storage hit @branlwyd discussed, I think this approach will also allow Janus to stop dealing with the `report_aggregations.prep_msg` column, because the stores `Transition` now contains enough information to reconstruct it.

I think `report_aggregations.prep_state` and `report_aggregations.prep_msg` will be merged into something like `report_aggregations.transition`, since the new ping-pong API deals in types that merge the `prep_state` & `prep_msg`.

src/vdaf/dummy.rs
tgeoghegan: Moot because `rounds` was moved into the prep step closure.

branlwyd (approved):

src/topology/ping_pong.rs
tgeoghegan: That would be more clear, but you can't do that in a trait declaration because [`associated type defaults are unstable`](https://github.com/rust-lang/rust/issues/29661). We have to do it in the trait impl block instead. And since we provide a blanket `PingPongTopology` impl for all `vdaf::Aggregator` impls, it doesn't really matter to clients where these associated types are defined.

tgeoghegan (commented):
> I like this abstraction. The only note I have is that a DAP implementation's stored state type will now be called Transition; this transition can fallibly compute a State. This naming scheme might be a bit confusing to folks new to this API, but since we're stuck with State to match the VDAF spec, I think we can't do much better here. (raising this in case someone else has a better idea)

I agree. I think that when we decide how to slice this up into functions and types, there is a tension here between the objectives of doing something that is useful or convenient for implementations and doing something that lines up recognizably with the spec. If we want to do better here, I think we'd have to revisit the decision to put the ping-pong topology into VDAF. If it were in DAP, then a project like Janus has more freedom.

>I think report_aggregations.prep_state and report_aggregations.prep_msg will be merged into something like report_aggregations.transition, since the new ping-pong API deals in types that merge the prep_state & prep_msg.

Yes, that's what's panning out in https://github.com/divviup/janus/pull/1741 (which I have not yet updated to reflect the latest changes in this PR). I had hoped that `last_prep_step` could also go away, but the `Transition` doesn't quite capture everything we need there. I think we can do better with `last_prep_step`, but I'll pursue that discussion in the Janus PR, and in fact may punt improvements to `last_prep_step` to a later change since janus/1741 is going to be rather large and complex as it is.

src/topology/ping_pong.rs
```diff
@@ -0,0 +1,865 @@
...
+{
+    /// Specialization of [`State`] for this VDAF.
+    type State;
+    /// Specialization of [`ContinuedValue`] for this VDAF.
+    type ContinuedValue;
+    /// Specializaton of [`Transition`] for this VDAF.
+    type Transition;
+
+    /// Initialize leader state using the leader's input share. Corresponds to
+    /// `ping_pong_leader_init` in the forthcoming `draft-irtf-cfrg-vdaf-07`.
```
cjpatton, line 343: nit: For consistency, maybe
```suggestion
    /// `ping_pong_leader_init` in [VDAF]`.
```
and add the link below. Also, since we're cutting 07 imminently, drop "forthcoming".

Here and below.

src/topology/ping_pong.rs
```diff
@@ -0,0 +1,865 @@
...
+    Helper,
+}
+
+/// Corresponds to `struct Message` in [VDAF's Ping-Pong Topology][VDAF]. All of the fields of the
+/// variants are opaque byte buffers. This is because the ping-pong routines take responsibility for
+/// decoding preparation shares and messages, which usually requires having the preparation state.
+///
+/// [VDAF]: https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-vdaf-06#section-5.8
+#[derive(Clone, PartialEq, Eq)]
+pub enum Message {
```
cjpatton, line 63: Elsewhere we use distinctive names that reflect the module in which the type is defined.
```suggestion
pub enum PingPongMessage {
```

src/topology/ping_pong.rs
```diff
@@ -0,0 +1,865 @@
...
+    /// `State::Continued`) should be used by the leader along with the next [`Message`] received
+    /// from the helper as input to [`Self::continued`] to advance to the next round.
+    fn leader_initialize(
+        &self,
+        verify_key: &[u8; VERIFY_KEY_SIZE],
+        agg_param: &Self::AggregationParam,
+        nonce: &[u8; NONCE_SIZE],
+        public_share: &Self::PublicShare,
+        input_share: &Self::InputShare,
+    ) -> Result<(Self::State, Message), PingPongError>;
```
cjpatton, line 356: Why do we only define the method here, then implement it for `Aggregator` below? Couldn't we just make these derived methods?

src/topology/ping_pong.rs
```diff
@@ -0,0 +1,865 @@
...
+
+        Ok(Transition {
+            previous_prepare_state: prep_state,
+            current_prepare_message,
+        })
+    }
+
+    fn continued(
+        &self,
+        role: Role,
```
cjpatton, line 506: It's a bit awkward to use `Role` to distinguish roles here, but for initialization we the name of the method we call is used to distinguish. We should be consistent.

My suggestion would be to split `continued` into `leader_continued` and `helper_continued`, that way we don't have to expose `Role` in the public API.

src/topology/ping_pong.rs
```diff
@@ -0,0 +1,865 @@
...
+            )
+            .unwrap()
+            .evaluate(&helper)
+            .unwrap();
+
+        // 2 round VDAF, round 1: helper should continue.
+        assert_matches!(helper_state, State::Continued(_));
+
+        let leader_state = leader
+            .continued(Role::Leader, leader_state, &helper_message)
```
cjpatton, line 673: What about "initialize" and _"continue"_? Seems odd to have different parts of speech.

cjpatton (commented):
Excellent work. I see no issues in the implementation, all my comments are relatively minor. One high-level thing to carefully consider: Which parts of the new API surface are essential, and which aren't?

src/topology/ping_pong.rs
```diff
@@ -0,0 +1,865 @@
...
+        const VERIFY_KEY_SIZE: usize,
+        const NONCE_SIZE: usize,
+        A: Aggregator<VERIFY_KEY_SIZE, NONCE_SIZE>,
+    > Transition<VERIFY_KEY_SIZE, NONCE_SIZE, A>
+{
+    /// Evaluate this transition to obtain a new [`State`] and a [`Message`] which should be
+    /// transmitted to the peer.
+    #[allow(clippy::type_complexity)]
+    pub fn evaluate(
+        &self,
```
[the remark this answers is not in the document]
tgeoghegan: I'd like to do that, but that ends up being impractical. What I observed in Janus is that I need to have a live reference to both the `Transition` and the `Message` returned from `evaluate` so I can store the former in a database and transmit the latter to the peer. So if we made `evaluate` consume `self`, we wouldn't be getting rid of clones, we'd just be pushing them into the caller.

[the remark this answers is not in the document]
branlwyd: I think this is a good idea; there is currently some API impedance-mismatch in Janus' usage of this method that means we need to call `evaluate` and then still keep the `Transition` around, but I think we need to resolve that to avoid calling `evaluate` multiple times on the same `Transition`.

src/topology/ping_pong.rs
divergentdave: The alternative would be to provide default methods inside the extension trait definition. This might be feasible, since the trait definition and the blanket implementation have the same bounds on their generic parameters. Note that the body of the blanket implementation block would then just have associated type definitions. Six of one, half a dozen of the other, I think.

src/vdaf.rs
cjpatton: Why is `ReportAggregationState` `PartialEq`?

If `PrepareState` needs to be `PartialEq`, then we need to make sure the implementation is constant-time. That way we make it less likely that we leak bits of the state via a timing channel.

src/topology/ping_pong.rs
```diff
@@ -0,0 +1,865 @@
...
+use crate::{
+    codec::{decode_u32_items, encode_u32_items, CodecError, Decode, Encode, ParameterizedDecode},
+    vdaf::{Aggregator, PrepareTransition, VdafError},
+};
+use std::fmt::Debug;
+
+/// Errors emitted by this module.
+#[derive(Debug, thiserror::Error)]
+pub enum PingPongError {
+    /// Error running prepare_init
```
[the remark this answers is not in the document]
cjpatton: Thanks for that explanation, good enough for me.

src/topology/ping_pong.rs
```diff
@@ -0,0 +1,865 @@
...
+    A::PrepareState: Default,
+    A::PrepareMessage: Default,
+{
+    fn default() -> Self {
+        Self {
+            previous_prepare_state: A::PrepareState::default(),
+            current_prepare_message: A::PrepareMessage::default(),
+        }
+    }
+}
```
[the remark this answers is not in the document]
cjpatton: I see. It seems reasonable for this particular use case, but `Default` seems icky in general. I think of this trait as providing the instance for the struct that is "natural" for all use cases. (E.g., `false` is a good default value of `bool`, `0` is a good default for `u8`, etc.) 

If this is is intended for "test-util", then I'd put it behind "test-util" as well.

src/topology/ping_pong.rs
cjpatton: I'm not sure I understand David's explanation here. To be clear, instead of

```rust
trait PingPong {
  fn cool();
}

impl PingPong for Vdaf {
  fn cool() {
    println!("guy");
  }
}
```

I was suggesting something like this: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021

This seems a bit cleaner to me, but I'm happy to defer to you guys.

src/topology/ping_pong.rs
cjpatton: We could also update the spec, if you'd like to resolve the inconsistency that way!

src/topology/ping_pong.rs
cjpatton: Whoopsies, here's a link to the playground: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=7a9e4dd238dab4056dbee25312e3068c

src/topology/ping_pong.rs
```diff
@@ -0,0 +1,963 @@
...
+        } else {
+            return Err(PingPongError::StateMismatch("finished", "continue"));
+        };
+
+        let (prep_msg, next_peer_prep_share) = match inbound {
+            PingPongMessage::Initialize { .. } => {
+                return Err(PingPongError::StateMismatch(
+                    "continue",
+                    inbound.state_name(),
+                ));
```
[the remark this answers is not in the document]
tgeoghegan: Done; please let me know what enum variant and method names you'd prefer if this isn't satisfactory.

supply-chain/config.toml
```diff
@@ -187,6 +187,11 @@ criteria = "safe-to-deploy"
 version = "0.8.5"
 criteria = "safe-to-deploy"
 
+[[exemptions.rand_chacha]]
+version = "0.3.1"
+criteria = "safe-to-deploy"
+notes = "This dependency is only used if feature test-util is enabled, which should not be the case for production deployments."
```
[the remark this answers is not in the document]
tgeoghegan: Thanks; done!

src/topology/ping_pong.rs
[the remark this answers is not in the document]
tgeoghegan: There's no harm in calling `evaluate` multiple times, but I suppose I, too, like the idea of consuming `self` here. I'll tinker with this some more to see if I can make this work out.

src/topology/ping_pong.rs
tgeoghegan: Chris and I discussed offline and agreed that it's OK for the spec to say `leader_init` and this impl to say `leader_initialized`; conventions on abbreviations differ across languages and that's fine.

src/topology/ping_pong.rs
tgeoghegan: OK, so unpacking Chris' suggestion, there's two implications here.

One is that we could hoist the implementations of the `PingPongTopology` methods into the trait declaration instead of providing them in the blanket implementation. I think all that does is move some code around in the file and makes no observable difference to users of the crate (except, see next point).

The other implication is that if we remove the blanket impl of `PingPongTopology`, then individual VDAFs like `Prio3` and `Poplar1` have to opt into `PingPongTopology` by doing `impl PingPongTopology<generics> for MyCoolVdaf {}`.

So I think what we have to decide is:

Should VDAFs get ping-pong topology support automatically, for free, just by virtue of being VDAFs, **or** should a VDAF be required to explicitly opt into ping-pong (via an empty impl block)?

My 2c: the intent of the ping-pong topology in the spec is that it works for any VDAF, so a blanket implementation is appropriate. Let's also consider a future where there are VDAFs implemented outside of crate `prio`. If we add a new topology to `prio`, then it's not usable until each downstream crate that implements a VDAF does a release just to add an empty impl block. I think that just introduces extra delay and release chores for something like a DAP implementation that wants to use a new topology.

I think we should use the current implementation, because given the problem we have right now (one topology across two VDAFs, three if you count the dummy VDAF), then going one way or another makes no difference to users of the crate.

divergentdave: One further reason to keep the blanket impl: if there's no blanket impl, and the trait provides default implementations for each ping pong method, then other VDAFs could override the default methods with their own code when implementing the PingPongTopology trait, which is unnecessary flexibility and opens up the door for mistakes.

Assuming we still have a blanket impl, I agree the choice of where to put the code doesn't matter much.

src/topology/ping_pong.rs
```diff
@@ -0,0 +1,963 @@
...
+                found: "finished",
+                expected: "continued",
+            });
+        };
+
+        let (prep_msg, next_peer_prep_share) = match inbound {
+            PingPongMessage::Initialize { .. } => {
+                return Err(PingPongError::PeerMessageStateMismatch {
+                    found: inbound.variant(),
+                    expected: "continued",
```
divergentdave, line 609: nit: "continued" is a state, and "continue" is a message
```suggestion
                    expected: "continue",
```

src/vdaf.rs
cjpatton: Ack. I still think that comparing state should be constant time, but we can just file an issue for now.

src/topology/mod.rs
```diff
@@ -0,0 +1,7 @@
+// SPDX-License-Identifier: MPL-2.0
+
+//! Implementations of the aggregator communication topologies specified in [VDAF].
```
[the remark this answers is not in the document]
cjpatton: Bump

src/topology/mod.rs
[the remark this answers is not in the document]
cjpatton: Assuming you want to just ack this and resolve it.

src/vdaf.rs
tgeoghegan: https://github.com/divviup/libprio-rs/issues/722

src/topology/ping_pong.rs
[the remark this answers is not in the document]
tgeoghegan: I do not think we can avoid calling `evaluate` twice, at least not without forcing a helper to hang on to more data when it's in the waiting state. The point of `Transition` is to allow users of `prio` to choose whether to spend more compute (by calling `evaluate` twice) or more storage space (by calling `evaluate` once and then having to store relatively large prepare message shares).

As for making this method consume `self`: I don't think it's worth it. Doing that forces a lot of awkward clone gymnastics into the caller instead of doing them here. In the case where the `Transition` is dropped right away, I suspect the compiler will optimize the clones away anyway. Finally, there's no harm in calling `evaluate` more than once on the same `Transition`: you'll just get the exact same result back. So I'm not certain we're improving the ergonomics or safety of the API by doing that.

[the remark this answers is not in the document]
cjpatton: For Prio3 the PrepState includes the output share (IIRC). For something like Prio3SumVec, this may be quite large, perhaps a megabyte or so. I think it's worth figuring out what the compiler does and if we can easily avoid a copy here.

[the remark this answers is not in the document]
cjpatton: Correction: It includes the measurement share, which is short for the Helper (just a seed) and long for the Leader. For the Leader we can probably optimize this a bit more by only storing the truncated measurement share, but that still doesn't help much for Prio3SumVec.

divergentdave (approved):

PR #697, opened 2023-08-24
Rewrite Criterion benchmarks

divergentdave:
This rewrites the Criterion benchmarks for Prio2 and Prio3. Benchmarks operating at the FLP level are removed, and we instead benchmark each VDAF through the VDAF interface. Benchmarks are also added for `prepare_init()`, in addition to sharding. I put the multithreaded implementations of VDAFs in the same groups as their respective serial implementations, where applicable, so that the reports will give us plots comparing the two.

### DIFF
--- a/benches/speed_tests.rs
+++ b/benches/speed_tests.rs
@@ -1,32 +1,26 @@
 // SPDX-License-Identifier: MPL-2.0
 
-use criterion::{criterion_group, criterion_main, BenchmarkId, Criterion};
+use criterion::{black_box, criterion_group, criterion_main, BenchmarkId, Criterion};
 #[cfg(feature = "experimental")]
 use criterion::{BatchSize, Throughput};
 #[cfg(feature = "experimental")]
+use fixed::types::{I1F15, I1F31};
+#[cfg(feature = "experimental")]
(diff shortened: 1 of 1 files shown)


benches/speed_tests.rs
```diff
@@ -83,177 +82,430 @@ pub fn poly_mul(c: &mut Criterion) {
...
-        let prio3 = Prio3::new_fixedpoint_boundedl2_vec_sum(num_shares, len).unwrap();
-        let fp_num = fixed!(0.0001: I1F15);
-        let measurement = vec![fp_num; len];
-        group.bench_function(BenchmarkId::new("fixedpoint16_boundedl2_vec", len), |b| {
-            b.iter(|| {
-                prio3.shard(&measurement, &nonce).unwrap();
-            })
-        });
+    let mut group = c.benchmark_group("prio3histogram_shard");
+    for input_len in [10, 100, 1_000] {
```
inahga, line 266: IIRC our practical usage of Prio3Histogram includes larger histogram sizes, approaching 10^4 and 10^5. Could consider benchmarking larger lengths?

tgeoghegan (approved):

PR #704, opened 2023-08-25
Add method on `SeedStream` for deriving a vector

cjpatton:
The "expand_into_vec()" function in the VDAF spec is used to expand a seed into a vector of field elements. The `prio` crate has an internal implementation of this feature (it's needed for Prio3 and Poplar1), but no method is exposed in the public API. This change adds a derived method `SeedStream::into_vec()` that implements the "expand_into_vec()" functionality of the spec.

The method consumes a seed stream and outputs a vector of the desired length. The seed stream is consumed because of the implementation. `Prng` is used, which consumes the seed stream so that it can internally buffer the output. A non-consuming method can be implemented ("next_vec()" in the VDAF spec) in the future if needed.

While at it, this change simplifies some of the `Prio3` internals by using the new method.

### DIFF
--- a/src/vdaf/prg.rs
+++ b/src/vdaf/prg.rs
@@ -4,7 +4,11 @@
 //!
 //! [draft-irtf-cfrg-vdaf-06]: https://datatracker.ietf.org/doc/draft-irtf-cfrg-vdaf/06/
 
-use crate::vdaf::{CodecError, Decode, Encode};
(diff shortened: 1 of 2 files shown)


divergentdave (approved):

cjpatton:
Rebaed.

divergentdave (approved):

PR #721, opened 2023-08-31
Remove PrgAes128 and Field96

divergentdave:
This removes PrgAes128 and Field96, which have been deprecated for a while. We still keep SeedStreamAes128 for use in Prio2.


cjpatton (commented):
I would probably hold off on removing Field96. There may be use cases for a "happy medium" between 64 and 128. Of course, we could always re-generate when we need it.

tgeoghegan (approved):
We can always fish `Field96` out of git history should we need it again. In the meantime, the code is just a drag.

PR #734, opened 2023-09-06
Remove non-constant-time comparisons of secret values.

### LINKED ISSUE #722: Comparing `Vdaf::PrepareState` in constant time
Users of `prio` need to compare various objects related to VDAFs such as `Vdaf::PrepareState` to one another. For example, to check whether a given state transition occurred as expected in a test, or to check whether a database row already contains some incoming value.

In test settings, it's probably fine to use derived `PartialEq + Eq` implementations which will boil down to comparing slices of bytes. But in other contexts, we need to make sure that comparisons are done in constant time.

We could address this by auditing the `PartialEq` implementations on each VDAF implementation's associated types, replacing them with hand-written, constant time verisons as needed. But the risk there is that new VDAFs will use derived `PartialEq` implementations that aren't constant time. So perhaps we would want to remove all `PartialEq + Eq` trait bounds and instead define some custom `ConstantTimeEq` trait.

_Originally posted by @cjpatton in https://github.com/divviup/libprio-rs/pull/683#discussion_r1312065005_

branlwyd:
This affects the types representing:
  * Input shares
  * Preparation states
  * Output shares
  * Aggregate shares

(and their constituent types)

Mostly, the comparisons were either dropped entirely or updated to be test-only.  Input shares were instead given a constant-time equality implementation, as it is believed this is required for DAP implementations.

PartialEq & Eq implementations are still derived when the "test-util" feature is enabled. This is to ease testing for users of this library.

Closes #722.

### DIFF
--- a/src/vdaf.rs
+++ b/src/vdaf.rs
@@ -59,8 +59,6 @@ pub enum VdafError {
 
 /// An additive share of a vector of field elements.
 #[derive(Clone, Debug)]
-// Only derive equality checks in test code, as the content of this type is a secret.
-#[cfg_attr(feature = "test-util", derive(PartialEq, Eq))]
 pub enum Share<F, const SEED_SIZE: usize> {
     /// An uncompressed share, typically sent to the leader.
     Leader(Vec<F>),
@@ -80,6 +78,14 @@ impl<F: Clone, const SEED_SIZE: usize> Share<F, SEED_SIZE> {
         }
     }
 }
+
+impl<F: ConstantTimeEq, const SEED_SIZE: usize> PartialEq for Share<F, SEED_SIZE> {
+    fn eq(&self, other: &Self) -> bool {
+        self.ct_eq(other).into()
+    }
+}
+
+impl<F: ConstantTimeEq, const SEED_SIZE: usize> Eq for Share<F, SEED_SIZE> {}
 
 impl<F: ConstantTimeEq, const SEED_SIZE: usize> ConstantTimeEq for Share<F, SEED_SIZE> {
     fn ct_eq(&self, other: &Self) -> subtle::Choice {
@@ -355,10 +361,22 @@ impl<F: FieldElement> Encode for OutputShare<F> {
 /// This is suitable for VDAFs where both output shares and aggregate shares are vectors of field
 /// elements, and output shares need no special transformation to be merged into an aggregate share.
 #[derive(Clone, Debug, Serialize, Deserialize)]
-// Only derive equality checks in test code, as the content of this type is a secret.
-#[cfg_attr(feature = "test-util", derive(PartialEq, Eq))]
 
 pub struct AggregateShare<F>(Vec<F>);
+
+impl<F: ConstantTimeEq> PartialEq for AggregateShare<F> {
+    fn eq(&self, other: &Self) -> bool {
+        self.ct_eq(other).into()
+    }
+}
+
+impl<F: ConstantTimeEq> Eq for AggregateShare<F> {}
+
+impl<F: ConstantTimeEq> ConstantTimeEq for AggregateShare<F> {
+    fn ct_eq(&self, other: &Self) -> subtle::Choice {
+        self.0.ct_eq(&other.0)
+    }
+}
 
 impl<F: FieldElement> AsRef<[F]> for AggregateShare<F> {
     fn as_ref(&self) -> &[F] {
--- a/src/vdaf.rs
+++ b/src/vdaf.rs
@@ -18,6 +18,7 @@ use crate::{
 };
 use serde::{Deserialize, Serialize};
 use std::{fmt::Debug, io::Cursor};
+use subtle::{Choice, ConstantTimeEq};
 
 /// A component of the domain-separation tag, used to bind the VDAF operations to the document
 /// version. This will be revised with each draft with breaking changes.
@@ -57,7 +58,9 @@ pub enum VdafError {
 }
 
 /// An additive share of a vector of field elements.
-#[derive(Clone, Debug, Eq, PartialEq)]
+#[derive(Clone, Debug)]
+// Only derive equality checks in test code, as the content of this type is a secret.
+#[cfg_attr(feature = "test-util", derive(PartialEq, Eq))]
 pub enum Share<F, const SEED_SIZE: usize> {
     /// An uncompressed share, typically sent to the leader.
     Leader(Vec<F>),
@@ -74,6 +77,18 @@ impl<F: Clone, const SEED_SIZE: usize> Share<F, SEED_SIZE> {
         match self {
             Self::Leader(ref data) => Self::Leader(data[..len].to_vec()),
             Self::Helper(ref seed) => Self::Helper(seed.clone()),
+        }
+    }
+}
+
+impl<F: ConstantTimeEq, const SEED_SIZE: usize> ConstantTimeEq for Share<F, SEED_SIZE> {
+    fn ct_eq(&self, other: &Self) -> subtle::Choice {
+        // We allow short-circuiting on the type (Leader vs Helper) of the value, but not the types'
+        // contents.
+        match (self, other) {
+            (Share::Leader(self_val), Share::Leader(other_val)) => self_val.ct_eq(other_val),
+            (Share::Helper(self_val), Share::Helper(other_val)) => self_val.ct_eq(other_val),
+            _ => Choice::from(0),
         }
     }
 }
@@ -310,7 +325,7 @@ pub trait Aggregatable: Clone + Debug + From<Self::OutputShare> {
 }
 
 /// An output share comprised of a vector of field elements.
-#[derive(Clone, Debug, PartialEq, Eq)]
+#[derive(Clone, Debug)]
 pub struct OutputShare<F>(Vec<F>);
 
 impl<F> AsRef<[F]> for OutputShare<F> {
@@ -339,7 +354,10 @@ impl<F: FieldElement> Encode for OutputShare<F> {
 ///
 /// This is suitable for VDAFs where both output shares and aggregate shares are vectors of field
 /// elements, and output shares need no special transformation to be merged into an aggregate share.
-#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[derive(Clone, Debug, Serialize, Deserialize)]
+// Only derive equality checks in test code, as the content of this type is a secret.
+#[cfg_attr(feature = "test-util", derive(PartialEq, Eq))]
+
 pub struct AggregateShare<F>(Vec<F>);
 
 impl<F: FieldElement> AsRef<[F]> for AggregateShare<F> {
--- a/src/vdaf.rs
+++ b/src/vdaf.rs
@@ -334,6 +334,20 @@ pub trait Aggregatable: Clone + Debug + From<Self::OutputShare> {
 #[derive(Clone, Debug)]
 pub struct OutputShare<F>(Vec<F>);
 
+impl<F: ConstantTimeEq> PartialEq for OutputShare<F> {
+    fn eq(&self, other: &Self) -> bool {
+        self.ct_eq(other).into()
+    }
+}
+
+impl<F: ConstantTimeEq> Eq for OutputShare<F> {}
+
+impl<F: ConstantTimeEq> ConstantTimeEq for OutputShare<F> {
+    fn ct_eq(&self, other: &Self) -> Choice {
+        self.0.ct_eq(&other.0)
+    }
+}
+
 impl<F> AsRef<[F]> for OutputShare<F> {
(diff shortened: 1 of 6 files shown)


branlwyd:
A few comparisons in constant-time equality implementations were left non-constant-time; only the content of secret values was considered as required for constant-time comparisons. Specifically:

* The `Share` type's `ConstantTimeEq` implementation will short-circuit on the type of the share (Leader vs Helper); this is thought to be OK since the content of the secrets inside is still compared in constant time.
* The `Prio3InputShare` type's `ConstantTimeEq` implementation will short-circuit on the presence/absence of `joint_rand_blind`; since this is a property of the type of VDAF in use, it is not considered a secret. The content (if any) of `joint_rand_blind` is compared in constant time, as well as all other fields.

Please verify these assertions as part of review.

cjpatton:
@branlwyd that seems like the right definition. Basically what our goal is to make runtime depend as little as possible on secrets. For other things the side channel is not important. We don't much care if the attacker knows that the party doing the computation is the Leader or Helper, for example.

src/vdaf/poplar1.rs
```diff
@@ -128,6 +130,24 @@ pub struct Poplar1InputShare<const SEED_SIZE: usize> {
...
 
+impl<const SEED_SIZE: usize> ConstantTimeEq for Poplar1InputShare<SEED_SIZE> {
+    fn ct_eq(&self, other: &Self) -> Choice {
+        // We short-circuit on the length of corr_inner being different. Only the content is
+        // protected.
+        if self.corr_inner.len() != other.corr_inner.len() {
+            return Choice::from(0);
+        }
+
+        let mut rslt = self.idpf_key.ct_eq(&other.idpf_key)
```
cjpatton, line 141: nit: I think elsewhere we're using `res` or `result` for a generic result. (I haven't checked.)
```suggestion
        let mut res = self.idpf_key.ct_eq(&other.idpf_key)
```

src/vdaf.rs
```diff
@@ -57,7 +58,9 @@ pub enum VdafError {
 }
 
 /// An additive share of a vector of field elements.
-#[derive(Clone, Debug, Eq, PartialEq)]
+#[derive(Clone, Debug)]
+// Only derive equality checks in test code, as the content of this type is a secret.
+#[cfg_attr(feature = "test-util", derive(PartialEq, Eq))]
```
cjpatton, line 63: I'm concerned that this API might be too permissive. The intent is to fence the unsafe comparison to testing, but IIUC, the user can still shoot themself in the foot if they're not careful. Let's say I enable "test-util" in my crate that I also deploy to production; then I get to use `Eq` in my production code, right?

cjpatton (commented):
Implementations of `ConstantTimeEq` look good. I'm a bit concerned about API safety, see inline comment.

src/vdaf/prio3.rs
```diff
@@ -620,6 +621,22 @@ impl<const SEED_SIZE: usize> Encode for Prio3PublicShare<SEED_SIZE> {
     }
 }
 
+impl<const SEED_SIZE: usize> PartialEq for Prio3PublicShare<SEED_SIZE> {
+    fn eq(&self, other: &Self) -> bool {
+        // Handle case that both join_rand_parts are populated.
+        if let Some(self_joint_rand_parts) = &self.joint_rand_parts {
```
divergentdave, line 627: I think this would be easier to reason about if it was written as a match block on `(&self.joint_rand_parts, &other.joint_rand_parts)`.

Moreover, since the `(Some(_), Some(_))` branch is already using `ConstantTimeEq`, we could implement `ConstantTimeEq for Prio3PublicShare<SEED_SIZE>`, then have this `PartialEq` dispatch to it and convert the `Choice` to a `bool`.

src/vdaf/prio3.rs
```diff
@@ -1111,7 +1154,13 @@ where
     ) -> Result<PrepareTransition<Self, SEED_SIZE, 16>, VdafError> {
         if self.typ.joint_rand_len() > 0 {
             // Check that the joint randomness was correct.
-            if step.joint_rand_seed.as_ref().unwrap() != msg.joint_rand_seed.as_ref().unwrap() {
+            if (!step
+                .joint_rand_seed
+                .as_ref()
+                .unwrap()
+                .ct_eq(msg.joint_rand_seed.as_ref().unwrap()))
```
divergentdave, line 1161: FYI there's a provided `ct_ne()` method, which will let us drop the separate `!`.

I think sidechannels from this comparison are less sensitive because it always passes if the client and aggregators are honest, all aggregators know the joint randomness seed if the client is honest, and the client's report shares determine both sides of the comparison. We may as well keep it constant time for peace of mind.

src/vdaf.rs
divergentdave: One way we could resolve this is to write `ConstantTimeEq` implementations for the relevant structs/enums, and write `PartialEq` implementations that dispatch to them. (for an example of this pattern, see `IdpfPublicShare`) This would provide safe APIs for both downstream dependents and for test code, avoid the need for the `test-util` Cargo feature, and still give us the convenience benefit of being able to use `assert_eq!()`.

src/vdaf/poplar1.rs
```diff
@@ -128,6 +130,24 @@ pub struct Poplar1InputShare<const SEED_SIZE: usize> {
...
+            & self.corr_seed.ct_eq(&other.corr_seed)
+            & self.corr_leaf.ct_eq(&other.corr_leaf);
+        for (x, y) in self.corr_inner.iter().zip(other.corr_inner.iter()) {
```
divergentdave, line 144: We can avoid manually checking lengths and iterating by using the blanket [`impl<T: ConstantTimeEq> ConstantTimeEq for [T]`](https://docs.rs/subtle/latest/subtle/trait.ConstantTimeEq.html#impl-ConstantTimeEq-for-%5BT%5D). Just as the code here does, it short-circuits on slice length. Then, this method body could be simplified to a single four-way `&` expression.

src/vdaf.rs
cjpatton: That sounds reasonable to me.

branlwyd: The concern that led to this implementation was (from https://github.com/divviup/libprio-rs/issues/722):
> But the risk there is that new VDAFs will use derived PartialEq implementations that aren't constant time. So perhaps we would want to remove all PartialEq + Eq trait bounds and instead define some custom ConstantTimeEq trait.

That is, if we expose `PartialEq + Eq` by default (and expect users of the library to use these traits for comparison), we risk future VDAFs implementing variable-time versions of these traits. This would lead to any DAP implementations built on these VDAFs silently becoming insecure, in a way that would be particularly hard to notice.

OTOH, I like the idea of safe-by-default implementations and especially not requiring callers to know about `subtle` at all to use the library. I have moved everything relevant over to a constant-time implmementation of `PartialEq + Eq`.


> Let's say I enable "test-util" in my crate that I also deploy to production; then I get to use Eq in my production code, right?

Correct; if you enable a feature called "test-util" in your production binaries, IMO you are responsible for what comes next. :)

src/vdaf/poplar1.rs
branlwyd: I tried this, unfortunately it doesn't work with an error like:

```
error[E0599]: the method `ct_eq` exists for struct `Vec<[Field64; 2]>`, but its trait bounds were not satisfied
   --> src/vdaf/poplar1.rs:159:31
    |
159 |             & self.corr_inner.ct_eq(&other.corr_inner)
    |                               ^^^^^ method cannot be called on `Vec<[Field64; 2]>` due to unsatisfied trait bounds
    |
    = note: the following trait bounds were not satisfied:
            `[Field64; 2]: ConstantTimeEq`
            which is required by `[[Field64; 2]]: ConstantTimeEq`
```

I _think_ the reason `[Field64; 2]` doesn't satisfy `ConstantTimeEq` is that it can't be automatically converted to a `[Field64]`. `subtle`'s `const-generics` feature seems like it should help -- but it actually only provides a blanket implementation of `ConditionallySelectable` for `[T; N]`, no `ConstantTimeEq` implementation. I don't see why one could be written, but this would require waiting on a release of the `subtle` crate.

branlwyd:
I added tests for all of the manually-implemented equality traits. They test `PartialEq`, but since all of the relevant `PartialEq` implementations fall through to `ConstantTimeEq`, this is sufficient to cover both traits.

src/vdaf/poplar1.rs
branlwyd: I filed https://github.com/dalek-cryptography/subtle/pull/114, we'll see if it lands. The last release was 6 months ago so we may be waiting awhile in any case.

src/vdaf/prio3.rs
```diff
@@ -620,6 +621,18 @@ impl<const SEED_SIZE: usize> Encode for Prio3PublicShare<SEED_SIZE> {
     }
 }
 
+impl<const SEED_SIZE: usize> PartialEq for Prio3PublicShare<SEED_SIZE> {
+    fn eq(&self, other: &Self) -> bool {
+        match (&self.joint_rand_parts, &other.joint_rand_parts) {
```
divergentdave, line 626: nit: This could use `option_ct_eq()` as well.

src/vdaf/poplar1.rs
```diff
@@ -308,6 +381,44 @@ enum SketchState<F> {
...
+                    is_leader: self_is_leader,
+                },
+                SketchState::RoundOne {
+                    A_share: other_a_share,
+                    B_share: other_b_share,
+                    is_leader: other_is_leader,
+                },
+            ) => {
+                let self_is_leader = Choice::from(*self_is_leader as u8);
+                let other_is_leader = Choice::from(*other_is_leader as u8);
```
cjpatton, line 409: Short-circuiting on the is_leader should be fine, too. In fact, I would do this just to make clear to the reader what is secret and what is not.

src/vdaf/prio3.rs
```diff
@@ -726,8 +755,9 @@ where
     }
 }
 
-#[derive(Clone, Debug, Eq, PartialEq)]
+#[derive(Clone, Debug)]
 /// Message broadcast by each [`Aggregator`] in each round of the Preparation phase.
+
```
divergentdave, line 760: nit: extra line
```suggestion
```

src/vdaf/prio3.rs
```diff
@@ -620,6 +621,18 @@ impl<const SEED_SIZE: usize> Encode for Prio3PublicShare<SEED_SIZE> {
     }
 }
 
+impl<const SEED_SIZE: usize> PartialEq for Prio3PublicShare<SEED_SIZE> {
+    fn eq(&self, other: &Self) -> bool {
```
cjpatton, line 625: For consistency, note that you short-circuit on whether joint randomness is used or not.

divergentdave (approved):

src/vdaf/prio3.rs
```diff
@@ -620,6 +621,18 @@ impl<const SEED_SIZE: usize> Encode for Prio3PublicShare<SEED_SIZE> {
...
+        match (&self.joint_rand_parts, &other.joint_rand_parts) {
+            (Some(self_jrps), Some(other_jrps)) => self_jrps.ct_eq(other_jrps).into(),
```
cjpatton, line 627: For consistency and readability, s/jrps/joint_rand_parts/g

cjpatton (approved):
Thanks for putting your back into making this API-safe. It was a lot more work than we initially expected, but I think this is a really great improvement. Two high-level comments:

1. For consistency, I think we should short-circuit any comparisons that don't need to be constant time. The only things that need to be constant time are secrets and secret shares. (See inline comments for specific suggestions.)
2. It's important to keep in mind that constant time code is "best effort" and doesn't guarantee that side channels are eliminated. One day some academic lab might take a radio receiver or electron microscope to libprio-rs and find they can pull off an attack if both Aggregators are on old Pentium laptops in the same room or something. The documentation and comments should reflect that we have done our best to mitigate some timing side channels, but others may exist. I don't have a specific suggestion here, so maybe this is just something to keep in mind.

branlwyd:
> For consistency, I think we should short-circuit any comparisons that don't need to be constant time. The only things that need to be constant time are secrets and secret shares. (See inline comments for specific suggestions.)

Done, but I'm relying on y'all for which fields specifically are safe to compare in variable-time.

> It's important to keep in mind that constant time code is "best effort" and doesn't guarantee that side channels are eliminated. One day some academic lab might take a radio receiver or electron microscope to libprio-rs and find they can pull off an attack if both Aggregators are on old Pentium laptops in the same room or something. The documentation and comments should reflect that we have done our best to mitigate some timing side channels, but others may exist. I don't have a specific suggestion here, so maybe this is just something to keep in mind.

Agreed; I think this should go in something like a "security considerations" section, once one is written.

PR #736, opened 2023-09-06
xof: Replace `SeedStream` trait with `RngCore`

### LINKED ISSUE #644: Migrate from SeedStream to RngCore

cjpatton:
Closes #644.

Most of our `SeedStream` implementors also implement `RngCore`, which requires a method with the same semantics as `SeedStream::fill()`. Further, the blanket trait `Rng` provides a method with the same name that does pretty much the same thing (but is slightly more generic).

To avoid confusion going forward, drop `SeedSstream` and use `RngCore` instead. Accordingly:

* Implement `RngCore` for `SeedStreamFixedKeyAes128` (the only `SeedStream` that does not already implement `RngCore`)

* Move `SeedStream::into_vec()` it its own trait, `IntoVec`, and provide a blanket implementation for `RngCore`.

### DIFF
--- a/src/vdaf/xof.rs
+++ b/src/vdaf/xof.rs
@@ -91,13 +91,14 @@ impl<const SEED_SIZE: usize> Decode for Seed<SEED_SIZE> {
     }
 }
 
-/// A stream of bytes derived from a seed.
-pub trait SeedStream: Sized {
-    /// Fill `buf` with the next `buf.len()` bytes of output.
-    fn fill(&mut self, buf: &mut [u8]);
-
+/// Trait for deriving a vector of field elements.
+pub trait IntoFieldVec: RngCore + Sized {
     /// Generate a finite field vector from the seed stream.
-    fn into_vec<F: FieldElement>(self, length: usize) -> Vec<F> {
+    fn into_field_vec<F: FieldElement>(self, length: usize) -> Vec<F>;
+}
+
+impl<S: RngCore> IntoFieldVec for S {
+    fn into_field_vec<F: FieldElement>(self, length: usize) -> Vec<F> {
         Prng::from_seed_stream(self).take(length).collect()
     }
 }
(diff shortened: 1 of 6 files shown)


cjpatton:
Rebased.

prng.rs
divergentdave, line 1: nit: extra file

src/vdaf/xof.rs
```diff
@@ -91,12 +91,13 @@ impl<const SEED_SIZE: usize> Decode for Seed<SEED_SIZE> {
...
-/// A stream of bytes derived from a seed.
-pub trait SeedStream: Sized {
-    /// Fill `buf` with the next `buf.len()` bytes of output.
-    fn fill(&mut self, buf: &mut [u8]);
-
+/// Trait for deriving a vector of field elements.
+pub trait IntoVec: RngCore + Sized {
     /// Generate a finite field vector from the seed stream.
+    fn into_vec<F: FieldElement>(self, length: usize) -> Vec<F>;
+}
```
divergentdave, line 98: This name is a little close to Rust's naming conventions for conversions. For example, slices have an `into_vec()` method that copies into an owned `Vec`. How about `IntoFieldVec`/`into_field_vec()` instead?

divergentdave (approved):

src/vdaf/xof.rs
cjpatton: Done.

PR #737, opened 2023-09-07
Unroll first iteration of fft.

armfazh:
Reduces a round of multiplications times 1.

| poly_mul/fft/n | Improvement|
|--|--|
| 1    | -15.324%|
| 30   | -14.770%|
| 60   | -10.638%|
| 90   | -11.503%|
| 120  |  -9.163%|
| 150  |  -9.752%|
| 10^3 |  -7.856%|
| 10^4 |  -6.788%|
| 10^5 |  -5.771%|

### DIFF
--- a/src/fft.rs
+++ b/src/fft.rs
@@ -58,15 +58,26 @@ pub fn discrete_fourier_transform<F: FftFriendlyFieldElement>(
         w = F::one();
         let r = F::root(l).unwrap();
         let y = 1 << (l - 1);
-        for i in 0..y {
-            for j in 0..(size / y) >> 1 {
-                let x = (1 << l) * j + i;
+        let chunk = (size / y) >> 1;
+
+        // unrolling first iteration of i-loop.
+        for j in 0..chunk {
+            let x = j << l;
+            let u = outp[x];
+            let v = outp[x + y];
+            outp[x] = u + v;
+            outp[x + y] = u - v;
+        }
+
+        for i in 1..y {
+            w *= r;
+            for j in 0..chunk {
+                let x = (j << l) + i;
                 let u = outp[x];
                 let v = w * outp[x + y];
                 outp[x] = u + v;
                 outp[x + y] = u - v;
             }
-            w *= r;
         }
     }
 


cjpatton:
Did you try benchmarking for larger polynomials? Where this really counts is 1000s to 100,000s. If we get diminishing returns, then this optimization might not be worth it.

Also, what is the underlying principle that makes this faster? Compilers often try to unroll loops, but maybe this particular case is probably hard for the compiler to automatically unroll?

Would unrolling more iterations help? What do other FFT implementations that you're aware of do?

armfazh:
> Did you try benchmarking for larger polynomials? Where this really counts is 1000s to 100,000s. If we get diminishing returns, then this optimization might not be worth it.

Table of timings is updated. Improvement lowers a N grows, but still some gains are observed.

> Also, what is the underlying principle that makes this faster? 

Removes a round of multiplications when `w=1`, so it's unnecessary.

>Compilers often try to unroll loops, but maybe this particular case is probably hard for the compiler to automatically unroll?

Compilers unroll variables that usually fit in registers, here the variables are structs so it's difficult to do enrolling at a higher lever of abstraction.

> Would unrolling more iterations help? What do other FFT implementations that you're aware of do?

Not sure if it helps. This particular unrolling is to avoid some multiplications, and not to relieve pressure to register allocation.

src/fft.rs
```diff
@@ -58,15 +58,26 @@ pub fn discrete_fourier_transform<F: FftFriendlyFieldElement>(
         w = F::one();
         let r = F::root(l).unwrap();
         let y = 1 << (l - 1);
-        for i in 0..y {
-            for j in 0..(size / y) >> 1 {
-                let x = (1 << l) * j + i;
+        let chunk = (size / y) >> 1;
+
+        // unrolling first iteration of i-loop.
```
cjpatton, line 63:
```suggestion
        // Unroll the first iteration of `i`-loop.
```

src/fft.rs
```diff
@@ -58,15 +58,26 @@ pub fn discrete_fourier_transform<F: FftFriendlyFieldElement>(
...
+            let u = outp[x];
+            let v = outp[x + y];
+            outp[x] = u + v;
+            outp[x + y] = u - v;
+        }
+
+        for i in 1..y {
+            w *= r;
+            for j in 0..chunk {
+                let x = (j << l) + i;
```
cjpatton, line 75: It looks to me like `x` is computed differently. Previously we had $x = 2^\ell \cdot j + i$; now it is $x = j^2 + i$. It's been a while since i had FFT in my head, but why does the new math work?

cjpatton (changes requested):
We have more than a few FFTs for FLP proof generation/verification so I think even a 5% improvement per FFT will pay off. However I want to make sure the code is equivalent before we merge. (See inline comment.)

src/fft.rs
divergentdave: It's equivalent, note that the shift amount is the letter l, not one. The one change this line makes is replacing multiplication by the result of left shifting one with just a left shift.

cjpatton: Ohohh, misread `l` as `1` 😬

cjpatton (approved):

divergentdave (approved):

PR #744, opened 2023-09-11
Align `Aggregator` trait with spec

### LINKED ISSUE #670: `Vdaf::prepare_preprocess` lacks `Vdaf::AggregationParam` argument
Trait `Aggregator: Vdaf` has:

```
/// Preprocess a round of preparation shares into a single input to [`Aggregator::prepare_step`].
    fn prepare_preprocess<M: IntoIterator<Item = Self::PrepareShare>>(
        &self,
        inputs: M,
    ) -> Result<Self::PrepareMessage, VdafError>;
```

However, VDAF-06 defines `Vdaf.prep_shares_to_prep(agg_param: AggParam, prep_shares: Vec[Bytes]) -> Bytes`. ([VDAF](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-vdaf-06#section-5.2-6.3)). Note the `agg_param` argument that does not appear in our Rust implementation.

We get away with this because `Prio3.prep_shares_to_prep` never uses the aggregation parameter. But I'm more confused by the Poplar1 side: the reference implementation of `Poplar1.prep_shares_to_prep` takes responsibility for decoding field elements from bytes and so needs to know what level of the tree it's on, but libprio's implementation seems to have long since decoded everything. I'm not certain what's going on, but it's difficult to reconcile it with the specification.

Even if there's some clever trick here that makes Poplar1 work out, it seems like crate `prio`'s interface should match up with the specification.

tgeoghegan:
`Vdaf::prepare_preprocess` was missing the `&Vdaf::AggregationParam` argument specified in VDAF. This adds that argument to the trait method and its implementations. While we're at it, to further align with the spec, we add new methods `prepare_shares_to_prepare_message` and `prepare_next`, to align with the spec's `prep_shares_to_prep` and `prep_next`, as aliases for the existing `prepare_preprocess` and `prepare_step`. The existing methods are marked as deprecated.

Resolves #670

### DIFF
--- a/src/vdaf.rs
+++ b/src/vdaf.rs
@@ -208,6 +208,10 @@ pub trait Vdaf: Clone + Debug {
 pub trait Client<const NONCE_SIZE: usize>: Vdaf {
     /// Shards a measurement into a public share and a sequence of input shares, one for each
     /// Aggregator.
+    ///
+    /// Implements `Vdaf::shard` from [VDAF].
+    ///
+    /// [VDAF]: https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-vdaf-07#section-5.1
     fn shard(
         &self,
         measurement: &Self::Measurement,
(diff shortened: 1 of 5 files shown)


src/vdaf.rs
```diff
@@ -246,9 +253,36 @@ pub trait Aggregator<const VERIFY_KEY_SIZE: usize, const NONCE_SIZE: usize>: Vda
...
+    ///
+    /// Implements `Vdaf.prep_shares_to_prep` from [VDAF].
+    ///
+    /// # Notes
+    ///
+    /// [`Self::prepare_shares_to_prepare_message`] is preferable since its name better matches the
+    /// specification.
+    ///
+    /// [VDAF]: https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-vdaf-07#section-5.2
+    #[deprecated(
```
branlwyd, line 266: IMO, since this PR will be part of an effective-major-version bump, we may as well just drop the old methods. Backwards compatibility is not a concern, and if we keep these methods we'll need to keep them until the next effective-major-version bump.

branlwyd (approved):

src/vdaf.rs
tgeoghegan: This way it's easier to give people a hint of what to move to. Filed #745 to remove them later.

PR #755, opened 2023-09-19
Add function to find optimal chunk length

divergentdave:
This adds a function to find the optimal chunk length for either Prio3Histogram or Prio3SumVec, given its measurement length. I came up with a search algorithm with runtime logarithmic in the measurement size. It checks a small number of candidate parameter choices, while still getting an optimal result.

My plan is for this to be used primarily in our UI, to hide some complexity from users. We may also want to use this internally in some codepaths in Janus, i.e. for Taskprov, until it is updated for VDAF-07. This will support divviup/janus#1900.

Future work: If we want to change Prio3FixedPointBoundedL2VecSum to either expose chunk lengths as VDAF parameters, or always use optimal chunk lengths, note that the one of the two chunk lengths would need to be determined by a different search, because one of the gadgets has a different arity, and thus different proof length.

### DIFF
--- a/src/vdaf/prio3.rs
+++ b/src/vdaf/prio3.rs
@@ -1401,6 +1401,57 @@ where
     }
 }
 
+/// This is a polyfill for `usize::ilog2()`, which is only available in Rust 1.67 and later. It is
+/// based on the implementation in the standard library. It can be removed when the MSRV has been
+/// advanced past 1.67.
+///
+/// # Panics
+///
+/// This function will panic if `input` is zero.
+fn ilog2(input: usize) -> u32 {
+    if input == 0 {
+        panic!("Tried to take the logarithm of zero");
+    }
+    (usize::BITS - 1) - input.leading_zeros()
+}
+
+/// Finds the optimal choice of chunk length for [`Prio3Histogram`] or [`Prio3SumVec`], given its
+/// encoded measurement length. For [`Prio3Histogram`], the measurement length is equal to the
+/// length parameter. For [`Prio3SumVec`], the measurement length is equal to the product of the
+/// length and bits parameters.
+pub fn optimal_chunk_length(measurement_length: usize) -> usize {
+    if measurement_length <= 1 {
+        return 1;
+    }
+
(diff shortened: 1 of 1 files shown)


src/vdaf/prio3.rs
```diff
@@ -1401,6 +1401,57 @@ where
...
+        panic!("Tried to take the logarithm of zero");
+    }
+    (usize::BITS - 1) - input.leading_zeros()
+}
+
+/// Finds the optimal choice of chunk length for [`Prio3Histogram`] or [`Prio3SumVec`], given its
+/// encoded measurement length. For [`Prio3Histogram`], the measurement length is equal to the
+/// length parameter. For [`Prio3SumVec`], the measurement length is equal to the product of the
+/// length and bits parameters.
+pub fn optimal_chunk_length(measurement_length: usize) -> usize {
```
inahga, line 1422: Maybe nitpicky--can we have constructors for Prio3Histogram and Prio3SumVec that do the optimal calculation for us?

e.g.
```rust
impl Prio3SumVec {
    pub fn new_sum_vec_with_optimal_chunk_size( // feel free to editorialize this function name
        num_aggregators: u8,
        bits: usize,
        len: usize,
    ) -> Result<Self, VdafError> {
        Prio3::new(
            num_aggregators,
            SumVec::new(bits, len, optimal_chunk_length(bits * len))?,
        )
    }
}
```

Same for Prio3Histogram except call would be `optimal_chunk_length(len)`. Prevents the developer from making a mistake in calculating the measurement size.

inahga (approved):

tgeoghegan (approved):

src/vdaf/prio3.rs
divergentdave: There are some cases where there are multiple equally-optimal chunk lengths that produce the same proof length. If another implementation made a different choice in such a case, or if we changed this search algorithm in the future, then we'd run the risk of different parties disagreeing on the chunk length, which would lead to proof verification failures. So long as the chunk length is still a VDAF parameter, I think we shouldn't encourage selecting the parameter in more than one place by bundling this logic into the VDAF constructors.

PR #758, opened 2023-09-23
Adding Fp128 implementation based on fiat-crypto.

### LINKED ISSUE #757: Use fiat-crypto for Fp128 and other fields
[fiat-crypto](https://github.com/mit-plv/fiat-crypto) code generator is currently used for Fp255, however it can also be used for other fields, such as Fp128.

armfazh:
This is an implementation of Fp128 using [fiat-crypto](https://github.com/mit-plv/fiat-crypto). The code generated needs a wrapper to implement the traits required by FieldElement.

The downside of this implementation is the performance observed. The latency for field multiplication is almost twice as the current human-produced code (#165 ).

Closes #757

## Timings

| poly_mul/fft Fp128 | Current | fiat-crypto | Change |
|--|--|--|--|
| 1   |  180.14 ns  | 281.47 ns | +57.701% | 
| 30  |  5.7912 µs  | 10.398 µs | +79.487% | 
| 60  |  12.925 µs  | 24.852 µs | +87.113% | 
| 90  |  29.935 µs  | 53.975 µs | +80.685% | 
| 120 |  28.688 µs  | 59.180 µs | +101.55% | 
| 150 |  64.779 µs  | 127.07 µs | +97.426% |

## Analisis
The critical part are the primitives: addcarryx, subborrowx, and mulx which are implemented using native-rust code. 

Usually these primitives should be replaced by the compiler emitting a single assembler instruction (ADC,SBB,MUL, respectively).

**Potential Solution:**

It is possible to implement the primitives using intrinsics such as: [carrying_add](https://doc.rust-lang.org/std/primitive.u64.html#method.carrying_add), [borrowing_sub](https://doc.rust-lang.org/std/primitive.u64.html#method.borrowing_sub), and 
[widening_mul](https://doc.rust-lang.org/std/primitive.u64.html#method.widening_mul). However, they are only available in the _nightly_ version.
I tested this approach in nightly, however there is still a performance regression.


cjpatton:
Thanks @armfazh for working on this. Can you paste benchmarks for Nightly, just so we have those numbers as well?

armfazh:
> Thanks @armfazh for working on this. Can you paste benchmarks for Nightly, just so we have those numbers as well?

See #781

PR #767, opened 2023-09-26
nit: Replace "SHAKE128" with "cSHAKE128" in docs

cjpatton:


src/vdaf/xof.rs
```diff
@@ -216,7 +216,7 @@ impl Xof<16> for XofShake128 {
     }
 }
 
-/// The key stream produced by the cSHAKE128 XOF.
+/// The seed stream produced by the SHAKE128.
```
divergentdave, line 219: This doesn't read right, we should do one of the following
```suggestion
/// The seed stream produced by the SHAKE128 XOF.
```
```suggestion
/// The seed stream produced by SHAKE128.
```

divergentdave (approved):
FYI the PR title is backwards

PR #769, opened 2023-09-26
Add derivation of precomputed values for fields

divergentdave:
This adds a Sage script that demonstrates how to compute all the values in `src/fp.rs` for each supported FFT-friendly field. It starts from the choices of prime modulus and generator element (two coming from `draft-irtf-cfrg-vdaf`, and one coming from Prio2 code), reproduces all the values we have in that file currently, and documents what each value means and how they are defined. Note that some fields are pre-transformed into Montgomery representation.


documentation/field_parameters.sage
```diff
@@ -0,0 +1,117 @@
...
+        """
+        return [
+            self.to_montgomery(
+                pow(
+                    self.generator_element,
+                    2 ^ (self.num_roots - i),
+                    self.modulus,
+                )
+            )
+            for i in range(min(self.num_roots, 20) + 1)
```
inahga, line 86: Where does the `min(self.num_roots, 20)` come from, i.e. why do we need at least 20 roots?

documentation/field_parameters.sage
```diff
@@ -0,0 +1,117 @@
...
+FIELDS = [
+    Field(
+        "FieldPrio2",
+        2 ^ 20 * 4095 + 1,
+        3925978153,
+    ),
+    Field(
+        "Field64",
+        2 ^ 32 * 4294967295 + 1,
+        pow(7, 4294967295, 2 ^ 32 * 4294967295 + 1),
```
inahga, line 99: How were the generators originally chosen (I get that they're defined in the spec)? Just some arbitrary value [that fit this criteria](https://www.ietf.org/archive/id/draft-irtf-cfrg-vdaf-07.html#name-fft-friendly-fields)?
> To be FFT-friendly, the order of this subgroup MUST be a power of 2. In addition, the size of the subgroup dictates how large interpolated polynomials can be. It is RECOMMENDED that a generator is chosen with order at least 2^20.

inahga (commented):
Just some naive questions, nothing actionable.

documentation/field_parameters.sage
divergentdave: I wasn't able to figure out how the generator used with `FieldPrio2` was selected, but it has the right order, so it works. For the other two, the definitions use a neat trick wherein they start with an arbitrary group member with an order that's too high, and then knock off the unneeded cofactors through exponentiation until they get an element with the desired order. (If $g$ has order 6, then by the definition of the order of an element, $g^6 = e$, where $e$ is the identity, and $g^2 \neq e$ and $g^3 \neq e$. Then we can set $h = g^3$, and by substituting we know $h^2 = (g^3)^2 = g^6 = e$. By assumption, $h = g^3 \neq e$, thus $h$ has order two.) Choosing an arbitrary group element in a group this big has a large chance of selecting one with an order equal to the group's order, and 7 worked out as a starting place for both cases. If it hadn't, we could try again with something else.

```
sage: Field64 = GF(2^32 * 4294967295 + 1)
sage: factor(Field64.unit_group().order())
2^32 * 3 * 5 * 17 * 257 * 65537
sage: factor(Field64(7).multiplicative_order())
2^32 * 3 * 5 * 17 * 257 * 65537
sage: factor((Field64(7) ^ 4294967295).multiplicative_order())
2^32
sage: Field128 = GF(2^66 * 4611686018427387897 + 1)
sage: factor(Field128.unit_group().order())
2^66 * 3 * 3491 * 440340496364689
sage: factor(Field128(7).multiplicative_order())
2^66 * 3 * 3491 * 440340496364689
sage: factor((Field128(7) ^ 4611686018427387897).multiplicative_order())
2^66
```

documentation/field_parameters.sage
divergentdave: This is mainly here to match the Rust-side data structure, which uses `[u128; 21]`. If a field had a smaller power-of-two-sized multiplicative subgroup, such that we couldn't provide as many roots, then that would impact utility by putting a ceiling on how large of an FFT we could do. 20 should be plenty for most cases, as that lets us call gadgets about a million times (~2^20). We haven't had a need for more roots, and it seems unlikely for now that we will in practice.

armfazh:
In #758 , I added documentation about the generation of the constants for field128:
See this chunk of comments.
https://github.com/armfazh/libprio-rs/blob/fiat_fp128/src/field/field128.rs#L327-L358

> How were the generators originally chosen:

This `a = GF(p).primitive_element()` gives the smallest element that is a primitive element, and from it, one can get a generator by raising to the cofactor.

tgeoghegan (approved):

PR #776, opened 2023-10-02
Fix ignored clippy::needless_range_loop lints

divergentdave:
This rewrites various loops in Prio2 and Prio3 code to address ignored `clippy::needless_range_loop` lints. This improves some benchmarks by 2%-3%. None of the benchmarks regressed.


tgeoghegan (approved):
Chris argued that the C-style loops are easier to resolve against the papers, but I prefer to embrace Rust style (and fervently pray that cryptography papers and specifications will stop using C or Python inspired pseudocode). And I don't think we can argue with the performance improvement.

Issue #643, opened 2023-07-13
Field: Expose methods for converting to/from bitvector encoding


cjpatton:
These methods are use internally by `FixedPointBoundedL2Norm` and `Sum`. Since the 0/1 check sub-circuit is so common, my guess is that this will be needed by other VDAFs. Exposing these methods would encourage folks to use the same bit order, which could help support interoperability.

PR #819, opened 2023-11-03
Add helpers to get 0/1 as integer associated types

divergentdave:
This adds new methods `zero_integer()` and `one_integer()` to `FieldElementWithInteger`. These replace an existing pattern of constructing zero or one with a round trip through the Montgomery representation, by constructing it as a field element first, and then converting it back. This preparatory work for addressing #643.

I considered and rejected creating a new trait for integers, and putting these methods there, because adding a new trait bound to the `Integer` associated type would be a breaking change. We could reconsider this API design the next time we make significant breaking changes.

### DIFF
--- a/src/field.rs
+++ b/src/field.rs
@@ -194,6 +194,12 @@ pub trait FieldElementWithInteger: FieldElement + From<Self::Integer> {
 
     /// Returns the prime modulus `p`.
     fn modulus() -> Self::Integer;
+
+    /// Returns the integer representation of the additive identity.
+    fn zero_integer() -> Self::Integer;
+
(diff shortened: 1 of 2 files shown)


tgeoghegan (commented):
In the interest of limiting the public API commitment, would it be possible to provide the public helpers in #820 while keeping the new `zero_integer` and `one_integer` methods private? i.e. put them on a new trait `FieldElementWithIntegerPrivate` (a mouthful, I know) that we don't export.

divergentdave:
Sure, we can always make it public later, maybe via a new `Integer` trait laid out above

PR #820, opened 2023-11-03
Add public bitvector helpers

### LINKED ISSUE #643: Field: Expose methods for converting to/from bitvector encoding

divergentdave:
This is stacked on #819.

This closes #643, by adding encoding and decoding methods for bitvector representations of integers. Internal uses of `fill_with_bitvector_representation()` and the existing `encode_into_bitvector_representation()` are rewritten to use the new iterator-based method.

### DIFF
--- a/src/field.rs
+++ b/src/field.rs
@@ -199,13 +199,18 @@ pub trait FieldElementWithInteger: FieldElement + From<Self::Integer> {
     ///
     /// Returns an error if `input` cannot be represented with `bits` many bits, or if `bits`
     /// is larger than the bit width of the field's modulus.
-    fn encode_into_bitvector_representation(
+    fn encode_as_bitvector(
         input: Self::Integer,
         bits: usize,
     ) -> Result<BitvectorRepresentationIter<Self>, FieldError> {
+        // Check if `bits` is too large for this field.
         if !Self::valid_integer_bitlength(bits) {
             return Err(FieldError::ModulusOverflow);
         }
+
+        // Check if the input value can be represented in the requested number of bits by shifting
+        // it. The above check on `bits` ensures this shift won't panic due to the shift width
+        // being too large.
         if input >> bits != Self::zero_integer() {
             return Err(FieldError::InputSizeMismatch);
         }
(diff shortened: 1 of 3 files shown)


src/field.rs
```diff
@@ -200,78 +200,86 @@ pub trait FieldElementWithInteger: FieldElement + From<Self::Integer> {
...
+    /// is larger than the bit width of the field's modulus.
+    fn encode_into_bitvector_representation(
+        input: Self::Integer,
+        bits: usize,
+    ) -> Result<BitvectorRepresentationIter<Self>, FieldError> {
+        if !Self::valid_integer_bitlength(bits) {
+            return Err(FieldError::ModulusOverflow);
         }
-
-        // If `i` is still not zero, this means that it cannot be encoded by `bits` bits.
```
tgeoghegan, line 224: I thought this comment was helpful. Can we have it back, please?

src/field.rs
```diff
@@ -200,78 +200,86 @@ pub trait FieldElementWithInteger: FieldElement + From<Self::Integer> {
...
-        let mut i = *input;
-
-        let one = Self::one_integer();
-        for bit in output.iter_mut() {
-            let w = Self::from(i & one);
-            *bit = w;
-            i = i >> 1;
+    /// Returns an error if `input` cannot be represented with `bits` many bits, or if `bits`
+    /// is larger than the bit width of the field's modulus.
+    fn encode_into_bitvector_representation(
```
tgeoghegan, line 209: I wonder if we could drop `_representation` here and in the decode method. Is `encode_into_bitvector` clear enough?

src/field.rs
```diff
@@ -200,78 +200,86 @@ pub trait FieldElementWithInteger: FieldElement + From<Self::Integer> {
...
-    }
-
-    /// Decode the bitvector-represented value `input` into a simple representation as a single
-    /// field element.
+    /// Note that this decoding operation is linear, so it can be applied to secret shares of an
+    /// encoded integer, and if the results are summed up, it will be equal to the encoded integer.
     ///
     /// # Errors
     ///
     /// This function errors if `2^input.len() - 1` does not fit into the field `Self`.
```
tgeoghegan, line 239: nit: IIUC this condition is symmetric to "Returns an error if `input` cannot be represented with `bits` many bits, or if `bits` is larger than the bit width of the field's modulus" on `encode_into_bitvector_representation`, and it'd be nice if they were written in the same terms (e.g. relative to the width of the modulus of the field).

src/field.rs
```diff
@@ -200,78 +200,86 @@ pub trait FieldElementWithInteger: FieldElement + From<Self::Integer> {
...
         Ok(decoded)
     }
+}
+
+/// This iterator returns a sequence of field elements that are equal to zero or one, representing
+/// some integer in two's complement form. See
+/// [`FieldElementWithInteger::encode_into_bitvector_representation`].
+// Note that this is implemented with a separate struct, instead of using the map combinator,
+// because return_position_impl_trait_in_trait is not yet stable.
+#[derive(Debug)]
```
tgeoghegan, line 262: Any reason not to derive `Clone` here? I feel like I always regret not deriving `Clone` if it's trivial to do so.

tgeoghegan (approved):
Some nits but LGTM!

src/field.rs
divergentdave: I went with `encode_as_bitvector()` and `decode_bitvector()`.

tgeoghegan (approved):

PR #846, opened 2023-11-22
xof: Make the `SeedStreamAes128` constructor public

cjpatton:


divergentdave (approved):
Sounds good to me. It's strange that we had `SeedStreamAes128` public, but made constructing one impossible since removing `PrgAes128`.

inahga:
n.b. looking at `SeedStreamSha3` and `SeedStreamFixedKeyAes128` they don't appear to be publicly constructable either.

divergentdave:
Yes, those two aren't directly constructable, but they are produced by their respective Xof (née Prg) types, after hashing inputs.

inahga:
Ah I see, I didn't catch that `SeedStreamAes128` didn't have a matching `Xof` type.

Issue #839, opened 2023-11-20
Implement draft-irtf-cfrg-vdaf-08


divergentdave:
- [x] Switch to TurboSHAKE
- [x] Support multiple proofs in Prio3
- [x] Poplar1: Include nonce in more XOF calls
- [x] IDPF: Steal control bit from seed's XOF output
- [x] Update VERSION
- [x] Update test vectors

PR #850, opened 2023-11-27
Prio3: Add support for multiple proofs

cjpatton:
Partially addresses #839.

As of draft-irtf-cfrg-vdaf-08 it's possible to configure to Prio3 so that the Client generates (and the Aggregators verify) multiple FLPs for the same measurement. This commit adds support for this feature. While at it, this commit de-duplicates some of the code used to derive field elements from seeds.

 This change breaks compatibility with the vdaf-07 test vectors. Thus this commit disables the test vector tests for Prio3 as well.


src/vdaf/prio3.rs
```diff
@@ -350,10 +352,11 @@ where
...
     /// underlying type.
-    pub fn new(num_aggregators: u8, typ: T) -> Result<Self, VdafError> {
+    pub fn new(num_aggregators: u8, num_proofs: u8, typ: T) -> Result<Self, VdafError> {
         check_num_aggregators(num_aggregators)?;
         Ok(Self {
             num_aggregators,
+            num_proofs,
             typ,
             phantom: PhantomData,
         })
```
inahga, line 368: This makes it possible to create a Prio3 instantiation with `num_proofs = 0`.

Interestingly it turns out to work, i.e. no panics, evaluates the function correctly but with no proofs nor proof validation. But it leaves open the footgun of a user instantiating a "DAF" instead of a "VDAF".

cjpatton:
I'll rebase on top of #851.

src/vdaf/prio3.rs
```diff
@@ -1211,19 +1262,21 @@ where
...
+        for verifier in verifiers.chunks(self.typ.verifier_len()) {
+            match self.typ.decide(verifier) {
+                Ok(true) => (),
+                Ok(false) => {
+                    return Err(VdafError::Uncategorized(
+                        "proof verifier check failed".into(),
+                    ))
+                }
+                Err(err) => return Err(VdafError::from(err)),
             }
```
inahga, line 1275:
```suggestion
            if !self.typ.decide(verifier)? {
                return Err(VdafError::Uncategorized(
                    "proof verifier check failed".into(),
                ));
            }
```
nit: looks like we derive `FlpError` in `VdafError`, so we can use the `?` operator to clean this up a bit.

cjpatton: Nice!

src/vdaf/prio3.rs
```diff
@@ -369,19 +372,71 @@ where
         self.typ.verifier_len()
     }
 
+    fn num_proofs(&self) -> usize {
+        self.num_proofs.try_into().unwrap()
+    }
```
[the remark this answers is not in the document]
cjpatton: Ahh, well `usize` might not be `u64` (it could be `u32`, say, on a 32-bit processor, right?). I assumed this would be fallable because the behavior depends somewhat on the target architecture. Happily I was wrong.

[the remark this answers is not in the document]
cjpatton: Done, and I've also inlined this funciton.

divergentdave (approved):
Looks good! I checked benchmark performance, and all differences were within the noise threshold, so no worries there.

inahga (approved):
:tada:

divergentdave:
FYI, I was working through re-enabling Prio3 test vectors, and I noticed that the Prio3SumVecWithMultiproof_0.json file was generated using an algorithm ID of 0xFFFFFFFF. That may or may not be an oversight in the spec, but note that this implementation keeps using the same Prio3SumVec algorithm ID passed through from the SumVec Type impl. (We don't need to block merging this PR for this issue, and can fix the algorithm ID or test vector file in a followup)

cjpatton:
> FYI, I was working through re-enabling Prio3 test vectors, and I noticed that the Prio3SumVecWithMultiproof_0.json file was generated using an algorithm ID of 0xFFFFFFFF. That may or may not be an oversight in the spec, but note that this implementation keeps using the same Prio3SumVec algorithm ID passed through from the SumVec Type impl. (We don't need to block merging this PR for this issue, and can fix the algorithm ID or test vector file in a followup)

We didn't intend to allocate codepoints for these, so the algorithm ID is fine. In fact, generating test vectors for these was an oversight on my part (https://github.com/cfrg/draft-irtf-cfrg-vdaf/pull/315). I suggest we remove them.

cjpatton:
Rebased and squashed.

PR #851, opened 2023-11-27
Update XOFs to use TurboSHAKE128

divergentdave:
This implements XofTurboShake128, updates XofFixedKeyAes128, updates test vectors that were only affected by XOF changes, and temporarily ignores other VDAF test vectors. Part of #839.


tests/discrete_gauss.rs
```diff
@@ -43,6 +43,10 @@ fn discrete_gauss_reference() {
             "test_vectors/discrete_gauss_9999999999999999999999.json"
         )))
         .unwrap(),
+        serde_json::from_str(include_str!(concat!(
+            "test_vectors/discrete_gauss_2.342.json"
+        )))
+        .unwrap(),
```
inahga, line 49: nit: use of `concat!` on lines 32-49 is unnecessary

inahga (approved):

cjpatton (approved):
Looks good to me. Thanks for all the work you've done here ... 🤞 this will be the last time we change the XOF.

FWIW I don't see any changes to `Idpf`, despite the suggestion in the top comment.

divergentdave:
True, but since I've changed XofFixedKeyAes128 here, it's no longer compatible with VDAF-07 either. Mainly I just wanted to update all the document references in one go, since so many of them were near affected code in one way or another.

PR #855, opened 2023-11-29
Prio3: Decouple the algorithm ID from the FLP circuit

### LINKED ISSUE #854: Prio3: Codepoints for algorithm variants
The codepoint for Prio3 (i.e, `Vdaf::ID`) is determined by the FLP circuit. But it may be desirable to define custom codepoints for Prio3 variants that use a standard circuit, like `Sum`, but a different `Xof`, `Field`, or choice for the number of FLP proofs.

I don't think we can easily accommodate this without API changes. Two ideas come to mind:
1. Add a `const ID: u32` generic to `Prio3` and set `Vdaf::ID` to `ID`. That way we just specify the codepoint at the top level. This would be sufficient but would be a pretty invasive change.
2. Replace the associated type `Vdaf::ID` with something like `Vdaf { fn algorithm_id(&self) -> u32 }`.

Are folks amendable to making a change here? If so, which of these options seems better? Any other ideas?

cjpatton:
Closes #854.

This requires an API-breaking change.

### DIFF
--- a/src/flp.rs
+++ b/src/flp.rs
@@ -122,9 +122,6 @@ pub enum FlpError {
 /// as a vector of field elements and how validity of the encoded measurement is determined.
 /// Validity is determined via an arithmetic circuit evaluated over the encoded measurement.
 pub trait Type: Sized + Eq + Clone + Debug {
-    /// The Prio3 VDAF identifier corresponding to this type.
-    const ID: u32;
-
     /// The type of raw measurement to be encoded.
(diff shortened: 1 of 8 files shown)


divergentdave (approved):
I like taking the algorithm_id as a parameter in Prio3::new(), at that point the circuit, XOF, and field are all determined.

cjpatton:
Force-pushed with clippy fix.

PR #857, opened 2023-11-29
Implement `From<Vec<F>>` for `AggregateShare<F>`

cjpatton:


divergentdave (approved):
See also #173 and #383

PR #900, opened 2024-01-11
Improve FLP testing and add tooling to "test-util"

### LINKED ISSUE #898: Add `flp_validity_test` to test-utils
This is useful for testing FLP types defined outside of this repository.

cjpatton:
Closes #898.

Implement a new struct, `flp::test_util::FlpTest`, for driving generic tests for FLPs. This tool subsumes most of the functionality of `flp_validity_test()`. The irregular length tests have been dropped such that the `valid()` is no longer required to do bounds checking itself.

Remove `flp_validity_test()` and replace each use with `FlpTest`.

### DIFF
--- a/src/flp.rs
+++ b/src/flp.rs
@@ -111,11 +111,6 @@ pub enum FlpError {
     /// An error happened during noising.
     #[error("differential privacy error: {0}")]
     DifferentialPrivacy(#[from] crate::dp::DpError),
-
-    /// Unit test error.
-    #[cfg(test)]
-    #[error("test failed: {0}")]
-    Test(String),
 }
 
 /// A type. Implementations of this trait specify how a particular kind of measurement is encoded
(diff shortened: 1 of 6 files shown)


src/flp.rs
```diff
@@ -111,11 +111,6 @@ pub enum FlpError {
     /// An error happened during noising.
     #[error("differential privacy error: {0}")]
     DifferentialPrivacy(#[from] crate::dp::DpError),
-
-    /// Unit test error.
-    #[cfg(test)]
-    #[error("test failed: {0}")]
-    Test(String),
```
cjpatton, line 118: Question: The new tool replaces errors with panics. Is this useful? If you think it would be better to return errors, let me know and I'll revert.

divergentdave: This sounds great, immediately panicking is more appropriate for tests than returning an error (backtraces will be more informative). Plus, I'm always happy to remove conditional compilation from an enum's definition.

src/flp/types/fixedpoint_l2.rs
```diff
@@ -792,52 +792,49 @@ mod tests {
...
-                ]),
-                num_shares: 3,
-            },
-        )
-        .unwrap();
+        FlpTest {
+            name: None,
+            flp: &vsum,
+            input: &input,
+            expected_output: Some(&vec![
```
divergentdave, line 799: Clippy is flagging this as a useless `vec![]`. I think the same could be said about some of the expected output arguments passed to `expect_valid()`, but I'm not sure why Clippy didn't flag those. Maybe generic parameters interfere with the analysis?

src/flp.rs
```diff
@@ -752,6 +747,252 @@ pub(crate) fn gadget_poly_len(gadget_degree: usize, wire_poly_len: usize) -> usi
...
+                    .unwrap();
+                assert!(
+                    !self.flp.decide(&verifier).unwrap(),
+                    "{name}: proof mutant {} deemed valid",
+                    i
+                );
+            }
+
+            // Try mutating the joint randomness.
+            if self.test_tweaked_joint_rand {
```
divergentdave, line 943: I'd like to see some test cases exercise this, as currently none do. Alternately, we could remove this configuration boolean, and instead automatically perform these checks whenever the FLP asks for a non-empty joint randomness.

cjpatton: Hmm, we don't have a circuit right now for which this test passes, at least if we mutate every one of the joint random inputs. I'll just delete it.

divergentdave: Hmm, I tried enabling this and running `flp::types::tests::test_sum_vec_serial`, and it passed for me. For that VDAF, we have only one joint randomness element.

src/flp.rs
```diff
@@ -752,6 +747,225 @@ pub(crate) fn gadget_poly_len(gadget_degree: usize, wire_poly_len: usize) -> usi
     gadget_degree * (wire_poly_len - 1) + 1
 }
 
+/// Utilities for testing FLPs.
+#[cfg(feature = "test-util")]
+pub mod test_utils {
```
tgeoghegan, line 752: Would you use this outside of `prio`? Which is to say, does `pub(crate)` suffice? Same goes for the various `pub` items defined in this module.

cjpatton: I intend to use this outside of `prio` (see linked issue)

src/flp.rs
cjpatton: Yeah it's a matter of whether a gadget call depends on the value: if it does, this test is useful; if it doiesn't, the test won't pass.

cjpatton:
`cargo build --no-default-features` was failing because it needed `random_vector()` to be in the empty feature set. This is currently only included with "crypto-dependencies". I decided to resolve this by re-implemnting this function with TurboSHAKE128.

src/prng.rs
```diff
@@ -275,7 +280,7 @@ mod tests {
             Prng::from_seed_stream(XofTurboShake128::seed_stream(&seed, b"", b""));
         let want_buffer = want.buffer.clone();
 
-        let got: Prng<FieldPrio2, _> = want.into_new_field();
+        let got: Prng<Field64, _> = want.into_new_field();
```
divergentdave, line 283: I think this should use a field other than Field64, to maintain the test's semantics.

src/flp.rs
tgeoghegan: Good enough for me, thank you for confirming.

divergentdave (approved):

src/prng.rs
cjpatton: Ugh !!! Missed that.

Replaced with `Field128`, and I renamed the test to make the semantics a little more clear.

tgeoghegan (approved):

PR #901, opened 2024-01-11
Add XofHmacSha256Aes128

cjpatton:
An XOF based on HMAC-SHA256 and AES128. This XOF does not appear in the VDAF spec.


src/vdaf/xof.rs
```diff
@@ -446,6 +450,34 @@ impl RngCore for SeedStreamFixedKeyAes128 {
     }
 }
 
+/// XOF based on HMAC-SHA256 and AES128. This XOF is not part of the VDAF spec.
+#[derive(Clone, Debug)]
+pub struct XofHmacSha256Aes128(Hmac<Sha256>);
```
divergentdave, line 455: This struct and the impl block below need to be guarded with `#[cfg(feature = "crypto-dependencies")]` as well for a `--no-default-features` build to succeed.

divergentdave (approved):
LGTM, modulo conditional compilation fixes. It's kinda neat that this reuses `SeedStreamAes128`, with better key derivation.

divergentdave:
I briefly looked into whether XofHmacSha256Aes128 could be dropped into the experimental Prio2 VDAF we define, but unfortunately there's one difference: the XOF hashes a DST length byte and a DST first, before hashing the binder input, while the Prio2 VDAF only hashes the nonce. This means there's some minor duplication on the order of 3 lines of code, but oh well.

Cargo.toml
```diff
@@ -56,7 +56,7 @@ default = ["crypto-dependencies"]
 experimental = ["bitvec", "fiat-crypto", "fixed", "num-bigint", "num-rational", "num-traits", "num-integer", "num-iter", "rand"]
 multithreaded = ["rayon"]
 prio2 = ["crypto-dependencies", "hmac", "sha2"]
```
tgeoghegan, line 58: Since `crypto-dependencies` now pulls in `"hmac", "sha2"`, I think we can remove them from here?
```suggestion
prio2 = ["crypto-dependencies"]
```

tgeoghegan (approved):

Issue #947, opened 2024-02-16
Implement the Mastic VDAF


cjpatton:
Mastic is a new VDAF (https://github.com/jimouris/draft-mouris-cfrg-mastic) designed as a replacement for Poplar1 that is more efficient and provides more functionality. It is comprised of two components:

1. FLP, as in the `flp` module
2. VIDPF, an extension of IDPF (the `idpf` module) that still needs to be implemented

Security analysis for Mastic is underway (https://eprint.iacr.org/2024/221), but is incomplete as of this writing. We're confident in the composition of FLP and VIDPF, but the parameters for each are subject to change. In the meantime, it should be considered experimental and only made available via the "experimental" feature. Likewise, we should not deprecate Poplar1 until Mastic is finished and we have high confidence in its security.

cjpatton:
Right now the plan is for @armfazh to implement VIDPF (already underway, see #933), then have @hannahdaviscrypto implement the end-to-end protocol.

cjpatton:
After PR #1107, the following tasks till need to be completed:

- [ ] encode the counter in the output and use it during unsharding (https://github.com/divviup/libprio-rs/pull/1107/files?diff=split&w=1#r1835001615)
- [ ] counter and path consistency checks (https://github.com/divviup/libprio-rs/pull/1107/files?diff=split&w=1#r1823571345)
- [ ] Align the spec with the implementation (https://github.com/jimouris/draft-mouris-cfrg-mastic/issues/80, https://github.com/jimouris/draft-mouris-cfrg-mastic/issues/81)
- [ ] Prune unused code in `bt` and `vidpf`
- [ ] Add test vectors

divergentdave:
- [ ] We'll need to define a serialization format for `MasticPrepareState` in order to productionize Mastic.

PR #954, opened 2024-02-22
Implement Verifiable Incremental Distributed Point Function (VIDPF).

armfazh:
Partially addresses #947.

Implements Verifiable Incremental Distributed Point Function (VIDPF).

The VIDPF construction is specified in [[draft-mouris-cfrg-mastic]] and builds on techniques from [[MST23]] and [[CP22]] to lift an IDPF to a VIDPF.
    
[CP22]: https://eprint.iacr.org/2021/580
[MST23]: https://eprint.iacr.org/2023/080
[draft-mouris-cfrg-mastic]: https://datatracker.ietf.org/doc/draft-mouris-cfrg-mastic/02/

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -33,3 +33,9 @@ mod polynomial;
 mod prng;
 pub mod topology;
 pub mod vdaf;
+#[cfg(all(feature = "crypto-dependencies", feature = "experimental"))]
+#[cfg_attr(
+    docsrs,
+    doc(cfg(all(feature = "crypto-dependencies", feature = "experimental")))
+)]
+pub mod vidpf;
--- a/src/vidpf.rs
+++ b/src/vidpf.rs
@@ -1,0 +1,827 @@
+// SPDX-License-Identifier: MPL-2.0
+
+//! Verifiable Incremental Distributed Point Function (VIDPF).
+//!
+//! The VIDPF construction is specified in [[draft-mouris-cfrg-mastic]] and builds
+//! on techniques from [[MST23]] and [[CP22]] to lift an IDPF to a VIDPF.
+//!
+//! [CP22]: https://eprint.iacr.org/2021/580
+//! [MST23]: https://eprint.iacr.org/2023/080
+//! [draft-mouris-cfrg-mastic]: https://datatracker.ietf.org/doc/draft-mouris-cfrg-mastic/02/
+
+use core::{
+    iter::zip,
+    ops::{Add, AddAssign, BitXor, BitXorAssign, Index, Sub},
+};
+
+use bitvec::field::BitField;
+use rand_core::RngCore;
+use std::io::Cursor;
+use subtle::{Choice, ConditionallyNegatable, ConditionallySelectable};
+
+use crate::{
+    codec::{CodecError, Encode, ParameterizedDecode},
+    field::FieldElement,
+    idpf::{
+        conditional_select_seed, conditional_swap_seed, conditional_xor_seeds, xor_seeds,
+        IdpfInput, IdpfValue,
+    },
+    vdaf::xof::{Seed, Xof, XofFixedKeyAes128, XofTurboShake128},
+};
+
+/// VIDPF-related errors.
+#[derive(Debug, thiserror::Error)]
+#[non_exhaustive]
+pub enum VidpfError {
+    /// Error when key's identifier are equal.
+    #[error("key's identifier should be different")]
+    SameKeyId,
+
+    /// Error when level does not fit in a 32-bit number.
+    #[error("level is not representable as a 32-bit integer")]
+    LevelTooBig,
+
+    /// Error during VIDPF evaluation: tried to access a level index out of bounds.
+    #[error("level index out of bounds")]
+    IndexLevel,
+
+    /// Error when weight's length mismatches the length in weight's parameter.
+    #[error("invalid weight length")]
+    InvalidWeightLength,
+
+    /// Failure when calling getrandom().
+    #[error("getrandom: {0}")]
+    GetRandom(#[from] getrandom::Error),
+}
+
+/// Represents the domain of an incremental point function.
+pub type VidpfInput = IdpfInput;
+
+/// Represents the codomain of an incremental point function.
+pub trait VidpfValue: IdpfValue + Clone {}
+
+/// A VIDPF instance.
+pub struct Vidpf<W: VidpfValue, const NONCE_SIZE: usize> {
+    /// Any parameters required to instantiate a weight value.
+    weight_parameter: W::ValueParameter,
+}
+
+impl<W: VidpfValue, const NONCE_SIZE: usize> Vidpf<W, NONCE_SIZE> {
+    /// Creates a VIDPF instance.
+    ///
+    /// # Arguments
+    ///
+    /// * `weight_parameter`, any parameters required to instantiate a weight value.
+    pub const fn new(weight_parameter: W::ValueParameter) -> Self {
+        Self { weight_parameter }
+    }
+
+    /// The [`Vidpf::gen`] method splits an incremental point function `F` into two private keys
+    /// used by the aggregation servers, and a common public share.
+    ///
+    /// The incremental point function is defined as `F`: [`VidpfInput`] --> [`VidpfValue`]
+    /// such that:
+    ///
+    /// ```txt
+    /// F(x) = weight, if x is a prefix of the input.
+    /// F(x) = 0,      if x is not a prefix of the input.
+    /// ```
+    ///
+    /// # Arguments
+    ///
+    /// * `input`, determines the input of the function.
+    /// * `weight`, determines the input's weight of the function.
+    /// * `nonce`, used to cryptographically bind some information.
+    pub fn gen(
+        &self,
+        input: &VidpfInput,
+        weight: &W,
+        nonce: &[u8; NONCE_SIZE],
+    ) -> Result<(VidpfPublicShare<W>, [VidpfKey; 2]), VidpfError> {
+        let keys = [
+            VidpfKey::gen(VidpfServerId::S0)?,
+            VidpfKey::gen(VidpfServerId::S1)?,
+        ];
+        let public = self.gen_with_keys(&keys, input, weight, nonce)?;
+        Ok((public, keys))
+    }
+
+    /// [`Vidpf::gen_with_keys`] works as the [`Vidpf::gen`] method, except that two different
+    /// keys must be provided.
+    fn gen_with_keys(
+        &self,
+        keys: &[VidpfKey; 2],
+        input: &VidpfInput,
+        weight: &W,
+        nonce: &[u8; NONCE_SIZE],
+    ) -> Result<VidpfPublicShare<W>, VidpfError> {
+        if keys[0].id == keys[1].id {
+            return Err(VidpfError::SameKeyId);
+        }
+
+        let mut s_i = [keys[0].value, keys[1].value];
+        let mut t_i = [Choice::from(keys[0].id), Choice::from(keys[1].id)];
+
+        let n = input.len();
+        let mut cw = Vec::with_capacity(n);
+        let mut cs = Vec::with_capacity(n);
+
+        for level in 0..n {
+            let alpha_i = Choice::from(u8::from(input.get(level).ok_or(VidpfError::IndexLevel)?));
+
+            // If alpha_i == 0 then
+            //     (same_seed, diff_seed) = (right_seed, left_seed)
+            // else
+            //     (same_seed, diff_seed) = (left_seed, right_seed)
+            let seq_0 = Self::prg(&s_i[0], nonce);
+            let (same_seed_0, diff_seed_0) = &mut (seq_0.right_seed, seq_0.left_seed);
+            conditional_swap_seed(same_seed_0, diff_seed_0, alpha_i);
+
+            let seq_1 = Self::prg(&s_i[1], nonce);
+            let (same_seed_1, diff_seed_1) = &mut (seq_1.right_seed, seq_1.left_seed);
+            conditional_swap_seed(same_seed_1, diff_seed_1, alpha_i);
+
+            // If alpha_i == 0 then
+            //    diff_control_bit = left_control_bit
+            // else
+            //    diff_control_bit = right_control_bit
+            let diff_control_bit_0 = Choice::conditional_select(
+                &seq_0.left_control_bit,
+                &seq_0.right_control_bit,
+                alpha_i,
+            );
+            let diff_control_bit_1 = Choice::conditional_select(
+                &seq_1.left_control_bit,
+                &seq_1.right_control_bit,
+                alpha_i,
+            );
+
+            let s_cw = xor_seeds(same_seed_0, same_seed_1);
+            let t_cw_l =
+                seq_0.left_control_bit ^ seq_1.left_control_bit ^ alpha_i ^ Choice::from(1);
+            let t_cw_r = seq_0.right_control_bit ^ seq_1.right_control_bit ^ alpha_i;
+            let t_cw_diff = Choice::conditional_select(&t_cw_l, &t_cw_r, alpha_i);
+
+            let s_tilde_i_0 = conditional_xor_seeds(diff_seed_0, &s_cw, t_i[0]);
+            let s_tilde_i_1 = conditional_xor_seeds(diff_seed_1, &s_cw, t_i[1]);
+
+            t_i[0] = diff_control_bit_0 ^ (t_i[0] & t_cw_diff);
+            t_i[1] = diff_control_bit_1 ^ (t_i[1] & t_cw_diff);
+
+            let w_i_0;
+            let w_i_1;
+            (s_i[0], w_i_0) = self.convert(s_tilde_i_0, nonce);
+            (s_i[1], w_i_1) = self.convert(s_tilde_i_1, nonce);
+
+            let mut w_cw = w_i_1 - w_i_0 + weight.clone();
+            w_cw.conditional_negate(t_i[1]);
+
+            let cw_i = VidpfCorrectionWord {
+                seed: s_cw,
+                left_control_bit: t_cw_l,
+                right_control_bit: t_cw_r,
+                weight: w_cw,
+            };
+            cw.push(cw_i);
+
+            let pi_tilde_0 = Self::node_proof(input, level, &s_i[0])?;
+            let pi_tilde_1 = Self::node_proof(input, level, &s_i[1])?;
+            let cs_i = xor_proof(pi_tilde_0, &pi_tilde_1);
+            cs.push(cs_i);
+        }
+
+        Ok(VidpfPublicShare { cw, cs })
+    }
+
+    /// [`Vidpf::eval`] evaluates the entire `input` and produces a share of the
+    /// input's weight.
+    pub fn eval(
+        &self,
+        key: &VidpfKey,
+        public: &VidpfPublicShare<W>,
+        input: &VidpfInput,
+        nonce: &[u8; NONCE_SIZE],
+    ) -> Result<VidpfValueShare<W>, VidpfError> {
+        let mut state = VidpfEvalState::init_from_key(key);
+        let mut share = W::zero(&self.weight_parameter);
+
+        let n = input.len();
+        for level in 0..n {
+            (state, share) = self.eval_next(key.id, public, input, level, &state, nonce)?;
+        }
+
+        Ok(VidpfValueShare {
+            share,
+            proof: state.proof,
+        })
+    }
+
+    /// [`Vidpf::eval_next`] evaluates the `input` at the given level using the provided initial
+    /// state, and returns a new state and a share of the input's weight at that level.
+    fn eval_next(
+        &self,
+        id: VidpfServerId,
+        public: &VidpfPublicShare<W>,
+        input: &VidpfInput,
+        level: usize,
+        state: &VidpfEvalState,
+        nonce: &[u8; NONCE_SIZE],
+    ) -> Result<(VidpfEvalState, W), VidpfError> {
+        let cw = public.cw.get(level).ok_or(VidpfError::IndexLevel)?;
+
+        let seq_tilde = Self::prg(&state.seed, nonce);
+
+        let t_i = state.control_bit;
+        let sl = conditional_xor_seeds(&seq_tilde.left_seed, &cw.seed, t_i);
+        let sr = conditional_xor_seeds(&seq_tilde.right_seed, &cw.seed, t_i);
+        let tl = seq_tilde.left_control_bit ^ (t_i & cw.left_control_bit);
+        let tr = seq_tilde.right_control_bit ^ (t_i & cw.right_control_bit);
+
+        let x_i = Choice::from(u8::from(input.get(level).ok_or(VidpfError::IndexLevel)?));
+        let s_tilde_i = conditional_select_seed(x_i, &[sl, sr]);
+
+        let next_control_bit = Choice::conditional_select(&tl, &tr, x_i);
+        let (next_seed, w_i) = self.convert(s_tilde_i, nonce);
+
+        let zero = <W as IdpfValue>::zero(&self.weight_parameter);
+        let mut y = <W as IdpfValue>::conditional_select(&zero, &cw.weight, next_control_bit);
+        y += w_i;
+        y.conditional_negate(Choice::from(id));
+
+        let pi_i = &state.proof;
+        let cs_i = public.cs.get(level).ok_or(VidpfError::IndexLevel)?;
+        let pi_tilde = Self::node_proof(input, level, &next_seed)?;
+        let h2_input = xor_proof(
+            conditional_xor_proof(pi_tilde, cs_i, next_control_bit),
+            pi_i,
+        );
+        let next_proof = xor_proof(Self::node_proof_adjustment(h2_input), pi_i);
+
+        let next_state = VidpfEvalState {
+            seed: next_seed,
+            control_bit: next_control_bit,
+            proof: next_proof,
+        };
+
+        Ok((next_state, y))
+    }
+
+    fn prg(seed: &VidpfSeed, nonce: &[u8]) -> VidpfPrgOutput {
+        let mut rng = XofFixedKeyAes128::seed_stream(&Seed(*seed), VidpfDomainSepTag::PRG, nonce);
+
+        let mut left_seed = VidpfSeed::default();
+        let mut right_seed = VidpfSeed::default();
+        rng.fill_bytes(&mut left_seed);
+        rng.fill_bytes(&mut right_seed);
+        // Use the LSB of seeds as control bits, and clears the bit,
+        // i.e., seeds produced by `prg` always have their LSB = 0.
+        // This ensures `prg` costs two AES calls only.
+        let left_control_bit = Choice::from(left_seed[0] & 0x01);
+        let right_control_bit = Choice::from(right_seed[0] & 0x01);
+        left_seed[0] &= 0xFE;
+        right_seed[0] &= 0xFE;
+
+        VidpfPrgOutput {
+            left_seed,
+            left_control_bit,
+            right_seed,
+            right_control_bit,
+        }
+    }
+
+    fn convert(&self, seed: VidpfSeed, nonce: &[u8; NONCE_SIZE]) -> (VidpfSeed, W) {
+        let mut rng =
+            XofFixedKeyAes128::seed_stream(&Seed(seed), VidpfDomainSepTag::CONVERT, nonce);
+
+        let mut out_seed = VidpfSeed::default();
+        rng.fill_bytes(&mut out_seed);
+        let value = <W as IdpfValue>::generate(&mut rng, &self.weight_parameter);
+
+        (out_seed, value)
+    }
+
+    fn node_proof(
+        input: &VidpfInput,
+        level: usize,
+        seed: &VidpfSeed,
+    ) -> Result<VidpfProof, VidpfError> {
+        let mut shake = XofTurboShake128::init(seed, VidpfDomainSepTag::NODE_PROOF);
+        for chunk128 in input
+            .index(..=level)
+            .chunks(128)
+            .map(BitField::load_le::<u128>)
+            .map(u128::to_le_bytes)
+        {
+            shake.update(&chunk128);
+        }
+        shake.update(
+            &u16::try_from(level)
+                .map_err(|_e| VidpfError::LevelTooBig)?
+                .to_le_bytes(),
+        );
+        let mut rng = shake.into_seed_stream();
+
+        let mut proof = VidpfProof::default();
+        rng.fill_bytes(&mut proof);
+
+        Ok(proof)
+    }
+
+    fn node_proof_adjustment(mut proof: VidpfProof) -> VidpfProof {
+        let mut rng = XofTurboShake128::seed_stream(
+            &Seed(Default::default()),
+            VidpfDomainSepTag::NODE_PROOF_ADJUST,
+            &proof,
+        );
+        rng.fill_bytes(&mut proof);
+
+        proof
+    }
+}
+
+/// Contains the domain separation tags for invoking different oracles.
+struct VidpfDomainSepTag;
+impl VidpfDomainSepTag {
+    const PRG: &'static [u8] = b"Prg";
+    const CONVERT: &'static [u8] = b"Convert";
+    const NODE_PROOF: &'static [u8] = b"NodeProof";
+    const NODE_PROOF_ADJUST: &'static [u8] = b"NodeProofAdjust";
+}
+
+/// Private key of an aggregation server.
+pub struct VidpfKey {
+    id: VidpfServerId,
+    value: [u8; 16],
+}
+
+impl VidpfKey {
+    /// Generates a key at random.
+    ///
+    /// # Errors
+    /// Triggers an error if the random generator fails.
+    pub(crate) fn gen(id: VidpfServerId) -> Result<Self, VidpfError> {
+        let mut value = [0; 16];
+        getrandom::getrandom(&mut value)?;
+        Ok(Self { id, value })
+    }
+}
+
+/// Identifies the two aggregation servers.
+#[derive(Clone, Copy, PartialEq, Eq)]
+pub(crate) enum VidpfServerId {
+    /// S0 is the first server.
+    S0,
+    /// S1 is the second server.
+    S1,
+}
+
+impl From<VidpfServerId> for Choice {
+    fn from(value: VidpfServerId) -> Self {
+        match value {
+            VidpfServerId::S0 => Self::from(0),
+            VidpfServerId::S1 => Self::from(1),
+        }
+    }
+}
+
+/// Adjusts values of shares during the VIDPF evaluation.
+#[derive(Debug)]
+struct VidpfCorrectionWord<W: VidpfValue> {
+    seed: VidpfSeed,
+    left_control_bit: Choice,
+    right_control_bit: Choice,
+    weight: W,
+}
+
+/// Common public information used by aggregation servers.
+#[derive(Debug)]
+pub struct VidpfPublicShare<W: VidpfValue> {
+    cw: Vec<VidpfCorrectionWord<W>>,
+    cs: Vec<VidpfProof>,
+}
+
+/// Contains the values produced during input evaluation at a given level.
+pub struct VidpfEvalState {
+    seed: VidpfSeed,
+    control_bit: Choice,
+    proof: VidpfProof,
+}
+
+impl VidpfEvalState {
+    fn init_from_key(key: &VidpfKey) -> Self {
+        Self {
+            seed: key.value,
+            control_bit: Choice::from(key.id),
+            proof: VidpfProof::default(),
+        }
+    }
+}
+
+/// Contains a share of the input's weight together with a proof for verification.
+pub struct VidpfValueShare<W: VidpfValue> {
+    /// Secret share of the input's weight.
+    pub share: W,
+    /// Proof used to verify the share.
+    pub proof: VidpfProof,
+}
+
+/// Proof size in bytes.
+const VIDPF_PROOF_SIZE: usize = 32;
+
+/// Allows to validate user input and shares after evaluation.
+type VidpfProof = [u8; VIDPF_PROOF_SIZE];
+
+fn xor_proof(mut lhs: VidpfProof, rhs: &VidpfProof) -> VidpfProof {
+    zip(&mut lhs, rhs).for_each(|(a, b)| a.bitxor_assign(b));
+    lhs
+}
+
+fn conditional_xor_proof(mut lhs: VidpfProof, rhs: &VidpfProof, choice: Choice) -> VidpfProof {
+    zip(&mut lhs, rhs).for_each(|(a, b)| a.conditional_assign(&a.bitxor(b), choice));
+    lhs
+}
+
+/// Feeds a pseudorandom generator during evaluation.
+type VidpfSeed = [u8; 16];
+
+/// Contains the seeds and control bits produced by [`Vidpf::prg`].
+struct VidpfPrgOutput {
+    left_seed: VidpfSeed,
+    left_control_bit: Choice,
+    right_seed: VidpfSeed,
+    right_control_bit: Choice,
+}
+
+/// Represents an array of field elements that implements the [`VidpfValue`] trait.
+#[derive(Debug, PartialEq, Eq, Clone)]
+pub struct VidpfWeight<F: FieldElement>(Vec<F>);
+
+impl<F: FieldElement> From<Vec<F>> for VidpfWeight<F> {
+    fn from(value: Vec<F>) -> Self {
+        Self(value)
+    }
+}
+
+impl<F: FieldElement> VidpfValue for VidpfWeight<F> {}
+
+impl<F: FieldElement> IdpfValue for VidpfWeight<F> {
+    /// The parameter determines the number of field elements in the vector.
+    type ValueParameter = usize;
+
+    fn generate<S: RngCore>(seed_stream: &mut S, length: &Self::ValueParameter) -> Self {
+        Self(
+            (0..*length)
+                .map(|_| <F as IdpfValue>::generate(seed_stream, &()))
+                .collect(),
+        )
+    }
+
+    fn zero(length: &Self::ValueParameter) -> Self {
+        Self((0..*length).map(|_| <F as IdpfValue>::zero(&())).collect())
+    }
+
+    /// Panics if weight lengths are different.
+    fn conditional_select(lhs: &Self, rhs: &Self, choice: Choice) -> Self {
+        assert_eq!(
+            lhs.0.len(),
+            rhs.0.len(),
+            "{}",
+            VidpfError::InvalidWeightLength
+        );
+
+        Self(
+            zip(&lhs.0, &rhs.0)
+                .map(|(a, b)| <F as IdpfValue>::conditional_select(a, b, choice))
+                .collect(),
+        )
+    }
+}
+
+impl<F: FieldElement> ConditionallyNegatable for VidpfWeight<F> {
+    fn conditional_negate(&mut self, choice: Choice) {
+        self.0.iter_mut().for_each(|a| a.conditional_negate(choice));
+    }
+}
+
+impl<F: FieldElement> Add for VidpfWeight<F> {
+    type Output = Self;
+
+    /// Panics if weight lengths are different.
+    fn add(self, rhs: Self) -> Self::Output {
+        assert_eq!(
+            self.0.len(),
+            rhs.0.len(),
+            "{}",
+            VidpfError::InvalidWeightLength
+        );
+
+        Self(zip(self.0, rhs.0).map(|(a, b)| a.add(b)).collect())
+    }
+}
+
+impl<F: FieldElement> AddAssign for VidpfWeight<F> {
+    /// Panics if weight lengths are different.
+    fn add_assign(&mut self, rhs: Self) {
+        assert_eq!(
+            self.0.len(),
+            rhs.0.len(),
+            "{}",
+            VidpfError::InvalidWeightLength
+        );
+
+        zip(&mut self.0, rhs.0).for_each(|(a, b)| a.add_assign(b));
+    }
+}
+
+impl<F: FieldElement> Sub for VidpfWeight<F> {
+    type Output = Self;
+
+    /// Panics if weight lengths are different.
+    fn sub(self, rhs: Self) -> Self::Output {
+        assert_eq!(
+            self.0.len(),
+            rhs.0.len(),
+            "{}",
+            VidpfError::InvalidWeightLength
+        );
+
+        Self(zip(self.0, rhs.0).map(|(a, b)| a.sub(b)).collect())
+    }
+}
+
+impl<F: FieldElement> Encode for VidpfWeight<F> {
+    fn encode(&self, bytes: &mut Vec<u8>) -> Result<(), CodecError> {
+        for e in &self.0 {
+            F::encode(e, bytes)?;
+        }
+        Ok(())
+    }
+
+    fn encoded_len(&self) -> Option<usize> {
+        Some(self.0.len() * F::ENCODED_SIZE)
+    }
+}
+
+impl<F: FieldElement> ParameterizedDecode<<Self as IdpfValue>::ValueParameter> for VidpfWeight<F> {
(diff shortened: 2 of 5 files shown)


src/vidpf.rs
```diff
@@ -0,0 +1,735 @@
+// Copyright (c) 2024 Cloudflare, Inc.
```
cjpatton, line 1: We're not the copyright holders, ISRG is.
```suggestion
```

src/lib.rs
```diff
@@ -33,3 +33,4 @@ mod polynomial;
 mod prng;
 pub mod topology;
 pub mod vdaf;
+pub mod vidpf;
```
cjpatton, line 36: This should be behind the "experimental" feature flag, as `idpf` is.

src/vidpf.rs
```diff
@@ -0,0 +1,735 @@
...
+// SPDX-License-Identifier: MPL-2.0
+
+//! Implementation of a Verifiable Incremental Distributed Point Function (VIDPF).
+//!
+//! The VIDPF construction is specified in [[draft-mouris-cfrg-mastic]] and builds
+//! on techniques from [[MST23]] and [[CP22]] to lift an IDPF to a VIDPF.
+//!
+//! [CP22]: https://doi.org/10.1007/978-3-031-06944-4_6
```
cjpatton, line 9: eprint links are preferred.
```suggestion
//! [CP22]: https://eprint.iacr.org/2021/580
```

src/vidpf.rs
```diff
@@ -0,0 +1,735 @@
...
+use crate::{
+    codec::Decode,
+    field::{FieldElement, FieldElementExt},
+    vdaf::xof::{Seed as XofSeed, Xof},
+};
+
+/// Creates a new instance of a VIDPF.
+pub fn new_vidpf<P, R>(sec_param: usize, prng_ctor: P, bind_info: Vec<u8>) -> VidpfInstance<P, R>
+where
+    P: GetPRNG,
```
cjpatton, line 33: fyi: Rust convention is to not capitalize acronyms (unlike Go)
```suggestion
    P: GetPrng,
```

src/vidpf.rs
```diff
@@ -0,0 +1,735 @@
...
+}
+
+/// Fill populates a struct from a PRNG source.
+pub trait Fill {
+    /// fill reads as many bytes as needed from a PRNG source to fill the [Self] struct.
+    fn fill(&mut self, r: &mut Box<dyn RngCore>);
+}
+
+/// GetPRG
+pub trait GetPRNG {
```
cjpatton, line 329: It looks like the purpose of this trait is to translate an XOF seed into an `RngCore`. However, `SeedStream` implements `RngCore`, so you should be able just use the seed stream instead of wrapping it.

For example, `hash_two()` can be rewritten as follows:
```rust
    use rand::Rng;

    fn hash_two(&self, proof: &Proof) -> Proof {
        let dst = "vidpf proof adjustment".as_bytes();
        let seed = self.initial_seed();
        let binder = &proof.0;
        let mut rng = XofTurboShake128::seed_stream(
            &crate::vdaf::xof::Seed(seed.0.try_into().unwrap()),
            dst,
            binder,
        );
        let mut out_proof = self.initial_proof();
        rng.fill(&mut out_proof.0[..]);

        out_proof
    }
```

src/vidpf.rs
```diff
@@ -0,0 +1,735 @@
...
+    type Output = Self;
+
+    fn bitxor(self, rhs: Self) -> Self::Output {
+        ControlBit(self.0 ^ rhs.0)
+    }
+}
+
+#[derive(Debug, Clone)]
+/// Seed
+pub struct Seed(Vec<u8>);
```
cjpatton, line 517: We should be able to use `crate::vdaf::xof::Seed` instead.

src/vidpf.rs
```diff
@@ -0,0 +1,735 @@
...
+    R: Range,
+    for<'a> &'a R: RangeOps<R>,
+{
+    /// y is the sharing of the output.
+    pub y: R,
+    /// pi is a proof used to verify the share.
+    pub pi: Proof,
+}
+
+/// RangeOps
```
cjpatton, line 308: What is the puprose of `RangeOps`, `Range`, and `Fill`? Are these redefined from the standard library? Probably we don't actually need to re-define them here.

cjpatton (commented):
Some inital comments, focused mainly on rust conventions.

src/vidpf.rs
armfazh: adopted, it wasn't clear whether all the Xof have the same security properties. Consider to add a marker to distinguish.

src/vidpf.rs
armfazh: there are no contribution guidelines for this project, btw.

src/vidpf.rs
armfazh: done

src/vidpf.rs
armfazh: Range defines the codomain of the PDF function F:{0,1}*-->Range.
Fill was removed now.

src/vidpf.rs
```diff
@@ -0,0 +1,667 @@
...
+    for<'a> &'a R: RangeOps<R>,
+{
+    /// The gen method splits the point function `F(alpha) = beta` in two private keys
+    /// used by the aggregation servers, and a common public key.
+    /// The `beta` value must be a non-zero value.
+    ///
+    /// # Errors
+    /// Triggers an error if the random generator fails, or when beta is zero.
+    pub fn gen(&self, alpha: &[u8], beta: &R) -> Result<(PublicKey<R>, Key, Key), VdafError> {
+        if beta.is_zero() {
```
cjpatton, line 70: Why shouldn't `beta` be zero?

src/vidpf.rs
```diff
@@ -0,0 +1,667 @@
...
+
+        let mut s_i = [Seed::from(&k0), Seed::from(&k1)];
+        let mut t_i = [ControlBit::from(&k0), ControlBit::from(&k1)];
+
+        let n = 8 * alpha.len();
+        let mut cw = Vec::with_capacity(n);
+        let mut cs = Vec::with_capacity(n);
+
+        for i in 0..n {
+            let alpha_i = ControlBit::from((alpha[i / 8] >> (i % 8)) & 0x1);
```
cjpatton, line 86: Question: would bit operations be faster than `/` and `%`, or is the compiler smart enough to recognize it can do bit fiddling here?

(**Don't just blindly take the suggestion below, I'm not 100% I got the bit fiddling right!**)
```suggestion
            let alpha_i = ControlBit::from((alpha[i >> 3] >> (i & 0xff)) & 0x1);
```

src/vidpf.rs
```diff
@@ -0,0 +1,667 @@
...
+        if beta.is_zero() {
+            return Err(VidpfError::InvalidParameter("beta cannot be zero".into()).into());
+        }
+
+        // Key Generation.
+        let k0 = Key::gen(ServerID::S0)?;
+        let k1 = Key::gen(ServerID::S1)?;
+
+        let mut s_i = [Seed::from(&k0), Seed::from(&k1)];
+        let mut t_i = [ControlBit::from(&k0), ControlBit::from(&k1)];
```
cjpatton, line 79: I'ts not clear from context that `s_i` is set to the server's key share. (Likewise for `t_i`.) Let's get rid of the `From` and make this explicit.
```suggestion
        let mut s_i = [k0.value, k1.value];
        let mut t_i = [k0.id, k1.id];
```

src/vidpf.rs
```diff
@@ -0,0 +1,667 @@
...
+
+    fn prg(&self, seed: &Seed) -> Sequence {
+        let dst = b"100";
+        let mut rng = XofFixedKeyAes128::seed_stream(&seed.0, dst, &self.bind_info);
+
+        let mut sl = Seed::default();
+        let mut sr = Seed::default();
+        rng.fill_bytes(&mut sl.0 .0);
+        rng.fill_bytes(&mut sr.0 .0);
+        // Use LSB of seed for ControlBits -- saves a call to the prng,
```
cjpatton, line 226: Let's see if we can make this comment a little more useful. "call to prng" is a little vague, I think. What we're trying to do here concretely is get ensure that `prg()` costs exactly two calls to AES.

Also, the sentence needs to end with a period, not a comma.
```suggestion
        // Use LSB of seed for ControlBits -- saves a call to the prng,
```

src/vidpf.rs
```diff
@@ -0,0 +1,667 @@
...
+    }
+
+    /// verify checks whether the proofs are equal and that the shares add up to beta.
+    #[must_use]
+    pub fn verify(&self, a: &Share<R>, b: &Share<R>, beta: &R) -> bool {
+        a.y.add(&b.y) == *beta && a.pi.0 == b.pi.0
+    }
+
+    fn prg(&self, seed: &Seed) -> Sequence {
+        let dst = b"100";
```
cjpatton, line 219: Let's start thinking about how to align this with the draft. I think we're going to want to pick domain separation tags similar to how we do for IDPF: https://github.com/divviup/libprio-rs/blob/main/src/idpf.rs#L411-L420

src/vidpf.rs
```diff
@@ -0,0 +1,667 @@
...
+///
+/// The function is defined as F: {0.1}* --> [Range].
+pub struct Instance<R>
+where
+    // Represents the codomain of a Distributed Point Function (DPF).
+    R: Range,
+    for<'a> &'a R: RangeOps<R>,
+{
+    // Used to cryptographically bind some information.
+    bind_info: Vec<u8>,
```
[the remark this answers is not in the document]
cjpatton: Looking ahead, this is going to be the "nonce" in Mastic.

src/vidpf.rs
```diff
@@ -0,0 +1,667 @@
...
+        let n = 8 * alpha.len();
+        let mut cw = Vec::with_capacity(n);
+        let mut cs = Vec::with_capacity(n);
+
+        for i in 0..n {
+            let alpha_i = ControlBit::from((alpha[i / 8] >> (i % 8)) & 0x1);
+
+            let Sequence(sl_0, tl_0, sr_0, tr_0) = self.prg(&s_i[0]);
+            let Sequence(sl_1, tl_1, sr_1, tr_1) = self.prg(&s_i[1]);
+
```
cjpatton, line 90: A code comment here mapping would be helpful. "If `alpha_i == 0` then set s_same_0 to ...; otherwise, set .... Likewise for `s_same_1`."

It's not clear in context I think why we negate `alpha_i`, especially without having the documentation for `ConditionallySelectable` handy.

src/vidpf.rs
```diff
@@ -0,0 +1,667 @@
...
+            let s_same_0 = &Seed::conditional_select(&sl_0, &sr_0, !alpha_i.0);
+            let s_same_1 = &Seed::conditional_select(&sl_1, &sr_1, !alpha_i.0);
+
+            let s_cw = s_same_0 ^ s_same_1;
+            let t_cw_l = tl_0 ^ tl_1 ^ alpha_i ^ ControlBit::from(1);
+            let t_cw_r = tr_0 ^ tr_1 ^ alpha_i;
+            let t_cw_diff = ControlBit::conditional_select(&t_cw_l, &t_cw_r, alpha_i.0);
+
+            let s_diff_0 = &Seed::conditional_select(&sl_0, &sr_0, alpha_i.0);
+            let s_diff_1 = &Seed::conditional_select(&sl_1, &sr_1, alpha_i.0);
```
cjpatton, line 100: nit: I'd move these up to just after declaring `s_same_0` and `s_same_1`.

src/vidpf.rs
```diff
@@ -0,0 +1,667 @@
...
+        let mut out = Proof::default();
+        zip(&mut out.0, zip(self.0, rhs.0)).for_each(|(c, (a, b))| *c = a ^ b);
+
+        out
+    }
+}
+
+#[derive(Debug)]
+/// `PublicKey` is the common public key of aggregation servers.
+pub struct PublicKey<R>
```
cjpatton, line 348: This is called the "public share" in IDPF parlance.
```suggestion
/// `PublicShare` is the common public key of aggregation servers.
pub struct PublicShare<R>
```

src/vidpf.rs
```diff
@@ -0,0 +1,667 @@
...
+    }
+}
+
+impl ConditionallySelectable for ControlBit {
+    fn conditional_select(a: &Self, b: &Self, choice: Choice) -> Self {
+        Self((a.0 & !choice) | (b.0 & choice))
+    }
+}
+
+impl BitAnd<&Seed> for ControlBit {
```
cjpatton, line 413: This is delicate code... if we get rid of the `vidpf::Seed` wrapper around `XofSeed`, then it might be possible to reuse code from the `idpf` module.

If we cannot reuse code, make sure to add some unit tests for this.

src/vidpf.rs
```diff
@@ -0,0 +1,667 @@
...
+    R: Range,
+    for<'a> &'a R: RangeOps<R>,
+{
+    /// The gen method splits the point function `F(alpha) = beta` in two private keys
+    /// used by the aggregation servers, and a common public key.
+    /// The `beta` value must be a non-zero value.
+    ///
+    /// # Errors
+    /// Triggers an error if the random generator fails, or when beta is zero.
+    pub fn gen(&self, alpha: &[u8], beta: &R) -> Result<(PublicKey<R>, Key, Key), VdafError> {
```
cjpatton, line 69: Note: For validating test vectors, we'll need a way to run `Vidpf::gen()` using randomness provided by the caller. See `Idpf::gen()` for details.

src/vidpf.rs
```diff
@@ -0,0 +1,667 @@
...
+            let s_diff_1 = &Seed::conditional_select(&sl_1, &sr_1, alpha_i.0);
+            let s_tilde_i_0 = s_diff_0 ^ (t_i[0] & s_cw.borrow()).borrow();
+            let s_tilde_i_1 = s_diff_1 ^ (t_i[1] & s_cw.borrow()).borrow();
+
+            let t_diff_0 = ControlBit::conditional_select(&tl_0, &tr_0, alpha_i.0);
+            let t_diff_1 = ControlBit::conditional_select(&tl_1, &tr_1, alpha_i.0);
+            t_i[0] = t_diff_0 ^ (t_i[0] & t_cw_diff);
+            t_i[1] = t_diff_1 ^ (t_i[1] & t_cw_diff);
+
+            let (w_i_0, w_i_1): (R, R);
```
cjpatton, line 109: The compiler can infer the types of the variables declared here.

Generally it's better to avoid declaring types unless the compiler needs you to. That way, if the types change, we have a little less refactoring to do.
```suggestion
            let (w_i_0, w_i_1);
```

src/vidpf.rs
```diff
@@ -0,0 +1,667 @@
...
+                t_cw_l,
+                t_cw_r,
+                w_cw,
+            };
+            cw.push(cw_i);
+
+            let pi_0 = &Self::hash_one(alpha, i, &s_i[0]);
+            let pi_1 = &Self::hash_one(alpha, i, &s_i[1]);
+            let pi_i = pi_0 ^ pi_1;
+            cs.push(pi_i);
```
cjpatton, line 127: nit: consistency (up to this point we are using the same variable names as the paper
```suggestion
            let pi_tilde_0 = &Self::hash_one(alpha, i, &s_i[0]);
            let pi_tidle_1 = &Self::hash_one(alpha, i, &s_i[1]);
            let cs_i = pi_tilde_0 ^ pi_tilde_1;
            cs.push(cs_i);
```

cjpatton (commented):
Really nice work so far. I've gotten through gen(); next up is convert(), hash_one(), and then evaluation.

src/vidpf.rs
armfazh: changed, it makes things easier.

src/vidpf.rs
armfazh: updated

src/vidpf.rs
armfazh: no need, as it's using bitvector now.

src/vidpf.rs
```diff
@@ -0,0 +1,667 @@
...
+}
+
+/// `ServerID` used to identify the two aggregation servers.
+#[derive(Debug, Clone, Copy)]
+pub enum ServerID {
+    /// S0 is the first server.
+    S0 = 0,
+    /// S1 is the second server.
+    S1 = 1,
+}
```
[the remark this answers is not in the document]
armfazh: the ServerId needs to be public as is present in the Key, which is also public.

src/vidpf.rs
```diff
@@ -0,0 +1,667 @@
...
+            self.1 ^ rhs.1,
+            &self.2 ^ &rhs.2,
+            self.3 ^ rhs.3,
+        )
+    }
+}
+
+#[derive(Debug, Clone, Copy)]
+/// Stores a boolean value that controls correction operations.
+pub struct ControlBit(Choice);
```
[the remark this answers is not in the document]
armfazh: this trick really helped a lot. thanks

src/vidpf.rs
```diff
@@ -0,0 +1,676 @@
...
+pub enum ServerId {
+    /// S0 is the first server.
+    S0 = 0,
+    /// S1 is the second server.
+    S1 = 1,
+}
+
+/// OutputShare is a share of the output value produced during the evaluation tree.
+#[derive(Debug)]
+pub struct OutputShare<V>
```
cjpatton, line 345: I think this should be called a `ValueShare` in order to avoid conflating this with the concept of a "VDAF output share".

src/vidpf.rs
```diff
@@ -0,0 +1,676 @@
...
+    pub fn eval(
+        &self,
+        alpha: &VidpfInput,
+        key: &Key,
+        pk: &PublicShare<V>,
+        parameter: &V::ValueParameter,
+    ) -> OutputShare<V> {
+        let n = alpha.len();
+        assert!(pk.cw.len() >= n, "bad public key size of cw field");
+        assert!(pk.cs.len() >= n, "bad public key size of cs field");
```
cjpatton, line 239: Avoid evaluating these assertions in the release build.
```suggestion
        debug_assert!(pk.cw.len() >= n, "bad public key size of cw field");
        debug_assert!(pk.cs.len() >= n, "bad public key size of cs field");
```

src/vidpf.rs
```diff
@@ -0,0 +1,676 @@
...
+
+        let mut s_i = key.value;
+        let mut t_i = ControlBit::from(key.id as u8);
```
cjpatton, line 242: `as` is dangerous here: the compiler may not notice if we change the type of `key.id` in a way that has side-effects. The type-safe choice is to implement `From<ServerId>` for `ControlBit`:
```rust
impl From<ServerId> for ControlBit {
    fn from(value: ServerId) -> Self {
        todo!()
    }
}
```

src/vidpf.rs
```diff
@@ -0,0 +1,676 @@
...
+
+    /// [Vidpf::eval] queries the evaluation tree with `alpha` input and produces a share of the output value.
+    ///
+    /// # Panics
+    /// Panics when the public key does not contain as many correction words as the length
+    /// of the `alpha` input in bits.
+    #[must_use]
+    pub fn eval(
+        &self,
+        alpha: &VidpfInput,
```
cjpatton, line 232: We're virtually always going to need a sequence of candidate prefixes, not just one.
```suggestion
        prefixes: impl Iterator<Item = VidpfInput>,
```

src/vidpf.rs
```diff
@@ -0,0 +1,676 @@
...
+        self.gen_with_keys([k0, k1], alpha, beta, parameter)
+    }
+
+    /// The [Vidpf::gen_with_keys] method works as the [gen] method, except keys were already generated.
+    fn gen_with_keys(
+        &self,
+        keys: [Key; 2],
+        alpha: &VidpfInput,
+        beta: &V,
+        parameter: &V::ValueParameter,
```
cjpatton, line 106: Let's move the parameter to `self` and set it via `Vidpf::new()`. Likewise for `eval()`.

src/vidpf.rs
cjpatton: Bump: If you don't want to work on this in this PR, let me know and I"ll find some place to track the TODO.

cjpatton (commented):
Still working my way through `eval()`, but I wanted to give you some comments to catch up on in the meantime. Looking good!

1. Be careful about what we expose in the public API.
2. I'd like to find more meaningful names than "hash_one" and "hash_two". What about something like "node_proof" and "node_proof_adjustment" respectively? This is more or less what we came up in the draft.
3. Documentation: Be consistent about notation for the struct name in the code comment. All of the following are use: `Name`, `[Name]`, and `Name`. Note that the are all rendered differently in `cargo doc`.

src/vidpf.rs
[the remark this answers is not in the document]
cjpatton: The fields of a public struct can be private.

src/vidpf.rs
```diff
@@ -0,0 +1,676 @@
...
+impl VidpfInput {
+    fn iter_controlbit(&self) -> impl DoubleEndedIterator<Item = ControlBit> + '_ {
+        self.iter().rev().map(|a| ControlBit::from(a as u8))
+    }
+}
+
+/// Represents the codomain of a point function.
+pub trait VidpfValue: IdpfValue + Eq
+where
+    for<'a> &'a Self: Add<Output = Self> + Sub<Output = Self>,
```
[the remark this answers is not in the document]
divergentdave: This HRTB can be removed on the traits and structs, and we only need it on the `impl<V> Vidpf<V>` block. (for arithmetic in `gen_with_keys()`, `eval_next()`, `verify()`, and thus their callers) Putting the where bound on the trait or structs means they have to be repeated on more impl blocks, and on generic structs.

src/vidpf.rs
```diff
@@ -0,0 +1,676 @@
...
+    // Used to cryptographically bind some information.
+    nonce: Vec<u8>,
+
+    phantom: PhantomData<V>,
+}
+
+impl<V> Vidpf<V>
+where
+    V: VidpfValue,
+    for<'a> &'a V: Add<Output = V> + Sub<Output = V>,
```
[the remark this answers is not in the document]
divergentdave: The IDPF module avoids all of this by passing around owned field elements, etc., and using Add on that. When it uses a borrowed value out of the correction words, it does so using `conditional_select()`, which copies from a reference to an owned value as part of its operation.

Some additions could be rewritten in a similar way; to the extent that VIDPF requires using values of this type more than once, it would be possible to clone first, and then use Add on owned values. I tried replacing the HRTB in the where clause with `Clone + for<'a> Add<&'a Self, Output = Self> bounds on `VidpfValue` itself, and there were three callsites I had to tweak. The calculation of `w_cw` can be rearranged so that `w_i_1` is on the left hand side, and that is an owned value coming out of `convert()`. `beta` can stay as a borrowed value on the right hand side. The calculation of `y_i` could be changed to conditionally select between zero and the correction word, and then add `w_i` to that. This is how the IDPF module does the calculation, and it takes ownership of w_i, using it just once, and copies out of the w_cw borrow. Lastly, `verify()` probably just needs a clone on the left hand argument.

src/vidpf.rs
```diff
@@ -0,0 +1,676 @@
...
+        beta: &V,
+        parameter: &V::ValueParameter,
+    ) -> Result<(PublicShare<V>, Key, Key), VidpfError> {
+        // Key Generation.
+        let k0 = Key::gen(ServerId::S0)?;
+        let k1 = Key::gen(ServerId::S1)?;
+        self.gen_with_keys([k0, k1], alpha, beta, parameter)
+    }
+
+    /// The [Vidpf::gen_with_keys] method works as the [gen] method, except keys were already generated.
```
divergentdave, line 100: nit: fixed-with font for method references, and fixing a `cargo doc --document-private-items` warning
```suggestion
    /// The [`Vidpf::gen_with_keys`] method works as the [`Vidpf::gen`] method, except keys were already generated.
```

src/vidpf.rs
```diff
@@ -0,0 +1,676 @@
...
+            cs.push(cs_i);
+        }
+
+        Ok((PublicShare { cw, cs }, k0, k1))
+    }
+
+    /// [Vidpf::eval_next] queries one bit `(alpha_i)` in the evaluation tree at the level
+    /// corresponding to the current state `(s_i,t_i,cw_i)`. This evaluation updates
+    /// the state `(s_i,t_i)` and returns a new secret-shared value `(y)`.
+    #[must_use]
```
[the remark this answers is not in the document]
armfazh: deleted

src/vidpf.rs
```diff
@@ -0,0 +1,676 @@
...
+pub type VidpfInput = IdpfInput;
+
+impl VidpfInput {
+    fn iter_controlbit(&self) -> impl DoubleEndedIterator<Item = ControlBit> + '_ {
+        self.iter().rev().map(|a| ControlBit::from(a as u8))
+    }
+}
+
+/// Represents the codomain of a point function.
+pub trait VidpfValue: IdpfValue + Eq
```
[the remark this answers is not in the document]
armfazh: it wasn't required.

src/vidpf.rs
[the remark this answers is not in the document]
armfazh: updated, those HRTB not required at all.

src/vidpf.rs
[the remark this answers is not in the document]
armfazh: thanks for the tip

src/vidpf.rs
```diff
@@ -0,0 +1,676 @@
...
+    /// Triggers an error if the random generator fails.
+    pub fn gen(id: ServerId) -> Result<Self, getrandom::Error> {
+        let mut value = [0; 16];
+        getrandom::getrandom(&mut value)?;
+        Ok(Self { id, value })
+    }
+}
+
+/// [ControlBit] stores a boolean value that controls correction operations.
+pub type ControlBit = Choice;
```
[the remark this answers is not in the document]
armfazh: removed

src/vidpf.rs
```diff
@@ -0,0 +1,676 @@
...
+
+        (sl, tl, sr, tr)
+    }
+
+    fn convert(&self, seed: &Seed, parameter: &V::ValueParameter) -> (Seed, V) {
+        let dst = b"101";
+        let mut rng = XofFixedKeyAes128::seed_stream(&XofSeed(*seed), dst, &self.nonce);
+
+        let mut out_seed = Seed::default();
+        rng.fill_bytes(&mut out_seed);
```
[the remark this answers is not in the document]
armfazh: it turns out that calling fill_bytes directly is cheaper

src/vidpf.rs
```diff
@@ -0,0 +1,676 @@
...
+}
+
+// Proof size in bytes.
+const VIDPF_PROOF_SIZE: usize = 32;
+
+/// [Proof] is used to validate user input after tree evaluation.
+#[derive(Debug, PartialEq, Eq, Default)]
+pub struct Proof([u8; VIDPF_PROOF_SIZE]);
+
+fn xor_proof(lhs: &Proof, rhs: &Proof) -> Proof {
```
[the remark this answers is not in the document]
armfazh: done

src/vidpf.rs
armfazh: I think this api will be useful later, when using cache between invocations
so for now, let's process one input each time

src/vidpf.rs
```diff
@@ -0,0 +1,644 @@
...
+        let sl = conditional_xor_seeds(&seq_tilde.0, s_cw, t_i);
+        let sr = conditional_xor_seeds(&seq_tilde.2, s_cw, t_i);
+        let tl = seq_tilde.1 ^ (t_i & *t_cw_l);
+        let tr = seq_tilde.3 ^ (t_i & *t_cw_r);
+
+        let x_i = Choice::from(u8::from(input[level]));
+        let s_tilde_i = conditional_select_seed(x_i, &[sl, sr]);
+
+        let t_new = Choice::conditional_select(&tl, &tr, x_i);
+        let (seed_new, w_i) = self.convert(&s_tilde_i);
```
cjpatton, line 235: nit: Align with the function's verbiage
```suggestion
        let next_t = Choice::conditional_select(&tl, &tr, x_i);
        let (next_seed, w_i) = self.convert(&s_tilde_i);
```

src/vidpf.rs
```diff
@@ -0,0 +1,644 @@
...
+        let pi_i = &state.proof;
+        let cs_i = &public.cs.get(level).ok_or(VidpfError::IndexLevel)?;
+        let pi_tilde = Self::node_proof(input, level, &seed_new)?;
+        let h2_input = pi_tilde.conditional_xor(t_new, cs_i).xor(pi_i);
+        let pi_new = Self::node_proof_adjustment(h2_input).xor(pi_i);
+
+        let state_new = EvalState {
+            level: level + 1,
+            seed: seed_new,
+            t_control_bit: t_new,
```
cjpatton, line 250: How about just "control bit"?
```suggestion
            control_bit: next_control_bit,
```

src/vidpf.rs
```diff
@@ -0,0 +1,644 @@
...
+
+        let mut out_seed = Seed::default();
+        rng.fill_bytes(&mut out_seed);
+        let value = Weight::generate(&mut rng, &self.weight_parameter);
+
+        (out_seed, value)
+    }
+
+    fn node_proof(input: &VidpfInput, level: usize, seed: &Seed) -> Result<Proof, VidpfError> {
+        let mut binder = input.prefix(level).to_bytes();
```
cjpatton, line 296: This causes an allocation, which is probably not necessary. It should be possible to iterate over the bytes of `input[..level]` and use https://docs.rs/prio/latest/prio/vdaf/xof/trait.Xof.html#tymethod.update to update the XOF state.

src/vidpf.rs
```diff
@@ -0,0 +1,644 @@
...
+
+        proof
+    }
+}
+
+/// Contains the domain separation tags for invoking different oracles.
+struct DomainSepTag;
+// [todo] Fix dst for the VIDPF protocol.
+impl DomainSepTag {
+    const PRNG: &'static [u8] = b"Prng";
```
cjpatton, line 327:
```suggestion
    const PRG: &'static [u8] = b"Prg";
```

src/vidpf.rs
```diff
@@ -0,0 +1,644 @@
...
+
+        Ok((state_new, y))
+    }
+
+    /// [`Vidpf::open`] recovers the input's weight from additive shares produced
+    /// by aggregation servers.
+    ///
+    /// # Errors
+    /// Triggers an error when internal proofs are not equal.
+    pub fn open(a: ValueShare<Weight>, b: ValueShare<Weight>) -> Result<Weight, VidpfError> {
```
cjpatton, line 262: This is unsafe to put in the public API since the user is only meant to "open" aggregate shares.

Also, elsewhere we treat shares as a length-2 array.

```suggestion
    /// Triggers an error when internal proofs are not equal.
    #[cfg(test)]
    pub(crate) fn open(value_shares: [ValueShare<Weight>; 2]) -> Result<Weight, VidpfError> {
```

src/vidpf.rs
```diff
@@ -0,0 +1,644 @@
...
+            XofFixedKeyAes128::seed_stream(&XofSeed(*seed), DomainSepTag::PRNG, &self.nonce);
+
+        let mut sl = Seed::default();
+        let mut sr = Seed::default();
+        rng.fill_bytes(&mut sl);
+        rng.fill_bytes(&mut sr);
+        // Use the LSB of seeds as control bits -- This ensures [prg] costs two AES calls only.
+        let tl = Choice::from(sl[0] & 0x1);
+        let tr = Choice::from(sr[0] & 0x1);
+
```
cjpatton, line 280: We also need to clear the bit from `sl` and `sr` that we stole for `tl` and `tr`: https://github.com/cfrg/draft-irtf-cfrg-vdaf/blob/main/poc/idpf_poplar.py#L181-L182

src/vidpf.rs
```diff
@@ -0,0 +1,644 @@
...
+        self
+    }
+}
+
+/// [`Seed`] is used to feed a pseudorandom generator during evaluation.
+type Seed = [u8; 16];
+
+/// [`Weight`] represents a vector of field elements that implements the [`VidpfWeight`] trait.
+#[derive(Debug, PartialEq, Eq)]
+pub struct Weight<F: FieldElement, const N: usize>([F; N]);
```
cjpatton, line 434: I don't think we should have `N` as a const generic. Generally the length of the weight is known only at runtime because it's determined by the type of data we want to aggregate; making `N` generic forces us to choose this at compile time.

That said, I think it would be useful to have `impl<F: FieldElement> VidpfWeight` for `Weight<F>`. where
```rust
pub struct Weight<F>(Vec<F>);
```

src/vidpf.rs
```diff
@@ -0,0 +1,644 @@
...
+    }
+
+    #[test]
+    fn vidpf_correctness_at_last_level() {
+        let input = VidpfInput::from_bytes(&[0xFF]);
+        let weight = TestWeight::new([21.into(), 22.into(), 23.into()]);
+        let (vidpf, [alice_key, bob_key], public) = vidpf_gen_setup(&input, &weight);
+
+        let alice_share = vidpf.eval(&alice_key, &public, &input).unwrap();
+        let bob_share = vidpf.eval(&bob_key, &public, &input).unwrap();
```
cjpatton, line 555: For consistency with other test code in this repo.

Here and below.

```suggestion
        let value_share_0 = vidpf.eval(&alice_key, &public, &input).unwrap();
        let value_share_1 = vidpf.eval(&bob_key, &public, &input).unwrap();
```

src/vidpf.rs
```diff
@@ -0,0 +1,644 @@
...
+        );
+        rng.fill_bytes(&mut proof.0);
+
+        proof
+    }
+}
+
+/// Contains the domain separation tags for invoking different oracles.
+struct DomainSepTag;
+// [todo] Fix dst for the VIDPF protocol.
```
cjpatton, line 325: Remove this and use the issue to track remaining work. (Per @tgeoghegan request: we have from time to time let TODOs in code rot.)

```suggestion
```

cjpatton (commented):
Looking very good! It looks to me like you've properly translated the paper. Just one minor thing to fix in `prg()` to match the (eventual) draft.

The only other high level thing is that we want a default implementation of `VidpfWeight` for `Vec<F>` instead of `[F; N]`.

src/vidpf.rs
armfazh: this improved the timings by half.

src/vidpf.rs
armfazh: decided to remove this method.

src/vidpf.rs
armfazh: updated, I believe there are still differences with the python implementation, but I will resolve them in another PR.

src/vidpf.rs
armfazh: done

src/vidpf.rs
```diff
@@ -0,0 +1,644 @@
...
+        Self {
+            nonce,
+            weight_parameter,
+        }
+    }
+
+    /// The [`Vidpf::gen`] method splits an incremental point function `F` in two private keys
+    /// used by the aggregation servers, and a common public share.
+    ///
+    /// The incremental point function is defined as `F`: [`VidpfInput`] --> [`VidpfWeight`]
```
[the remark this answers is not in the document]
armfazh: well, this makes the entire document a unicode file.

src/vidpf.rs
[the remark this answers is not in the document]
armfazh: Noted. Also, it's likely than there are some other discrepancies between this implementation and the Python impl.

src/vidpf.rs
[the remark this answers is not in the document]
divergentdave: No problem, sounds good. FYI Rust source files are [defined to be encoded in UTF-8](https://doc.rust-lang.org/reference/input-format.html).

src/vidpf.rs
```diff
@@ -0,0 +1,743 @@
...
+
+        let mut left_seed = Seed::default();
+        let mut right_seed = Seed::default();
+        rng.fill_bytes(&mut left_seed);
+        rng.fill_bytes(&mut right_seed);
+        // Use the LSB of seeds as control bits, and clears the bit,
+        // i.e., seeds produced by `prg` always have their LSB = 0.
+        // This ensures `prg` costs two AES calls only.
+        let left_control_bit = Choice::from(left_seed[0] & 0x1);
+        let right_control_bit = Choice::from(right_seed[0] & 0x1);
```
cjpatton, line 284: nit: Slightly more obvious which bit we're stealing (`0x01 ^ 0xfe == 0xff`)
```suggestion
        let left_control_bit = Choice::from(left_seed[0] & 0x01);
        let right_control_bit = Choice::from(right_seed[0] & 0x01);
```

src/vidpf.rs
```diff
@@ -0,0 +1,743 @@
...
+        (out_seed, value)
+    }
+
+    fn node_proof(input: &VidpfInput, level: usize, seed: &Seed) -> Result<Proof, VidpfError> {
+        let mut shake = XofTurboShake128::init(seed, DomainSepTag::NODE_PROOF);
+        for chunk128 in input
+            .index(..=level)
+            .chunks(128)
+            .map(BitField::load_le::<u128>)
+            .map(u128::to_le_bytes)
```
cjpatton, line 312: We can make this slightly more compact and possibly save a Keccak evaluation
```suggestion
            .chunks(8)
            .map(BitField::load_le::<u8>)
```

src/vidpf.rs
```diff
@@ -0,0 +1,743 @@
...
+    ///
+    /// * `input`, determines the input of the function.
+    /// * `weight`, determines the input's weight of the function.
+    /// * `nonce`, used to cryptographically bind some information.
+    pub fn gen(
+        &self,
+        input: &VidpfInput,
+        weight: &W,
+        nonce: &[u8; NONCE_SIZE],
+    ) -> Result<(Key, Key, PublicShare<W>), VidpfError> {
```
cjpatton, line 100: nit: For symmetry with the `Vdaf` API, we might make the return type: `PublicShare<W>, [Key; 2]`

src/vidpf.rs
```diff
@@ -0,0 +1,743 @@
...
+    right_control_bit: Choice,
+}
+
+/// [`Weight`] represents an array of field elements that implements the [`VidpfWeight`] trait.
+#[derive(Debug, PartialEq, Eq)]
+pub struct Weight<F: FieldElement>(Vec<F>);
+
+impl<F: FieldElement> Weight<F> {
+    /// Creates a [`Weight`] from an array of field elements with length specified in parameters.
+    pub fn new_with_params(params: &WeightParameter<F>, array: &[F]) -> Result<Self, VidpfError> {
```
cjpatton, line 460:
```suggestion
    pub fn new_with_param(param: &WeightParameter<F>, array: &[F]) -> Result<Self, VidpfError> {
```

src/vidpf.rs
```diff
@@ -0,0 +1,743 @@
...
+        if params.length != array.len() {
+            return Err(VidpfError::InvalidWeightLength);
+        }
+
```
cjpatton, line 464: This length check is redundant since you `get()` by range below.
```suggestion
```

src/vidpf.rs
```diff
@@ -0,0 +1,743 @@
...
+        let slice = array
+            .get(0..params.length)
+            .ok_or(VidpfError::InvalidWeightLength)?;
+
+        Ok(Self(Vec::from(slice)))
+    }
+
+    /// Panics if weight lengths are different.
+    fn check_eq_length(a: &Self, b: &Self) {
```
cjpatton, line 473: Code smell: It's a bit odd to have a function whose only function is to panic.

It looks like you're trying to enforce a precondition in certain places: that the inputs to a function have the same length.

First, it would be a bit more clear if you replaced `Self::chunk_eq_length(a b)` with `assert_eq!(a, b, "vidpf: weights with different number of field elements");`

Second, I wonder if we should use `debug_assert_eq!()` instead of `assert_eq!()` such that the condition is not evaluated in the release code.

src/vidpf.rs
```diff
@@ -0,0 +1,743 @@
...
+        )
+    }
+
+    fn zero(parameter: &Self::ValueParameter) -> Self {
+        Self(
+            (0..parameter.length)
+                .map(|_| <F as IdpfValue>::zero(&parameter.field_parameter))
+                .collect(),
+        )
+    }
```
cjpatton, line 508: An explicit parameter would simplify this code substantially.
```suggestion
impl<F: FieldElement> VidpfWeight for Weight<F> {}

impl<F: FieldElement> IdpfValue for Weight<F> {
    type ValueParameter = usize;

    fn generate<S: RngCore>(seed_stream: &mut S, length: &usize) -> Self {
        Self(
            (0..*length)
                .map(|_| <F as IdpfValue>::generate(seed_stream, &()))
                .collect(),
        )
    }

    fn zero(length: &usize) -> Self {
        Self((0..*length).map(|_| F::zero()).collect())
    }

    fn conditional_select(lhs: &Self, rhs: &Self, choice: Choice) -> Self {
        Self::check_eq_length(lhs, rhs);
        Self(
            zip(&lhs.0, &rhs.0)
                .map(|(a, b)| <F as IdpfValue>::conditional_select(a, b, choice))
                .collect(),
        )
    }
}
```

src/vidpf.rs
```diff
@@ -0,0 +1,743 @@
...
+    /// Panics if weight lengths are different.
+    fn sub(self, rhs: Self) -> Self::Output {
+        Self::check_eq_length(&self, &rhs);
+        Self(zip(self.0, rhs.0).map(|(a, b)| a.sub(b)).collect())
+    }
+}
+
+impl<F: FieldElement> Encode for Weight<F> {
+    fn encode(&self, bytes: &mut Vec<u8>) -> Result<(), CodecError> {
+        encode_u32_items(bytes, &(), &self.0)
```
cjpatton, line 565: Remove the length prefix.

For decoding we can provide the expected length as a decoding parameter.

```rust
impl<F: FieldElement> ParameterizedDecode<usize> for Weight<F> {
    fn decode_with_param(length: usize, bytes: &mut Cursor<&[u8]>) -> Result<Self, CodecError> {
        todo!()
    }
}
```

src/vidpf.rs
armfazh: this is quite the opposite, we want to call SHAKE fewer times, not updating byte to byte. 
Also note that the blocksize of SHAKE is large (~168 bytes), so calls with 128 bits (16 bytes) is more efficient.

src/vidpf.rs
armfazh: This condition needs to be evaluated on runtime; otherwise, without the checking, calculations will be performed incorrectly without noticing.
Open to other suggestions.

src/vidpf.rs
armfazh: it does, thanks.

src/vidpf.rs
armfazh: good idea, however, removing the prefix makes impossible to implement `Decode` which is required by `IdpfValue` trait. -- unless I'm missing something.

Also the blanket implementation for `ParameterizedDecode for T where T:Decode`  avoids to define another implementation.

* Tests are failing for codec:Encode/Decode.

src/vidpf.rs
cjpatton: I would just replace `Self::chunk_eq_length(a, b)` with `assert_eq!(a,b)`.

src/vidpf.rs
cjpatton: I see. It's your call! FYI: we are likely to see input lengths of anywhere from 32 to 256 bits. We should also plan for 512 bits I think.

armfazh:
I've addressed your comments, and added more tests for coverage.

src/vidpf.rs
```diff
@@ -0,0 +1,828 @@
...
+    /// Proof used to verify the share.
+    pub proof: Proof,
+}
+
+/// Proof size in bytes.
+const VIDPF_PROOF_SIZE: usize = 32;
+
+/// [`Proof`] is used to validate user input and shares after evaluation.
+#[derive(Debug, PartialEq, Eq, Default)]
+pub struct Proof([u8; VIDPF_PROOF_SIZE]);
```
cjpatton, line 419: This should probably be crate-private. Also, consider whether this could be just a type alias for `[u8; VIDPF_PROOF_SIZE]` rather than a wrapper for it.

```suggestion
pub(crate) struct Proof([u8; VIDPF_PROOF_SIZE]);
```

src/vidpf.rs
```diff
@@ -0,0 +1,828 @@
...
+}
+
+/// [`PublicShare`] is the common public information used by aggregation servers.
+pub struct PublicShare<W: VidpfWeight> {
+    cw: Vec<CorrectionWord<W>>,
+    cs: Vec<Proof>,
+}
+
+/// [`EvalState`] contains the values produced during input evaluation at a given level.
+pub struct EvalState {
```
cjpatton, line 390:
```suggestion
pub struct VidpfEvalState {
```

src/vidpf.rs
```diff
@@ -0,0 +1,828 @@
...
+struct DomainSepTag;
+impl DomainSepTag {
+    const PRG: &'static [u8] = b"Prg";
+    const CONVERT: &'static [u8] = b"Convert";
+    const NODE_PROOF: &'static [u8] = b"NodeProof";
+    const NODE_PROOF_ADJUST: &'static [u8] = b"NodeProofAdjust";
+}
+
+/// [`Key`] is the private key of an aggregation server.
+pub struct Key {
```
cjpatton, line 340:
```suggestion
pub struct VidpfKey {
```

src/vidpf.rs
```diff
@@ -0,0 +1,828 @@
...
+        Self {
+            seed: key.value,
+            control_bit: Choice::from(key.id),
+            proof: Proof::default(),
+        }
+    }
+}
+
+/// [`ValueShare`] contains a share of the input's weight together with a proof for verification.
+pub struct ValueShare<W: VidpfWeight> {
```
cjpatton, line 407:
```suggestion
pub struct VidpfValueShare<W: VidpfWeight> {
```

cjpatton (approved):
Looks good to me. All of my remaining comments are debatable gripes about the API. Take these as suggestions, and let @divergentdave make the final call.

1. For consistency with the rest of this crate, each struct should begin with `Vidpf`. (See inline comments.)
1. Each docuco mment for a struct begins with a refereence to that struct. This gets rendered into a hyperlink, which looks strange in `cargo doc`:
image

src/vidpf.rs
```diff
@@ -0,0 +1,822 @@
...
+            [VidpfKey; 2],
+            [u8; TEST_NONCE_SIZE],
+        ) {
+            let vidpf = Vidpf::new(TEST_WEIGHT_LEN);
+            let (public, keys) = vidpf.gen(input, weight, TEST_NONCE).unwrap();
+            (vidpf, public, keys, *TEST_NONCE)
+        }
+
+        #[test]
+        #[should_panic]
```
divergentdave, line 612: Since this panic is coming from an unwrap in test code, we could improve this by instead writing `let err = vidpf.gen_with_keys(...).unwrap_err(); assert_matches!(err, Error::SameKeyId);`. This would prevent false negatives if the test panicked for some other reason.

src/vidpf.rs
```diff
@@ -0,0 +1,822 @@
...
+            c.conditional_negate(Choice::from(0));
+            let mut d = a.clone();
+            d.conditional_negate(Choice::from(1));
+            let zero = TestWeight::zero(&TEST_WEIGHT_LEN);
+
+            assert_eq!(c + d, zero, "a: {:?}", a);
+        }
+
+        #[test]
+        #[should_panic]
```
divergentdave, line 779: We can tighten these tests by matching against the panic message.
```suggestion
        #[should_panic = "invalid weight length"]
```

src/vidpf.rs
```diff
@@ -0,0 +1,822 @@
...
+        fn add_panics() {
+            let [w0, w1] = incompatible_weights();
+            let _ = w0 + w1;
+        }
+
+        #[test]
+        #[should_panic]
+        fn add_assing_panics() {
```
divergentdave, line 787: nit: typo
```suggestion
        fn add_assign_panics() {
```

divergentdave (approved):

src/vidpf.rs
armfazh: I adopted this idea, however `assert_matches` is in nightly.

divergentdave: We have a dev dependency on the `assert_matches` crate, which provides such a macro. It's nice that it's getting uplifted into std!

PR #956, opened 2024-02-23
Add a tool for assessing the IDPF aggregation parameter size

cjpatton:


Cargo.toml
```diff
@@ -34,6 +34,7 @@ sha2 = { version = "0.10.8", optional = true }
 sha3 = "0.10.8"
 subtle = "2.5.0"
 thiserror = "1.0"
+zipf = { version = "7.0.1", optional = true }
```
cjpatton, line 37: This is a bummer :( Is there any way to tell `cargo vet` that this is only used for "test-util" and should be ignored?

divergentdave: No, but adding an exemption is easy, and we started off with that for some such dependencies. OTOH, it's only ~300 lines, it should be easy to re-audit it for safe-to-deploy.

cjpatton: OK. I have absolutely no idea how this works, just let me know what I need to do.

binaries/src/bin/idpf_agg_param_size.rs
```diff
@@ -0,0 +1,46 @@
...
+use rand::prelude::*;
+
+fn main() {
+    let bits = 256;
+    let measurement_count = 10_000;
+    let threshold = ((measurement_count as f64) * 0.01) as usize; // 1%
+    let zipf_support = 128;
+    let zipf_exponent = 1.03;
+
+    println!("Generating inputs and computing the prefix tree. This make take some time...");
```
tgeoghegan, line 16:
```suggestion
    println!("Generating inputs and computing the prefix tree. This may take some time...");
```

tgeoghegan (approved):
I think you'll need to rebase on #961 and then CI should pass.

Cargo.toml
cjpatton: Thanks for #961.

PR #970, opened 2024-03-04
dummy vdaf: make aggregate results predictable

tgeoghegan:
Users of the dummy VDAF need to be able to predict what it will output so they can verify that they have executed it correctly. To that end, summations performed by the dummy VDAF are now done modulo `u8::MAX`. Additionally, it's nice for aggregations run with distinct aggregation parameters to yield distinct aggregate results, so we now sum in the aggregation parameter when unsharding aggregate shares. Finally, we add `vdaf::dummy::expected_aggregate_result` as a convenience function.

### DIFF
--- a/src/vdaf.rs
+++ b/src/vdaf.rs
@@ -464,9 +464,9 @@ impl<F: FieldElement> Encode for AggregateShare<F> {
 #[cfg(feature = "test-util")]
 #[cfg_attr(docsrs, doc(cfg(feature = "test-util")))]
 pub mod test_utils {
-    use crate::codec::{Encode, ParameterizedDecode};
-
     use super::{Aggregatable, Aggregator, Client, Collector, PrepareTransition, VdafError};
+    use crate::codec::{Encode, ParameterizedDecode};
+    use rand::prelude::*;
 
     /// Execute the VDAF end-to-end and return the aggregate result.
     pub fn run_vdaf<V, M, const SEED_SIZE: usize>(
@@ -478,14 +478,9 @@ pub mod test_utils {
         V: Client<16> + Aggregator<SEED_SIZE, 16> + Collector,
         M: IntoIterator<Item = V::Measurement>,
     {
-        use rand::prelude::*;
-        let mut rng = thread_rng();
-        let mut verify_key = [0; SEED_SIZE];
-        rng.fill(&mut verify_key[..]);
-
         let mut sharded_measurements = Vec::new();
         for measurement in measurements.into_iter() {
-            let nonce = rng.gen();
+            let nonce = random();
             let (public_share, input_shares) = vdaf.shard(&measurement, &nonce)?;
 
             sharded_measurements.push((public_share, nonce, input_shares));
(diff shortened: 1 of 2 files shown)


src/vdaf.rs
```diff
@@ -483,12 +483,37 @@ pub mod test_utils {
         let mut verify_key = [0; SEED_SIZE];
         rng.fill(&mut verify_key[..]);
```
inahga, line 484: The verify key isn't necessary here, since we don't use it for sharding.

src/vdaf.rs
```diff
@@ -483,12 +483,37 @@ pub mod test_utils {
...
 
-        let mut agg_shares: Vec<Option<V::AggregateShare>> = vec![None; vdaf.num_aggregators()];
-        let mut num_measurements: usize = 0;
+        let mut sharded_measurements = Vec::new();
         for measurement in measurements.into_iter() {
-            num_measurements += 1;
             let nonce = rng.gen();
```
inahga, line 488:
```suggestion
            let nonce = random();
```

nit: lets us get rid of the `rng` variable, after removing `verify_key`. I would also suggest we hoist up `use rand::prelude::*` to the mod level, for some dedupe between the two `run_vdaf` functions.

src/vdaf/dummy.rs
```diff
@@ -331,3 +333,86 @@ impl Encode for AggregateShare {
...
 }
+
+/// Returns the aggregate result that the dummy VDAF would compute over the provided measurements,
+/// for the provided aggregation parameter.
+pub fn expected_aggregate_result<M>(aggregation_parameter: u8, measurements: M) -> u64
+where
+    M: IntoIterator<Item = u8>,
+{
+    (measurements.into_iter().map(u64::from).sum::<u64>()) % u64::from(u8::MAX)
+        + u64::from(aggregation_parameter)
```
inahga, line 344:
```suggestion
    measurements.into_iter().map(u64::from).sum::<u64>() % (u64::from(u8::MAX) + 1)
        + u64::from(aggregation_parameter)
```

src/vdaf/dummy.rs
```diff
@@ -331,3 +333,86 @@ impl Encode for AggregateShare {
...
+mod tests {
+    use super::*;
+    use crate::vdaf::{test_utils::run_vdaf_sharded, Client};
+    use rand::prelude::*;
+
+    fn run_test(rounds: u32, aggregation_parameter: u8) {
+        let vdaf = Vdaf::new(rounds);
+        let mut verify_key = [0; 0];
+        thread_rng().fill(&mut verify_key[..]);
+        let measurements = [1, 2, 3, 4, 5];
```
inahga, line 357: The tests pass even without the above change since the measurements aren't large enough exercise the modulo. We could get clever with randomness and multiple rounds of testing, but I think it's sufficient to just make the measurements larger.
```suggestion
        let measurements = [10, 20, 30, 40, 50, 60, 70, 80, 90, 100];
```

src/vdaf/dummy.rs
inahga: We might also benefit from extracting `u64::from(u8::MAX) + 1` into its own constant.
```rust
const MODULUS: u64 = u8::MAX as u64 + 1;
```

src/vdaf/dummy.rs
tgeoghegan: Doing this also revealed a bug where I used the wrong modulus in `expected_aggregate_result`. Thanks!

inahga (approved):

Issue #925, opened 2024-01-31
IdpfCache implementation using a binary tree


divergentdave:
Chris suggested adding an IdpfCache implementation that uses a binary tree, as the cached values are arranged in a tree already. Note that this will have to sparsely populated, because the full binary tree may get impractically large.

cjpatton:
I suspect this will be the most efficient option for VIDPF.

armfazh:
In #954 , I implemented a binary tree for this purpose.

PR #978, opened 2024-03-21
vidpf: Implements a binary tree for caching VIPDF multiple evaluations.

armfazh:
Implements a binary tree for caching multiple VIDPF evaluations.

This is a generic append-only binary tree. Its API is minimal to support caching VIDPF evaluations. Tree is serializable.

Related #925

### DIFF
--- a/src/bt.rs
+++ b/src/bt.rs
@@ -1,0 +1,492 @@
+// SPDX-License-Identifier: MPL-2.0
+
+//! An append-only binary tree.
+//!
+//! ## Properties:
+//! - Binary tree: nodes have either 0, 1, or 2 child nodes.
+//! - Append-only: tree only grows, values can only be inserted and queried.
+//! - Serializable: tree can be coverted to and from bytes.
+//!
+//! ## Creation
+//! Use [`BinaryTree::default`] to create a binary tree, note that there must be
+//! one value at the root node.
+//!
+//! ## Insertion
+//! Given a value and a binary path, [`BinaryTree::insert`] stores the value
+//! at the end of the path.
+//!
+//! ## Query
+//! Given a binary path, [`BinaryTree::get`] returns a reference to the value
+//! stored in the node located at the end of the path. The empty path returns
+//! the value at the root node. No value is returned if the end of the path
+//! is unreachable.
+//!
+//! ## Serialization
+//! A tree can be serializad to and from bytes using [`BinaryTree::encode`]
+//! and [`BinaryTree::decode_with_param`] functions, respectively. One-byte
+//! markers are used allowing to store sparse trees with a lower overhead.
+//!
+//! ## Example
+//! This binary tree can be created with the following code:
+//!
+//! ```txt
+//!               ()=@
+//!          /              \
+//!       (0)=a           (1)=b
+//!     /      \         /      \
+//! (00)=c  (01)=d   (10)=e  (11)=f
+//! ```
+//!
+//! use prio::bt::BinaryTree;
+//! use bitvec::{bits,prelude::Lsb0};
+//! let mut tree = BinaryTree::default();
+//! tree.insert(bits!(), '@');
+//! tree.insert(bits!(0), 'a');
+//! tree.insert(bits!(1), 'b');
+//! tree.insert(bits!(0, 0), 'c');
+//! tree.insert(bits!(0, 1), 'd');
+//! tree.insert(bits!(1, 0), 'e');
+//! tree.insert(bits!(1, 1), 'f');
+
+// TODO(#947): Remove these lines once the module gets used by Mastic implementation.
+#![allow(dead_code)]
+#![allow(unused_variables)]
+
+use core::fmt::Debug;
+use std::io::Cursor;
+
+use bitvec::slice::BitSlice;
+
+use crate::codec::{CodecError, Decode, Encode, ParameterizedDecode};
+
+/// Errors triggered by binary tree operations.
+#[derive(Debug, thiserror::Error)]
+#[non_exhaustive]
+pub enum BinaryTreeError<V> {
+    /// Error when inserting in a node with existing value.
+    #[error("node already contains a value")]
+    InsertNonEmptyNode(V),
+    /// Error when an operation cannot reach the node at the end of a path.
+    #[error("unreachable node in the tree")]
+    UnreachableNode(V),
+}
+
+/// Used to indicate a traversal path on the binary tree.
+pub type Path = BitSlice;
+
+type SubTree<V> = Option<Box<Node<V>>>;
+
+/// Represents a node of a binary tree.
+pub struct Node<V> {
+    value: V,
+    left: SubTree<V>,
+    right: SubTree<V>,
+}
+
+impl<V> Node<V> {
+    fn new(value: V) -> Self {
+        Self {
+            value,
+            left: None,
+            right: None,
+        }
+    }
+
+    /// Inserts the value at the end of the path.
+    ///
+    /// This function traverses the tree from this node (`self`) until reaching the
+    /// node at the end of the path. If the node is unreachable or already
+    /// contains a value, an error wrapping the value is returned. Otherwise,
+    /// a new node containing the value is inserted.
+    ///
+    /// # Returns
+    /// - `Ok(())` when the node is inserted at the end of the path.
+    /// - `Err(InsertNonEmptyNode(value))` when the subtree already contains a
+    /// value at the end of the path.
+    /// - `Err(UnreachablePath(value))` when the end of the path is unreachable.
+    ///
+    /// In the error cases, no insertion occurs and the value is returned to
+    /// the caller of this function.
+    pub fn insert(&mut self, path: &Path, value: V) -> Result<(), BinaryTreeError<V>> {
+        enum Ref<'a, V> {
+            This(&'a mut Node<V>),
+            Other(&'a mut Option<Box<Node<V>>>),
+        }
+
+        // Finds the node at the end of the path.
+        let mut node = Ref::This(self);
+        for bit in path.iter() {
+            node = match node {
+                Ref::This(n) => Ref::Other(if !bit { &mut n.left } else { &mut n.right }),
+                Ref::Other(Some(n)) => Ref::Other(if !bit { &mut n.left } else { &mut n.right }),
+                Ref::Other(None) => {
+                    return Err(BinaryTreeError::UnreachableNode(value));
+                }
+            };
+        }
+
+        // Checks whether the node already has a value,
+        // If so, returns an error wrapping the value that could not be inserted.
+        // otherwise, a new node is created containing the value to be inserted.
+        match node {
+            Ref::This(_) | Ref::Other(Some(_)) => {
+                return Err(BinaryTreeError::InsertNonEmptyNode(value))
+            }
+            Ref::Other(empty_node) => {
+                *empty_node = Some(Box::new(Node::new(value)));
+            }
+        }
+
+        Ok(())
+    }
+
+    /// Gets a reference to the value located at the end of the path.
+    ///
+    /// This function traverses the tree from this node (`self`) until reaching the
+    /// node at the end of the path. It returns [None], if the node is
+    /// unreachable or nonexistent. Otherwise, it returns a reference to the
+    /// value stored in the node.
+    pub fn get(&self, path: &Path) -> Option<&V> {
+        let mut node = self;
+        for bit in path {
+            match if !bit { &node.left } else { &node.right } {
+                None => return None,
+                Some(next_node) => node = next_node,
+            };
+        }
+        Some(&node.value)
+    }
+
+    /// Gets a mutable reference to the node located at the end of the path.
+    ///
+    /// This function traverses the tree from this node (`self`) until reaching the
+    /// node at the end of the path. It returns [None], if the node is
+    /// unreachable or nonexistent. Otherwise, it returns a mutable reference
+    /// to the node.
+    pub fn get_node(&mut self, path: &Path) -> Option<&mut Node<V>> {
+        let mut node = self;
+        for bit in path {
+            match if !bit {
+                &mut node.left
+            } else {
+                &mut node.right
+            } {
+                None => return None,
+                Some(next_node) => node = next_node,
+            };
+        }
+        Some(node)
+    }
+}
+
+/// Represents an append-only binary tree.
+pub struct BinaryTree<V> {
+    root: SubTree<V>,
+}
+
+impl<V> BinaryTree<V> {
+    /// Inserts the value at the end of the path.
+    ///
+    /// This function traverses the tree from the root node until reaching the
+    /// node at the end of the path. If the node is unreachable or already
+    /// contains a value, an error wrapping the value is returned. Otherwise,
+    /// a new node containing the value is inserted.
+    ///
+    /// # Returns
+    /// - `Ok(())` when the node is inserted at the end of the path.
+    /// - `Err(InsertNonEmptyNode(value))` when the tree already contains a
+    /// value at the end of the path.
+    /// - `Err(UnreachablePath(value))` when the end of the path is unreachable.
+    ///
+    /// In the error cases, no insertion occurs and the value is returned to
+    /// the caller of this function.
+    pub fn insert(&mut self, path: &Path, value: V) -> Result<(), BinaryTreeError<V>> {
+        if let Some(node) = &mut self.root {
+            node.insert(path, value)
+        } else if path.is_empty() {
+            self.root = Some(Box::new(Node::new(value)));
+            Ok(())
+        } else {
+            Err(BinaryTreeError::UnreachableNode(value))
+        }
+    }
+
+    /// Gets a reference to the value located at the end of the path.
+    ///
+    /// This function traverses the tree from the root node until reaching the
+    /// node at the end of the path. It returns [None], if the node is
+    /// unreachable or nonexistent. Otherwise, it returns a reference to the
+    /// value stored in the node.
+    pub fn get(&self, path: &Path) -> Option<&V> {
+        self.root.as_ref().and_then(|node| node.get(path))
+    }
+
+    /// Gets a mutable reference to the node located at the end of the path.
+    ///
+    /// This function traverses the tree from the root node until reaching the
+    /// node at the end of the path. It returns [None], if the node is
+    /// unreachable or nonexistent. Otherwise, it returns a mutable reference
+    /// to the node.
+    pub fn get_node(&mut self, path: &Path) -> Option<&mut Node<V>> {
+        self.root.as_mut().and_then(|node| node.get_node(path))
+    }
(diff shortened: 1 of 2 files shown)


src/vidpf.rs
```diff
@@ -337,6 +338,41 @@ impl<W: VidpfValue, const NONCE_SIZE: usize> Vidpf<W, NONCE_SIZE> {
 
         proof
     }
+
+    /// Evaluates multiple user reports.
+    pub fn multi_eval(
```
divergentdave, line 343: I'm not sure of the use of this function, and whether it should be part of the public API or test-only. Generally, I don't think there's any computation that can be shared across multiple client reports, so it doesn't make much difference whether we loop over reports in the library or in the application. On the other hand, if this is anticipating batched verification, that would be useful to have in the library (though we wouldn't be able to make use of it within the VDAF abstraction).

src/vidpf.rs
```diff
@@ -452,6 +489,13 @@ struct VidpfPrgOutput {
     right_control_bit: Choice,
 }
 
+/// Clients use this structure to send reports to each of the Servers.
+pub struct ClientReport<V: VidpfValue> {
```
divergentdave, line 493: This probably doesn't need to be public, I figure it will just be used for tests here. Higher-level protocols such as DAP will be responsible for bundling the public share, input share, and nonce together in their own way, and that will include additional fields from Mastic in the shares as well.

src/vidpf.rs
```diff
@@ -337,6 +338,41 @@ impl<W: VidpfValue, const NONCE_SIZE: usize> Vidpf<W, NONCE_SIZE> {
...
+                    let mut tree = BinaryTree::new(state.clone());
+
+                    let input = VidpfInput::from(prefix.clone());
+                    let nonce_array = &nonce[..NONCE_SIZE].try_into().unwrap(); // todo: make this conversion reliable
+
+                    (state, _) =
+                        self.eval_next(key.id, public, &input, k - 1, &state, nonce_array)?;
+
+                    // State is cached in the path indicated by the prefix.
+                    tree.insert(&prefix, state);
```
divergentdave, line 369: Note that this is only evaluating the VIDPF at the first level, because it starts afresh with the initial state each time. The `BinaryTree` data structure only has one state inserted, and it isn't read from. To get the benefit of caching, we'd want to create a tree per client report, and use it to look up previously computed states while evaluating the same report at multiple prefixes. (and at multiple levels, though the multiple levels wouldn't all happen in one invocation in a realistic distributed trust setting)

src/bt.rs
```diff
@@ -0,0 +1,426 @@
...
+                    list_nodes[parent].node.right = new_node
+                } else {
+                    list_nodes[parent].node.left = new_node
+                }
+            } else {
+                out = Some(elem.node);
+            }
+        }
+
+        out.ok_or(CodecError::Other("failed to decode node".into()))
```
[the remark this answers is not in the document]
armfazh: I've simplified the code for this function, thanks for the pointers.

src/vidpf.rs
armfazh: this is true, it's not actually used as is. 
I was trying to exemplify the use of the tree for caching. 
The tree will be used when mastic is in fact implemented.

src/vidpf.rs
armfazh: agree, I will let the high-level protocol to use the binary tree.

src/bt.rs
```diff
@@ -0,0 +1,426 @@
...
+            let node = iter.take().unwrap();
+
+            if !bit {
+                match node.left {
+                    None => {
+                        node.left = new_node;
+                        break;
+                    }
+                    Some(_) => iter = node.left.as_deref_mut(),
+                };
```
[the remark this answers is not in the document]
armfazh: > We could either return an error or panic if we find an empty left/right pointer too early, or make node.value an option, and keep building the rest of the path's nodes before trying to store the value (this may affect the serialization format).

I opted by the former.

src/bt.rs
```diff
@@ -0,0 +1,358 @@
...
+        } else {
+            *node = Some(Box::new(Node::new(value)));
+            Ok(None)
+        }
+    }
+
+    /// Gets a reference to the value located at the end of the path.
+    /// [`None`] is returned if no values are found during the traversal of the tree
+    /// along the path.
+    pub fn get(&self, path: &Path) -> Option<&V> {
```
cjpatton, line 136: It might be a good idea to return a reference to the node rather than just the value. That way we can start an insertion from this point if we wanted to. I think we'll end up wanting to do this in VIDPF.

src/bt.rs
```diff
@@ -0,0 +1,358 @@
...
+        Ok(())
+    }
+}
+
+/// Marker used to distinguish between full and empty nodes.
+#[repr(u8)]
+#[derive(Debug)]
+enum CodecMarker {
+    None,
+    Node,
```
cjpatton, line 250:
```suggestion
    Leaf,
    Inner,
```

src/bt.rs
```diff
@@ -0,0 +1,358 @@
...
+                }
+            }
+        }
+
+        Ok(())
+    }
+}
+
+impl<V: Decode> Decode for BinaryTree<V> {
+    fn decode(bytes: &mut Cursor<&[u8]>) -> Result<Self, CodecError> {
```
cjpatton, line 180: I strongly suspect that `V` will need to be `ParamerizedDecode` for VIDPF.
```suggestion
impl<V: ParameterizedDecode<P>, P> ParameterizedDecode<P> for BinaryTree<V, P> {
    fn decode_with_param(param: &P, bytes: &mut Cursor<&[u8]>) -> Result<Self, CodecError> {
```

src/bt.rs
```diff
@@ -0,0 +1,358 @@
...
+                    stack.push(&mut node.left);
+                }
+            };
+        }
+
+        Ok(Self { root })
+    }
+}
+
+impl<V: Display> Display for BinaryTree<V> {
```
cjpatton, line 203: Holy crap this is pretty!! However the output will contain secret shares when used with VIDPF. We are treating secret shared data as something you don't want to inadvertently log. Thus implementation of `Debug` or `Display` should either not contain any secret shares or should be fenced behind the "test-util" feature such that it's less likely to be compiled in production.

We should also decide whether we're OK with special characters or would rather stick to ascii. I'm agnostic, but @tgeoghegan may have thoughts here for example.

cjpatton (commented):
The binary tree bits look great. Mostly minor comments, plus one feature request.

src/bt.rs
```diff
@@ -0,0 +1,358 @@
+// SPDX-License-Identifier: MPL-2.0
+
+//! An append-only binary tree.
+//!
+//! ## Properties:
+//! - Binary tree: nodes have either 0, 1, or 2 child nodes.
+//! - Append-only: tree only grows, values can only be replaced and queried.
```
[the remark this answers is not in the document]
armfazh: correct

src/bt.rs
```diff
@@ -0,0 +1,358 @@
...
+    /// Creates an empty binary tree.
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Inserts the value at the end of the path, possibly replacing an existing value.
+    ///
+    /// # Returns
+    /// - `Ok(None)`, the node was inserted at the end of the path.
+    /// - `Ok(Some(old))`, the `old` value at the end of the path was replaced.
```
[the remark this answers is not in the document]
armfazh: dropping replacement, not needed

src/bt.rs
armfazh: added a new method.

src/bt.rs
```diff
@@ -0,0 +1,358 @@
...
+            right: None,
+        }
+    }
+}
+/// Represents an append-only binary tree.
+pub struct BinaryTree<V> {
+    root: SubTree<V>,
+}
+
+impl<V> BinaryTree<V> {
```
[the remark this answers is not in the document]
armfazh: Now it's implemented

src/bt.rs
```diff
@@ -0,0 +1,423 @@
...
+    /// # Returns
+    /// - `Ok(())`, the node was inserted at the end of the path.
+    /// - `Err(InsertNonEmptyNode(value))`, the node already contains a value.
+    /// - `Err(UnreachablePath(value))`, the end of the path is unreachable.
+    ///
+    /// In the error cases, no insertion occurs and the value is returned to its owner.
+    pub fn insert(&mut self, path: &Path, value: V) -> Result<(), BinaryTreeError<V>> {
+        let mut iter = path.iter();
+
+        if let Some(first_bit) = iter.next() {
```
cjpatton, line 102: Clenaer control flow:

```rust
let Some(first_bit) = iter.next() else {
    return Err(BinaryTreeError::InsertNonEmptyNode(value));
};
```

src/bt.rs
```diff
@@ -0,0 +1,423 @@
...
+            let mut node = if !first_bit {
+                &mut self.left
+            } else {
+                &mut self.right
+            };
```
cjpatton, line 107: It looks like we're unrolling the loop for the first bit. Is this necessary? Could we initially set `node` to `self`?

src/bt.rs
```diff
@@ -0,0 +1,423 @@
...
+            }
+        } else {
+            Err(BinaryTreeError::InsertNonEmptyNode(value))
+        }
+    }
+
+    /// Gets a reference to the value located at the end of the path.
+    /// [`None`] is returned if no values are found during the traversal of the tree
+    /// along the path.
+    pub fn get(&self, path: &Path) -> Option<&V> {
```
cjpatton, line 139: Could method use `get_node()` as a subroutine?

src/bt.rs
```diff
@@ -0,0 +1,423 @@
...
+pub type Path = BitSlice;
+
+type SubTree<V> = Option<Box<Node<V>>>;
+
+/// Represents a node of a binary tree.
+pub struct Node<V> {
+    value: V,
+    left: SubTree<V>,
+    right: SubTree<V>,
+}
```
cjpatton, line 80: This seems like the most natural way to implement a binary tree in Rust, using `Box` to emulate pointers in C/Go/etc. However, I wonder if it's the most efficient thing we could do.

Alternatively, we could keep the tree in a a vector of nodes and use indices into the vector for the children:

```rust
pub struct Node<V> {
    value: V,
    left: usize,
    right: usize,
}

pub struct Tree<V>(Vec<Node<V>>);
```

This would make the API a bit messier (when traversing the tree from one node to another, you need to have a reference to the tree), but this messiness would be crate-private. The benefit, I suspect, would be fewer allocations. I have no idea if this would be worth it however.

Thoughts, @divergentdave?

src/bt.rs
cjpatton: cc/ @divergentdave: Should `Debug` implementations have unicode characters in them?

src/lib.rs
```diff
@@ -15,6 +15,8 @@
 //! [vdaf]: https://datatracker.ietf.org/doc/draft-irtf-cfrg-vdaf/05/
 
 pub mod benchmarked;
+#[cfg(feature = "experimental")]
+pub mod bt;
```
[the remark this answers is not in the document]
cjpatton: Bump
```suggestion
pub(crate) mod bt;
```

src/bt.rs
armfazh: fixed it,

src/bt.rs
armfazh: unfortunately not, due to mutability.

src/bt.rs
armfazh: That's a good idea, however, I would do it in a different PR, so one can contrast timings.

src/bt.rs
armfazh: I have changed the unicode symbols for ascii chars.
Coincidentally, printing, for example a `BinaryTree<u32>`, produces a valid YAML output.

src/bt.rs
armfazh: good one

src/lib.rs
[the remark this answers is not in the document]
armfazh: Agree that module must be private, however at the current state, it breaks CI since warnings are treated as errors.

src/bt.rs
branlwyd: Drive-by:

If the tree is unlikely to be sparse (or if the tree is sparse, but an acceptable sparse-vector implementation is available, or perhaps even a map from `usize` to `V`), the representation could be:

```rust
pub struct Tree<V>(Vec<Option<V>>);
/* or */
pub struct Tree<V>(HashMap<usize, Option<V>>);
```

... using an implicit rule to determine the children of a node: the value for the root node is stored at index 0; the children of the node whose value is stored at index `i` have their values stored at indices `2*i + 1` and `2*i + 2`.

This maintains the likely cache-locality/lowered-allocation wins of the node-referenced-by-index tree, while permitting fewer bad tree representations (i.e. it would not be possible to construct a "tree" with a cycle in it), and slightly simplifying tree traversal.

Insertion could _almost_ be reduced to a simple index-lookup-and-set based on an index that has the same bitwise representation as the path being inserted to, except that this tree implementation currently requires all intermediate nodes to have a value. So insertion would need to check all of the intermediate nodes as well; if that requirement was relaxed, the need to check intermediate nodes could be dropped.

src/lib.rs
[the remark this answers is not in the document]
divergentdave: I think I'd prefer `#[allow(unused_code)]` and a TODO pointing to #947 over making the module public.

src/bt.rs
divergentdave: Yes, I expect that using one big allocation as an offset-indexed arena of nodes could improve performance by improving cache locality, and amortizing malloc calls. It would be helpful if we could use heuristics from the application to size the `Vec`, to avoid unnecessary copying when resizing, but `Decode` doesn't lend itself to this.

For practical reasons, this tree will need to be sparse, as we could easily have a tree depth of over 100.

src/bt.rs
```diff
@@ -0,0 +1,435 @@
...
+
+    /// Inserts the value at the end of the path.
+    ///
+    /// # Returns
+    /// - `Ok(())`, the node was inserted at the end of the path.
+    /// - `Err(InsertNonEmptyNode(value))`, the node already contains a value.
+    /// - `Err(UnreachablePath(value))`, the end of the path is unreachable.
+    ///
+    /// In the error cases, no insertion occurs and the value is returned to its owner.
+    pub fn insert(&mut self, path: &Path, value: V) -> Result<(), BinaryTreeError<V>> {
```
divergentdave, line 99: I think this could use some more explanation, particularly that `Node::insert()` expects only part of a node's index, and particularly just enough bits to navigate from `self` to another node. (as opposed to `BinaryTree::insert()`, which takes the entire node index, since it is starting from the root) So, this would be a suffix of a prefix of a Mastic input ultimately.

src/bt.rs
```diff
@@ -0,0 +1,435 @@
...
+        } else {
+            Err(BinaryTreeError::UnreachableNode(value))
+        }
+    }
+
+    /// Gets a reference to the value located at the end of the path.
+    /// [`None`] is returned if no values are found during the traversal of the tree
+    /// along the path.
+    pub fn get(&self, path: &Path) -> Option<&V> {
+        self.root.as_deref().map(|node| node.get(path)).unwrap()
```
divergentdave, line 196: I think this should be an and_then rather than a map and unwrap. Particularly, `BinaryTree::default().get(&bits!())` should return None rather than panic.

src/bt.rs
```diff
@@ -0,0 +1,435 @@
...
+    }
+
+    /// Gets a mutable reference to the node located at the end of the path.
+    /// [`None`] is returned if no values are found during the traversal of the tree
+    /// along the path.
+    pub fn get_node(&mut self, path: &Path) -> Option<&mut Node<V>> {
+        self.root
+            .as_deref_mut()
+            .map(|node| node.get_node(path))
+            .unwrap()
```
divergentdave, line 206: Same here

src/bt.rs
armfazh: updated, hope is clearer now.

src/bt.rs
armfazh: good catch

src/bt.rs
cjpatton: @armfazh  proposes to land this PR as-is and consider this change later on. I'm cool with that if you all are.

cjpatton (approved):

src/bt.rs
```diff
@@ -0,0 +1,487 @@
...
+        match CodecMarker::decode(bytes)? {
+            CodecMarker::Leaf => Ok(Self::default()),
+            CodecMarker::Inner => {
+                bytes.seek(SeekFrom::Start(0))?;
+                let root = Some(Box::new(Node::decode_with_param(param, bytes)?));
+                Ok(Self { root })
+            }
+        }
+    }
+}
```
divergentdave, line 349: The asymmetry between encoding and decoding, and the seek, make me a bit uneasy. This is ultimately because the encoding of a Node must start with a 1 byte, so BinaryTree uses that byte to signal a non-empty root node, and writes/reads its own 0 byte otherwise for an empty node. But, in the 1 case, that first byte has to be read by the `BinaryTree` impl once, then read again by the `Node` impl after rewinding.

I'd like to get rid of the seeking to make future maintenance easier, particularly in case we make further big changes to the codec traits. Moreover, if a `BinaryTree<V>` were included as a field in a larger message, seeking to the start of the buffer would be incorrect -- we'd instead need to save `bytes.stream_position()` before reading the `CodecMarker`, and rewind by seeking to that position.

I suggest either eliminating the codec impls on `Node<V>`, combining them with these, and then initializing the `BinaryTree` impl's encoding stack with `vec![self.root.as_ref()]`, or moving common codec logic into private functions that are used by impls for both `Node<V>` and `BinaryTree<V>`. If we keep codec impls on `Node<V>`, I'd suggest redefining the serialization format such that a node doesn't start with a `CodecMarker`, but a `CodecMarker` still appears before each child slot. This might require changing what enum is stored in the stack, but maybe there's a more clever way to handle this.

armfazh: it wasn't so difficult to remove the seek call after all while keeping serialization of `Node<V>`.

src/bt.rs
```diff
@@ -0,0 +1,487 @@
...
+//! let mut tree = BinaryTree::default();
+//! tree.insert(bits!(), '@');
+//! tree.insert(bits!(0), 'a');
+//! tree.insert(bits!(1), 'b');
+//! tree.insert(bits!(0, 0), 'c');
+//! tree.insert(bits!(0, 1), 'd');
+//! tree.insert(bits!(1, 0), 'e');
+//! tree.insert(bits!(1, 1), 'f');
+
+// Remove these lines once the module gets used by Mastic implementation.
```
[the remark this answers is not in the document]
armfazh: done

divergentdave (approved):

PR #985, opened 2024-03-26
Run `cargo update`.

branlwyd:
Many transitive dependencies had minor- or patch-version updates.


branlwyd:
Closing -- lots of auditing required, and some of updates apparently don't work at MSRV (did not investigate further).

PR #1009, opened 2024-04-22
CI runs only on divviup repository

armfazh:
This prevents from running CI jobs in forks (which may have CI limitations).


divergentdave:
What CI limitations are you running into? Would disabling actions entirely in the fork work? (in Settings -> Actions -> General -> Disable actions)

PR #1013, opened 2024-04-29
Shared ZK implementation

hannahdaviscrypto:
Partially addresses #947.

A module for computing fully linear proofs on data that is already secret shared, using the Prio3 method to derive joint randomness from the input shares via an XOF.

### DIFF
--- a/src/flp/szk.rs
+++ b/src/flp/szk.rs
@@ -1,0 +1,609 @@
+// SPDX-License-Identifier: MPL-2.0
+
+//! A wrapper for the [FLP](crate::flp) proof system as it is used in Mastic.
+//!
+//! [`Szk`] is a wrapper struct for the FLP proof system (accessible through the [`Type`] trait)
+//! that is designed to accept inputs that are already secret-shared between two aggregators
+//! and returns proofs that are also shared between the two aggregators. The underlying FLP
+//! accepts inputs that are vectors over a finite field, and its random coins and proofs are also
+//! vectors over the same field. In contrast, only the leader's SZK proof share contains any finite
+//! field elements, and the initial coins are bytestrings ([`Seed`]s). The wrapper struct defined
+//! here uses an [`Xof`] (to be modeled as a random oracle) to sample coins and the helper's proof share,
+//! following a strategy similar to [`Prio3`](crate::vdaf::prio3::Prio3).
+
+use crate::{
+    codec::{CodecError, Encode},
+    field::{FftFriendlyFieldElement, FieldElement},
+    flp::{FlpError, Type},
+    prng::{Prng, PrngError},
+    vdaf::xof::{IntoFieldVec, Seed, Xof, XofTurboShake128},
+};
+use std::{borrow::Cow, marker::PhantomData};
+
+// Domain separation tags
+const DST_PROVE_RANDOMNESS: u16 = 0;
+const DST_PROOF_SHARE: u16 = 1;
+const DST_QUERY_RANDOMNESS: u16 = 2;
+const DST_JOINT_RAND_SEED: u16 = 3;
+const DST_JOINT_RAND_PART: u16 = 4;
+const DST_JOINT_RANDOMNESS: u16 = 5;
+
+const MASTIC_VERSION: u8 = 0;
+
+/// Errors propagated by methods in this module.
+#[derive(Debug, thiserror::Error)]
+#[non_exhaustive]
+pub enum SzkError {
+    #[error("Szk decide error: {0}")]
+    /// Returned for errors in Szk verification step
+    Decide(String),
+
+    #[error("Szk query error: {0}")]
+    /// Returned for errors in query evaluation
+    Query(String),
+
+    /// Returned if an FLP operation encountered an error.
+    #[error("Flp error: {0}")]
+    Flp(#[from] FlpError),
+
+    /// PRNG error.
+    #[error("prng error: {0}")]
+    Prng(#[from] PrngError),
+
+    /// Codec error.
+    #[error("codec error: {0}")]
+    Codec(#[from] CodecError),
+}
+
+/// Contains an FLP proof share, and if joint randomness is needed, the blind
+/// used to derive it and the other party's joint randomness part.
+#[derive(Clone)]
+pub enum SzkProofShare<F: FftFriendlyFieldElement, const SEED_SIZE: usize> {
+    /// Leader's proof share is uncompressed. The first Seed is a blind, second
+    /// is a joint randomness part.
+    Leader {
+        uncompressed_proof_share: Vec<F>,
+        leader_blind_and_helper_joint_rand_part: Option<(Seed<SEED_SIZE>, Seed<SEED_SIZE>)>,
+    },
+    /// The Helper uses one seed for both its compressed proof share and as the blind for its joint
+    /// randomness.
+    Helper {
+        proof_share_seed_and_blind: Seed<SEED_SIZE>,
+        leader_joint_rand_part: Option<Seed<SEED_SIZE>>,
+    },
+}
+
+/// A tuple containing the state and messages produced by an SZK query.
+#[derive(Clone)]
+pub(crate) struct SzkQueryShare<F, const SEED_SIZE: usize> {
+    joint_rand_part: Option<Seed<SEED_SIZE>>,
+    verifier: SzkVerifier<F>,
+}
+
+/// The state that needs to be stored by an Szk verifier between query() and decide()
+pub(crate) struct SzkQueryState<const SEED_SIZE: usize> {
+    joint_rand_seed: Option<Seed<SEED_SIZE>>,
+}
+
+/// Verifier type for the SZK proof.
+pub type SzkVerifier<F> = Vec<F>;
+
+/// Main struct encapsulating the shared zero-knowledge functionality. The type
+/// T is the underlying FLP proof system. P is the XOF used to derive all random
+/// coins (it should be indifferentiable from a random oracle for security.)
+pub struct Szk<T, P, const SEED_SIZE: usize>
+where
+    T: Type,
+    P: Xof<SEED_SIZE>,
+{
+    typ: T,
+    algorithm_id: u32,
+    phantom: PhantomData<P>,
+}
+
+#[cfg(test)]
+impl<T: Type> Szk<T, XofTurboShake128, 16> {
+    /// Create an instance of [`Szk`] using [`XofTurboShake128`].
+    pub fn new_turboshake128(typ: T, algorithm_id: u32) -> Self {
+        Szk::new(typ, algorithm_id)
+    }
+}
+
+impl<T, P, const SEED_SIZE: usize> Szk<T, P, SEED_SIZE>
+where
+    T: Type,
+    P: Xof<SEED_SIZE>,
+{
+    /// Construct an instance of this sharedZK proof system with the underlying
+    /// FLP.
+    pub fn new(typ: T, algorithm_id: u32) -> Self {
+        Self {
+            typ,
+            algorithm_id,
+            phantom: PhantomData,
+        }
+    }
+
+    fn domain_separation_tag(&self, usage: u16) -> [u8; 8] {
+        let mut dst = [0u8; 8];
+        dst[0] = MASTIC_VERSION;
+        dst[1] = 0; // algorithm class
+        dst[2..6].copy_from_slice(&(self.algorithm_id).to_be_bytes());
+        dst[6..8].copy_from_slice(&usage.to_be_bytes());
+        dst
+    }
+
+    /// Derive a vector of random field elements for consumption by the FLP
+    /// prover.
+    fn derive_prove_rand(&self, prove_rand_seed: &Seed<SEED_SIZE>) -> Vec<T::Field> {
+        P::seed_stream(
+            prove_rand_seed,
+            &self.domain_separation_tag(DST_PROVE_RANDOMNESS),
+            &[],
+        )
+        .into_field_vec(self.typ.prove_rand_len())
+    }
+
+    fn derive_joint_rand_part(
+        &self,
+        aggregator_blind: &Seed<SEED_SIZE>,
+        measurement_share: &[T::Field],
+        nonce: &[u8; 16],
+    ) -> Result<Seed<SEED_SIZE>, SzkError> {
+        let mut xof = P::init(
+            aggregator_blind.as_ref(),
+            &self.domain_separation_tag(DST_JOINT_RAND_PART),
+        );
+        xof.update(nonce);
+        // Encode measurement_share (currently an array of field elements) into
+        // bytes and include it in the XOF input.
+        let mut encoding_buffer = Vec::with_capacity(T::Field::ENCODED_SIZE);
+        for field_elem in measurement_share {
+            field_elem.encode(&mut encoding_buffer)?;
+            xof.update(&encoding_buffer);
+            encoding_buffer.clear();
+        }
+        Ok(xof.into_seed())
+    }
+
+    fn derive_joint_rand_seed(
+        &self,
+        leader_joint_rand_part: &Seed<SEED_SIZE>,
+        helper_joint_rand_part: &Seed<SEED_SIZE>,
+    ) -> Seed<SEED_SIZE> {
+        let mut xof = P::init(
+            &[0; SEED_SIZE],
+            &self.domain_separation_tag(DST_JOINT_RAND_SEED),
+        );
+        xof.update(&leader_joint_rand_part.0);
+        xof.update(&helper_joint_rand_part.0);
+        xof.into_seed()
+    }
+
+    fn derive_joint_rand_and_seed(
+        &self,
+        leader_joint_rand_part: &Seed<SEED_SIZE>,
+        helper_joint_rand_part: &Seed<SEED_SIZE>,
+    ) -> (Seed<SEED_SIZE>, Vec<T::Field>) {
+        let joint_rand_seed =
+            self.derive_joint_rand_seed(leader_joint_rand_part, helper_joint_rand_part);
+        let joint_rand = P::seed_stream(
+            &joint_rand_seed,
+            &self.domain_separation_tag(DST_JOINT_RANDOMNESS),
+            &[],
+        )
+        .into_field_vec(self.typ.joint_rand_len());
+
+        (joint_rand_seed, joint_rand)
+    }
+
+    fn derive_helper_proof_share(&self, proof_share_seed: &Seed<SEED_SIZE>) -> Vec<T::Field> {
+        Prng::from_seed_stream(P::seed_stream(
+            proof_share_seed,
+            &self.domain_separation_tag(DST_PROOF_SHARE),
+            &[],
+        ))
+        .take(self.typ.proof_len())
+        .collect()
+    }
+
+    fn derive_query_rand(&self, verify_key: &[u8; SEED_SIZE], nonce: &[u8; 16]) -> Vec<T::Field> {
+        let mut xof = P::init(
+            verify_key,
+            &self.domain_separation_tag(DST_QUERY_RANDOMNESS),
+        );
+        xof.update(nonce);
+        xof.into_seed_stream()
+            .into_field_vec(self.typ.query_rand_len())
+    }
+
+    pub(crate) fn has_joint_rand(&self) -> bool {
+        self.typ.joint_rand_len() > 0
+    }
+
+    fn prove(
+        &self,
+        leader_input_share: &[T::Field],
+        helper_input_share: &[T::Field],
+        encoded_measurement: &[T::Field],
+        rand_seeds: [Seed<SEED_SIZE>; 2],
+        leader_seed_opt: Option<Seed<SEED_SIZE>>,
+        nonce: &[u8; 16],
+    ) -> Result<[SzkProofShare<T::Field, SEED_SIZE>; 2], SzkError> {
+        let [prove_rand_seed, helper_seed] = rand_seeds;
+        // If joint randomness is used, derive it from the two input shares,
+        // the seeds used to blind the derivation, and the nonce. Pass the
+        // leader its blinding seed and the helper's joint randomness part, and
+        // pass the helper the leader's joint randomness part. (The seed used to
+        // derive the helper's proof share is reused as the helper's blind.)
+        let (leader_blind_and_helper_joint_rand_part, leader_joint_rand_part, joint_rand) =
+            if let Some(leader_seed) = leader_seed_opt.clone() {
+                let leader_joint_rand_part =
+                    self.derive_joint_rand_part(&leader_seed, leader_input_share, nonce)?;
+                let helper_joint_rand_part =
+                    self.derive_joint_rand_part(&helper_seed, helper_input_share, nonce)?;
+                let (_joint_rand_seed, joint_rand) = self
+                    .derive_joint_rand_and_seed(&leader_joint_rand_part, &helper_joint_rand_part);
+                (
+                    Some((leader_seed, helper_joint_rand_part)),
+                    Some(leader_joint_rand_part),
+                    joint_rand,
+                )
+            } else {
+                (None, None, Vec::new())
+            };
+
+        let prove_rand = self.derive_prove_rand(&prove_rand_seed);
+        let mut leader_proof_share =
+            self.typ
+                .prove(encoded_measurement, &prove_rand, &joint_rand)?;
+
+        // Generate the proof shares.
+        for (x, y) in leader_proof_share
+            .iter_mut()
+            .zip(self.derive_helper_proof_share(&helper_seed))
+        {
+            *x -= y;
+        }
+
+        // Construct the output messages.
+        let leader_proof_share = SzkProofShare::Leader {
+            uncompressed_proof_share: leader_proof_share,
+            leader_blind_and_helper_joint_rand_part,
+        };
+        let helper_proof_share = SzkProofShare::Helper {
+            proof_share_seed_and_blind: helper_seed.clone(),
+            leader_joint_rand_part,
+        };
+        Ok([leader_proof_share, helper_proof_share])
+    }
+
+    fn query(
+        &self,
+        input_share: &[T::Field],
+        proof_share: SzkProofShare<T::Field, SEED_SIZE>,
+        verify_key: &[u8; SEED_SIZE],
+        nonce: &[u8; 16],
+    ) -> Result<(SzkQueryShare<T::Field, SEED_SIZE>, SzkQueryState<SEED_SIZE>), SzkError> {
+        let query_rand = self.derive_query_rand(verify_key, nonce);
+        let flp_proof_share = match proof_share {
+            SzkProofShare::Leader {
+                ref uncompressed_proof_share,
+                ..
+            } => Cow::Borrowed(uncompressed_proof_share),
+            SzkProofShare::Helper {
+                ref proof_share_seed_and_blind,
+                ..
+            } => Cow::Owned(self.derive_helper_proof_share(proof_share_seed_and_blind)),
+        };
+
+        let (joint_rand, joint_rand_seed, joint_rand_part) = if self.has_joint_rand() {
+            let ((joint_rand_seed, joint_rand), host_joint_rand_part) = match proof_share {
+                SzkProofShare::Leader {
+                    uncompressed_proof_share: _,
+                    leader_blind_and_helper_joint_rand_part,
+                } => match leader_blind_and_helper_joint_rand_part {
+                    Some((seed, helper_joint_rand_part)) => {
+                        match self.derive_joint_rand_part(&seed, input_share, nonce) {
+                            Ok(leader_joint_rand_part) => (
+                                self.derive_joint_rand_and_seed(
+                                    &leader_joint_rand_part,
+                                    &helper_joint_rand_part,
+                                ),
+                                leader_joint_rand_part,
+                            ),
+                            Err(e) => return Err(e),
+                        }
+                    }
+                    None => {
+                        return Err(SzkError::Query(
+                            "leader_blind_and_helper_joint_rand_part should be set".to_string(),
+                        ))
+                    }
+                },
+                SzkProofShare::Helper {
+                    proof_share_seed_and_blind,
+                    leader_joint_rand_part,
+                } => match leader_joint_rand_part {
+                    Some(leader_joint_rand_part) => match self.derive_joint_rand_part(
+                        &proof_share_seed_and_blind,
+                        input_share,
+                        nonce,
+                    ) {
+                        Ok(helper_joint_rand_part) => (
+                            self.derive_joint_rand_and_seed(
+                                &leader_joint_rand_part,
+                                &helper_joint_rand_part,
+                            ),
+                            helper_joint_rand_part,
+                        ),
+                        Err(e) => return Err(e),
+                    },
+                    None => {
+                        return Err(SzkError::Query(
+                            "leader_joint_rand_part should be set".to_string(),
+                        ))
+                    }
+                },
+            };
+            (
+                joint_rand,
+                Some(joint_rand_seed),
+                Some(host_joint_rand_part),
+            )
+        } else {
+            (Vec::new(), None, None)
+        };
+        let verifier_share = self.typ.query(
+            input_share,
+            flp_proof_share.as_ref(),
+            &query_rand,
+            &joint_rand,
+            2,
+        )?;
+        Ok((
+            SzkQueryShare {
+                joint_rand_part,
+                verifier: verifier_share,
+            },
+            SzkQueryState { joint_rand_seed },
+        ))
+    }
+
+    /// Returns true if the verifier message indicates that the input from which
+    /// it was generated is valid.
+    fn decide(
+        &self,
+        verifier: &[T::Field],
+        leader_joint_rand_part_opt: Option<Seed<SEED_SIZE>>,
+        helper_joint_rand_part_opt: Option<Seed<SEED_SIZE>>,
+        joint_rand_seed_opt: Option<Seed<SEED_SIZE>>,
+    ) -> Result<bool, SzkError> {
+        // Check if underlying FLP proof validates
+        let check_flp_proof = self.typ.decide(verifier)?;
+        if !check_flp_proof {
+            return Ok(false);
+        }
+        // Check that joint randomness was properly derived from both
+        // aggregators' parts
+        match (
+            joint_rand_seed_opt,
+            leader_joint_rand_part_opt,
+            helper_joint_rand_part_opt,
+        ) {
+            (Some(joint_rand_seed), Some(leader_joint_rand_part), Some(helper_joint_rand_part)) => {
+                let expected_joint_rand_seed =
+                    self.derive_joint_rand_seed(&leader_joint_rand_part, &helper_joint_rand_part);
+                Ok(joint_rand_seed == expected_joint_rand_seed)
+            }
+            (None, None, None) => Ok(true),
+            (_, _, _) => Err(SzkError::Decide(
+                "at least one of the input seeds is missing".to_string(),
+            )),
+        }
+    }
+}
+
+mod tests {
+    use super::*;
+    use crate::field::Field128 as TestField;
+    use crate::field::{random_vector, FieldElementWithInteger};
+    use crate::flp::types::{Count, Sum};
+    use crate::flp::Type;
+    use rand::{thread_rng, Rng};
+
+    fn generic_szk_test<T: Type>(typ: T, encoded_measurement: &[T::Field], valid: bool) {
+        let mut nonce = [0u8; 16];
+        let mut verify_key = [0u8; 16];
+        let algorithm_id = 5;
+        let szk_typ = Szk::new_turboshake128(typ.clone(), algorithm_id);
+        thread_rng().fill(&mut verify_key[..]);
+        thread_rng().fill(&mut nonce[..]);
+        let prove_rand_seed = Seed::<16>::generate().unwrap();
+        let helper_seed = Seed::<16>::generate().unwrap();
+        let leader_seed_opt = if szk_typ.has_joint_rand() {
+            Some(Seed::<16>::generate().unwrap())
+        } else {
+            None
+        };
+        let helper_input_share: Vec<T::Field> = random_vector(szk_typ.typ.input_len()).unwrap();
+        let mut leader_input_share = encoded_measurement.to_owned();
+        for (x, y) in leader_input_share.iter_mut().zip(&helper_input_share) {
+            *x -= *y;
+        }
+
+        let proof_shares = szk_typ.prove(
+            &leader_input_share,
+            &helper_input_share,
+            encoded_measurement,
+            [prove_rand_seed, helper_seed],
+            leader_seed_opt,
+            &nonce,
+        );
+
+        let [l_proof_share, h_proof_share] = proof_shares.unwrap();
+        let (l_query_share, l_query_state) = szk_typ
+            .query(
+                &leader_input_share,
+                l_proof_share.clone(),
+                &verify_key,
+                &nonce,
+            )
+            .unwrap();
+        let (h_query_share, h_query_state) = szk_typ
+            .query(&helper_input_share, h_proof_share, &verify_key, &nonce)
+            .unwrap();
+
+        let mut verifier = l_query_share.clone().verifier;
+
+        for (x, y) in verifier.iter_mut().zip(h_query_share.clone().verifier) {
+            *x += y;
+        }
+        let h_jr_part = h_query_share.clone().joint_rand_part;
+        let h_jr_seed = h_query_state.joint_rand_seed;
+        let l_jr_part = l_query_share.joint_rand_part;
+        let l_jr_seed = l_query_state.joint_rand_seed;
+        if let Ok(leader_decision) = szk_typ.decide(
+            &verifier,
+            l_jr_part.clone(),
+            h_jr_part.clone(),
+            l_jr_seed.clone(),
+        ) {
+            assert_eq!(
+                leader_decision, valid,
+                "Leader incorrectly determined validity",
+            );
+        } else {
+            panic!("Leader failed during decision");
+        };
+        if let Ok(helper_decision) = szk_typ.decide(
+            &verifier,
+            l_jr_part.clone(),
+            h_jr_part.clone(),
+            h_jr_seed.clone(),
+        ) {
+            assert_eq!(
+                helper_decision, valid,
+                "Helper incorrectly determined validity",
+            );
+        } else {
+            panic!("Helper failed during decision");
+        };
+
+        //test mutated jr seed
+        if szk_typ.has_joint_rand() {
+            let joint_rand_seed_opt = Some(Seed::<16>::generate().unwrap());
+            if let Ok(leader_decision) = szk_typ.decide(
+                &verifier,
+                l_jr_part.clone(),
+                h_jr_part.clone(),
+                joint_rand_seed_opt,
+            ) {
+                assert!(!leader_decision, "Leader accepted wrong jr seed");
+            };
+        };
+
+        //test mutated verifier
+        let mut verifier = l_query_share.verifier;
+
+        for (x, y) in verifier.iter_mut().zip(h_query_share.clone().verifier) {
+            *x += y + T::Field::from(
+                <T::Field as FieldElementWithInteger>::Integer::try_from(7).unwrap(),
+            );
+        }
+
+        let leader_decision = szk_typ
+            .decide(
+                &verifier,
+                l_jr_part.clone(),
+                h_jr_part.clone(),
+                l_jr_seed.clone(),
+            )
+            .unwrap();
+        assert!(!leader_decision, "Leader validated after proof mutation");
+
+        // test mutated input share
+        let mut mutated_input = leader_input_share.clone();
+        mutated_input[0] *=
+            T::Field::from(<T::Field as FieldElementWithInteger>::Integer::try_from(23).unwrap());
+        let (mutated_query_share, mutated_query_state) = szk_typ
+            .query(&mutated_input, l_proof_share.clone(), &verify_key, &nonce)
+            .unwrap();
+        let mut verifier = mutated_query_share.verifier;
+
+        for (x, y) in verifier.iter_mut().zip(h_query_share.clone().verifier) {
+            *x += y;
+        }
+
+        let mutated_jr_seed = mutated_query_state.joint_rand_seed;
+        let mutated_jr_part = mutated_query_share.joint_rand_part;
+        if let Ok(leader_decision) = szk_typ.decide(
+            &verifier,
+            mutated_jr_part.clone(),
+            h_jr_part.clone(),
+            mutated_jr_seed,
+        ) {
+            assert!(!leader_decision, "Leader validated after input mutation");
+        };
+
+        // test mutated proof share
+        let (mut mutated_proof, leader_blind_and_helper_joint_rand_part) = match l_proof_share {
+            SzkProofShare::Leader {
+                uncompressed_proof_share,
+                leader_blind_and_helper_joint_rand_part,
+            } => (
+                uncompressed_proof_share.clone(),
+                leader_blind_and_helper_joint_rand_part,
+            ),
+            _ => (vec![], None),
+        };
+        mutated_proof[0] *=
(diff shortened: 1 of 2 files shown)


hannahdaviscrypto:
@cjpatton Would love if you could take a look at this, thank you!

src/szk.rs
```diff
@@ -0,0 +1,459 @@
...
+///Helper's seed is compressed using the same blind as its joint randomness
+Helper(Option<&'a Seed<SEED_SIZE>>, Option<Seed<SEED_SIZE>>)
+}
+
+/// Verifier type for Szk Proof
+pub type SzkVerifier<F> = Vec<F>;
+
+
+///Main struct encapsulating the shared zero-knowledge functionality.
+/// The type T is the undelrying FLP proof system
```
divergentdave, line 74: nit: these sentences would run together after Markdown processing
```suggestion
/// The type T is the underlying FLP proof system.
```

src/szk.rs
```diff
@@ -0,0 +1,459 @@
...
+    typ: T,
+    dst: [u8; 8],
+    phantom: PhantomData<P>,
+}
+
+impl<T: Type> Szk<T, XofTurboShake128, 16> {
+    /// Create an instance of [`Poplar1`] using [`XofTurboShake128`]. The caller provides the bit length of
+    /// each measurement (`BITS` as defined in [[draft-irtf-cfrg-vdaf-08]]).
+    ///
+    /// [draft-irtf-cfrg-vdaf-08]: https://datatracker.ietf.org/doc/draft-irtf-cfrg-vdaf/08/
```
divergentdave, line 91:
```suggestion
    /// Create an instance of [`Szk`] using [`XofTurboShake128`]. The caller provides the bit length of
    /// each measurement.
```

src/szk.rs
```diff
@@ -0,0 +1,459 @@
...
+where
+    T: Type,
+    P: Xof<SEED_SIZE>,
+{
+    /// Construct an instance of this sharedZK proof system with the underlying FLP
+    /// and a space to hold XOF state for coin derivation.
+    pub fn new(
+        typ: T,
+        dst: [u8; 8]
+    ) -> Self {
```
divergentdave, line 107: I'd prefer not exposing the bare DST in public API. Also, I think referring to this as "XOF state" is misleading, since that sounds like it's about the internal TurobSHAKE sponge buffer. If the constructor just took in the algorithm ID and stored it, and new DSTs were constructed afresh rather than copying from a template, that should be plenty fast for our purposes, and it would be more readable and misuse-resistant.

src/szk.rs
```diff
@@ -0,0 +1,459 @@
...
+    fn derive_joint_rands(
+        &self,
+        leader_joint_rand_part: &Option<Seed<SEED_SIZE>>,
+        helper_joint_rand_part: &Option<Seed<SEED_SIZE>>,
+    ) -> (Seed<SEED_SIZE>, Vec<T::Field>) {
+        if self.typ.joint_rand_len() <= 0 {
+            return (Seed::from_bytes([0; SEED_SIZE]), Vec::<T::Field>::with_capacity(0));
+        };
+        let joint_rand_seed = self.derive_joint_rand_seed(leader_joint_rand_part.as_ref().expect("jr_len > 0"),
+                                                          helper_joint_rand_part.as_ref().expect("jr_len > 0"));
```
divergentdave, line 188: I think this could be simplified a bit if it only handled the case where joint randomness is used, and its callers passed seeds instead of `Option`s of seeds. It would be nice to remove the branch that returns a zero seed. I think this will ultimately require changing the signature of `query()` to return an Option<Seed> for the joint randomness seed. There again, it'd be nice to not have a case where a seed is all zeros.

src/szk.rs
```diff
@@ -0,0 +1,459 @@
...
+        )
+        .into_field_vec(self.typ.prove_rand_len())
+    }
+
+    fn derive_joint_rand_part(
+        &self,
+        aggregator_blind: &Seed<SEED_SIZE>,
+        measurement_share: &[T::Field],
+        nonce: &[u8; 16],
+    ) -> Result<Seed<SEED_SIZE>, SzkError> {
```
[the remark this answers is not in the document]
hannahdaviscrypto: Yes, we plan to update the spec to match after the implementation is finished.

src/szk.rs
```diff
@@ -0,0 +1,459 @@
...
+}
+
+/// Contains an FLP proof share, and if joint randomness is needed, the blind
+/// used to derive it and the other party's joint randomness part.
+pub enum SzkProofShare<'a, const SEED_SIZE: usize, F: FftFriendlyFieldElement>{
+///Leader's proof share is uncompressed.
+///The first Seed is a blind, second is a joint randomness part.
+Leader(Vec<F>, Option<&'a Seed<SEED_SIZE>>, Option<Seed<SEED_SIZE>>),
+///Helper's seed is compressed using the same blind as its joint randomness
+Helper(Option<&'a Seed<SEED_SIZE>>, Option<Seed<SEED_SIZE>>)
```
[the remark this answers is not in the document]
hannahdaviscrypto: You're correct, the first seed is always present. I was trying to match types between the leader seed & helper seed, but I've found a workaround.

src/szk.rs
hannahdaviscrypto: That's fair, I'll add the algorithm ID to the constructor. The XOF space referenced in the documentation string isn't DST, though, it's the phantom data.

divergentdave: Ah, okay. I think we can skip mentioning the `PhantomData` itself in the doc comment, it's mainly an implementation detail to make sure that you can't use multiple XOFs with the same object, and/or to let type inference figure out the XOF type by following dataflow out from the constructor.

src/szk.rs
```diff
@@ -0,0 +1,469 @@
...
+/// P is the XOF used to derive all random coins (it should be indifferentiable
+/// from a random oracle for security.)
+#[cfg(test)]
+pub struct Szk<T, P, const SEED_SIZE: usize>
+where
+    T: Type,
+    P: Xof<SEED_SIZE>,
+{
+    typ: T,
+    algorithm_id: u32,
```
tgeoghegan, line 113: Where are these algorithm IDs assigned? Perhaps an enum would be a good idea?

src/szk.rs
```diff
@@ -0,0 +1,469 @@
...
+impl<T, P, const SEED_SIZE: usize> Szk<T, P, SEED_SIZE>
+where
+    T: Type,
+    P: Xof<SEED_SIZE>,
+{
+    /// Construct an instance of this sharedZK proof system with the underlying FLP.
+    pub fn new(typ: T, algorithm_id: u32) -> Self {
+        Self {
+            typ,
+            algorithm_id,
```
tgeoghegan, line 136: Should we have a check for whether `algorithm_id` is valid, and have this return some error if not? n.b. if we have `enum Algorithm`, then the algorithm ID would be correct by construction.

cjpatton (commented):
Kicking this off with a round of Rust-related comments. I'll get to the crypto bits once the code starts to smooth out. Looking good so far!

src/szk.rs
```diff
@@ -0,0 +1,469 @@
...
+    *    if self.typ.joint_rand_len() == 0 {
+    *        // One seed for the helper's proof share and one for proving
+    *        // randomness.
+    *        2 * SEED_SIZE
+    *    } else {
+    *        // One seed for the leader, one for the helper, and one for proving
+    *        // randomness.
+    *        3 * SEED_SIZE
+    *    }
+    }*/
```
tgeoghegan, line 258: Delete commented out code

src/szk.rs
```diff
@@ -0,0 +1,469 @@
...
+
+    #[cfg(test)]
+    fn prove<'a>(
+        &'a self,
+        leader_input_share: &[T::Field],
+        helper_input_share: &[T::Field],
+        encoded_measurement: &[T::Field],
+        random_seeds: &'a [Seed<SEED_SIZE>],
```
tgeoghegan, line 266: It seems like the leader seed is optional, based on whether the underlying FLP has joint randomness. I think this function would be more clear with arguments:
```suggestion
        prove_rand_seed: &'a Seed<SEED_SIZE>,
        helper_seed: &'a Seed<SEED_SIZE>,
        leader_seed: Option<&'a [Seed<SEED_SIZE>>,
```

src/szk.rs
```diff
@@ -0,0 +1,469 @@
...
+    pub fn output_len(&self) -> usize {
+        self.typ.output_len()
+    }
+
+    /// The verifier length of the underlying FLP.
+    pub fn verifier_len(&self) -> usize {
+        self.typ.verifier_len()
+    }
+    #[cfg(test)]
+    fn adjust_domsep_tag(&self, usage: u16) -> [u8; 8] {
```
tgeoghegan, line 151: It looks like `usage` must be one of the `DST_` constants above, but this isn't enforced. I think this would be more clear with an `enum Usage`, because then you'd only define variants for usages that exist.

src/szk.rs
```diff
@@ -0,0 +1,469 @@
...
+/// used to derive it and the other party's joint randomness part.
+pub enum SzkProofShare<'a, const SEED_SIZE: usize, F: FftFriendlyFieldElement> {
+    ///Leader's proof share is uncompressed.
+    ///The first Seed is a blind, second is a joint randomness part.
+    Leader(Vec<F>, Option<&'a Seed<SEED_SIZE>>, Option<Seed<SEED_SIZE>>),
+    ///Helper's seed is compressed using the same blind as its joint randomness
+    Helper(&'a Seed<SEED_SIZE>, Option<Seed<SEED_SIZE>>),
+}
+
+/// A tuple of two proof shares: one for each aggregator
```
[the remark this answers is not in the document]
divergentdave: I think the tuple was introduced to help with moving shares out of the argument. You can do the same with an array if you destructure the whole thing at once, i.e. `let [leader, helper] = shares;`.

src/szk.rs
```diff
@@ -0,0 +1,469 @@
...
+const DST_JOINT_RAND_PART: u16 = 4;
+#[cfg(test)]
+const DST_JOINT_RANDOMNESS: u16 = 5;
+// I don't know how to access the crate version number .
+#[cfg(test)]
+const VERSION: u8 = 8;
+
+/// Errors propagated by methods in this module.
+#[derive(Debug, thiserror::Error)]
+#[non_exhaustive]
```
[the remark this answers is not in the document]
divergentdave: I prefer non_exhaustive on error enums because it gives us room to introduce new error variants without it being a breaking change. We've had to add this to other error types, so starting with the `#[non_exhaustive]` attribute here may save us later.

src/szk.rs
```diff
@@ -0,0 +1,469 @@
...
+        adjusted_dst[0] = VERSION;
+        adjusted_dst[1] = 0; // algorithm class
+        adjusted_dst[2..6].copy_from_slice(&(self.algorithm_id).to_be_bytes());
+        adjusted_dst[6..8].copy_from_slice(&usage.to_be_bytes());
+        adjusted_dst
+    }
+
+    /// Derive a vector of random field elements for consumption by the FLP prover
+    /// Question: is it ok to have the binder value be zero here? This is different from
+    /// both Prio3 (where it's the number of proofs) and the Mastic paper (where it's empty).
```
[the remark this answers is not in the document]
hannahdaviscrypto: I thought the underlying function required a non-empty binder. Since it works with it as empty, all's well.

src/szk.rs
```diff
@@ -0,0 +1,469 @@
...
+const DST_QUERY_RANDOMNESS: u16 = 2;
+#[cfg(test)]
+const DST_JOINT_RAND_SEED: u16 = 3;
+#[cfg(test)]
+const DST_JOINT_RAND_PART: u16 = 4;
+#[cfg(test)]
+const DST_JOINT_RANDOMNESS: u16 = 5;
+// I don't know how to access the crate version number .
+#[cfg(test)]
+const VERSION: u8 = 8;
```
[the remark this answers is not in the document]
hannahdaviscrypto: I do mean the latter. There's already a constant defined in the VDAF module, which it would be nice to not duplicate, but I couldn't figure out how to import it.

src/szk.rs
```diff
@@ -0,0 +1,469 @@
...
+    #[error("prove error: {0}")]
+    Prove(String),
+
+    /// Calling [`Type::query`] returned an error.
+    #[error("query error: {0}")]
+    Query(String),
+
+    /// Calling [`Type::decide`] returned an error.
+    #[error("decide error: {0}")]
+    Decide(String),
```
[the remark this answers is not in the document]
hannahdaviscrypto: These variants aren't for errors in the Type methods. There is already a variant that wraps the FlpError. But SZK has its own Prove, Query, and Decide methods that can throw errors unrelated to the underlying Type; these variants are to catch those.

src/szk.rs
hannahdaviscrypto: They're passed in from whatever higher-level VDAF is using the SZK (i.e., this will eventually be the Mastic algorithm ID).

src/szk.rs
hannahdaviscrypto: I don't think so, the algorithm ID is just for domain separation between VDAFs, and SZK shouldn't have any awareness of what algorithm is calling it. It just needs to store the algorithm ID because it gets hashed as part of joint randomness derivation.

src/szk.rs
hannahdaviscrypto: Sure

src/szk.rs
hannahdaviscrypto: As above, this gets a little complicated since there might be other usages in the same VDAF that SZK is not aware of.

src/szk.rs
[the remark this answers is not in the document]
cjpatton: I see. If we get rid of the lifetime (see other comment) then clippy may not yell as much.

src/lib.rs
```diff
@@ -29,8 +31,11 @@ mod fp;
     doc(cfg(all(feature = "crypto-dependencies", feature = "experimental")))
 )]
 pub mod idpf;
+#[cfg(feature = "experimental")]
```
divergentdave, line 34: This should be backed out too

src/szk.rs
```diff
@@ -0,0 +1,469 @@
...
+    Prng(#[from] PrngError),
+
+    /// Codec error.
+    #[error("codec error: {0}")]
+    Codec(#[from] CodecError),
+}
+
+/// Contains an FLP proof share, and if joint randomness is needed, the blind
+/// used to derive it and the other party's joint randomness part.
+pub enum SzkProofShare<'a, const SEED_SIZE: usize, F: FftFriendlyFieldElement> {
```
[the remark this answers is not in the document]
hannahdaviscrypto: I don't know the answer to "should we support many aggregators in SZK", but I think it's doable while still keeping Leader/Helper variants--there would just be multiple shares of type Helper.

src/szk.rs
```diff
@@ -0,0 +1,444 @@
...
+        &self,
+        leader_joint_rand_part: &Seed<SEED_SIZE>,
+        helper_joint_rand_part: &Seed<SEED_SIZE>,
+    ) -> (Seed<SEED_SIZE>, Vec<T::Field>) {
+        let joint_rand_seed =
+            self.derive_joint_rand_seed(leader_joint_rand_part, helper_joint_rand_part);
+        let joint_rands = P::seed_stream(
+            &joint_rand_seed,
+            &self.adjust_domsep_tag(DST_JOINT_RANDOMNESS),
+            &[0],
```
divergentdave, line 188: As with proof randomness, I think we could use an empty binder string here too. It should work either way, as this is the only XOF call with this usage constant, but the empty string is easier to explain.
```suggestion
            &[],
```

src/szk.rs
```diff
@@ -0,0 +1,444 @@
...
+
+        (joint_rand_seed, joint_rands)
+    }
+
+    /// Again the binder is zero here, since the helper's agg_id and the number of proofs doesn't vary.
+    fn derive_helper_proofs_share(&self, proofs_share_seed: &Seed<SEED_SIZE>) -> Vec<T::Field> {
+        Prng::from_seed_stream(P::seed_stream(
+            proofs_share_seed,
+            &self.adjust_domsep_tag(DST_PROOF_SHARE),
+            &[0],
```
divergentdave, line 200: Same as above
```suggestion
    fn derive_helper_proofs_share(&self, proofs_share_seed: &Seed<SEED_SIZE>) -> Vec<T::Field> {
        Prng::from_seed_stream(P::seed_stream(
            proofs_share_seed,
            &self.adjust_domsep_tag(DST_PROOF_SHARE),
            &[],
```

src/szk.rs
```diff
@@ -0,0 +1,444 @@
...
+                ..
+            } => Cow::Owned(self.derive_helper_proofs_share(proof_share_seed_and_blind)),
+        };
+
+        let (joint_rand, jr_seed_and_part) = if self.typ.joint_rand_len() > 0 {
+            let (seed, other_joint_rand_part) = match proof {
+                SzkProofShare::Leader {
+                    uncompressed_proof_share: _,
+                    blind_and_joint_rand_part,
+                } => blind_and_joint_rand_part.expect("jr_len > 0 so there is joint_randomness"),
```
divergentdave, line 296: FWIW, we should return an error either here or during deserialization of proof shares if the extra seeds are missing when they shouldn't be. We can take care of this during deserialization so long as we include enough information in a decoding parameter, so this panic is fine for now.

src/szk.rs
```diff
@@ -0,0 +1,444 @@
...
+                } => (
+                    proof_share_seed_and_blind,
+                    joint_rand_part.expect("jr_len > 0 so there is joint randomness"),
+                ),
+            };
+            let own_joint_rand_part = self.derive_joint_rand_part(&seed, input_share, nonce)?;
+            let (leader_joint_rand_part, helper_joint_rand_part) = match agg_id {
+                0 => (&own_joint_rand_part, &other_joint_rand_part),
+                1 => (&other_joint_rand_part, &own_joint_rand_part),
+                _ => return Err(SzkError::Query("bad agg ID".to_string())),
```
divergentdave, line 309: nit: This is unreachable since agg_id is set to zero or one above. We could tidy this up by replacing `agg_id` with `is_leader: bool`.

divergentdave (commented):
I noticed that you rebased or cherry-picked main onto this feature branch. This doesn't work well because each new change from main increases the diff between the feature branch and the last common ancestor commit, which means more chances for conflicts. Ideally, the diff should only have the new `src/szk.rs` file, and two lines in `src/lib.rs` declaring the new module.

There are a few ways to get out of this loop. Merging main into this feature branch would give you a new last common ancestor commit, thus simplifying the diff, though leaving the branch with a complicated history. Rebasing in the other direction, i.e. rebasing the feature branch onto the latest commit from main, would provide the cleanest end result history, though it may take some care. The commits that already appear on main would need to be dropped: rebasing can often do this automatically, but if not this can be manually controlled in an interactive rebase. There may be conflicts to resolve along the way as well, particularly for fixup commits touching existing parts of the library. I think you could alternatively squash the whole feature branch into one commit by first doing a soft reset, cleaning up the changes in the index by unstaging changes from cherry-picked main commits, and then creating a new commit from the index's contents.

src/szk.rs
```diff
@@ -0,0 +1,444 @@
...
+            &self.adjust_domsep_tag(DST_JOINT_RAND_PART),
+        );
+        // Encode measurement_share (currently an array of field elements) into bytes
+        // and include it in the XOF input.
+        let mut bytes = vec![0u8; T::Field::ENCODED_SIZE];
+        for field_elem in measurement_share {
+            field_elem.encode(&mut bytes)?;
+            xof.update(&bytes)
+        }
+        xof.update(nonce);
```
cjpatton, line 160: Proving robustness might require requires extracting the nonce from the attacker's RO query. In that spirit, it's a good idea to set up inputs to the XOF so that we can decode the individual components. I think we're fine in this case since we always know the length of `measurement_share`. Still, I might hash the variable length thing last.

src/szk.rs
```diff
@@ -0,0 +1,444 @@
...
+        let mut xof = P::init(
+            &[0; SEED_SIZE],
+            &self.adjust_domsep_tag(DST_JOINT_RAND_SEED),
+        );
+        xof.update(&leader_joint_rand_part.0);
+        xof.update(&helper_joint_rand_part.0);
+        xof.into_seed()
+    }
+
+    fn derive_joint_rands(
```
cjpatton, line 178: Similar reasoning as for `derive_prove_rands()`.
```suggestion
    fn derive_joint_rand(
```

src/szk.rs
```diff
@@ -0,0 +1,444 @@
...
+            &self.adjust_domsep_tag(DST_JOINT_RANDOMNESS),
+            &[0],
+        )
+        .into_field_vec(self.typ.joint_rand_len());
+
+        (joint_rand_seed, joint_rands)
+    }
+
+    /// Again the binder is zero here, since the helper's agg_id and the number of proofs doesn't vary.
+    fn derive_helper_proofs_share(&self, proofs_share_seed: &Seed<SEED_SIZE>) -> Vec<T::Field> {
```
cjpatton, line 196:
```suggestion
    fn derive_helper_proof_share(&self, proofs_share_seed: &Seed<SEED_SIZE>) -> Vec<T::Field> {
```

src/szk.rs
```diff
@@ -0,0 +1,444 @@
...
+        Prng::from_seed_stream(P::seed_stream(
+            proofs_share_seed,
+            &self.adjust_domsep_tag(DST_PROOF_SHARE),
+            &[0],
+        ))
+        .take(self.typ.proof_len())
+        .collect()
+    }
+
+    fn derive_query_rands(&self, verify_key: &[u8; SEED_SIZE], nonce: &[u8; 16]) -> Vec<T::Field> {
```
cjpatton, line 206:
```suggestion
    fn derive_query_rand(&self, verify_key: &[u8; SEED_SIZE], nonce: &[u8; 16]) -> Vec<T::Field> {
```

src/szk.rs
```diff
@@ -0,0 +1,444 @@
...
+        &self,
+        leader_input_share: &[T::Field],
+        helper_input_share: &[T::Field],
+        encoded_measurement: &[T::Field],
+        rand_seeds: [Seed<SEED_SIZE>; 2],
+        leader_seed_opt: Option<Seed<SEED_SIZE>>,
+        nonce: &[u8; 16],
+    ) -> Result<[SzkProofShare<SEED_SIZE, T::Field>; 2], SzkError> {
+        let [prove_rand_seed, helper_seed] = rand_seeds;
+        let (blind_and_joint_rand_part_for_leader, helper_joint_rand_part, joint_rand) =
```
cjpatton, line 223: nit: "blind and part for leader" and "helper part" are grammatically inconsistent.

src/szk.rs
```diff
@@ -0,0 +1,444 @@
...
+            if self.typ.joint_rand_len() > 0 {
+                let leader_seed = leader_seed_opt.clone().unwrap();
```
cjpatton, line 225: Rule of thumb: Try to avoid `unwrap()` if possible, as this causes the code to panic. In this situation it's fairly clear that the code will never actually panic, but it's good practice to make your code more readable and free of edge cases.
```suggestion
            if let Some(leader_seed) = leader_seed_opt {
```

src/szk.rs
```diff
@@ -0,0 +1,444 @@
...
+                    Some((leader_seed, helper_joint_rand_part)),
+                    Some(leader_joint_rand_part),
+                    jr,
+                )
+            } else {
+                (None, None, Vec::<T::Field>::new())
+            };
+
+        let prove_rand = self.derive_prove_rands(&prove_rand_seed);
+        let mut leader_proofs_share =
```
cjpatton, line 242:
```suggestion
        let mut leader_proof_share =
```

src/szk.rs
```diff
@@ -0,0 +1,444 @@
...
+            self.typ
+                .prove(encoded_measurement, &prove_rand, &joint_rand)?;
+
+        // Generate the proof shares.
+        for (x, y) in leader_proofs_share
```
cjpatton, line 247:
```suggestion
        for (x, y) in leader_proof_share
```

src/szk.rs
```diff
@@ -0,0 +1,444 @@
...
+        }
+
+        // Prep the output messages.
+        let leader_proof_share = SzkProofShare::Leader {
+            uncompressed_proof_share: leader_proofs_share,
+            blind_and_joint_rand_part: blind_and_joint_rand_part_for_leader,
+        };
+        let helper_proof_share = SzkProofShare::Helper {
+            proof_share_seed_and_blind: helper_seed.clone(),
+            joint_rand_part: helper_joint_rand_part,
```
cjpatton, line 261: Question: The Helper is supposed to get the Leader's part, correct? (The Aggregators confirm in the decision step that the Client correctly sent its co-Aggregator's part.)

src/szk.rs
```diff
@@ -0,0 +1,444 @@
...
+        };
+        Ok([leader_proof_share, helper_proof_share])
+    }
+
+    fn query(
+        &self,
+        input_share: &[T::Field],
+        proof: SzkProofShare<SEED_SIZE, T::Field>,
```
cjpatton, line 269: I'd call this the `proof_share` and the proof share for the FLP below the "flp_proof_share".

src/szk.rs
```diff
@@ -0,0 +1,444 @@
...
+                    joint_rand_part,
+                } => (
+                    proof_share_seed_and_blind,
+                    joint_rand_part.expect("jr_len > 0 so there is joint randomness"),
+                ),
+            };
+            let own_joint_rand_part = self.derive_joint_rand_part(&seed, input_share, nonce)?;
+            let (leader_joint_rand_part, helper_joint_rand_part) = match agg_id {
+                0 => (&own_joint_rand_part, &other_joint_rand_part),
+                1 => (&other_joint_rand_part, &own_joint_rand_part),
```
cjpatton, line 308: I like "host" for "own" and "peer" for "other".

src/szk.rs
```diff
@@ -0,0 +1,444 @@
...
+        h_jr_part: Option<Seed<SEED_SIZE>>,
+        joint_rand_seed_opt: Option<Seed<SEED_SIZE>>,
+    ) -> Result<bool, FlpError> {
+        // Check if underlying FLP proof validates
+        let check_flp_proof = self.typ.decide(verifier)?;
+        if !check_flp_proof {
+            return Ok(false);
+        }
+        // Check that joint randomness was properly derived from both aggregators' parts
+        if self.typ.joint_rand_len() > 0 {
```
cjpatton, line 341: Same here: Use `if let Some(_) = joint_rand_seed_opt` to avoid unwrapping.

src/szk.rs
```diff
@@ -0,0 +1,444 @@
...
+        let mut nonce = [0u8; 16];
+        let mut verify_key = [0u8; 16];
+        let algorithm_id = 5;
+        let szk_sum = Szk::new_turboshake128(sum, algorithm_id);
+        let mut coins = [0u8; 48];
+        thread_rng().fill(&mut verify_key[..]);
+        thread_rng().fill(&mut nonce[..]);
+        thread_rng().fill(&mut coins[..]);
+        let mut random = coins.chunks_exact(16);
+        let prove_rand_seed = Seed::<16>::from_bytes(
```
cjpatton, line 373: An easier way to generate a seed is [`Seed::generate()`](https://docs.rs/prio/latest/prio/vdaf/xof/struct.Seed.html#method.generate).

src/szk.rs
```diff
@@ -0,0 +1,444 @@
...
+
+        let l_query_share = szk_sum
+            .query(&leader_input_share, l_proof_share, &verify_key, &nonce)
+            .unwrap();
+
+        let h_query_share = szk_sum
+            .query(&helper_input_share, h_proof_share, &verify_key, &nonce)
+            .unwrap();
+
+        // Note to self: verifier destroyed here
```
cjpatton, line 416: Did you intend to leave this comment?

src/szk.rs
```diff
@@ -0,0 +1,444 @@
...
+            *x += y;
+        }
+        let (h_jr_seed, h_joint_rand) = match h_query_share.joint_rand_seed_and_part {
+            Some([jr_seed, jr_part]) => (Some(jr_seed), Some(jr_part)),
+            None => (None, None),
+        };
+        let (l_jr_seed, l_joint_rand) = match l_query_share.joint_rand_seed_and_part {
+            Some([jr_seed, jr_part]) => (Some(jr_seed), Some(jr_part)),
+            None => (None, None),
+        };
```
cjpatton, line 429: Hmm, this work was already done `query()` correct?

An API that will be more useful for Mastic is one that returns the "expected" joint randomness seed along side the query share. In this test we would want to make sure that the aggregators compute the same expected seed; then we would pass this to `decide()`.

src/szk.rs
```diff
@@ -0,0 +1,444 @@
...
+}
+
+mod tests {
+    use super::*;
+    use crate::field::Field128 as TestField;
+    use crate::flp::types::Sum;
+    use rand::{thread_rng, Rng};
+
+    #[test]
+    fn test_sum() {
```
cjpatton, line 361: I'd like to see a little bit more test coverage.
1. Let's test a `Type` that uses joint randomness and another type that doesn't. That way we ensure all code paths are covered.
2. We should have some negative tests as well, e.g., tweaking: (a) an input share; (b) a proof share; (c) a query share; (d) or the expected joint randomness seed.

To make this easier, it's a good idea to define here a "tester" function that is generic in the `Type`. Then you'd have a `#[test]` for each variant you want to test.

cjpatton (commented):
Looking really good. I have one substantive question (inline); all others I think are easy to resolve. I would expect just one more round of review from me.

src/szk.rs
hannahdaviscrypto: I'm not quite understanding. The only thing this block of code is doing is converting the expected  jr seed and jr part from one option as they're stored in the query share (c.f. Tim's comments) into two options so that we can "send" the jr part between leader and helper as a prep message without the jr seed. The mastic query API does already return the expected seed; but decide is performing the test you describe since the aggregator doesn't have the confirmed jr part from the other aggregator until that point.

cjpatton: Hmm, perhaps I'm misunderstanding how the protocol is supposed to work. I believe it's this (from the perspective of the Leader):

1. Leader computes its verifier share, the Leader joint rand part, and the joint rand seed. It needs to confirm the joint rand seed is correct. 
2. The Helper sends its verifier share and the Helper joint rand part.
3. The Leader checks that joint rand seed is correct using the Leader's and Helper's joint rand parts.
4. The Leader checks the verifier.

Is this correct?

If so, then my point is: at some point `query()` needs to admit the joint rand seed it used to compute the verifier share, correct?

hannahdaviscrypto: That's correct, yes. Step 1 is the Leader's run of query(); it returns a query share containing its verifier share, the joint rand part, and the joint rand seed. These will need to be extracted separately because the first two (vf and jr part) will get sent to the Helper and the last one (jr seed) will get cached and passed to decide() along with the Helper's messages. That's what's happening in the tagged code. Step 2 is the Helper's run of query(), which does the same thing. 

The Step 3 check that the joint rand seed is correct happens in decide(), and it is based on both the Leader and Helper's jr parts. Step 4 also happens in decide(), although I usually do it before step 3 (the order should not matter). 

If you're asking whether query() needs to take the joint rand seed as input, the answer is no, because it derives the jr seed for the first time from the jr part it is given in the input share. If you're asking whether it needs to return the jr seed, it already does. Are you suggesting that query() needs to return the jr seed as part of a tuple instead of a query_share struct? I think that was another Clippy-hates-complex-data-types decision.

src/szk.rs
```diff
@@ -0,0 +1,444 @@
...
+use crate::{
+    codec::{CodecError, Encode},
+    field::{FftFriendlyFieldElement, FieldElement},
+    flp::{FlpError, Type},
+    prng::{Prng, PrngError},
+    vdaf::xof::{IntoFieldVec, Seed, Xof, XofTurboShake128},
+};
+use std::{borrow::Cow, marker::PhantomData};
+
+/// Domain separation tags
```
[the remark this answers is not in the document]
tgeoghegan: While Chris is correct in general, nobody but him calls this a "doc*u*comment". Everyone else on the planet says "doc comment".

https://doc.rust-lang.org/rust-by-example/meta/doc.html#doc-comments

src/szk.rs
```diff
@@ -0,0 +1,444 @@
...
+    algorithm_id: u32,
+    phantom: PhantomData<P>,
+}
+
+impl<T: Type> Szk<T, XofTurboShake128, 16> {
+    /// Create an instance of [`Szk`] using [`XofTurboShake128`].
+    pub fn new_turboshake128(typ: T, algorithm_id: u32) -> Self {
+        Szk::new(typ, algorithm_id)
+    }
+}
```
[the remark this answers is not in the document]
hannahdaviscrypto: I need to specify an XOF for testing.

src/szk.rs
hannahdaviscrypto: This one's a little different because it's deriving both the joint randomness and the joint randomness seed.

src/szk.rs
hannahdaviscrypto: Yes, and they do. the swap happens in lines 223-224 currently.

src/szk.rs
```diff
@@ -0,0 +1,565 @@
...
+            verifier: verifier_share,
+        })
+    }
+
+    /// Returns true if the verifier message indicates that the input from which
+    /// it was generated is valid.
+    fn decide(
+        &self,
+        verifier: &[T::Field],
+        l_jr_part: Option<Seed<SEED_SIZE>>,
```
cjpatton, line 328: nit: In general, it's a good idea to use consistent variable names across function calls. This make it easier for the reader to identify related concepts across the code.

Here and below.
```suggestion
        leader_joint_rand_part: Option<Seed<SEED_SIZE>>,
```

src/szk.rs
```diff
@@ -0,0 +1,565 @@
...
+
+        match (joint_rand_seed_opt, l_jr_part, h_jr_part) {
+            (Some(joint_rand_seed), Some(leader_joint_rand_part), Some(helper_joint_rand_part)) => {
+                if joint_rand_seed
+                    != self.derive_joint_rand_seed(&leader_joint_rand_part, &helper_joint_rand_part)
+                {
+                    Ok(false)
+                } else {
+                    Ok(true)
+                }
```
cjpatton, line 348: A little more readable (less control flow)

```suggestion
                let expected_joint_rand_seed = self.derive_joint_rand_seed(&leader_joint_rand_part, &helper_joint_rand_part);
                Ok(joint_rand_seed == expected_joint_rand_seed)
```

src/szk.rs
```diff
@@ -0,0 +1,565 @@
...
+        joint_rand_seed_opt: Option<Seed<SEED_SIZE>>,
+    ) -> Result<bool, SzkError> {
+        // Check if underlying FLP proof validates
+        let check_flp_proof = self.typ.decide(verifier)?;
+        if !check_flp_proof {
+            return Ok(false);
+        }
+        // Check that joint randomness was properly derived from both
+        // aggregators' parts
+
```
cjpatton, line 339: There shouldn't be a space between a code block and a comment that refers to it.
```suggestion
```

src/szk.rs
```diff
@@ -0,0 +1,565 @@
...
+                if joint_rand_seed
+                    != self.derive_joint_rand_seed(&leader_joint_rand_part, &helper_joint_rand_part)
+                {
+                    Ok(false)
+                } else {
+                    Ok(true)
+                }
+            }
+            (None, None, None) => Ok(true),
+            (_, _, _) => Ok(false),
```
cjpatton, line 351: This should never happen. I would return an `Err()`. I think it may even be appropriate to panic here.
```suggestion
            (_, _, _) => panic!("at least one the input seeds is missing"),
```

src/szk.rs
```diff
@@ -0,0 +1,565 @@
...
+            &verifier,
+            l_jr_part.clone(),
+            h_jr_part.clone(),
+            mutated_jr_seed,
+        ) {
+            assert!(!leader_decision, "Leader validated after proof mutation");
+        };
+    }
+
+    #[test]
```
cjpatton, line 558: Add a test for `Count` as well so that we try with and without joint randomness.

src/szk.rs
```diff
@@ -0,0 +1,565 @@
...
+        let helper_seed = Seed::<16>::generate().unwrap();
+        let leader_seed_opt = if typ.joint_rand_len() > 0 {
+            Some(Seed::<16>::generate().unwrap())
+        } else {
+            None
+        };
+        let helper_input_share: Vec<T::Field> =
+            Prng::from_seed_stream(XofTurboShake128::seed_stream(&helper_seed, &[2u8; 8], &[]))
+                .take(szk_typ.typ.input_len())
+                .collect();
```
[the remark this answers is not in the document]
hannahdaviscrypto: We need to expand the random vector from a specific seed that can be passed to the aggregator, so I don't think we can use random_vector, which samples and then destoys its own seed.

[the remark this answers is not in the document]
divergentdave: That's true in Prio3, and in Mastic the IDPF will do something similar, but in this test the source of the helper input share vector doesn't matter, since it's external to `Szk`. I noticed that this expansion uses `0202020202020202` as a domain separation tag, while, nothing else does.

src/szk.rs
hannahdaviscrypto: There is one, right? line 558

src/szk.rs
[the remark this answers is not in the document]
hannahdaviscrypto: I see, thanks!

src/szk.rs
```diff
@@ -0,0 +1,565 @@
...
+        helper_input_share: &[T::Field],
+        encoded_measurement: &[T::Field],
+        rand_seeds: [Seed<SEED_SIZE>; 2],
+        leader_seed_opt: Option<Seed<SEED_SIZE>>,
+        nonce: &[u8; 16],
+    ) -> Result<[SzkProofShare<T::Field, SEED_SIZE>; 2], SzkError> {
+        let [prove_rand_seed, helper_seed] = rand_seeds;
+        let leader_seed_opt_copy = leader_seed_opt.clone();
+        let (leader_blind_and_joint_rand_part, helper_joint_rand_part, joint_rand) =
+            if let Some(leader_seed) = leader_seed_opt_copy {
```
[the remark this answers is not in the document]
cjpatton: You're right!

src/szk.rs
```diff
@@ -0,0 +1,574 @@
...
+    ) -> Result<[SzkProofShare<T::Field, SEED_SIZE>; 2], SzkError> {
+        let [prove_rand_seed, helper_seed] = rand_seeds;
+        let leader_seed_opt_copy = leader_seed_opt.clone();
+        // If joint randomness is used, derive it from the two input shares,
+        // the seeds used to blind the derivation, and the nonce. Pass the
+        // leader its blinding seed and the helper's joint randomness part, and
+        // pass the helper the leader's joint randomness part. (The seed used to
+        // derive the helper's proof share is reused as the helper's blind)
+        let (leader_blind_and_helper_joint_rand_part, leader_joint_rand_part, joint_rand) =
+            if let Some(leader_seed) = leader_seed_opt_copy {
```
cjpatton, line 222: Less variables makes the code easier to read.
```suggestion
        // If joint randomness is used, derive it from the two input shares,
        // the seeds used to blind the derivation, and the nonce. Pass the
        // leader its blinding seed and the helper's joint randomness part, and
        // pass the helper the leader's joint randomness part. (The seed used to
        // derive the helper's proof share is reused as the helper's blind)
        let (leader_blind_and_helper_joint_rand_part, leader_joint_rand_part, joint_rand) =
            if let Some(leader_seed) = leader_seed_opt.clone() {
```

src/szk.rs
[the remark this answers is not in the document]
cjpatton: Bump.

src/szk.rs
```diff
@@ -0,0 +1,469 @@
+//! Szk is a wrapper class for the FLP proof system (accessible through the `Type` trait)
```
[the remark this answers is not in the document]
cjpatton: I think that's a good idea.

src/szk.rs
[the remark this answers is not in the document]
cjpatton: I suggest we keep it as is. Let's not implement the generality until we actually use it.

src/szk.rs
cjpatton: I think we should return an error here.

src/szk.rs
```diff
@@ -0,0 +1,574 @@
...
+}
+
+/// A tuple containing the state and messages produced by an SZK query.
+#[derive(Clone)]
+pub(crate) struct SzkQueryShare<F, const SEED_SIZE: usize> {
+    joint_rand_seed_and_part: Option<[Seed<SEED_SIZE>; 2]>,
+    verifier: SzkVerifier<F>,
+}
+
+/// Verifier type for the SZK Proof
```
cjpatton, line 74:
```suggestion
/// Verifier type for the SZK proof.
```

src/szk.rs
cjpatton: Ah, I see. If you want a distinguished name, then I would go with something more verbose, like derive_joint_rand_and_seed() or something.

src/szk.rs
cjpatton: Yup, you're right! Sorry :/

src/szk.rs
```diff
@@ -0,0 +1,565 @@
...
+                ..
+            } => Cow::Borrowed(uncompressed_proof_share),
+            SzkProofShare::Helper {
+                ref proof_share_seed_and_blind,
+                ..
+            } => Cow::Owned(self.derive_helper_proof_share(proof_share_seed_and_blind)),
+        };
+
+        let (joint_rand, jr_seed_and_part) = if self.typ.joint_rand_len() > 0 {
+            let (seed, peer_joint_rand_part) = match proof_share {
```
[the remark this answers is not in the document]
cjpatton: Bump.

src/szk.rs
```diff
@@ -0,0 +1,574 @@
...
+        leader_seed_opt: Option<Seed<SEED_SIZE>>,
+        nonce: &[u8; 16],
+    ) -> Result<[SzkProofShare<T::Field, SEED_SIZE>; 2], SzkError> {
+        let [prove_rand_seed, helper_seed] = rand_seeds;
+        let leader_seed_opt_copy = leader_seed_opt.clone();
+        // If joint randomness is used, derive it from the two input shares,
+        // the seeds used to blind the derivation, and the nonce. Pass the
+        // leader its blinding seed and the helper's joint randomness part, and
+        // pass the helper the leader's joint randomness part. (The seed used to
+        // derive the helper's proof share is reused as the helper's blind)
```
cjpatton, line 220:
```suggestion
        // derive the helper's proof share is reused as the helper's blind.)
```

src/szk.rs
```diff
@@ -0,0 +1,574 @@
...
+        let (leader_blind_and_helper_joint_rand_part, leader_joint_rand_part, joint_rand) =
+            if let Some(leader_seed) = leader_seed_opt_copy {
+                let leader_joint_rand_part =
+                    self.derive_joint_rand_part(&leader_seed, leader_input_share, nonce)?;
+                let helper_joint_rand_part =
+                    self.derive_joint_rand_part(&helper_seed, helper_input_share, nonce)?;
+                let (_jrs, jr) =
```
cjpatton, line 227:
```suggestion
                let (_joint_rand_seed, joint_rand) =
```

src/szk.rs
```diff
@@ -0,0 +1,574 @@
...
+                    self.derive_joint_rand(&leader_joint_rand_part, &helper_joint_rand_part);
+                (
+                    Some((leader_seed, helper_joint_rand_part)),
+                    Some(leader_joint_rand_part),
+                    jr,
+                )
+            } else {
+                (None, None, Vec::<T::Field>::new())
```
cjpatton, line 235: style: Let the compiler infer the type here.
```suggestion
                (None, None, Vec::new())
```

src/szk.rs
```diff
@@ -0,0 +1,574 @@
...
+            *x -= y;
+        }
+
+        // Construct the output messages.
+        let leader_proof_share = SzkProofShare::Leader {
+            uncompressed_proof_share: leader_proof_share,
+            leader_blind_and_helper_joint_rand_part,
+        };
+        let helper_proof_share = SzkProofShare::Helper {
+            proof_share_seed_and_blind: helper_seed.clone(),
```
cjpatton, line 257: Unnecessary clone

```suggestion
            proof_share_seed_and_blind: helper_seed,
```

src/szk.rs
```diff
@@ -0,0 +1,574 @@
...
+            } => Cow::Owned(self.derive_helper_proof_share(proof_share_seed_and_blind)),
+        };
+
+        let (joint_rand, joint_rand_seed_and_part) = if self.typ.joint_rand_len() > 0 {
+            let (seed, peer_joint_rand_part) = match proof_share {
+                SzkProofShare::Leader {
+                    uncompressed_proof_share: _,
+                    leader_blind_and_helper_joint_rand_part,
+                } => leader_blind_and_helper_joint_rand_part
+                    .expect("jr_len > 0 so there is joint_randomness"),
```
cjpatton, line 294: "jr_len" is not a variable name. A more helpful panic message would be something like this: "leader_blind_and_helper_joint_rand_part should be set"

Likewise below.

src/szk.rs
```diff
@@ -0,0 +1,574 @@
...
+        };
+        let verifier_share = self.typ.query(
+            input_share,
+            flp_proof_share.as_ref(),
+            &query_rand,
+            &joint_rand,
+            2,
+        )?;
+        Ok(SzkQueryShare {
+            joint_rand_seed_and_part,
```
cjpatton, line 323: Re: https://github.com/divviup/libprio-rs/pull/1013#discussion_r1602074882

I think this is where @divergentdave and I are confused. I think both of us are reading this code as if `SzkQueryShare` is a message that is transmitted to the peer. Is this true?

If so, do we actually need to send BOTH the joint rand seed AND our joint rand part to the peer? The way Prio3 works is that each aggregator stores the joint rand seed and broadcasts its joint rand part only. That way each aggregator can re-compute the joint rand seed from the broadcast parts and make sure it matches the stored value. The upside is that we avoid transmitting more bytes (`SEED_SIZE`) than we need to.

If you don't intend for joint rand seed to be transmitted, then it's a good idea to separate the bits that are stored and the bits that are transmitted at the API level. Concretely, you might modify the return type of this function to return the query share and the joint rand seed to be stored.

src/szk.rs
```diff
@@ -0,0 +1,565 @@
...
+
+    /// Returns true if the verifier message indicates that the input from which
+    /// it was generated is valid.
+    fn decide(
+        &self,
+        verifier: &[T::Field],
+        l_jr_part: Option<Seed<SEED_SIZE>>,
+        h_jr_part: Option<Seed<SEED_SIZE>>,
+        joint_rand_seed_opt: Option<Seed<SEED_SIZE>>,
+    ) -> Result<bool, SzkError> {
```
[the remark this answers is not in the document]
hannahdaviscrypto: Adjusted re: comments below, the API now returns one SzkQueryShare to be sent to the partner and one SzkQueryState for caching.

src/szk.rs
hannahdaviscrypto: Resolved. SzkQueryShare was not originally be meant to sent to the peer; only the verifier share and joint rand part should be set. There are now two structs: SzkQueryShare for sending and SzkQueryState for storing.

cjpatton (approved):
👏 LGTM. Before merging:
1. Let's make sure @divergentdave is happy
2. We need to squash this into a single commit.

src/flp/szk.rs
```diff
@@ -0,0 +1,600 @@
...
+            leader_joint_rand_part_opt,
+            helper_joint_rand_part_opt,
+        ) {
+            (Some(joint_rand_seed), Some(leader_joint_rand_part), Some(helper_joint_rand_part)) => {
+                let expected_joint_rand_seed =
+                    self.derive_joint_rand_seed(&leader_joint_rand_part, &helper_joint_rand_part);
+                Ok(joint_rand_seed == expected_joint_rand_seed)
+            }
+            (None, None, None) => Ok(true),
+            (_, _, _) => panic!("at least one of the input seeds is missing"),
```
divergentdave, line 393: I think this should return an error as well, instead of panicking.

divergentdave (approved):
Looking good, I have a few small comments.

hannahdaviscrypto:
Squashed and ready to merge, pending CI checks!

PR #1017, opened 2024-04-30
prio3: Use `Cow` to hold expanded shares

cjpatton:
Idea from #1013.

All we need is a reference, so borrow the Leader's share and own the Helper's expanded share.


divergentdave (approved):
Nice cleanup!

tgeoghegan (approved):

PR #1058, opened 2024-05-17
Downgrade getrandom to fix transitive libc dependency

mendess:
getrandom 0.2.15 depends on libc 0.2.154 which has been yanked:
https://crates.io/crates/libc/versions

I recommend also yanking prio 0.16.4 after releasing this version

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -731,7 +731,7 @@ checksum = "eb9f9e6e233e5c4a35559a617bf40a4ec447db2e84c20b55a6f83167b7e57872"
 
 [[package]]
 name = "prio"
-version = "0.16.4"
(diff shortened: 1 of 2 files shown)


divergentdave (approved):
Looks good, though I don't think a transitive yank is necessary. prio 0.16.4 will eventually work once a new libc is released, and there's nothing wrong with the crate itself.

cjpatton (approved):
@divergentdave is there any chance you can cut today? Otherwise we'll have to downgrade to 0.16.3 for our next internal daphne release, which we'd prefer not to do.

divergentdave:
Sure, I'll cut it once #1059 is landed

Issue #209, opened 2022-04-26
compiling libprio-rs to wasm fails to run


siyengar:
Compiling libprio to wasm fails to run. On further investigation it appears that the main reason for this is that there are symbols from ring's key agreement libraries that do not support the wasm target and explicitly disable it https://github.com/briansmith/ring/blob/155231fb017acaaa94a044f124bb34a777d115ef/src/lib.rs#L82

I experimented with a few solutions to this and it looks like one of the easiest solutions to this is to create a clone of the top level method that performs encryption of shares that does not perform encryption of shares. It might then be possible to use a pure js library to actually encrypt those shares in a format compatible with prio. Happy to put up a PR for this.

cjpatton:
Innnteresting. I have been compiling to wasm for a bit now without much trouble, though I haven't tried to run the "Prio v2" stuff. Can you provide a bit more detail on your target and configuration? Is it failing to compile or failing at runtime?

siyengar:
Have you tried initializing the wasm lib from js, or are you simply compiling? Wasm-pack will almost always succeed, but then error during initialization because there are unresolved symbols. 

Let me put up a quick repro rust project to show what I'm talking about. It's possible that you're using different top level methods than I am which could cause things to work for you. Wasm-pack is really good at eliding methods that are unused.

divergentdave:
We dropped our dependency on ring a while ago, and #1059 made wasm-bindgen use cases a bit easier. Compiling an executable for the wasm32-wasip1 target succeeds, and it only has a small number of imports from "wasi_snapshot_preview1" so I'm confident there aren't any more symbol issues. I'm thus going to close this issue.

Issue #955, opened 2024-02-23
Do we still need `getrandom`?


cjpatton:
I recall I initially pushed for using `getrandom` in lieu of `rand` in order to have compatibility with WASM. First, I'm not sure this was ever technically necessary. Second, we no longer need WASM for Daphne:
https://github.com/cloudflare/daphne/blob/b59507a0092e1267e7ec7c3f99c33b59bd1deb34/daphne_worker_test/Cargo.toml#L31-L34

What do you think about removing the dependency?
cc/ @mendess

divergentdave:
`rand` and `rand_core` rely on the `getrandom` crate with default features enabled, and that backs the `OsRng` implementation, so I think it's six of one or half a dozen of the other.

cjpatton:
Right, but can we avoid using it directly here? I.e., could we replace `getrandom()` with `thread_rng().fill()`?

tgeoghegan:
Dropping a direct dependency would be nice, but it sounds like this would have no other functional effect. On the other hand, we have a couple error enums with variants:

```rust
 /// Failure when calling getrandom().
 #[error("getrandom: {0}")]
 GetRandom(#[from] getrandom::Error)
```

That means that dropping `getrandom` is a public API change. Not one I'm opposed to per se, but we'd have to wait for `prio` 0.17.

cjpatton:
I think it's worth doing, and yeah this could wait until 0.17.

divergentdave:
Note that dropping our direct dependency on getrandom would now require removing our `wasm-compat` feature added in #1059. The [README for the `rand` crate](https://github.com/rust-random/rand/blob/master/README.md#wasm-support) says that, for wasm32-unknown-unknown in a JavaScript environment, you have to turn on the `getrandom/js` feature separately from your `rand` dependency.

cjpatton:
Looks like we're good to drop the "wasm-compat" feature in daphne: https://github.com/cloudflare/daphne/pull/613

PR #1059, opened 2024-05-17
Add a wasm-compat feature for getrandom

mendess:
If you're using `prio` from wasm you have to include `getrandom = { version "*", feature = ["js"] }` in your cargo toml, even if you don't make use of `getrandom` in your code, just so you can activate the "js" feature. 

I think it would be nicer if prio let me activate the feature without this hack.

Stacked on top of #1058

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -403,8 +403,10 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c4567c8db10ae91089c99af84c68c38da3ec2f087c3f82960bcdbf3656b6f4d7"
 dependencies = [
  "cfg-if",
+ "js-sys",
  "libc",
  "wasi",
+ "wasm-bindgen",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


divergentdave (approved):
LGTM. Closes #209, see also #955.

Separately, are you able to build for a WASI target instead of wasm32-unknown-unknown? That would remove the need for wasm-bindgen glue.

divergentdave:
We should document this new feature in the README:

```markdown
| `wasm-compat` | No | Enables the `getrandom/js` feature. This is necessary for `wasm32-unknown-unknown` targets, when in a JavaScript environment. | ✅ |
```

cjpatton:
> wasm32-unknown-unknown?

This target is required for the workers runtime, at least for now. There is experimental WASI support, but we haven't had a chance to test it thoroughly.

mendess:
> We should document this new feature in the README:
> 
> ```
> | `wasm-compat` | No | Enables the `getrandom/js` feature. This is necessary for `wasm32-unknown-unknown` targets, when in a JavaScript environment. | ✅ |
> ```

Docs added

divergentdave (approved):
Thanks! I'm going to push another commit updating `supply-chain/imports.lock`, to pull in one audit with a stronger criteria, and this'll be good to go.

Issue #694, opened 2023-08-22
Support exactly adding differential privacy to functions with square root irrational sensitivity


ooovi:
As discussed in [these PR comments](https://github.com/divviup/libprio-rs/pull/578#discussion_r1289772553), it may be desirable to allow adding differential privacy to functions with a sensitivity that is an irrational square root of a rational number.

This will require [computing the floor of the square root of a rational number exactly](https://github.com/IBM/discrete-gaussian-differential-privacy/blob/cb190d2a990a78eff6e21159203bc888e095f01b/discretegauss.py#L101), as well as providing [a way to create a distribution for a `DifferentialPrivacyStrategy`](https://github.com/dpsa4fl/libprio-rs/blob/dd50c944d153632028fd9f267f6d9ac42c26a817/src/dp.rs#L119-L122C3) from the square of the sensitivity instead of the sensitivity itself.

This can be worked around by rounding the sensitivity up, which will retain the privacy guarantee. We might want to analyse the loss of accuracy depending on the magnitude of the sensitivity and rational number precision.

PR #1072, opened 2024-06-05
Add support for DP noise during aggregation for Prio3SumVec, Prio3Histogram

### LINKED ISSUE #1068: Implement Differential Privacy for Prio3SumVec
We are currently using Prio3SumVec to submit reports to DAP and we are looking to introduce a small amount of noise to the results using differential privacy. We are flexible in how the noise is added, but as a starting point we think approximate/epsilon-delta DP, using I1-sensitivity, and the discrete Laplacian would be enough for our experiment.

Ideally this would be backwards compatible and we would not need to update our libprio-rs integration when submitting reports, and only need to pass in new parameters when calling the collector.

We are only looking to add a small amount of noise to the aggregated result counts  (i.e. +/- 1 or 2). Any implementation that provides this should suffice for our current use case.

divergentdave:
This PR implements support for differential privacy via aggregator randomization for two more Prio3 VDAFs, Prio3SumVec and Prio3Histogram. I tacked on Prio3Histogram because it was easy to do so, and it has nice utility behavior thanks to the mutual exclusion property enforced by the circuit. These implementations combine a pure DP budget and the discrete Laplace distribution. We already have a routine to sample the discrete Laplace, as it is a subroutine in our discrete Gaussian sampler, so this PR additionally packages it up in a new public type.

I still want to add some tests to smoke test the `AggregatorWithNoise` interface for these, and perhaps a distributional test on the noised results. I could probably extract the common noise-adding loop here and in `fixedpoint_l2`, in the spirit of DRY.

This resolves #1068.


divergentdave:
FWIW, draft-wang-ppm-differential-privacy provides a combination of Prio3Histogram with discrete Gaussian noise to achieve approximate differential privacy. We could also implement that, in parallel with the strategy provided here, but it would be blocked on first addressing #694, which would not be too hard. (it needs a floored square root implemented via binary search and a new constructor for the discrete Gaussian distribution)

src/flp/types/dp.rs
```diff
@@ -0,0 +1,179 @@
...
+// impose this bound on `FieldElementWithInteger`, which would allow simple blanket implementations
+// here.
+impl<S> TypeWithNoise<PureDpDiscreteLaplace> for SumVec<Field64, S>
+where
+    S: ParallelSumGadget<Field64, Mul<Field64>> + Eq + 'static,
+{
+    fn add_noise_to_result(
+        &self,
+        dp_strategy: &PureDpDiscreteLaplace,
+        agg_result: &mut [Self::Field],
```
junyechen1996, line 24: I agree with this interface, but `add_noise_to_agg_share` seems more appropriate as the function name. However, my understanding of the original issue (https://github.com/divviup/libprio-rs/issues/1068) is to add noise to the final aggregate result at collector stage, so this argument is more like `agg_result: &mut [Self::Field::Integer]`? Obviously that doesn't hold up the DAP/VDAF trust model, because ideally we want to add noise at aggregator stage (at least independently to start with).

src/dp/distributions.rs
```diff
@@ -287,6 +287,73 @@ impl DifferentialPrivacyStrategy for DiscreteGaussianDpStrategy<ZCdpBudget> {
...
+    type Distribution = DiscreteLaplace;
+    type Sensitivity = Ratio<BigUint>;
+
+    fn from_budget(budget: Self::Budget) -> Self {
+        DiscreteLaplaceDpStrategy { budget }
+    }
+
+    /// Create a new sampler for the discrete Laplace distribution with a scale parameter calibrated
+    /// to provide `epsilon`-differential privacy when added to the result of an integer-valued
+    /// function with l1-sensitivity `sensitivity`, following Lemma 29 from [[CKS20]]
```
junyechen1996, line 346: I noticed in Lemma 29, the query function outputs an integer $q: X^n \to \mathbb{Z}$. Since we are adding discrete Laplace independently on each coordinate, does this calculation still apply?

CKS20 does provide a multivariate version for discrete Gaussian in Theorem 14, but not for discrete Laplace

src/flp/types/dp.rs
```diff
@@ -0,0 +1,179 @@
...
+
+        // Generate noise for each vector coordinate and apply it.
+        let modulus = BigInt::from(F::modulus());
+        for entry in agg_result.iter_mut() {
+            // Generate noise.
+            let noise = sampler.sample(rng);
+
+            // Project it into the field by taking the modulus, converting to a fixed-precision
+            // integer, and converting to a field element.
+            let noise_wrapped = noise.mod_floor(&modulus);
```
junyechen1996, line 88: Question: does this function throw an exception if for example, `noise` is more than `modulus`?

Another consideration is: because `noise` could be negative, we may need to reserve some tail end of the field modulus for negative values in aggregate result, so there may need to be a reasonable upper bound on the magnitude of `noise`, perhaps `|noise| < modulus/2`?

src/dp/distributions.rs
divergentdave: Hmm, good point. Proposition 1 in [Calibrating Noise to Sensitivity in Private Data Analysis](https://people.csail.mit.edu/asmith/PS/sensitivity-tcc-final.pdf) shows the same thing, but with the continuous Laplace distribution. I think the mechanism is correct, but this will need a better reference.

src/flp/types/dp.rs
divergentdave: No, this succeeds unconditionally, and would wrap around on very large noise values. The rationale, (see https://github.com/divviup/libprio-rs/pull/578#discussion_r1226543791 and https://github.com/divviup/libprio-rs/pull/578#discussion_r1263075241) is that adding wrapped noise in the field, then unsharding to produce an integer, is equivalent to a trusted curator adding noise to the true sum using arbitrary-precision integers, and then taking the modulus as a post-processing step. (and post-processing preserves DP guarantees) I should bake this reasoning into a comment when I extract this loop into a helper function.

The `mod_floor()` operation does wrap negative noise values around to positive values below `modulus`; we're fine with this overlapping with the extreme upper tail of the noise distribution.

junyechen1996: I see, makes sense. Out of curiosity, how do you plan to interpret the aggregate result with noise in this case? Given a reasonably large field, and assuming number of clients cannot exceed the field modulus, how much of the tail end of the field modulus do you plan to reserve for noise?

src/dp/distributions.rs
```diff
@@ -287,6 +287,73 @@ impl DifferentialPrivacyStrategy for DiscreteGaussianDpStrategy<ZCdpBudget> {
...
+    /// Create a new sampler for the discrete Laplace distribution with the given scale parameter.
+    /// Returns an error if the scale parameter is zero or if it has a denominator of zero.
+    pub fn new(scale: Ratio<BigUint>) -> Result<Self, DpError> {
+        if scale.denom().is_zero() {
+            return Err(DpError::ZeroDenominator);
+        }
+        if scale.numer().is_zero() {
+            return Err(DpError::InvalidParameter(
+                "the scale of the discrete Laplace distribution must be nonzero".into(),
+            ));
```
tgeoghegan, line 310: It's sort of strange to me that a bad denominator gets its own special enum variant, but a bad numerator is a generic "invalid parameter". Do we expect the caller to do anything particular in the presence of either error? If not, then just returning `DpError::InvalidParameter` in both cases seems fine, and maybe we could remove `::ZeroDenominator` the next time we break public API.

src/flp/types/dp.rs
divergentdave: I hadn't thought about it too much. I suppose the optimal choice will depend on the batch size (and the VDAF's type and parameters).

src/dp/distributions.rs
divergentdave: Opened #1074 to track this

src/dp/distributions.rs
divergentdave: I looked around for a good cite for applying the discrete Laplace mechanism/geometric mechanism to a query that produces $\mathbb{Z}^n$, but came up short. I added an intuitive argument here, instead.

src/dp.rs
```diff
@@ -102,6 +106,21 @@ impl ZCdpBudget {
 
 impl DifferentialPrivacyBudget for ZCdpBudget {}
 
+/// Pure differential privacy budget. (&epsilon;-DP or (&epsilon;, 0)-DP)
```
martinthomson, line 109: What you have right now isn't pure, though maybe it is close enough as to be statistically indistinguishable.

The conversion of the output (which is a signed bignum!) to the field includes a modulus, which means that the infinitely long tails on the distribution are wrapped continuously.  In effect, you are creating a new truncated distribution that is close to a discrete Laplace.

As a result of the large size of the field, $\delta$ is probably close enough to zero that it won't matter.  That said, if you want certainty about your $\delta$, maybe a better approach is to truncate the distribution so that you can make a firmer commitment regarding $\delta$ (there are a number of papers looking at truncated distributions that you could draw from).

src/flp/types/dp.rs
```diff
@@ -0,0 +1,327 @@
...
+        &self,
+        dp_strategy: &PureDpDiscreteLaplace,
+        agg_result: &mut [F],
+        rng: &mut R,
+    ) -> Result<(), FlpError>
+    where
+        R: Rng,
+    {
+        // Compute the l1-sensitivity of the aggregation function (assuming the substitution-DP
+        // model). The worst case is when one individual's measurement changes such that each vector
```
martinthomson, line 69: I'm not sure that substitution-DP is relevant to this case.  Your goal is to measure the contribution of a single submission.  A submission could be $[0..2^b)$ on all dimensions equally, which leads to this outcome.  But that does not imply substitution.  If you have substitution, then you could have $\\{A=0 , B=2^b-1\\}$ or $\\{A=2^b-1 , B=0\\}$.  That means sensitivity under substitution effectively doubles, but I don't see how substitution applies to DAP.

src/flp/types/dp.rs
```diff
@@ -0,0 +1,327 @@
...
+        // element flips from 0 to 2^bits - 1, or vice versa. Then, the l1 distance from the initial
+        // query result to the new query result will be (2^bits - 1) * length.
+        let two = BigUint::from(2u64);
+        let bits = BigUint::from(self.bits);
+        let length = BigUint::from(self.len);
+        let sensitivity = (Pow::pow(two, &bits) - BigUint::one()) * length;
```
martinthomson, line 75: Isn't this going to generate too much noise?  The sensitivity for each dimension is $2^b-1$, which requires IID noise for each dimension proportional to $2^b-1$, not $\ell\cdot2^b-1$.  Unless I'm misreading your IID function, that is.

src/flp/types/dp.rs
```diff
@@ -0,0 +1,327 @@
...
+        dp_strategy: &PureDpDiscreteLaplace,
+        agg_result: &mut [F],
+        rng: &mut R,
+    ) -> Result<(), FlpError>
+    where
+        R: Rng,
+    {
+        // The l1-sensitivity of the aggregation function is two, assuming the substitution-DP
+        // model. Substituting a measurement may, at worst, cause one cell of the query result
+        // to be incremented by one, and another to be decremented by one.
```
martinthomson, line 138: Again, I'm not following why $\Delta_1 = 2$.  DAP doesn't operate on substitution.

src/flp/types/dp.rs
martinthomson: Separately, do you expect to support self.bits > 127?  Such a vector would be very expensive, but it would still be possible to do `let sensitivity = BigUint::from((1u128 << self.bits) - 1)`  or (clever) `let sensitivity = BigUint::from(u128::max >> 128.checked_sub(self.bits).ok_or(Error...)?)`.

martinthomson (commented):
I have questions.

src/dp.rs
divergentdave: I think this still achieves pure DP, due to the reasoning in the comment at the top of `add_iid_noise_to_field_vec()`. I can take a crack at writing a more traditional proof for this mechanism. I expect it should go through okay, because the PDF of the wrapped discrete Laplace distribution is equal to an infinite sum of PDFs of discrete Laplace distributions with offset means. Thus, it should be possible to pair off terms and show they are bounded above correctly.

One key difference between this mechanism and a mechanism using a traditional truncated noise distribution is that the support of the query output is always the same (zero to p-1) and doesn't shift when the true value of the function shifts.

src/flp/types/dp.rs
divergentdave: I relied on the reasoning in https://wangshan.github.io/draft-wang-ppm-differential-privacy/draft-wang-ppm-differential-privacy.html#appendix-B.2 for the choice of substitution-DP. The collector also learns the number of reports (which is needed for unsharding in some VDAFs, or for RAPPOR-style debiasing) and thus we can only get indistinguishable protocol runs if we assume substitution.

src/flp/types/dp.rs
divergentdave: Yes, this would add noise computed based on a sensitivity of $\ell \cdot (2^b-1)$ to each dimension. I think this is correct as far as I can see, and I think this is the main reason why other VDAFs, when composed with DP, benefit from tighter constraints on measurements. (i.e. Prio3Histogram here, Prio3FixedPointBoundedL2VecSum previously, or PINE)

If we split up one $\epsilon$-DP Prio3SumVec query into $\ell$ many Prio3Sum queries, we would have a pure DP budget of $\epsilon/\ell$ for each. The sensitivity of Prio3Sum is $2^b-1$. Thus, we'd add noise drawn from $Lap_\mathbb{Z}((2^b-1)/(\epsilon/\ell))=Lap_\mathbb{Z}(\ell\cdot(2^b-1)/\epsilon)$, which is the same as we'd get from this code as-is.

Another way I thought about this: if the adversary happens to know a priori that each dimension of some client's measurement is perfectly correlated or anti-correlated to the rest, then we'd need that factor of $\ell$ to avoid effectively giving them free queries beyond our budget.

divergentdave: We don't need to support self.bits > 127, because the output share would be able to wrap around Field128, and we haven't defined any bigger FFT-friendly fields as of yet, so that works.

src/dp.rs
martinthomson: I'm not convinced about that.  Consider a small prime field where $p$ is something small, like 31.  In that field, your value might be 10, with a discrete Laplace distribution that includes values below -10 or above 21.  Those will wrap, such that -10 contributes the same value as +21 (for any true value).

I've not seen an argument that supports the view that the probability density function over a limited domain like that is the same as an unbounded discrete Laplace.  Maybe it's OK.  Maybe it's close enough to OK that it doesn't matter when $p$ is very large, like it will be here.  But "pure" means perfect and that's a strong claim to make.

src/flp/types/dp.rs
martinthomson: I don't agree with that reasoning.  It will depend on the setting, but you have to assume that each contribution is unknown to the collector.  A histogram contribution could be a 0 just as much as it could be a 1.  That's the contribution that we're hiding, not that a given person is participating or not.  Substitution assumes that each report contributes something, which is not something we rely on.

My suggestion would be to externalize whether you want to use substitution as the basis of your sensitivity analysis.  That is, if someone believes that substitution is necessary in their setting, they can set epsilon accordingly (i.e., halve it).  Otherwise, settings where it is directly detrimental (i.e., the case we currently care about) will have to adjust in the other direction.

src/flp/types/dp.rs
```diff
@@ -0,0 +1,331 @@
...
+        R: Rng,
+    {
+        // Compute the l1-sensitivity of the aggregation function (assuming the substitution-DP
+        // model). The worst case is when one individual's measurement changes such that each vector
+        // element flips from 0 to 2^bits - 1, or vice versa. Then, the l1 distance from the initial
+        // query result to the new query result will be (2^bits - 1) * length.
+        let length = BigUint::from(self.len);
+        let sensitivity = BigUint::from(
+            1u128
+                .checked_shl(self.bits as u32)
```
martinthomson, line 74: Just so you know, because I've been stung by this one too....

`checked_shl` does not protect against the shifted value overflowing, it only checks that the shift operand (`self.bits` here) is in the right range.  That means that, for this example, it will fail only if `self.bits >= 128`, which is fine when the value is 1, but has a surprising effect if the value is larger.
`u128::MAX.checked_shr(u128::BITS - self.bits.try_into().ok_or(FlpError::...)?).ok_or(FlpError::...))` would be safer.  Or you could just use `>>`, because `checked_shr` is a trap.

src/flp/types/dp.rs
martinthomson: Just acknowledging that factor of $\ell$.  I've spent so long thinking about histograms, I'd forgotten how to think about vectors.  That's going to be a lot of noise, but ¯\\\_(ツ)\_/¯

src/dp.rs
kunal-talwar: This is a bit subtle, so let me try to clarify.

Differential Privacy satisfies the Post-Processing property  (or Data Processing Inequality). This means that if a mechanism is DP, you can apply any function to the output and that is still DP. Thus if the mechanism that adds Laplace noise over integers is DP, we can apply the transformation $T_q$ that maps an integer z to z mod q, that still preserves the privacy properties. This argument applies to computing the noisy noisy count over integers, and then applying the transform $T_q$, and so we are ok as long as we sample from Laplace over integers and apply $T_q$ to the noisy count. Morally, the reason this works is that the probability of any output $y$ is now the sum of probabilities of outputs $y, y+q, y-q, y+2q, y-2q,....$. Since each of those probabilities changes by at most a multiplicative $e^{\epsilon}$ between neighboring databases, their sum also can only change by a multiplicative $e^{\epsilon}$.

Note that this is different from truncating the noise itself, which would indeed need some delta adjustment, and your intuition above is indeed correct for that variant.

divergentdave: I just pushed a proof that the mechanism satisfies pure differential privacy, starting from the formulas. Here's a [rendered PDF](https://github.com/user-attachments/files/16072470/Pure.DP.Mechanism.pdf) for convenience; I authored this in LyX, and put the source file in the documentation directory.

src/flp/types/dp.rs
divergentdave: In the case of `Prio3Histogram`, the validity circuit enforces that every contribution adds 1 to one of the buckets, so there is no possible 'zero measurement' to exchange with. Using `Prio3MultihotCountVec` with `max_weight = 1` instead would permit contributions to add nothing.

We could allow a choice between substitution and "replacement with zero" (as a modification of deletion-DP) by rolling that choice into the `DifferentialPrivacyStrategy` trait as well. Then, the different strategy-specific implementations of `AggregatorWithNoise`/`TypeWithNoise` could use their own sensitivity formulas as appropriate.

junyechen1996: Agree with this reasoning for the choice of substitution-DP:

> The collector also learns the number of reports (which is needed for unsharding in some VDAFs, or for RAPPOR-style debiasing) and thus we can only get indistinguishable protocol runs if we assume substitution.

For unsharding in VDAF, `num_measurements` is an argument in the public API, and in DAP, `report_count` is in [`Collection`](https://www.ietf.org/archive/id/draft-ietf-ppm-dap-11.html#section-4.6.1-17), so the number of reports in the batch is public information.

> My suggestion would be to externalize whether you want to use substitution as the basis of your sensitivity analysis. That is, if someone believes that substitution is necessary in their setting, they can set epsilon accordingly (i.e., halve it).

If the code assumes substitution-DP in its sensitivity analysis, and if someone wants to use deletion-DP, they can double the target central epsilon?

src/dp.rs
martinthomson: Thanks both, the intuition and proof both helped, though I did not go through in detail, the inequality is the key piece I was looking for.

Question: is that the last thing holding the merge here?  I was not expecting to be the one holding this up.  We are very interested in having this land.

divergentdave: I think this and the neighboring dataset definition thread are sufficiently resolved, but I'd like to give @cjpatton a chance to review before I merge.

divergentdave:
I derived some additional traits on `DiscreteLaplaceDpStrategy` (matching the existing concentrated DP discrete Gaussian strategy) and added input validation to `PureDpBudget` based on experience using this over in divviup/janus#3302. I added mention of the sensitivity definition, including the concept of neighboring datasets, to the documentation of both DP strategies. I also rebased to fix an easy merge conflict.

src/flp/types/dp.rs
tgeoghegan: Should we take this suggestion?

src/flp/types/dp.rs
tgeoghegan: Could you add a TODO for the issue filed for the rename?

src/dp/distributions.rs
```diff
@@ -287,6 +289,89 @@ impl DifferentialPrivacyStrategy for DiscreteGaussianDpStrategy<ZCdpBudget> {
...
+where
+    B: DifferentialPrivacyBudget,
+{
+    budget: B,
+}
+
+/// A DP strategy using the discrete Laplace distribution, providing pure DP.
+///
+/// This uses l1-sensitivity, with the substitution definition of neighboring datasets.
+pub type PureDpDiscreteLaplace = DiscreteLaplaceDpStrategy<PureDpBudget>;
```
tgeoghegan, line 338: When naming this stuff, we should make sure to leave room for new DP schemes that may arrive soon, based on the [drafts mt and the Bens are working on](https://mailarchive.ietf.org/arch/msg/ppm/FvAUll-XRHhGvVh3VMyhY-GqgPE/).

Is this the only DP strategy we'd ever implement that is pure and samples the discrete Laplace distribution? IIUC (but I could easily be wrong), this strategy is where each aggregator independently adds sufficient noise to protect against either aggregator defecting, but the incoming schemes use MPC to meet privacy goals with less noise. So should this be `PureDpDiscreteLaplaceButWithMaybeTwiceAsMuchNoiseAsYouNeedSorryAboutThat`? (deliberately terrible name to force a better choice)

tgeoghegan (commented):
Reviewing on the basis of code quality and maintainability rather than on correctness of the DP stuff, which is well covered by other reviewers (thanks!).

documentation/Pure DP Mechanism.lyx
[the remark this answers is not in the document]
divergentdave: There are some instructions to do so using pandoc here: https://wiki.lyx.org/Tips/ConvertMarkdown. I'm going to try that out and see how GitHub handles the results.

src/dp.rs
```diff
@@ -95,13 +99,57 @@ impl ZCdpBudget {
     /// for a `rho`-ZCDP budget.
     ///
     /// [CKS20]: https://arxiv.org/pdf/2004.00010.pdf
+    // TODO(#1095): This should be fallible, and it should return an error if epsilon is zero.
```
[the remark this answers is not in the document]
divergentdave: It currently panics when adding noise, due to a division by zero. Ideally `prio` as a library crate shouldn't panic at all, and just return an error upon misconfiguration.

src/flp/types/dp.rs
divergentdave: This is #1073.

documentation/Pure DP Mechanism.lyx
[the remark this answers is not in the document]
divergentdave: I'm running into a number of issues with this.

* MathJax doesn't support \nicefrac
* GitHub Markdown has a different experimental syntax for "alerts", like `> [!NOTE]`, etc. Pandoc is using the same triple colon syntax for definitions and proofs here that Docusaurus uses for inset notes, etc.
* There seems to be a precedence issue in GitHub's Markdown parser, because underscores inside different `$` equations are getting grouped together, interpreted as italicizing the text they contain, and thus breaking equations. I've seen suggestions to escape the underscore with a backslash, but those come along with complaints that other tools don't handle this the same way, and thus render literal underscores inside equations.
* Decorations around the big sigmas and pis are not showing up right, this must be another MathJax incompatibility.
* The bibliographic reference did not get converted or rendered right, it may need to be turned into a footnote.

Ultimately, I think we'll need to pick one ecosystem, either pandoc/LaTeX or GitHub-flavored Markdown, and target it, as there are some fundamental incompatibilities in Markdown handling.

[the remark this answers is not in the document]
tgeoghegan: OK. I don't want to be too difficult here, because you put a lot of effort into good docs, and LaTeX is more natural for math at this level. Two further ideas:

- Can we get this to where it requires a LaTeX installaton but not LyX? Can a `.lyx` file be handled by LaTeX? If not, can we use a format that is LaTeX friendly?
- Could we check in a PDF of the document so that there's something in the repo people can read without needing additional tools?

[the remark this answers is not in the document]
divergentdave: Well, I worked through most of the way through these issues, but two remain. Some absolute value bars disappear on my browser at 100% zoom level, but reappear at 110%. In one equation, we have a sum in an exponent, and that is falling apart. It looks fine when I try to render it using the example JSBin link on mathjax.org, but here both bounds just show up below the sigma.

I can check in either a .tex file or a .pdf file. The PDF is only ~180kB, so I think that should be okay to check in, especially since it'll be the most straightforward to read.

src/dp/distributions.rs
divergentdave: I think the binomial noise in MPC mechanism would be different enough that it wouldn't need to interact with these traits, so I'm not too worried about namespace overlap. Plus, this strategy ultimately gets passed into an `AggregatorWithNoise` implementation, which makes it pretty explicit how noise is added.

src/flp/types/dp.rs
divergentdave: I think we're OK as is. Since we are shifting a constant one, there's no risk of the result getting silently truncated.

documentation/Pure DP Mechanism.lyx
```diff
@@ -0,0 +1,652 @@
...
+\end_layout
+
+\begin_layout Proof
+We will show Definition 1 holds for singleton events,
+ where 
+\begin_inset Formula $E$
+\end_inset
+
+ is a set of cardinality one,
+ then other events will follow by a union bound.
```
cjpatton, line 332: I don't have an intuition for this step. Is the union bound over the power set of $\mathbb{Z}^d$?

documentation/Pure DP Mechanism.lyx
```diff
@@ -0,0 +1,652 @@
...
+\begin_layout Proof
+Let 
+\begin_inset Formula $q=f\left(x\right)$
+\end_inset
+
+,
+ 
+\begin_inset Formula $q^{\prime}=f\left(x^{\prime}\right)$
+\end_inset
+
```
cjpatton, line 344: "For neighboring data sets $x, x'$"

documentation/Pure DP Mechanism.lyx
```diff
@@ -0,0 +1,652 @@
...
+,
+ and 
+\begin_inset Formula $q^{*}=\vec{\pi}_{\mathbb{F}_{p}}\left(f\left(x\right)\right)$
```
cjpatton, line 347: $\vec{\pi}$ appears to be new notation here

documentation/Pure DP Mechanism.lyx
```diff
@@ -0,0 +1,652 @@
...
+\[
+\forall j\in\left[d\right],y_{j}\in\mathbb{F}_{p},\mathbb{P}\left[M_{j}\left(x\right)=y_{j}\right]=\mathbb{P}\left[\pi_{\mathbb{F}_{p}}\left(Z_{j}\right)=y_{j}-q_{j}^{*}\right]
+\]
+
+\end_inset
+
+
+\begin_inset Formula 
+\[
+=\stackrel[k=-\infty]{\infty}{\sum}\mathbb{P}\left[Z_{j}=\pi_{\mathbb{Z}}\left(y_{j}\right)-q_{j}+kp\right]
```
cjpatton, line 387: Should this be
```suggestion
=\stackrel[k=-\infty]{\infty}{\sum}\mathbb{P}\left[Z_{j}=\pi_{\mathbb{Z}}\left(y_{j}\right)-(q_{j}+kp)\right]
```

documentation/Pure DP Mechanism.lyx
```diff
@@ -0,0 +1,652 @@
...
+\]
+
+\end_inset
+
+
+\begin_inset Formula 
+\[
+=\stackrel[k=-\infty]{\infty}{\sum}\frac{e^{\nicefrac{\varepsilon}{GS_{f}}}-1}{e^{\nicefrac{\varepsilon}{GS_{f}}}+1}e^{\nicefrac{-\varepsilon\left|\pi_{\mathbb{Z}}\left(y_{j}\right)-q_{j}+kp\right|}{GS_{f}}}
+\]
```
cjpatton, line 396: A common alternative to $e^X$ is $\exp(X)$. I think it would make this easier to read.

documentation/Pure DP Mechanism.lyx
```diff
@@ -0,0 +1,652 @@
+#LyX 2.4 created this file. For more info see https://www.lyx.org/
```
cjpatton, line 1: Put your name and email on this in case :)

documentation/Pure DP Mechanism.lyx
```diff
@@ -0,0 +1,652 @@
...
+\[
+\stackrel[j=1]{d}{\prod}\stackrel[k=-\infty]{\infty}{\sum}e^{-\frac{\varepsilon}{GS_{f}}\left|\pi_{\mathbb{Z}}\left(y_{j}\right)-q_{j}^{\prime}+kp\right|}\le e^{\varepsilon}\stackrel[j=1]{d}{\prod}\stackrel[k=-\infty]{\infty}{\sum}e^{-\frac{\varepsilon}{GS_{f}}\left|\pi_{\mathbb{Z}}\left(y_{j}\right)-q_{j}+kp\right|}
+\]
+
+\end_inset
+
+This shows that 
+\begin_inset Formula $\mathbb{P}\left[M\left(x^{\prime}\right)=y\right]\le e^{\varepsilon}\cdot\mathbb{P}\left[M\left(x\right)=y\right]$
+\end_inset
+
```
cjpatton, line 597: for any $x, y$.

cjpatton (commented):
I took at a look at the DP argument first: some editorial feedback, plus a couple of steps I didn't understand. I'll go on to the rest of the code now.

src/dp.rs
cjpatton: Thanks @kunal-talwar for giving us some intuition. I've gone the proof. There was a couple of steps I didn't see, but I'm confident David will clarify.

src/dp/distributions.rs
```diff
@@ -262,6 +262,8 @@ where
 }
 
 /// A DP strategy using the discrete gaussian distribution providing zero-concentrated DP.
+///
+/// This uses l2-sensitivity, with the substitution definition of neighboring datasets.
```
cjpatton, line 266: I'm not sure what's more common, "l2" or "L2"
```suggestion
/// This uses L2-sensitivity, with the substitution definition of neighboring datasets.
```

src/dp/distributions.rs
```diff
@@ -287,6 +289,89 @@ impl DifferentialPrivacyStrategy for DiscreteGaussianDpStrategy<ZCdpBudget> {
     }
 }
 
+/// Samples `BigInt` numbers according to the discrete Laplace distribution, with the given scale
+/// parameter. The distribution is defined over the integers, represented by arbitrary-precision
+/// integers. The sampling procedure follows [[CKS20]].
+///
+/// [CKS20]: https://arxiv.org/pdf/2004.00010.pdf
+pub struct DiscreteLaplace {
+    /// The scale parameter of the distribution.
```
cjpatton, line 298: Is this $t$ in your writeup?

src/dp/distributions.rs
```diff
@@ -287,6 +289,89 @@ impl DifferentialPrivacyStrategy for DiscreteGaussianDpStrategy<ZCdpBudget> {
...
+impl DiscreteLaplace {
+    /// Create a new sampler for the discrete Laplace distribution with the given scale parameter.
+    /// Returns an error if the scale parameter is zero or if it has a denominator of zero.
+    pub fn new(scale: Ratio<BigUint>) -> Result<Self, DpError> {
+        if scale.denom().is_zero() {
+            return Err(DpError::ZeroDenominator);
+        }
+        if scale.numer().is_zero() {
+            return Err(DpError::InvalidParameter(
+                "the scale of the discrete Laplace distribution must be nonzero".into(),
```
cjpatton, line 311: Does it need to be positive?

src/dp/distributions.rs
cjpatton: A binomial noise sampler w/o MPC seems conceivable, but I don't think we need to plan for this.

src/flp/types.rs
```diff
@@ -11,6 +11,9 @@ use std::fmt::{self, Debug};
 use std::marker::PhantomData;
 use subtle::Choice;
 
+#[cfg(feature = "experimental")]
```
cjpatton, line 14: Do we still want to put DP stuff behond the experimental flag? I guess the APIs haven't quite stabilized, which is one argument for doing it.

src/flp/types/dp.rs
cjpatton, line 1: You may want a copyright notice here.

src/flp/types/dp.rs
```diff
@@ -0,0 +1,331 @@
...
+    fn add_noise<R>(
+        &self,
+        dp_strategy: &PureDpDiscreteLaplace,
+        agg_result: &mut [F],
+        rng: &mut R,
+    ) -> Result<(), FlpError>
+    where
+        R: Rng,
+    {
+        // Compute the l1-sensitivity of the aggregation function (assuming the substitution-DP
```
cjpatton, line 67: Same comment as above
```suggestion
        // Compute the L1-sensitivity of the aggregation function (assuming the substitution-DP
```

src/flp/types/dp.rs
cjpatton: Does @martinthomson's suggestion imply specific client behavior, such as randomized response? For example, a client might submit a 1-hot vector, or a vector of all 0s, depending on the outcome of a coinflip?

Exposing in the API what notion of "neighboring databases" we consider seems fraught to me. It's perhaps safer if we don't ask the user to go through this analysis themselves. Of course, this is just applying a more general principle of good API design: with DP we may be in a situation where we really do need to expose a choice (and have an expert make that choice).

Question: Is there a way to be "conservative" here? In particular, is it the case that substitution-DP implies deletion-DP with the same bound, but perhaps at the cost of some amount of utility?

I understand there is time pressure to merge this, so if we merge as-is, I would suggest at least making sure it's documented that we're assuming the application requires substitution-DP.

src/flp/types/dp.rs
cjpatton, line 67: What's the difference between "general" and "l1-" sensitivity? The former is used in your writeup.

cjpatton (commented):
Code looks good! I'll await your response to the comments on the writeup before approving.

src/dp/distributions.rs
divergentdave: Yes, though I've seen t, b, and sigma used for it. I think I just want to refer to it as the "scale" in documentation for clarity.

src/dp/distributions.rs
divergentdave: It does, but since the argument's type is `Ratio<BigUint>`, negative values are ruled out already.

src/flp/types.rs
divergentdave: TypeWithNoise and AggregatorWithNoise are currently behind this feature, so this was the easiest change for now. I think removing the feature gate can wait for a later change.

src/flp/types/dp.rs
divergentdave: The neighboring dataset definitions used are documented on the `PureDpDiscreteLaplace` and `ZCdpDiscreteGaussian` items now.

The comparison between substitution versus deletion differs based on the query -- in this PR, you get the same numeric global sensitivity value either way for Prio3SumVec, whereas you get 2 or 1 for Prio3Histogram. I think it will always be the case that you get sensitivities that are greater or equal with substitution, since a substitution is equivalent to a deletion and an insertion. However, the different assumptions complicate things, because an epsilon value used in a substitution-DP mechanism measures against a different set of counterfactuals than an epsilon value used in a deletion-DP mechanism.

src/flp/types/dp.rs
divergentdave: I can't find "general" there, did you mean "global sensitivity"? The two are orthogonal. Global sensitivity, as opposed to smooth sensitivity, gives you one value for your query function, and takes the supremum over all pairs of neighboring datasets. IIRC, smooth sensitivity is parameterized by both the query function and the dataset upon which you evaluate it. The benefit of smooth sensitivity is that you can avoid bad utility due to steep changes if your data is never or rarely in the neighborhood of such issues, but the downside is that you now have to think about the sensitivity of your sensitivity, since the smooth sensitivity is data-dependent. L1-sensitivity and L2-sensitivity just refer to whether you use the taxicab metric or Euclidean metric to measure the distance between query results.

src/dp/distributions.rs
cjpatton: Yup, scale is a good term to use here. I just wanted to clarify. Perhaps mention in the write up that $t = \texttt{scale}$.

src/flp/types/dp.rs
cjpatton: Yes I meant global sensitivity.

This context is useful. I first read this as a discrepancy between what the write up says and what the comments say.

documentation/Pure DP Mechanism.lyx
```diff
@@ -0,0 +1,652 @@
...
+.
+\end_layout
+
+\begin_layout Definition
+A randomized algorithm 
+\begin_inset Formula $M:\mathcal{X}^{n}\rightarrow\mathcal{Y}$
+\end_inset
+
+ satisfies 
+\begin_inset Formula $\varepsilon$
```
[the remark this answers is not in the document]
cjpatton: PIck whichever you prefer.

documentation/Pure DP Mechanism.lyx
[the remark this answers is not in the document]
cjpatton: Well written LaTeX is very readable :)

cjpatton (approved):

tgeoghegan (approved):

PR #1077, opened 2024-06-13
Prio2: split precomputations to fix warnings

divergentdave:
This removes one precomputed array and moves another to a test-only struct. This fixes a lint with the 1.79.0 toolchain.


divergentdave:
Between Cargo feature flags and the test profile, different builds were actually using three different subsets of `PolyAuxMemory`'s fields. I eliminated that struct, and inlined its fields elsewhere.

tgeoghegan (approved):

PR #1080, opened 2024-06-14
bt: Implementing binary tree backed by a vector.

armfazh:
Related to #947.

Using a vector of nodes to represent an append-only binary tree.

E.g. This is the representation of a tree with three nodes. 
```mermaid
flowchart TB
    A --> B
    A --> C
```
In-memory storage:

|NodeRef| Node | Left | Right |
|--|--|--|--|
|0| A | 1 | 2|
|1| B | None  | None |
|2| C | None  | None |

Encoded representation: Nodes are serialized in a pre-order ordering.

The table of nodes starts with `NODES_CAPACITY=256` nodes pre-allocated in order to reduce frequent allocations.

## Benchmark Comparison

Code for the benchmark comparison: https://github.com/armfazh/libprio-rs/pull/3

**Task:** Insert `NUM_PATHS=1000` random paths of length N.

Old: Uses canonical pointers data structure (main).
New: Uses a table (`Vec`) to store nodes (this PR).

Root: Traverses the tree from the root node.
Node: Traverses the tree from the last node inserted. 

See **new** timings [below](https://github.com/divviup/libprio-rs/pull/1080#issuecomment-2225410553) in the thread of comments.

### DIFF
--- a/src/bt.rs
+++ b/src/bt.rs
@@ -132,16 +132,7 @@ impl<V> BinaryTree<V> {
     /// In the error cases, no insertion occurs and the value is returned to
     /// the caller of this function.
     pub fn insert(&mut self, path: &Path, value: V) -> Result<NodeRef, BinaryTreeError<V>> {
-        if self.nodes.is_empty() {
-            if path.is_empty() {
-                self.nodes.push(Node::new(value));
-                Ok(NodeRef::ROOT)
-            } else {
-                Err(BinaryTreeError::UnreachableNode(value))
-            }
-        } else {
-            self.insert_at(NodeRef::ROOT, path, value)
-        }
+        self.insert_at(NodeRef::ROOT, path, value)
     }
 
     /// Inserts the value at the location reached by traversing the tree
@@ -169,12 +160,14 @@ impl<V> BinaryTree<V> {
         path: &Path,
         value: V,
     ) -> Result<NodeRef, BinaryTreeError<V>> {
-        if !self.is_valid_node_ref(node_ref) {
+        if !(self.is_valid_node_ref(node_ref)
+            || (self.nodes.is_empty() && node_ref == NodeRef::ROOT))
+        {
             return Err(BinaryTreeError::InvalidNodeRef(value));
         }
 
         let new_index = self.nodes.len();
-        let mut node = &mut Some(node_ref.0);
+        let mut node = &mut (!self.nodes.is_empty()).then_some(node_ref.0);
         for bit in path.iter() {
             match *node {
                 None => return Err(BinaryTreeError::UnreachableNode(value)),
@@ -239,7 +232,7 @@ impl<V> BinaryTree<V> {
 
     /// Checks whether the node reference is valid with respect to the tree.
     fn is_valid_node_ref(&self, node_ref: NodeRef) -> bool {
-        !self.nodes.is_empty() && node_ref.0 < self.nodes.len()
+        node_ref.0 < self.nodes.len()
     }
 }
 
@@ -510,8 +503,8 @@ mod tests {
         t0.insert(bits!(1, 1, 1, 1), 10000).unwrap();
 
         let mut t1 = BinaryTree::<u32>::default();
-        let mut node_ref;
-        node_ref = t1.insert(bits!(), 1).unwrap();
+        let mut node_ref = NodeRef::ROOT;
+        node_ref = t1.insert_at(node_ref, bits!(), 1).unwrap();
         node_ref = t1.insert_at(node_ref, bits!(1), 10).unwrap();
         node_ref = t1.insert_at(node_ref, bits!(1), 100).unwrap();
         node_ref = t1.insert_at(node_ref, bits!(1), 1000).unwrap();
@@ -519,5 +512,31 @@ mod tests {
 
         assert_eq!(node_ref, NodeRef(4));
         assert_eq!(t0.get_encoded().unwrap(), t1.get_encoded().unwrap());
+    }
+
+    #[test]
+    fn insert_at_invalid_node_ref() {
+        // `insert_at` does not accept invalid node references, when the tree is empty.
+        let mut tree = BinaryTree::<u32>::default();
+        let invalid_node_ref = NodeRef(77);
+        assert!(!tree.is_valid_node_ref(invalid_node_ref));
+        assert_eq!(
+            tree.insert_at(invalid_node_ref, bits!(), 1)
+                .unwrap_err()
+                .to_string(),
+            BinaryTreeError::InvalidNodeRef(1).to_string()
+        );
+    }
+
+    #[test]
+    fn insert_at_root_ok() {
+        // `insert_at` does accept `NodeRef::ROOT`, when the tree is empty.
+        let mut tree = BinaryTree::<u32>::default();
+        let invalid_node_ref = NodeRef::ROOT;
+        assert!(!tree.is_valid_node_ref(invalid_node_ref));
+        assert_eq!(
+            tree.insert_at(invalid_node_ref, bits!(), 1).unwrap(),
+            NodeRef::ROOT
+        );
     }
 }
--- a/src/bt.rs
+++ b/src/bt.rs
@@ -8,18 +8,22 @@
 //! - Serializable: tree can be coverted to and from bytes.
 //!
 //! ## Creation
-//! Use [`BinaryTree::default`] to create a binary tree, note that there must be
-//! one value at the root node.
+//! Use [`BinaryTree::default`] to create a binary tree.
 //!
 //! ## Insertion
-//! Given a value and a binary path, [`BinaryTree::insert`] stores the value
-//! at the end of the path.
+//! Given a value and a binary path, [`BinaryTree::insert`] traverses the
+//! tree along the path starting from the root and inserts the value at the
+//! end of the path. [`BinaryTree::insert_at`] additionally takes a node
+//! reference that indicates the start of the traversal. No insertion occurs
+//! when the end of the path is unreachable or when the tree already contains
+//! a value at the specified location.
 //!
 //! ## Query
-//! Given a binary path, [`BinaryTree::get`] returns a reference to the value
-//! stored in the node located at the end of the path. The empty path returns
-//! the value at the root node. No value is returned if the end of the path
-//! is unreachable.
+//! Given a binary path, [`BinaryTree::get_node`] traverses the tree along
+//! the path starting from the root and returns a reference to the node
+//! located at the end of the path. No value is returned if the end of the
+//! path is unreachable. [`BinaryTree::get_value`] behaves similarly but
+//! instead returns a reference to the value stored in the node.
 //!
 //! ## Serialization
 //! A tree can be serializad to and from bytes using [`BinaryTree::encode`]
@@ -36,6 +40,7 @@
 //! (00)=c  (01)=d   (10)=e  (11)=f
 //! ```
 //!
+//! ```txt
 //! use prio::bt::BinaryTree;
 //! use bitvec::{bits,prelude::Lsb0};
 //! let mut tree = BinaryTree::default();
@@ -46,6 +51,7 @@
 //! tree.insert(bits!(0, 1), 'd');
 //! tree.insert(bits!(1, 0), 'e');
 //! tree.insert(bits!(1, 1), 'f');
+//! ```
 
 // TODO(#947): Remove these lines once the module gets used by Mastic implementation.
 #![allow(dead_code)]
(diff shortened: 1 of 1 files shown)


src/bt.rs
```diff
@@ -74,324 +73,217 @@ pub enum BinaryTreeError<V> {
...
-            }
-            Ref::Other(empty_node) => {
-                *empty_node = Some(Box::new(Node::new(value)));
+        if node.is_some() {
+            return Err(BinaryTreeError::InsertNonEmptyNode(value));
+        } else {
+            *node = Some(last);
+
+            if self.nodes.len() == self.nodes.capacity() {
+                self.nodes.reserve(Self::NODES_CAPACITY);
```
branlwyd, line 127: `Vec::push` already grows the capacity as needed, using an amortized strategy ensuring exponential growth of capacity. `Vec::reserve` falls through to the same codepath when it needs to grow the vector; while this reservation logic requests `NODES_CAPACITY = 256` rather than 1 more element each time, I think `Vec`'s [amortized-growth logic](https://doc.rust-lang.org/src/alloc/raw_vec.rs.html#464) means that we'll be growing the vector's underlying buffer by the same amount as if we just pushed each element without this reservation logic.

I think the logic to manually grow the vec by `NODES_CAPACITY` should be dropped, and we should just `push` each new element on, trusting the `Vec` implementation to provide appropriate performance/space tradeoff when growing the vector.

src/bt.rs
```diff
@@ -74,324 +73,217 @@ pub enum BinaryTreeError<V> {
...
-        Self {
-            value,
-            left: None,
-            right: None,
-        }
-    }
+/// Represents an append-only binary tree.
+pub struct BinaryTree<V> {
+    nodes: Vec<Node<V>>,
+    root: Option<usize>,
```
branlwyd, line 86: Is `root` necessary? We could use a convention like:
* If `nodes` is empty, the tree is empty & there is no root.
* If `nodes` is not empty, the root is at index 0.

(Indeed, I think there is no way to generate a tree whose root is not at index 0 -- the decoding code for `BinaryTree` already assumes this.)

src/bt.rs
```diff
@@ -74,324 +73,217 @@ pub enum BinaryTreeError<V> {
...
-            match elem {
-                None => CodecMarker::Leaf.encode(bytes)?,
-                Some(node) => {
-                    CodecMarker::Inner.encode(bytes)?;
-                    node.value.encode(bytes)?;
-                    stack.push(&node.right);
-                    stack.push(&node.left);
-                }
-            }
+impl Encode for Option<usize> {
```
branlwyd, line 185: Rather than defining `Encode` for `Option<usize>`, should we do this on an option of a newtype (perhaps named something like `Option<TreeIndex>`)? Otherwise, every user of the `Encode` trait will be locked into this implementation on `Option<usize>`, and IMO this is not a general-purpose implementation.

src/bt.rs
```diff
@@ -74,324 +73,217 @@ pub enum BinaryTreeError<V> {
...
-enum CodecMarker {
-    Leaf,
-    Inner,
+impl<V: Decode> Decode for BinaryTree<V> {
+    fn decode(bytes: &mut Cursor<&[u8]>) -> Result<Self, CodecError> {
+        let mut tree = Self::default();
+        let header = UsizeType::decode(bytes)?;
+        if header != 0 {
+            let len = header.into();
+            tree.nodes = Vec::with_capacity(len + Self::NODES_CAPACITY);
```
branlwyd, line 261: nit: instead of allocating a new vector (which will toss the vector of size `NODES_CAPACITY` allocated in the `Default` implementation), `reserve` `len` additional elements. This will perform no allocation if the storage space is large enough already, and a minimal allocation if it is not.

Alternatively, drop this line entirely, and let `push` resize the vector appropriately.

src/bt.rs
```diff
@@ -74,324 +73,217 @@ pub enum BinaryTreeError<V> {
...
-                    CodecMarker::Inner.encode(bytes)?;
-                    node.value.encode(bytes)?;
-                    stack.push(&node.right);
-                    stack.push(&node.left);
-                }
-            }
+impl Encode for Option<usize> {
+    fn encode(&self, bytes: &mut Vec<u8>) -> Result<(), CodecError> {
+        match *self {
+            None => UsizeType::encode(&0, bytes),
```
branlwyd, line 188: While this doesn't cause functional issues due to the way higher-level encoding/decoding functions are written, IMO it's questionable to encode `None` as a zero value; specifically, `None` and `Some(0)` are encoded in the same way, and `Some(0)` becomes `None` when roundtripped through encoding.

(This doesn't cause problems because, in practice, we never decode `Some(0)` for an encoding of a valid tree: decoding of `Option<usize>` is done only for the indices of child nodes, and the root is always at index 0, so index 0 will never be a valid child index.)

I think this is especially important to address if we decide to keep this as an encoding on `Option<usize>` rather than `Option<SomeNewtype>`, since in that case anyone using `Encode` will be using this implementation & they are unlikely to expect this behavior.

src/bt.rs
```diff
@@ -74,324 +73,217 @@ pub enum BinaryTreeError<V> {
...
 }
 
-/// Marker used to distinguish between full and empty nodes.
-#[repr(u8)]
-#[derive(Debug)]
-enum CodecMarker {
-    Leaf,
-    Inner,
+impl<V: Decode> Decode for BinaryTree<V> {
+    fn decode(bytes: &mut Cursor<&[u8]>) -> Result<Self, CodecError> {
```
branlwyd, line 256: A decoded `BinaryTree` might have:
* Cycles (i.e. node A is a parent of node B, and B is a parent of A)
* Dangling pointers (i.e. node indices which do not indicate a node)

Should we check for these? I am not certain how important the "integrity" of decoded trees is; at the least, dangling pointers could cause panics on indexing into the array at an index greater than the length of the array.

src/bt.rs
```diff
@@ -74,324 +73,217 @@ pub enum BinaryTreeError<V> {
...
     }
 
     /// Gets a mutable reference to the node located at the end of the path.
     ///
-    /// This function traverses the tree from this node (`self`) until reaching the
+    /// This function traverses the tree from the root node until reaching the
     /// node at the end of the path. It returns [None], if the node is
     /// unreachable or nonexistent. Otherwise, it returns a mutable reference
     /// to the node.
     pub fn get_node(&mut self, path: &Path) -> Option<&mut Node<V>> {
```
branlwyd, line 162: Q. What is the intended use of this function? It will return a mutable reference to a `Node`, but all of `Node`'s fields are private and `Node` has no publicly-accessible functionality other than an `Encode` implementation.

divergentdave: There used to be a `Node::insert()`, among others. The idea was that lookups and insertions starting from an intermediate node could skip the first part of tree traversal by retaining a reference to a node. I think this would have to be split out to a new type, like `NodeRef<'a, V>`, to work with the new arena-based approach without running afoul of borrowck.

src/bt.rs
```diff
@@ -74,324 +73,217 @@ pub enum BinaryTreeError<V> {
...
     /// Gets a reference to the value located at the end of the path.
     ///
-    /// This function traverses the tree from this node (`self`) until reaching the
+    /// This function traverses the tree from the root node until reaching the
     /// node at the end of the path. It returns [None], if the node is
     /// unreachable or nonexistent. Otherwise, it returns a reference to the
     /// value stored in the node.
     pub fn get(&self, path: &Path) -> Option<&V> {
-        let mut node = self;
+        let mut node = self.root;
```
branlwyd, line 147: nit: this tree-traversal code is more-or-less duplicated several times (`insert`, `get`, and `get_node`); can we factor this out to a single implementation that we call into at each location that needs to traverse a tree?

divergentdave: I expect we will need two traversal routines, one with & references and one with &mut references.

src/bt.rs
```diff
@@ -23,8 +23,7 @@
 //!
 //! ## Serialization
 //! A tree can be serializad to and from bytes using [`BinaryTree::encode`]
-//! and [`BinaryTree::decode_with_param`] functions, respectively. One-byte
-//! markers are used allowing to store sparse trees with a lower overhead.
+//! and [`BinaryTree::decode`] functions, respectively.
 //!
 //! ## Example
 //! This binary tree can be created with the following code:
```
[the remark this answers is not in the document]
armfazh: I've added the ` ```txt ` marker because the bt module is private.

src/bt.rs
armfazh: not really needed, thanks for pointing this out.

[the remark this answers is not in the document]
armfazh: See the updated code, which removes this inconsistency.

src/bt.rs
armfazh: ok, let's delegate the growth logic to rust Vec implementation

src/bt.rs
armfazh: I factored out the traversal, let me know what you think. Also due to mutability we need to (almost) duplicate code.

src/bt.rs
armfazh: Now, I am using `NodeRef` that store the index of the node in the tree. This reference is returned every time a node is inserted. In addition, the `insert_at` method can take a node reference so during insertion, the traversal can start from other node different than the root.

src/bt.rs
```diff
@@ -74,324 +73,217 @@ pub enum BinaryTreeError<V> {
...
+            nodes: Vec::with_capacity(Self::NODES_CAPACITY),
         }
     }
 }
 
-impl<V: Encode> Encode for Node<V> {
-    fn encode(&self, bytes: &mut Vec<u8>) -> Result<(), CodecError> {
-        self.value.encode(bytes)?;
+// Indicates the datatype used to serialize usize values.
+type UsizeType = u16;
```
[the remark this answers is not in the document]
armfazh: That type was removed now.

src/bt.rs
armfazh: This was removed as is now not needed with the new serialization method.

src/bt.rs
```diff
@@ -74,324 +73,217 @@ pub enum BinaryTreeError<V> {
...
+                .map_err(|_| CodecError::UnexpectedValue)?
+                .encode(bytes),
         }
+    }
 
-        Ok(())
+    fn encoded_len(&self) -> Option<usize> {
+        UsizeType::encoded_len(&0)
     }
 }
```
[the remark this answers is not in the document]
armfazh: removed now

src/bt.rs
```diff
@@ -74,324 +73,217 @@ pub enum BinaryTreeError<V> {
...
+    fn decode(bytes: &mut Cursor<&[u8]>) -> Result<Self, CodecError> {
+        Ok(Self {
+            value: V::decode(bytes)?,
+            left: Option::<usize>::decode(bytes)?,
+            right: Option::<usize>::decode(bytes)?,
+        })
     }
 }
 
 impl<V: Encode> Encode for BinaryTree<V> {
```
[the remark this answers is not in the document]
armfazh: now serialization is based on a pre-order traversal, which guarantee a specific order.
Note that NodeRef are valid while the tree is in memory.

src/bt.rs
armfazh: Correct, I changed the code to serialize using a pre-order traversal and no more checks are required.

src/bt.rs
armfazh: good catch.

src/bt.rs
```diff
@@ -91,310 +97,296 @@ impl<V> Node<V> {
...
+    /// - `Err(UnreachableNode(value))` when the end of the path is unreachable.
+    /// - `Err(InvalidNodeRef(value))` when the node reference is invalid.
     ///
     /// In the error cases, no insertion occurs and the value is returned to
     /// the caller of this function.
-    pub fn insert(&mut self, path: &Path, value: V) -> Result<(), BinaryTreeError<V>> {
-        enum Ref<'a, V> {
-            This(&'a mut Node<V>),
-            Other(&'a mut Option<Box<Node<V>>>),
+    pub fn insert_at(
```
branlwyd, line 166: nit: `insert_at` & `insert` are nearly identical, except that `insert` can insert into an empty tree while `insert_at(NodeRef::ROOT, /* pseudocode */ Path::EMPTY, value)` will fail. Can we factor `insert_at` such that `insert`'s body is just a call to `insert_at(NodeRef::ROOT, path, value)`?

To make this change, I think we'd need to special-case the root-insertion case in one or two places, but that's no worse than special-casing it in `insert` IMO.

src/bt.rs
```diff
@@ -91,310 +97,296 @@ impl<V> Node<V> {
...
-        } else {
-            Err(BinaryTreeError::UnreachableNode(value))
+    /// This function traverses the tree from the specified node until
+    /// reaching the node at the end of the path. It returns [None],
+    /// if the node is unreachable or nonexistent; otherwise, it returns
+    /// a reference to the node.
+    pub fn get_node_at(&self, node_ref: NodeRef, path: &Path) -> Option<NodeRef> {
+        self.is_valid_node_ref(node_ref).then_some(())?;
+
+        let new_index = self.nodes.len();
```
branlwyd, line 231: nit: `new_index` is unused?

src/bt.rs
```diff
@@ -91,310 +97,296 @@ impl<V> Node<V> {
...
-    ///
-    /// This function traverses the tree from the root node until reaching the
-    /// node at the end of the path. It returns [None], if the node is
-    /// unreachable or nonexistent. Otherwise, it returns a mutable reference
-    /// to the node.
-    pub fn get_node(&mut self, path: &Path) -> Option<&mut Node<V>> {
-        self.root.as_mut().and_then(|node| node.get_node(path))
+    /// Checks whether the node reference is valid with respect to the tree.
+    fn is_valid_node_ref(&self, node_ref: NodeRef) -> bool {
+        !self.nodes.is_empty() && node_ref.0 < self.nodes.len()
```
branlwyd, line 243: nit: I think the `!self.nodes.is_empty()` check is unnecessary: the `node_ref.0` is a `usize` and therefore nonnegative, so if `nodes` is empty, `node_ref.0 < self.nodes.len()` will be equivalent to `node_ref.0 < 0`, which is always false.

src/bt.rs
```diff
@@ -91,310 +97,296 @@ impl<V> Node<V> {
...
-                    node: left.as_deref(),
-                });
-            } else {
-                writeln!(f, "{name} <None>")?
-            };
+impl<V: Decode> Decode for BinaryTree<V> {
+    fn decode(bytes: &mut Cursor<&[u8]>) -> Result<Self, CodecError> {
+        #[derive(Default)]
+        struct Item {
+            index: Option<usize>,
```
branlwyd, line 304: nit: `index` -> `parent_index` (I think)?

src/bt.rs
armfazh: this is odd, why clippy didn't catch this issue

divergentdave: This is suppressed by the module-level `#![allow(unused_variables)]`.

src/bt.rs
armfazh: this has been refactored now

armfazh:
folks, this is ready for another round of review.

src/bt.rs
```diff
@@ -91,310 +97,290 @@ impl<V> Node<V> {
...
-            This(&'a mut Node<V>),
-            Other(&'a mut Option<Box<Node<V>>>),
+    pub fn insert_at(
+        &mut self,
+        node_ref: NodeRef,
+        path: &Path,
+        value: V,
+    ) -> Result<NodeRef, BinaryTreeError<V>> {
+        if !(self.is_valid_node_ref(node_ref)
+            || (self.nodes.is_empty() && node_ref == NodeRef::ROOT))
```
cjpatton, line 164: nit: I wonder if the second term in this `||` should be implied by the first? I.e., if the list of nodes is empty, and the refernece is to the root, then it's a valid reference, correct?

Put simply: consider moving `(self.nodes.is_empty() && node_ref == NodeRef::ROOT)` into `is_valid_node_ref()`.

src/bt.rs
```diff
@@ -91,310 +97,290 @@ impl<V> Node<V> {
...
-            Ref::Other(empty_node) => {
-                *empty_node = Some(Box::new(Node::new(value)));
-            }
+        if node.is_some() {
+            return Err(BinaryTreeError::InsertNonEmptyNode(value));
         }
 
-        Ok(())
+        *node = Some(new_index);
+        self.nodes.push(Node::new(value));
```
cjpatton, line 186: I'm probably just misreading the code, but I would expect that, somewhere in this function, we'd need to update the parent of the node we just inserted so that their node reference points to the node we just pushed. What am I missing?

cjpatton (commented):
This looks good to me, but before we land, we should make sure that this is actually faster. Please add some benchmarks and report the performance change in a comment.

src/bt.rs
divergentdave: Yes, `node` is a mutable reference to the `Option<usize>` from the parent node's `left` or `right` link. The name `node` may be a bit misleading.

branlwyd (approved):
Code LGTM; I agree with cjpatton that a benchmark is warranted.

simon-friedberger:
I think this code deserves a comment on why a generic binary tree from a crate wouldn't do or why this shouldn't be a separate crate. It looks like a fairly generic interface to have a binary tree without node removal.

armfazh:
I updated the description of the PR (top). Now it includes timings that compare the current implementation with the implementation in this PR. 

The savings are more significant when using large paths (i.e. large tree's height). Also, insertion in-place offers significant savings too. 

The code for benchmark is [here](https://github.com/armfazh/libprio-rs/pull/3).

cjpatton:
> The savings are more significant when using large paths (i.e. large tree's height). Also, insertion in-place offers significant savings too.

That doesn't seem to be corroborated by the timings you posted:


> `N=8`
> 
> | Insertion at | Old | New |
> |--|--|--|
> | Root | 199.47 µs | 185.05 µs | 
> | Node | 235.31 µs | 180.04 µs | 
> 
> `N=64`
> 
> | Insertion at | Old | New |
> |--|--|--|
> | Root | 5.6078 ms | 5.3975 ms | 
> | Node | 2.8733 ms | 1.0842 ms | 
> 
> `N=256`
> 
> | Insertion at | Old | New |
> |--|--|--|
> | Root | 66.765 ms | 93.256 ms | 
> | Node | 11.680 ms |  4.219 ms | 

From this table it looks like inserting from the root is significantly slower for the new code.

branlwyd:
It appears that insert-from-root only becomes slower once `N` becomes large enough. If `N` continues to increase, does the performance difference become larger?

armfazh:
I revisited the benchmarks and updated the bench script to compute the task correctly.

As @branlwyd pointed out, the time for inserting values consistently increases as the length of paths grows.

With these new measurements, it's clear that the new version (using `std::Vec`) is no better than the current code. 
One reason that explains this is because the expression `self.nodes[node]` is a runtime checked (unlike in C). I tried to use the `unsafe get_unchecked` method but only gives 15% savings, it's not closer to the current code.

**N** is length of path.

### Inserting at Root

| N | Old (Tree) | New (Vector) |
|--|--|--|
|  16  | 293.27 µs | 353.73 µs |
|  64  |   3.56 ms |   5.21 ms |
| 128  |  14.46 ms |  22.05 ms |
| 256  |  58.81 ms |  91.52 ms |
| 512  | 239.45 ms | 374.53 ms |

### Inserting at Node

| N | Old (Tree) | New (Vector) |
|--|--|--|
|  16  | 172.80 µs | 212.11 µs |
|  64  | 686.39 µs | 853.18 µs |
| 128  | 1.3668 ms | 1.6735 ms |
| 256  | 2.7321 ms | 3.3498 ms |
| 512  | 5.4535 ms | 6.6705 ms |

cjpatton:
Alrighty, this seems pretty definitive! @armfazh let's close this PR. Thanks for putting in the time to investigate this.

PR #1087, opened 2024-06-28
Mastic client implementation

hannahdaviscrypto:
Partially addresses #947.

Implements sharding for the Mastic VDAF, and adjusts the VIDPF and SZK modules in support of this development.

### DIFF
--- a/src/flp/szk.rs
+++ b/src/flp/szk.rs
@@ -12,17 +12,22 @@
 //! following a strategy similar to [`Prio3`](crate::vdaf::prio3::Prio3).
 
 use crate::{
-    codec::{CodecError, Encode},
-    field::{FftFriendlyFieldElement, FieldElement},
+    codec::{CodecError, Decode, Encode, ParameterizedDecode},
+    field::{decode_fieldvec, encode_fieldvec, FieldElement},
     flp::{FlpError, Type},
     prng::{Prng, PrngError},
     vdaf::xof::{IntoFieldVec, Seed, Xof, XofTurboShake128},
 };
-use std::{borrow::Cow, marker::PhantomData};
+#[cfg(test)]
+use std::borrow::Cow;
+use std::ops::BitAnd;
+use std::{io::Cursor, marker::PhantomData};
+use subtle::{Choice, ConstantTimeEq};
 
 // Domain separation tags
 const DST_PROVE_RANDOMNESS: u16 = 0;
 const DST_PROOF_SHARE: u16 = 1;
+#[allow(dead_code)]
 const DST_QUERY_RANDOMNESS: u16 = 2;
 const DST_JOINT_RAND_SEED: u16 = 3;
 const DST_JOINT_RAND_PART: u16 = 4;
@@ -57,51 +62,257 @@ pub enum SzkError {
 
 /// Contains an FLP proof share, and if joint randomness is needed, the blind
 /// used to derive it and the other party's joint randomness part.
-#[derive(Clone)]
-pub enum SzkProofShare<F: FftFriendlyFieldElement, const SEED_SIZE: usize> {
-    /// Leader's proof share is uncompressed. The first Seed is a blind, second
-    /// is a joint randomness part.
+#[derive(Debug, Clone)]
+pub enum SzkProofShare<F, const SEED_SIZE: usize> {
+    /// Leader's proof share is uncompressed.
     Leader {
+        /// Share of an FLP proof, as a vector of Field elements.
         uncompressed_proof_share: Vec<F>,
-        leader_blind_and_helper_joint_rand_part: Option<(Seed<SEED_SIZE>, Seed<SEED_SIZE>)>,
+        /// Set only if joint randomness is needed. The first Seed is a blind, second
+        /// is the helper's joint randomness part.
+        leader_blind_and_helper_joint_rand_part_opt: Option<(Seed<SEED_SIZE>, Seed<SEED_SIZE>)>,
     },
     /// The Helper uses one seed for both its compressed proof share and as the blind for its joint
     /// randomness.
     Helper {
+        /// The Seed that acts both as the compressed proof share and, optionally, as the blind.
         proof_share_seed_and_blind: Seed<SEED_SIZE>,
-        leader_joint_rand_part: Option<Seed<SEED_SIZE>>,
+        /// The leader's joint randomness part, if needed.
+        leader_joint_rand_part_opt: Option<Seed<SEED_SIZE>>,
     },
 }
 
+impl<F: FieldElement, const SEED_SIZE: usize> PartialEq for SzkProofShare<F, SEED_SIZE> {
+    fn eq(&self, other: &SzkProofShare<F, SEED_SIZE>) -> bool {
+        bool::from(self.ct_eq(other))
+    }
+}
+
+impl<F: FieldElement, const SEED_SIZE: usize> ConstantTimeEq for SzkProofShare<F, SEED_SIZE> {
+    fn ct_eq(&self, other: &SzkProofShare<F, SEED_SIZE>) -> Choice {
+        match (self, other) {
+            (
+                SzkProofShare::Leader {
+                    uncompressed_proof_share: s_proof,
+                    leader_blind_and_helper_joint_rand_part_opt: s_blind,
+                },
+                SzkProofShare::Leader {
+                    uncompressed_proof_share: o_proof,
+                    leader_blind_and_helper_joint_rand_part_opt: o_blind,
+                },
+            ) => s_proof[..]
+                .ct_eq(&o_proof[..])
+                .bitand(option_tuple_ct_eq(s_blind, o_blind)),
+            (
+                SzkProofShare::Helper {
+                    proof_share_seed_and_blind: s_seed,
+                    leader_joint_rand_part_opt: s_rand,
+                },
+                SzkProofShare::Helper {
+                    proof_share_seed_and_blind: o_seed,
+                    leader_joint_rand_part_opt: o_rand,
+                },
+            ) => s_seed.ct_eq(o_seed).bitand(option_ct_eq(s_rand, o_rand)),
+            _ => Choice::from(0),
+        }
+    }
+}
+
+impl<F: FieldElement, const SEED_SIZE: usize> Encode for SzkProofShare<F, SEED_SIZE> {
+    fn encode(&self, bytes: &mut Vec<u8>) -> Result<(), CodecError> {
+        match self {
+            SzkProofShare::Leader {
+                uncompressed_proof_share,
+                leader_blind_and_helper_joint_rand_part_opt,
+            } => (
+                encode_fieldvec(uncompressed_proof_share, bytes)?,
+                if let Some((blind, helper_joint_rand_part)) =
+                    leader_blind_and_helper_joint_rand_part_opt
+                {
+                    blind.encode(bytes)?;
+                    helper_joint_rand_part.encode(bytes)?;
+                },
+            ),
+            SzkProofShare::Helper {
+                proof_share_seed_and_blind,
+                leader_joint_rand_part_opt,
+            } => (
+                proof_share_seed_and_blind.encode(bytes)?,
+                if let Some(leader_joint_rand_part) = leader_joint_rand_part_opt {
+                    leader_joint_rand_part.encode(bytes)?;
+                },
+            ),
+        };
+        Ok(())
+    }
+
+    fn encoded_len(&self) -> Option<usize> {
+        match self {
+            SzkProofShare::Leader {
+                uncompressed_proof_share,
+                leader_blind_and_helper_joint_rand_part_opt,
+            } => Some(
+                uncompressed_proof_share.len() * F::ENCODED_SIZE
+                    + if let Some((blind, helper_joint_rand_part)) =
+                        leader_blind_and_helper_joint_rand_part_opt
+                    {
+                        blind.encoded_len()? + helper_joint_rand_part.encoded_len()?
+                    } else {
+                        0
+                    },
+            ),
+            SzkProofShare::Helper {
+                proof_share_seed_and_blind,
+                leader_joint_rand_part_opt,
+            } => Some(
+                proof_share_seed_and_blind.encoded_len()?
+                    + if let Some(leader_joint_rand_part) = leader_joint_rand_part_opt {
+                        leader_joint_rand_part.encoded_len()?
+                    } else {
+                        0
+                    },
+            ),
+        }
+    }
+}
+
+impl<F: FieldElement + Decode, const SEED_SIZE: usize> ParameterizedDecode<(bool, usize, bool)>
+    for SzkProofShare<F, SEED_SIZE>
+{
+    fn decode_with_param(
+        (is_leader, proof_len, requires_joint_rand): &(bool, usize, bool),
+        bytes: &mut Cursor<&[u8]>,
+    ) -> Result<Self, CodecError> {
+        if *is_leader {
+            Ok(SzkProofShare::Leader {
+                uncompressed_proof_share: decode_fieldvec::<F>(*proof_len, bytes)?,
+                leader_blind_and_helper_joint_rand_part_opt: if *requires_joint_rand {
+                    Some((
+                        Seed::<SEED_SIZE>::decode(bytes)?,
+                        Seed::<SEED_SIZE>::decode(bytes)?,
+                    ))
+                } else {
+                    None
+                },
+            })
+        } else {
+            Ok(SzkProofShare::Helper {
+                proof_share_seed_and_blind: Seed::<SEED_SIZE>::decode(bytes)?,
+                leader_joint_rand_part_opt: if *requires_joint_rand {
+                    Some(Seed::<SEED_SIZE>::decode(bytes)?)
+                } else {
+                    None
+                },
+            })
+        }
+    }
+}
+
 /// A tuple containing the state and messages produced by an SZK query.
-#[derive(Clone)]
-pub(crate) struct SzkQueryShare<F, const SEED_SIZE: usize> {
-    joint_rand_part: Option<Seed<SEED_SIZE>>,
-    verifier: SzkVerifier<F>,
+#[cfg(test)]
+#[derive(Clone, Debug)]
+pub struct SzkQueryShare<F: FieldElement, const SEED_SIZE: usize> {
+    joint_rand_part_opt: Option<Seed<SEED_SIZE>>,
+    flp_verifier: Vec<F>,
 }
 
 /// The state that needs to be stored by an Szk verifier between query() and decide()
-pub(crate) struct SzkQueryState<const SEED_SIZE: usize> {
-    joint_rand_seed: Option<Seed<SEED_SIZE>>,
+pub type SzkQueryState<const SEED_SIZE: usize> = Option<Seed<SEED_SIZE>>;
+
+#[cfg(test)]
+impl<F: FieldElement, const SEED_SIZE: usize> SzkQueryShare<F, SEED_SIZE> {
+    pub(crate) fn merge_verifiers(
+        mut leader_share: SzkQueryShare<F, SEED_SIZE>,
+        helper_share: SzkQueryShare<F, SEED_SIZE>,
+    ) -> SzkVerifier<F, SEED_SIZE> {
+        for (x, y) in leader_share
+            .flp_verifier
+            .iter_mut()
+            .zip(helper_share.flp_verifier)
+        {
+            *x += y;
+        }
+        SzkVerifier {
+            flp_verifier: leader_share.flp_verifier,
+            leader_joint_rand_part_opt: leader_share.joint_rand_part_opt,
+            helper_joint_rand_part_opt: helper_share.joint_rand_part_opt,
+        }
+    }
 }
 
 /// Verifier type for the SZK proof.
-pub type SzkVerifier<F> = Vec<F>;
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct SzkVerifier<F: FieldElement, const SEED_SIZE: usize> {
+    flp_verifier: Vec<F>,
+    leader_joint_rand_part_opt: Option<Seed<SEED_SIZE>>,
+    helper_joint_rand_part_opt: Option<Seed<SEED_SIZE>>,
+}
+
+impl<F: FieldElement, const SEED_SIZE: usize> Encode for SzkVerifier<F, SEED_SIZE> {
+    fn encode(&self, bytes: &mut Vec<u8>) -> Result<(), CodecError> {
+        encode_fieldvec(&self.flp_verifier, bytes)?;
+        if let Some(ref part) = self.leader_joint_rand_part_opt {
+            part.encode(bytes)?
+        };
+        if let Some(ref part) = self.helper_joint_rand_part_opt {
+            part.encode(bytes)?
+        };
+        Ok(())
+    }
+
+    fn encoded_len(&self) -> Option<usize> {
+        Some(
+            self.flp_verifier.len() * F::ENCODED_SIZE
+                + match self.leader_joint_rand_part_opt {
+                    Some(ref part) => part.encoded_len()?,
+                    None => 0,
+                }
+                + match self.helper_joint_rand_part_opt {
+                    Some(ref part) => part.encoded_len()?,
+                    None => 0,
+                },
+        )
+    }
+}
+
+impl<F: FieldElement + Decode, const SEED_SIZE: usize> ParameterizedDecode<(bool, usize)>
+    for SzkVerifier<F, SEED_SIZE>
+{
+    fn decode_with_param(
+        (requires_joint_rand, verifier_len): &(bool, usize),
+        bytes: &mut Cursor<&[u8]>,
+    ) -> Result<Self, CodecError> {
+        if *requires_joint_rand {
+            Ok(SzkVerifier {
+                flp_verifier: decode_fieldvec(*verifier_len, bytes)?,
+                leader_joint_rand_part_opt: Some(Seed::<SEED_SIZE>::decode(bytes)?),
+                helper_joint_rand_part_opt: Some(Seed::<SEED_SIZE>::decode(bytes)?),
+            })
+        } else {
+            Ok(SzkVerifier {
+                flp_verifier: decode_fieldvec(*verifier_len, bytes)?,
+                leader_joint_rand_part_opt: None,
+                helper_joint_rand_part_opt: None,
+            })
+        }
+    }
+}
 
 /// Main struct encapsulating the shared zero-knowledge functionality. The type
 /// T is the underlying FLP proof system. P is the XOF used to derive all random
 /// coins (it should be indifferentiable from a random oracle for security.)
+#[derive(Clone, Debug)]
 pub struct Szk<T, P, const SEED_SIZE: usize>
 where
     T: Type,
     P: Xof<SEED_SIZE>,
 {
-    typ: T,
+    /// The Type representing the specific FLP system used to prove validity of an input.
+    pub(crate) typ: T,
     algorithm_id: u32,
     phantom: PhantomData<P>,
 }
 
-#[cfg(test)]
 impl<T: Type> Szk<T, XofTurboShake128, 16> {
     /// Create an instance of [`Szk`] using [`XofTurboShake128`].
     pub fn new_turboshake128(typ: T, algorithm_id: u32) -> Self {
@@ -207,6 +418,7 @@ where
         .collect()
     }
 
+    #[cfg(test)]
     fn derive_query_rand(&self, verify_key: &[u8; SEED_SIZE], nonce: &[u8; 16]) -> Vec<T::Field> {
         let mut xof = P::init(
             verify_key,
@@ -221,7 +433,15 @@ where
         self.typ.joint_rand_len() > 0
     }
 
-    fn prove(
+    /// Used by a client to prove validity (according to an FLP system) of an input
+    /// that is both shared between the leader and helper
+    /// and encoded as a measurement. Has a precondition that leader_input_share
+    /// \+ helper_input_share = encoded_measurement.
+    /// leader_seed_opt should be set only if the underlying FLP system requires
+    /// joint randomness.
+    /// In this case, the helper uses the same seed to derive its proof share and
+    /// joint randomness.
+    pub(crate) fn prove(
         &self,
         leader_input_share: &[T::Field],
         helper_input_share: &[T::Field],
@@ -236,8 +456,8 @@ where
         // leader its blinding seed and the helper's joint randomness part, and
         // pass the helper the leader's joint randomness part. (The seed used to
         // derive the helper's proof share is reused as the helper's blind.)
-        let (leader_blind_and_helper_joint_rand_part, leader_joint_rand_part, joint_rand) =
-            if let Some(leader_seed) = leader_seed_opt.clone() {
+        let (leader_blind_and_helper_joint_rand_part_opt, leader_joint_rand_part_opt, joint_rand) =
+            if let Some(leader_seed) = leader_seed_opt {
                 let leader_joint_rand_part =
                     self.derive_joint_rand_part(&leader_seed, leader_input_share, nonce)?;
                 let helper_joint_rand_part =
@@ -269,16 +489,17 @@ where
         // Construct the output messages.
         let leader_proof_share = SzkProofShare::Leader {
             uncompressed_proof_share: leader_proof_share,
-            leader_blind_and_helper_joint_rand_part,
+            leader_blind_and_helper_joint_rand_part_opt,
         };
         let helper_proof_share = SzkProofShare::Helper {
-            proof_share_seed_and_blind: helper_seed.clone(),
-            leader_joint_rand_part,
+            proof_share_seed_and_blind: helper_seed,
+            leader_joint_rand_part_opt,
         };
         Ok([leader_proof_share, helper_proof_share])
     }
 
-    fn query(
+    #[cfg(test)]
+    pub(crate) fn query(
         &self,
         input_share: &[T::Field],
         proof_share: SzkProofShare<T::Field, SEED_SIZE>,
@@ -301,8 +522,8 @@ where
             let ((joint_rand_seed, joint_rand), host_joint_rand_part) = match proof_share {
                 SzkProofShare::Leader {
                     uncompressed_proof_share: _,
-                    leader_blind_and_helper_joint_rand_part,
-                } => match leader_blind_and_helper_joint_rand_part {
+                    leader_blind_and_helper_joint_rand_part_opt,
+                } => match leader_blind_and_helper_joint_rand_part_opt {
                     Some((seed, helper_joint_rand_part)) => {
                         match self.derive_joint_rand_part(&seed, input_share, nonce) {
                             Ok(leader_joint_rand_part) => (
@@ -323,8 +544,8 @@ where
                 },
                 SzkProofShare::Helper {
                     proof_share_seed_and_blind,
-                    leader_joint_rand_part,
-                } => match leader_joint_rand_part {
+                    leader_joint_rand_part_opt,
+                } => match leader_joint_rand_part_opt {
                     Some(leader_joint_rand_part) => match self.derive_joint_rand_part(
                         &proof_share_seed_and_blind,
                         input_share,
@@ -363,33 +584,31 @@ where
         )?;
         Ok((
             SzkQueryShare {
-                joint_rand_part,
-                verifier: verifier_share,
+                joint_rand_part_opt: joint_rand_part,
+                flp_verifier: verifier_share,
             },
-            SzkQueryState { joint_rand_seed },
+            joint_rand_seed,
         ))
     }
 
     /// Returns true if the verifier message indicates that the input from which
     /// it was generated is valid.
-    fn decide(
+    pub fn decide(
         &self,
-        verifier: &[T::Field],
-        leader_joint_rand_part_opt: Option<Seed<SEED_SIZE>>,
-        helper_joint_rand_part_opt: Option<Seed<SEED_SIZE>>,
-        joint_rand_seed_opt: Option<Seed<SEED_SIZE>>,
+        verifier: SzkVerifier<T::Field, SEED_SIZE>,
+        query_state: SzkQueryState<SEED_SIZE>,
     ) -> Result<bool, SzkError> {
         // Check if underlying FLP proof validates
-        let check_flp_proof = self.typ.decide(verifier)?;
+        let check_flp_proof = self.typ.decide(&verifier.flp_verifier)?;
         if !check_flp_proof {
             return Ok(false);
         }
         // Check that joint randomness was properly derived from both
         // aggregators' parts
         match (
-            joint_rand_seed_opt,
-            leader_joint_rand_part_opt,
-            helper_joint_rand_part_opt,
+            query_state,
+            verifier.leader_joint_rand_part_opt,
+            verifier.helper_joint_rand_part_opt,
         ) {
             (Some(joint_rand_seed), Some(leader_joint_rand_part), Some(helper_joint_rand_part)) => {
                 let expected_joint_rand_seed =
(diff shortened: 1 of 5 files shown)


src/vidpf.rs
```diff
@@ -108,7 +142,7 @@ impl<W: VidpfValue, const NONCE_SIZE: usize> Vidpf<W, NONCE_SIZE> {
 
     /// [`Vidpf::gen_with_keys`] works as the [`Vidpf::gen`] method, except that two different
     /// keys must be provided.
-    fn gen_with_keys(
+    pub fn gen_with_keys(
```
divergentdave, line 145: Could this be `pub(crate)` instead?

src/vidpf.rs
```diff
@@ -385,21 +455,130 @@ impl From<VidpfServerId> for Choice {
...
 }
 
+impl<'a, T, P, const SEED_SIZE: usize> ParameterizedDecode<&'a Mastic<T, P, SEED_SIZE>>
+    for VidpfCorrectionWord<VidpfWeight<T::Field>>
+where
+    T: Type,
+    P: Xof<SEED_SIZE>,
+{
+    fn decode_with_param(
+        decoding_parameter: &&'a Mastic<T, P, SEED_SIZE>,
```
divergentdave, line 473: We could get rid of the double reference by instead implementing `ParameterizedDecode<Mastic<T, P, SEED_SIZE>>`.

src/vidpf.rs
```diff
@@ -385,21 +455,130 @@ impl From<VidpfServerId> for Choice {
...
+            weight,
+        })
+    }
+}
+
+impl<W: VidpfValue> Encode for VidpfCorrectionWord<W> {
+    fn encode(&self, bytes: &mut Vec<u8>) -> Result<(), CodecError> {
+        for byte in self.seed {
+            bytes.push(byte);
+        }
```
divergentdave, line 502: Conversely to the `read_exact()` suggestion, loops like this could be optimized as follows.
```suggestion
        bytes.extend_from_slice(&self.seed);
```

src/mastic.rs
```diff
@@ -0,0 +1,331 @@
...
+/// Mastic aggregation parameter.
+///
+/// This includes the VIDPF tree level under evaluation, a set of prefixes to evaluate at that level,
+/// and, optionally, the aggregate results of prior levels.
+pub type MasticAggregationParam = Poplar1AggregationParam;
+
+/// Contains broadcast information shared between parties to support VIDPF correctness.
+pub type MasticPublicShare<V> = VidpfPublicShare<V>;
+
+/// Add necessary traits for MasticPublicShare here
```
divergentdave, line 65: nit: This comment looks like it's left over (currently it would show up as part of the MasticInputShare documentation)

src/mastic.rs
```diff
@@ -0,0 +1,331 @@
...
+impl<V: VidpfValue + Debug> Aggregatable for MasticOutputShare<V> {
+    type OutputShare = MasticOutputShare<V>;
+    /// Update an aggregate share by merging it with another (`agg_share`).
+    fn merge(&mut self, agg_share: &Self) -> Result<(), VdafError> {
+        if self.result.len() != agg_share.result.len() {
+            return Err(VdafError::Uncategorized(
+                "Attempted to merge two output shares with different agg_params".to_string(),
+            ));
+        };
+        // Would love to get rid of the below clone if possible.
```
divergentdave, line 121: I think we would have to add a `for<'a> Add<&'a Self, Output = Self>` bound to the `IdpfValue` trait to do so.

src/codec.rs
```diff
@@ -269,6 +269,78 @@ impl Encode for u64 {
         Some(8)
     }
 }
+impl<D: Decode, const BUFFER_SIZE: usize> Decode for [D; BUFFER_SIZE] {
+    fn decode(bytes: &mut Cursor<&[u8]>) -> Result<Self, CodecError> {
+        let mut v = Vec::with_capacity(BUFFER_SIZE);
+        for elem in v.iter_mut().take(BUFFER_SIZE) {
+            *elem = D::decode(bytes)?
+        }
```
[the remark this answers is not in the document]
hannahdaviscrypto: Got, it, thanks!

src/mastic.rs
cjpatton, line 1: This should be a submodule of `vdaf`. Also, you need a copyright string.

src/flp/szk.rs
```diff
@@ -57,29 +61,133 @@ pub enum SzkError {
...
+                    None
+                },
+            }),
+            1 => Ok(SzkProofShare::Helper {
+                proof_share_seed_and_blind: Seed::<SEED_SIZE>::decode(bytes)?,
+                leader_joint_rand_part_opt: if decoding_parameter.is_randomized() {
+                    Some(Seed::<SEED_SIZE>::decode(bytes)?)
+                } else {
+                    None
+                },
```
cjpatton, line 116: Checkout [`bool::then_some`](https://doc.rust-lang.org/std/primitive.bool.html#method.then_some)

src/mastic.rs
```diff
@@ -0,0 +1,331 @@
...
+    vdaf::{
+        poplar1::Poplar1AggregationParam,
+        xof::{Seed, Xof},
+        Aggregatable, Client, Vdaf, VdafError,
+    },
+    vidpf::{
+        Vidpf, VidpfInput, VidpfKey, VidpfPublicShare, VidpfServerId, VidpfValue, VidpfWeight,
+    },
+};
+use std::{fmt::Debug, io::Cursor, marker::PhantomData};
```
cjpatton, line 17:
```suggestion
use std::{fmt::Debug, io::Cursor, marker::PhantomData};

```

src/mastic.rs
```diff
@@ -0,0 +1,331 @@
...
+/// distributed point function.
+#[derive(Clone, Debug)]
+pub struct Mastic<T, P, const SEED_SIZE: usize>
+where
+    T: Type,
+    P: Xof<SEED_SIZE>,
+{
+    algorithm_id: u32,
+    szk: Szk<T, P, SEED_SIZE>,
+    pub(crate) vpf: Vidpf<VidpfWeight<T::Field>, 16>,
```
cjpatton, line 29: I think it's best to avoid introducing a new acronym.
```suggestion
    pub(crate) vidpf: Vidpf<VidpfWeight<T::Field>, 16>,
```

src/mastic.rs
```diff
@@ -0,0 +1,331 @@
...
+    pub bits: usize,
+    phantom: PhantomData<P>,
+}
+
+impl<T, P, const SEED_SIZE: usize> Mastic<T, P, SEED_SIZE>
+where
+    T: Type,
+    P: Xof<SEED_SIZE>,
+{
+    ///Creates a new instance of Mastic, with a specific label length and validity type.
```
cjpatton, line 40:
```suggestion
    /// Creates a new instance of Mastic, with a specific label length and validity type.
```

src/mastic.rs
```diff
@@ -0,0 +1,331 @@
...
+    ) -> Self {
+        Self {
+            algorithm_id,
+            szk,
+            vpf,
+            bits,
+            phantom: PhantomData,
+        }
+    }
+}
```
cjpatton, line 55:
```suggestion
}

```

src/mastic.rs
```diff
@@ -0,0 +1,331 @@
...
+    proofs_share: SzkProofShare<F, SEED_SIZE>,
+}
+impl<'a, T, P, const SEED_SIZE: usize> ParameterizedDecode<(&'a Mastic<T, P, SEED_SIZE>, usize)>
+    for MasticInputShare<T::Field, SEED_SIZE>
+where
+    T: Type,
+    P: Xof<SEED_SIZE>,
+{
+    fn decode_with_param(
+        decoding_parameter: &(&'a Mastic<T, P, SEED_SIZE>, usize),
```
cjpatton, line 83: You can unpack these arguments here, making the code more readable
```suggestion
        (mastic, whatever_this_thing_is_called): &(&'a Mastic<T, P, SEED_SIZE>, usize),
```

src/mastic.rs
```diff
@@ -0,0 +1,331 @@
...
+
+    fn encoded_len(&self) -> Option<usize> {
+        Some(self.vidpf_key.encoded_len()? + self.proofs_share.encoded_len()?)
+    }
+}
+/// Struct containing a vector of VIDPF outputs: one for each prefix. Likely needs to be changed in the next PR to include proofs as well.
+#[derive(Clone, Debug)]
+pub struct MasticOutputShare<V: VidpfValue> {
+    result: Vec<V>,
+}
```
cjpatton, line 111:
```suggestion
}

```

src/flp/szk.rs
```diff
@@ -57,29 +61,133 @@ pub enum SzkError {
...
+    for SzkProofShare<T::Field, SEED_SIZE>
+where
+    T: Type,
+    P: Xof<SEED_SIZE>,
+{
+    fn decode_with_param(
+        decoding_parameter: &Mastic<T, P, SEED_SIZE>,
+        bytes: &mut Cursor<&[u8]>,
+    ) -> Result<Self, CodecError> {
+        let role = <u8>::decode(bytes)?;
```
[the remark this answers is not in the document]
hannahdaviscrypto: The Mastic struct doesn't currently store an aggregator ID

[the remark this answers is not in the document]
divergentdave: The decoding parameter for input shares is defined as `(&'a Self, usize)` in the `Vdaf` trait definition, and the second field is the aggregator ID. We don't need to change the Mastic struct for this, we'd just need to plumb the same tuple as the decoding parameter here for `SzkProofShare`.

src/idpf.rs
```diff
@@ -158,7 +158,7 @@ pub trait IdpfValue:
     + Sized
 {
     /// Any run-time parameters needed to produce a value.
-    type ValueParameter;
+    type ValueParameter: Clone + std::fmt::Debug;
```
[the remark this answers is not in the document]
hannahdaviscrypto: I removed a couple, but most of the bounds that were added were to subfields of the Mastic struct, which has to be Debug to fit the VDAF trait. And I think it will be useful for many of them (VidpfKey, VidpfValues) to be Debuggable, especially since we don't need any extra implementations since FieldElements already implement Debug.

src/vdaf/mastic.rs
```diff
@@ -0,0 +1,473 @@
...
+    use crate::flp::types::Sum;
+    use rand::{thread_rng, Rng};
+
+    const TEST_WEIGHT_LEN: usize = 5;
+    const TEST_NONCE_SIZE: usize = 16;
+
+    #[test]
+    fn test_mastic_shard_sum() {
+        let algorithm_id = 6;
+        let sum_typ = Sum::<Field128>::new(5).unwrap();
```
divergentdave, line 399: We should have some tests using a circuit that uses joint randomness as well, for better coverage

src/vdaf/mastic.rs
```diff
@@ -0,0 +1,473 @@
...
+
+        let first_input = VidpfInput::from_bytes(&[15u8, 0u8, 1u8, 4u8][..]);
+
+        let mastic = Mastic::new(algorithm_id, sum_szk, sum_vidpf, 32);
+        let (_public, input_shares) = mastic.shard(&(first_input, 25u128), &nonce).unwrap();
+        let leader_input_share = &input_shares[0];
+        let helper_input_share = &input_shares[1];
+
+        let mut bytes = vec![];
+        leader_input_share.encode(&mut bytes).unwrap();
```
divergentdave, line 463: As with the other encoding tests, `get_encoded()` and `get_decoded()` would be more concise and a more strict test, respectively.

src/flp/szk.rs
```diff
@@ -57,29 +60,144 @@ pub enum SzkError {
...
+        /// The leader's joint randomness part, if needed.
+        leader_joint_rand_part_opt: Option<Seed<SEED_SIZE>>,
     },
 }
 
+impl<F: FieldElement + Decode, const SEED_SIZE: usize> ParameterizedDecode<(usize, usize, bool)>
+    for SzkProofShare<F, SEED_SIZE>
+{
+    fn decode_with_param(
+        (role, bits, is_random): &(usize, usize, bool),
```
cjpatton, line 87: What does "is_random" mean in this context? Do you mean something like "requires_joint_rand"?

src/flp/szk.rs
```diff
@@ -57,29 +60,144 @@ pub enum SzkError {
...
+                leader_blind_and_helper_joint_rand_part_opt: is_random.then_some((
+                    Seed::<SEED_SIZE>::decode(bytes)?,
+                    Seed::<SEED_SIZE>::decode(bytes)?,
+                )),
+            }),
+            1 => Ok(SzkProofShare::Helper {
+                proof_share_seed_and_blind: Seed::<SEED_SIZE>::decode(bytes)?,
+                leader_joint_rand_part_opt: is_random.then_some(Seed::<SEED_SIZE>::decode(bytes)?),
+            }),
+            _ => Err(CodecError::UnexpectedValue),
```
cjpatton, line 102: `UnexpectedValue` is really intended for malformed messages, not the decoding parameter. Perhaps replace `role: usize` with `is_leader: bool`, which would eliminate this case.

src/flp/szk.rs
```diff
@@ -57,29 +60,144 @@ pub enum SzkError {
...
+                    } else {
+                        0
+                    },
+            ),
+        }
+    }
+}
+
+impl<F: PartialEq, const SEED_SIZE: usize> PartialEq for SzkProofShare<F, SEED_SIZE> {
+    fn eq(&self, other: &SzkProofShare<F, SEED_SIZE>) -> bool {
```
cjpatton, line 166: Why do we need `PartialEq` for `SzkProofShare`? Can it be derived instead?
```rust
#[derive(PartialEq)]
pub SzkProofShare<F, const SEED_SIZE: usize> {
   ...
}
```

src/flp/szk.rs
```diff
@@ -57,29 +61,133 @@ pub enum SzkError {
...
+}
+
 /// A tuple containing the state and messages produced by an SZK query.
+#[cfg(test)]
 #[derive(Clone)]
 pub(crate) struct SzkQueryShare<F, const SEED_SIZE: usize> {
     joint_rand_part: Option<Seed<SEED_SIZE>>,
     verifier: SzkVerifier<F>,
 }
-
```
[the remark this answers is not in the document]
cjpatton: Bump

src/flp/szk.rs
```diff
@@ -91,17 +199,18 @@ pub type SzkVerifier<F> = Vec<F>;
...
 /// coins (it should be indifferentiable from a random oracle for security.)
+#[derive(Clone, Debug)]
 pub struct Szk<T, P, const SEED_SIZE: usize>
 where
     T: Type,
     P: Xof<SEED_SIZE>,
 {
-    typ: T,
+    /// The Type representing the specific FLP system used to prove validity of an input.
+    pub typ: T,
```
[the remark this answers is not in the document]
cjpatton: Bump: `pub(crate)` suffices. This is preferable to `pub` because it minimized the API surface.

src/flp/szk.rs
```diff
@@ -57,29 +60,144 @@ pub enum SzkError {
...
+        (role, bits, is_random): &(usize, usize, bool),
+        bytes: &mut Cursor<&[u8]>,
+    ) -> Result<Self, CodecError> {
+        match role {
+            0 => Ok(SzkProofShare::Leader {
+                uncompressed_proof_share: Vec::<F>::decode_with_param(bits, bytes)?,
+                leader_blind_and_helper_joint_rand_part_opt: is_random.then_some((
+                    Seed::<SEED_SIZE>::decode(bytes)?,
+                    Seed::<SEED_SIZE>::decode(bytes)?,
+                )),
```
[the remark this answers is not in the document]
hannahdaviscrypto: Swapped it for a lazily-evaluated version and added coverage.

src/flp/szk.rs
hannahdaviscrypto: We need PartialEq for MasticInputShare, which contains a SZKProofShare. But yes, deriving it does seem sufficient.

src/vidpf.rs
```diff
@@ -45,6 +46,10 @@ pub enum VidpfError {
     #[error("level index out of bounds")]
     IndexLevel,
 
+    /// Error when input's length is larger than the maximum.
+    #[error("invalid label length")]
+    InvalidLabelLength,
```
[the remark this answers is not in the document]
hannahdaviscrypto: I think I disagree on this one--all of VIDPF is only used in Mastic, and it would mean the VDAF would be passing Mastic errors when it doesn't pass Prio or Poplar errors. Plus the invalid label length is analogous to the invalid weight length and it's a little strange for them to appear at different levels of abstraction.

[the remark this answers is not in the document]
hannahdaviscrypto: Actually, VIDPF::eval should probably be checking that it doesn't try to evaluate longer inputs than it has correction words.

src/flp/szk.rs
```diff
@@ -581,19 +714,126 @@ mod tests {
...
+                &leader_input_share,
+                &helper_input_share,
+                &encoded_measurement[..],
+                [prove_rand_seed, helper_seed],
+                leader_seed_opt,
+                &nonce,
+            )
+            .unwrap();
+
+        let decoding_parameter = (0, szk_typ.proof_len(), true);
```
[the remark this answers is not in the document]
hannahdaviscrypto: Sorry, why is this inconsistent? Sum has nonzero joint randomness length.

src/vidpf.rs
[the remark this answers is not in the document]
hannahdaviscrypto: @divergentdave Bumping this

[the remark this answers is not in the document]
divergentdave: Now that we're using `vidpf::Error::InvalidLabelLength` in the VIDPF `eval()` method as well, I'm fine with this.

src/vidpf.rs
```diff
@@ -462,6 +626,12 @@ impl<F: FieldElement> From<Vec<F>> for VidpfWeight<F> {
     }
 }
 
+impl<F: FieldElement> VidpfWeight<F> {
+    pub(crate) fn as_slice(&self) -> &[F] {
+        &self.0[..]
+    }
+}
```
cjpatton, line 633: More rusty

```suggestion
impl<F> AsRef<[F]> for VidpfWeight<F> {
    fn as_ref(&self) -> &[F] {
        &self.0
    }
}
```

src/vdaf/mastic.rs
```diff
@@ -0,0 +1,727 @@
...
+    P: Xof<SEED_SIZE>,
+{
+    fn shard_with_random(
+        &self,
+        measurement_label: &VidpfInput,
+        measurement_weight: &VidpfWeight<T::Field>,
+        nonce: &[u8; 16],
+        vidpf_keys: [VidpfKey; 2],
+        szk_random: [Seed<SEED_SIZE>; 2],
+        opt_random: Option<Seed<SEED_SIZE>>,
```
cjpatton, line 267: "opt_random" is a somewhat ambiguous term in this context: we have some randomness for the SZK; what is an "OPT"?

src/vdaf/mastic.rs
```diff
@@ -0,0 +1,727 @@
...
+                .share;
+
+        let szk_proof_shares = self.szk.prove(
+            leader_measurement_share.as_slice(),
+            helper_measurement_share.as_slice(),
+            measurement_weight.as_slice(),
+            szk_random,
+            opt_random,
+            nonce,
+        )?;
```
cjpatton, line 322: In general, clone may result in a memory allocation, so we should avoid cloning if we can. Here it's easy, as you just need to unpack the array output by `self.szk.prove()`.

We should do the same thing for `vidpf_keys`, for tidiness if nothing else. Those clones are unlikely to result in allocations.

```suggestion
        let [leader_szk_proof_share, helper_szk_proof_share] = self.szk.prove(
            leader_measurement_share.as_slice(),
            helper_measurement_share.as_slice(),
            measurement_weight.as_slice(),
            szk_random,
            opt_random,
            nonce,
        )?;
        let [leader_vidpf_key, helepr_vidpf_key] = vidpf_keys;
```

src/vdaf/mastic.rs
```diff
@@ -0,0 +1,727 @@
...
+    }
+
+    fn encode_measurement(
+        &self,
+        measurement: &T::Measurement,
+    ) -> Result<VidpfWeight<T::Field>, VdafError> {
+        Ok(VidpfWeight::<T::Field>::from(
+            self.szk.typ().encode_measurement(measurement)?,
+        ))
+    }
```
cjpatton, line 349: As mentioned above, I think these indirections are probably not needed, and they hurt readability somewhat.

src/vdaf/mastic.rs
```diff
@@ -0,0 +1,727 @@
...
+                    nonce,
+                )?
+                .share;
+
+        let szk_proof_shares = self.szk.prove(
+            leader_measurement_share.as_slice(),
+            helper_measurement_share.as_slice(),
+            measurement_weight.as_slice(),
+            szk_random,
+            opt_random,
```
cjpatton, line 320: Can you remind why SZK needs this additional input when joint randomness is required?

src/vdaf/mastic.rs
```diff
@@ -0,0 +1,727 @@
...
+        }
+
+        let vidpf_keys = [
+            VidpfKey::gen(VidpfServerId::S0)?,
+            VidpfKey::gen(VidpfServerId::S1)?,
+        ];
+        let opt_random = match self.szk.has_joint_rand() {
+            true => Some(Seed::<SEED_SIZE>::generate()?),
+            false => None,
+        };
```
cjpatton, line 373: More idiomatic
```suggestion
        let opt_random = self
            .szk
            .has_joint_rand()
            .then_some(Seed::<SEED_SIZE>::generate()?);
```

src/vdaf/mastic.rs
```diff
@@ -0,0 +1,727 @@
...
+        let szk_random = [
+            Seed::<SEED_SIZE>::generate()?,
+            Seed::<SEED_SIZE>::generate()?,
+        ];
+
+        let encoded_measurement = self.encode_measurement(&measurement.1)?;
+        if encoded_measurement.as_slice().len() != *self.vidpf().weight_parameter() {
+            return Err(VdafError::Uncategorized(
+                "encoded_measurement is wrong length".to_string(),
```
cjpatton, line 382:
```suggestion
                "encoded_measurement is the wrong length".to_string(),
```

src/vdaf/mastic.rs
```diff
@@ -0,0 +1,727 @@
...
+}
+
+impl<T, P, const SEED_SIZE: usize> Client<16> for Mastic<T, P, SEED_SIZE>
+where
+    T: Type,
+    P: Xof<SEED_SIZE>,
+{
+    fn shard(
+        &self,
+        measurement: &(VidpfInput, T::Measurement),
```
cjpatton, line 359:
```suggestion
        (label, weight): &(VidpfInput, T::Measurement),
```

src/vidpf.rs
```diff
@@ -76,6 +82,11 @@ impl<W: VidpfValue, const NONCE_SIZE: usize> Vidpf<W, NONCE_SIZE> {
         Self { weight_parameter }
     }
 
+    /// Public accessor method for the weight_parameter field
+    pub(crate) fn weight_parameter(&self) -> &W::ValueParameter {
+        &self.weight_parameter
+    }
```
cjpatton, line 88: Similar point as elsewhere: it may be clearer to make `weight_parameter` `pub(crate)` and drop this accessor.

src/vidpf.rs
```diff
@@ -366,8 +381,48 @@ impl VidpfKey {
...
+            VidpfServerId::S1 => 1u8,
+        });
+        bytes.extend_from_slice(&self.value[..]);
+        Ok(())
+    }
+
+    fn encoded_len(&self) -> Option<usize> {
+        Some(17)
+    }
+}
```
cjpatton, line 413: It should always be the case that whoever is decoding a `VidpfKey` knows its role (i.e., the value of `VidpfServerId`), so I don't think this is needed. We should remove it because:
1. We can save a byte on the wire for each input share (this adds up)
2. We don't have to handle the case where the server tries to parse the wrong role

What we could do instead is add the `VidpfServerId` to the decoding parameter.

src/vidpf.rs
```diff
@@ -366,8 +381,48 @@ impl VidpfKey {
...
+
+impl PartialEq for VidpfKey {
+    fn eq(&self, other: &VidpfKey) -> bool {
+        if self.id == other.id {
+            return self.value == other.value;
+        };
+        false
+    }
+}
```
cjpatton, line 422: This implementation is incomplete because it doesn't compare the key bytes themselves.

Where do we need `PartialEq` for `VidpfKey`?

src/vidpf.rs
```diff
@@ -385,21 +440,130 @@ impl From<VidpfServerId> for Choice {
...
-#[derive(Debug)]
+#[derive(Clone, Debug)]
 struct VidpfCorrectionWord<W: VidpfValue> {
     seed: VidpfSeed,
     left_control_bit: Choice,
     right_control_bit: Choice,
     weight: W,
 }
 
+impl<W: VidpfValue> PartialEq for VidpfCorrectionWord<W> {
```
cjpatton, line 451: Let's follow `Idpf`'s lead here:
https://github.com/divviup/libprio-rs/blob/967cd629aee9fe375332d4e635b6209bfb717a34/src/idpf.rs#L763-L783

src/vidpf.rs
```diff
@@ -385,21 +440,130 @@ impl From<VidpfServerId> for Choice {
...
+        };
+        bytes.push(both_control_bits);
+        self.weight.encode(bytes)?;
+        Ok(())
+    }
+
+    fn encoded_len(&self) -> Option<usize> {
+        Some(17 + self.weight.encoded_len()?)
+    }
+}
```
cjpatton, line 516: Following the lead of the `idpf` module, we would not encode/decode correction words individually. Rather, we would encode/decode the public share so that we can pack the control bits as tightly as possible. I think we're going to do the same thing here: we're trying to shave as many bytes off the wire as we can. 
https://github.com/divviup/libprio-rs/blob/967cd629aee9fe375332d4e635b6209bfb717a34/src/idpf.rs#L675-L679

I propose we punt this to the next PR, since we have plenty to review here.

src/flp/szk.rs
```diff
@@ -57,29 +60,125 @@ pub enum SzkError {
...
+                },
+            })
+        } else {
+            Ok(SzkProofShare::Helper {
+                proof_share_seed_and_blind: Seed::<SEED_SIZE>::decode(bytes)?,
+                leader_joint_rand_part_opt: if *requires_joint_rand {
+                    Some(Seed::<SEED_SIZE>::decode(bytes)?)
+                } else {
+                    None
+                },
```
cjpatton, line 109: [then_some](https://doc.rust-lang.org/nightly/std/primitive.bool.html#method.then_some) works nicely here as well

src/codec.rs
```diff
@@ -270,6 +270,75 @@ impl Encode for u64 {
     }
 }
```
cjpatton, line 272: I mention this elsewhere, but: as a reader, I expect to see the encoder first, then the decoder.

src/codec.rs
```diff
@@ -270,6 +270,75 @@ impl Encode for u64 {
...
+    }
+
+    fn encoded_len(&self) -> Option<usize> {
+        let mut total = 0;
+        for item in self {
+            total += item.encoded_len()?
+        }
+        Some(total)
+    }
+}
```
cjpatton, line 298:
```suggestion
}

```

src/codec.rs
```diff
@@ -270,6 +270,75 @@ impl Encode for u64 {
...
+impl<D: Decode> ParameterizedDecode<usize> for Vec<D> {
+    fn decode_with_param(
+        decoding_parameter: &usize,
```
cjpatton, line 301:
```suggestion
        len: &usize,
```

src/codec.rs
cjpatton, line 1: We might be able to get away with not adding these new implementations of encoders and decoders, but they may be useful for other applications. What do you think about the changes in this file, @branlwyd ?

cjpatton (commented):
Looks good. Most of my comments are about readability or Rust conventions, just a couple of more substantive suggestions.

src/flp/szk.rs
divergentdave: Since there is a side effect to `Seed::decode(bytes)?`, we can't use `then_some()`. See https://github.com/divviup/libprio-rs/pull/1087#discussion_r1681833720. If we wanted to go the combinator route, I think `requires_joint_rand.then(|| Seed::<SEED_SIZE>::decode(bytes)).transpose()?` would work.

src/vdaf/mastic.rs
```diff
@@ -0,0 +1,727 @@
...
+    /// The length of the private label associated with any input.
+    bits: usize,
+}
+
+impl<T, P, const SEED_SIZE: usize> Mastic<T, P, SEED_SIZE>
+where
+    T: Type,
+    P: Xof<SEED_SIZE>,
+{
+    /// Creates a new instance of Mastic, with a specific label length and validity type.
```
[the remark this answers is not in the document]
hannahdaviscrypto: T doesn't just specify the type of weight, it also specifies the validity circuit it has to fulfill. But sure.

src/flp/szk.rs
hannahdaviscrypto: I'd like to use the same technique for both branches of the *is_leader if-statement, and the one above runs decode twice in a tuple, so using transpose()? instead of applying the ? operator to decode directly doesn't seem to work as intended.

src/codec.rs
hannahdaviscrypto: Whoops, I've been going alphabetically.

src/flp/szk.rs
divergentdave: I think that may be doable with question marks both inside and outside of the closure:

```rust
flag.then(|| {
   Ok((
        Seed::<SEED_SIZE>::decode(bytes)?,
        Seed::<SEED_SIZE>::decode(bytes)?,
    ))
}).transpose()?
```

src/vidpf.rs
hannahdaviscrypto: The VDAF interface requires PartialEq for MasticInputShare, which contains a VidpfKey.

src/flp/szk.rs
hannahdaviscrypto: It ended up also needing type annotations.

src/vdaf/mastic.rs
```diff
@@ -0,0 +1,727 @@
...
+    fn eq(&self, other: &MasticInputShare<F, SEED_SIZE>) -> bool {
+        self.vidpf_key == other.vidpf_key && self.proofs_share == other.proofs_share
+    }
+}
+
+/// Struct containing a vector of VIDPF outputs: one for each prefix.
+#[derive(Clone, Debug)]
+pub struct MasticOutputShare<V: VidpfValue> {
+    result: Vec<V>,
+}
```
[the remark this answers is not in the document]
hannahdaviscrypto: Technically yes, but we still have to reimplement most of the things since VDAF only implements them for vectors of single field elements and we have nested vectors.

src/vdaf/mastic.rs
```diff
@@ -0,0 +1,727 @@
...
+        };
+        // Would love to get rid of the below clone if possible.
+        for (a, o) in self.result.iter_mut().zip(output_share.result.iter()) {
+            *a += o.clone();
+        }
+        Ok(())
+    }
+}
+
+impl<'a, T, P, const SEED_SIZE: usize>
```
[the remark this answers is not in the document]
hannahdaviscrypto: yes

src/flp/szk.rs
```diff
@@ -124,6 +333,10 @@ where
         }
     }
 
+    pub(crate) fn typ(&self) -> &T {
+        &self.typ
+    }
```
cjpatton, line 338: Consuder making `typ` `pub(crate)` instead.

src/flp/szk.rs
```diff
@@ -221,7 +435,17 @@ where
         self.typ.joint_rand_len() > 0
     }
 
-    fn prove(
+    pub(crate) fn proof_len(&self) -> usize {
+        self.typ.proof_len()
+    }
```
cjpatton, line 440: Likewise: consider making `typ` `pub(crate)`.

src/flp/szk.rs
```diff
@@ -221,7 +435,17 @@ where
...
+
+    /// Used by a client to prove validity (according to an FLP system) of an input
+    /// that is both shared between the leader and helper
+    /// and encoded as a measurement. Has a precondition that leader_input_share
+    /// + helper_input_share = encoded_measurement.
+    ///     leader_seed_opt should be set only if the underlying FLP system requires
+    ///     joint randomness.
```
cjpatton, line 447:
```suggestion
    /// leader_seed_opt should be set only if the underlying FLP system requires
    /// joint randomness.
```

cjpatton: I'd also clarify that, if the helper's seed is reused for blinding the joint randomness computation.

src/vdaf/mastic.rs
```diff
@@ -0,0 +1,651 @@
...
+}
+
+/// Mastic aggregation parameter.
+///
+/// This includes the VIDPF tree level under evaluation and a set of prefixes to evaluate at that level.
+#[derive(Clone, Debug)]
+pub struct MasticAggregationParam {
+    /// aggregation parameter inherited from [`Poplar1`]: contains the level (attribute length) and a vector of attributes (IdpfInputs)
+    poplar_param: Poplar1AggregationParam,
+    /// Flag indicating whether the VIDPF weight needs to be validated using SZK.
```
cjpatton, line 71: It's worth documenting here under what conditions this would need to be done. In particular, the collector is allowed to start at any level of the tree.

src/vdaf/mastic.rs
```diff
@@ -0,0 +1,651 @@
...
+    }
+}
+
+/// Mastic aggregation parameter.
+///
+/// This includes the VIDPF tree level under evaluation and a set of prefixes to evaluate at that level.
+#[derive(Clone, Debug)]
+pub struct MasticAggregationParam {
+    /// aggregation parameter inherited from [`Poplar1`]: contains the level (attribute length) and a vector of attributes (IdpfInputs)
+    poplar_param: Poplar1AggregationParam,
```
cjpatton, line 70: I think the name "poplar param" is going to get confusing. I think I understand why you wrote the code this way: we have to be able use the implementation of `Encode` on `Poplar1AggregationParam`. A couple of options to consider:
1. rename `poplar_param` to something more generic, like `tree_position` or something?
2. inline `level` and `prefixes` here and refactor so that Poplar1 and Mastic use the same encoding/decoding logic

src/vdaf/mastic.rs
```diff
@@ -0,0 +1,651 @@
...
+    /// Flag indicating whether the VIDPF weight needs to be validated using SZK.
+    root_check_flag: bool,
```
cjpatton, line 72: For consistency with "require joint rand" elsewhere
```suggestion
    require_check_flag: bool,
```

src/vdaf/mastic.rs
```diff
@@ -0,0 +1,651 @@
...
+
+/// Mastic input share
+///
+/// Message sent by the [`Client`] to each Aggregator during the Sharding phase.
+#[derive(Clone, Debug)]
+pub struct MasticInputShare<F: FieldElement, const SEED_SIZE: usize> {
+    /// VIDPF key share.
+    vidpf_key: VidpfKey,
+
+    /// The proof share.
```
cjpatton, line 130: Why is this variable called "proofs_share"? We use this name in Prio3 to denote the fact that there may be multiple, independent proofs generated and verified. Is that also the case for Mastic?

src/vdaf/mastic.rs
```diff
@@ -0,0 +1,651 @@
...
+}
+
+impl<'a, T, P, const SEED_SIZE: usize> ParameterizedDecode<(&'a Mastic<T, P, SEED_SIZE>, usize)>
+    for MasticInputShare<T::Field, SEED_SIZE>
+where
+    T: Type,
+    P: Xof<SEED_SIZE>,
+{
+    fn decode_with_param(
+        (mastic, role): &(&'a Mastic<T, P, SEED_SIZE>, usize),
```
cjpatton, line 153: For consistency with the other `Vdaf` implementations
```suggestion
        (mastic, agg_id): &(&'a Mastic<T, P, SEED_SIZE>, usize),
```

src/vdaf/mastic.rs
```diff
@@ -0,0 +1,651 @@
...
+    }
+}
+
+/// Mastic output share
+///
+/// Contains a vector of VIDPF outputs: one for each prefix.
+pub type MasticOutputShare<V> = OutputShare<V>;
+
+/// Vector of VidpfWeights to be aggregated by Mastic aggregators
+pub type MasticAggregateShare<V> = MasticOutputShare<V>;
```
cjpatton, line 200: The type definition bleeds through the API boundary, so this looks a little odd.  Does this work? This would be more consistent with `Prio3`.
```suggestion
pub type MasticAggregateShare<V> = AggregateShare<V>;
```

src/vidpf.rs
```diff
@@ -366,8 +395,54 @@ impl VidpfKey {
     }
 }
 
+impl ConstantTimeEq for VidpfKey {
+    fn ct_eq(&self, other: &VidpfKey) -> Choice {
+        Choice::from(self.id)
+            .ct_eq(&Choice::from(other.id))
+            .bitand(self.value.ct_eq(&other.value))
```
cjpatton, line 402: It's OK to leak `id`.

src/vidpf.rs
```diff
@@ -366,8 +395,54 @@ impl VidpfKey {
...
+        Ok(VidpfKey {
+            id: if *is_leader {
+                VidpfServerId::S0
+            } else {
+                VidpfServerId::S1
+            },
+            value,
+        })
+    }
+}
```
cjpatton, line 434: footgun: I'm a little worried about this API betting misused down the line. The user might expect that the encoded VIDPF key includes the server ID, which it doesn't.

I think it would be a little safer to just inline this code where it's needed.

src/vidpf.rs
```diff
@@ -384,22 +459,138 @@ impl From<VidpfServerId> for Choice {
...
+            if bool::from(self.right_control_bit) {
+                3u8
+            } else {
+                2
+            }
+        } else if bool::from(self.right_control_bit) {
+            1
+        } else {
+            0
+        };
```
cjpatton, line 504: For IDPF, we don't encode each correction word individually because this wastes bytes. To make the encoding of the public share as compact as possible, we "pack" the control bits as tightly as possible:
https://github.com/cfrg/draft-irtf-cfrg-vdaf/blob/45b75e584d402496988130130b8755a742e5c096/poc/vdaf_poc/idpf_bbcggi21.py#L281-L284

Let's do the same thing for VIDPF.

src/codec.rs
```diff
@@ -270,6 +270,73 @@ impl Encode for u64 {
...
+    }
+
+    fn encoded_len(&self) -> Option<usize> {
+        let mut total = 0;
+        for item in self {
+            total += item.encoded_len()?
+        }
+        Some(total)
+    }
+}
```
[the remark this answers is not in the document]
hannahdaviscrypto: It wasn't that hard to just do everything directly, so I did that instead.

src/vdaf/mastic.rs
hannahdaviscrypto: Nope, fixed now.

divergentdave (approved):

src/vdaf/mastic.rs
```diff
@@ -0,0 +1,547 @@
...
+        let require_check = u8::decode(bytes)?;
+        let require_check_flag = require_check != 0;
+        Ok(Self {
+            level_and_prefixes,
+            require_check_flag,
+        })
+    }
+}
+
+/// Mastic public share
```
cjpatton, line 103: nit: Be consistent about adding a period after the first paragraph (here and below)
```suggestion
/// Mastic public share.
```

src/vdaf/mastic.rs
```diff
@@ -0,0 +1,547 @@
...
+    }
+
+    fn ct_ne(&self, other: &MasticInputShare<F, SEED_SIZE>) -> Choice {
+        self.ct_eq(other).not()
+    }
+}
+
+/// Mastic output share
+///
+/// Contains a vector of VIDPF outputs: one for each prefix.
```
cjpatton, line 209:
```suggestion
/// Contains a flattened vector of VIDPF outputs: one for each prefix.
```

src/vdaf/mastic.rs
```diff
@@ -0,0 +1,547 @@
...
+pub type MasticOutputShare<V> = OutputShare<V>;
+
+/// Vector of VidpfWeights to be aggregated by Mastic aggregators
```
cjpatton, line 212: This comment needs to be updated. Also, start with "Mastic aggregate share." like you've done for other types.

src/vdaf/mastic.rs
```diff
@@ -0,0 +1,542 @@
...
+        let (mastic, agg_param) = decoding_parameter;
+        let l = mastic
+            .vidpf
+            .weight_parameter
+            .checked_mul(agg_param.level_and_prefixes.prefixes().len())
+            .ok_or_else(|| CodecError::Other("multiplication overflow".into()))?;
+        let mut result = Vec::<T::Field>::with_capacity(l);
+        for _ in 0..agg_param.level_and_prefixes.prefixes().len() {
+            result.append(&mut decode_fieldvec(mastic.vidpf.weight_parameter, bytes)?);
+        }
```
divergentdave, line 255: See above
```suggestion
        let result = decode_fieldvec(l, bytes)?;
```

src/flp/szk.rs
```diff
@@ -221,7 +433,15 @@ where
         self.typ.joint_rand_len() > 0
     }
 
-    fn prove(
+    /// Used by a client to prove validity (according to an FLP system) of an input
+    /// that is both shared between the leader and helper
+    /// and encoded as a measurement. Has a precondition that leader_input_share
+    /// \+ helper_input_share = encoded_measurement.
```
cjpatton, line 439:
```suggestion
    /// + helper_input_share = encoded_measurement.
```

src/vidpf.rs
```diff
@@ -384,22 +435,110 @@ impl From<VidpfServerId> for Choice {
...
+            )?);
+        }
+        for _ in 0..*bits {
+            let mut proof = VidpfProof::from([0u8; VIDPF_PROOF_SIZE]);
+            bytes.read_exact(&mut proof)?;
+            cs.push(proof)
+        }
+        Ok(Self { cw, cs })
+    }
+}
```
cjpatton, line 538: Let's stub these out because we're going to need to rewrite them anyway (we have to split the control bits out of each word)
```suggestion
impl<W: VidpfValue> Encode for VidpfPublicShare<W> {
    fn encode(&self, _bytes: &mut Vec<u8>) -> Result<(), CodecError> {
        todo!()
    }

    fn encoded_len(&self) -> Option<usize> {
        todo!()
    }
}

impl<W: VidpfValue> ParameterizedDecode<(usize, W::ValueParameter)> for VidpfPublicShare<W> {
    fn decode_with_param(
        (_bits, _weight_parameter): &(usize, W::ValueParameter),
        _bytes: &mut Cursor<&[u8]>,
    ) -> Result<Self, CodecError> {
        todo!()
    }
}
```

cjpatton (approved):

src/flp/szk.rs
divergentdave: This backslash escape is necessary to avoid a lint and Markdown formatting issues, see https://github.com/divviup/libprio-rs/pull/1087#discussion_r1707576356. A bare plus sign at the beginning of a line is interpreted as the start of an unordered list.

PR #1097, opened 2024-07-17
Use 64-bit implementation for Fp32

armfazh:
Using 64-bit implementation for Fp32, previously, it was using the 128-bit implementation.

Another option is to use directly a dedicated 32-bit implementation, but I will propose that in PR #1098 .

```sh
$ cargo bench --bench speed_tests -- poly_mul
```

| poly_mul | n | Time| Change|
|--|--|--|--|
| fft    | 1   |  210.72 ns | -19.316% |
| fft    | 30  |  5.1285 µs | -39.153% |
| fft    | 60  |  11.299 µs | -40.479% |
| fft    | 90  |  24.815 µs | -42.182% |
| fft    | 120 |  24.746 µs | -42.175% |
| fft    | 150 |  53.449 µs | -42.518% |
| fft    | 255 |  54.293 µs | -42.240% |
| direct | 1   |  68.245 ns | -26.112% |
| direct | 30  |  4.4645 µs | -46.017% |
| direct | 60  |  17.331 µs | -44.944% |
| direct | 90  |  67.985 µs | -45.110% |
| direct | 120 |  70.084 µs | -44.915% |
| direct | 150 |  265.90 µs | -47.659% |
| direct | 255 |  268.17 µs | -44.727% |


src/fp.rs
```diff
@@ -279,17 +279,17 @@ fn modp(x: u128, p: u128) -> u128 {
     z.wrapping_add(m & p)
 }
 
-pub(crate) const FP32: FieldParameters = FieldParameters {
+pub(crate) const FP32: crate::fp64::FieldParameters64 = crate::fp64::FieldParameters64 {
     p: 4293918721, // 32-bit prime
     mu: 17302828673139736575,
-    r2: 1676699750,
-    g: 1074114499,
```
cjpatton, line 286: Changing the generator breaks interop, which is why I think a test is failing. Why change this? Does this benefit performance?

armfazh: no, it shouldn't change. I going to revert it to the original value.

divergentdave: I think this should change, but to a different value. This is because `g` needs to be the Montgomery representation of the generator, and changing big R changes the mapping from residue representation to Montgomery representation. I added the following to `documentation/field_parameters.sage`, to recalculate these parameters with big R equal to 2^64.

```sage
    Field(
        "Field32, u64",
        2 ^ 20 * 4095 + 1,
        3925978153,
        2 ^ 64,
        2 ^ 64,
    ),
```

I get the following output.

```
Field32, u64
p: 4293918721
mu: 17302828673139736575
r2: 2564090464
g: 45203785
num_roots: 20
bit_mask: 4294967295
roots:
[266338049,
 4027580672,
 1578227120,
 3779916562,
 273724820,
 3424176308,
 1517016107,
 1358722360,
 1304575291,
 4258821146,
 4139968002,
 1029526548,
 717934416,
 1564021295,
 3160561122,
 1320076770,
 3893444443,
 1375455849,
 849387224,
 3198638617,
 45203785]
```

armfazh: Correct, it has to change due to Montgomery representation, but keeping the same integer value. Making changes now.

src/field.rs
```diff
@@ -817,7 +817,7 @@ impl Integer for u128 {
 make_field!(
     /// Same as Field32, but encoded in little endian for compatibility with Prio v2.
```
divergentdave, line 818: FWIW this comment is out of date (had me confused for a bit). We changed all field encodings to be little-endian in #495, following a change in draft-irtf-cfrg-vdaf-05. Thus, Prio v2 and VDAF all use the same encoding.

armfazh:
Closing in favor of #1099

PR #1098, opened 2024-07-17
Creating a 32-bit implementation for Fp32

armfazh:
This is a 32-bit oriented implementation for GF(p) where p is a 32-bit prime.

Compared to #1097 , the timings are quite similar since I'm measuring both in a 64-bit machine. It would be interesting how they perform on a native 32-bit architecture.

| poly_mul | n | Time| Change|
|--|--|--|--|
|fft    |   1   | 205.57 ns | -18.674% |
|fft    |   30  | 5.1169 µs | -38.846% |
|fft    |   60  | 10.664 µs | -49.509% |
|fft    |   90  | 23.202 µs | -43.997% |
|fft    |   120 | 23.830 µs | -43.156% |
|fft    |   150 | 50.701 µs | -45.515% |
|fft    |   255 | 51.774 µs | -44.655% |
|direct |   1   | 68.716 ns | -26.442% |
|direct |   30  | 4.3859 µs | -46.988% |
|direct |   60  | 18.329 µs | -45.414% |
|direct |   90  | 69.479 µs | -46.361% |
|direct |   120 | 69.879 µs | -48.257% |
|direct |   150 | 279.00 µs | -45.679% |
|direct |   255 | 279.15 µs | -44.797% |

### DIFF
--- a/src/fp.rs
+++ b/src/fp.rs
@@ -263,6 +263,16 @@ impl FieldParameters {
 }
 
 #[inline(always)]
+pub(crate) fn lo32(x: u64) -> u64 {
+    x & ((1 << 32) - 1)
+}
+
+#[inline(always)]
+pub(crate) fn hi32(x: u64) -> u64 {
+    x >> 32
+}
+
+#[inline(always)]
 pub(crate) fn lo64(x: u128) -> u128 {
     x & ((1 << 64) - 1)
 }
@@ -278,20 +288,6 @@ fn modp(x: u128, p: u128) -> u128 {
     let m = 0u128.wrapping_sub(carry as u128);
     z.wrapping_add(m & p)
 }
-
-pub(crate) const FP32: FieldParameters = FieldParameters {
-    p: 4293918721, // 32-bit prime
-    mu: 17302828673139736575,
-    r2: 1676699750,
-    g: 1074114499,
-    num_roots: 20,
-    bit_mask: 4294967295,
-    roots: [
-        2564090464, 1729828257, 306605458, 2294308040, 1648889905, 57098624, 2788941825,
-        2779858277, 368200145, 2760217336, 594450960, 4255832533, 1372848488, 721329415,
-        3873251478, 1134002069, 7138597, 2004587313, 2989350643, 725214187, 1074114499,
-    ],
-};
 
 pub(crate) const FP128: FieldParameters = FieldParameters {
     p: 340282366920938462946865773367900766209, // 128-bit prime
@@ -344,6 +340,7 @@ pub(crate) mod tests {
     pub(crate) trait TestFieldParameters {
         fn p(&self) -> u128;
         fn g(&self) -> u128;
+        fn base(&self) -> u128;
         fn r2(&self) -> u128;
         fn mu(&self) -> u64;
         fn bit_mask(&self) -> u128;
@@ -367,6 +364,10 @@ pub(crate) mod tests {
 
         fn g(&self) -> u128 {
             self.g
+        }
+
+        fn base(&self) -> u128 {
+            1u128 << 64
         }
 
         fn r2(&self) -> u128 {
@@ -422,7 +423,7 @@ pub(crate) mod tests {
         }
 
         fn radix(&self) -> BigInt {
-            BigInt::from(1) << 128
+            BigInt::from(self.base()).pow(2) // radix = base^2 = (2^64)^2 = 2^128
         }
     }
 
@@ -452,16 +453,6 @@ pub(crate) mod tests {
     }
 
     #[test]
-    fn test_fp32_u128() {
-        all_field_parameters_tests(TestFieldParametersData {
-            fp: Box::new(FP32),
-            expected_p: 4293918721,
-            expected_g: 3925978153,
-            expected_order: 1 << 20,
-        });
-    }
-
-    #[test]
     fn test_fp128_u128() {
         all_field_parameters_tests(TestFieldParametersData {
             fp: Box::new(FP128),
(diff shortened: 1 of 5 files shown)


src/fp32.rs
```diff
@@ -0,0 +1,293 @@
...
+}
+
+#[inline(always)]
+fn modp(x: u32, p: u32) -> u32 {
+    let (z, carry) = x.overflowing_sub(p);
+    let m = 0u32.wrapping_sub(carry as u32);
+    z.wrapping_add(m & p)
+}
+
+pub(crate) const FP32: FieldParameters32 = FieldParameters32 {
```
divergentdave, line 190: Could you check in a corresponding change to `documentation/field_parameters.sage` to generate these values?

divergentdave:
I think that using u32 instead of u64 is worth it, even if the speed is similar, due to the reduced memory usage. I figure PINE parameters will tend to be on the larger side, so avoiding doubling the size of input and output shares means a significant memory pressure reduction.

src/fp32.rs
armfazh: I've added a field in the parameters structure, let me know what do you think.

src/fp.rs
```diff
@@ -344,6 +340,7 @@ pub(crate) mod tests {
     pub(crate) trait TestFieldParameters {
         fn p(&self) -> u128;
         fn g(&self) -> u128;
+        fn r_radix(&self) -> usize;
```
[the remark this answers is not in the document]
armfazh: Updated now.

divergentdave (approved):

src/fp.rs
```diff
@@ -279,22 +291,9 @@ fn modp(x: u128, p: u128) -> u128 {
...
-    roots: [
-        2564090464, 1729828257, 306605458, 2294308040, 1648889905, 57098624, 2788941825,
-        2779858277, 368200145, 2760217336, 594450960, 4255832533, 1372848488, 721329415,
-        3873251478, 1134002069, 7138597, 2004587313, 2989350643, 725214187, 1074114499,
-    ],
-};
-
 pub(crate) const FP128: FieldParameters = FieldParameters {
     p: 340282366920938462946865773367900766209, // 128-bit prime
+    base: 18446744073709551616,
```
cjpatton, line 296: Much more readable (base aligns with a word)
```suggestion
    base: 1 << 64,
```

src/fp.rs
```diff
@@ -12,7 +12,9 @@ pub(crate) const MAX_ROOTS: usize = 20;
 pub(crate) struct FieldParameters {
     /// The prime modulus `p`.
     pub p: u128,
-    /// `mu = -p^(-1) mod 2^64`.
+    /// `base = 2^64`.
```
cjpatton, line 15: Is this `2^64` for all `FieldParameters`? If so, why have a struct field for it? If not:
```suggestion
    /// `base`, e.g., `2^64` for `FP128`.
```

src/fp.rs
```diff
@@ -422,7 +426,7 @@ pub(crate) mod tests {
         }
 
         fn radix(&self) -> BigInt {
-            BigInt::from(1) << 128
+            BigInt::from(self.base).pow(2)
```
cjpatton, line 429: Is the base always a power of 2? If so, is this worth it? I suppose it's possible that `BigInt` 
```suggestion
            debug_assert_eq!(self.base, (self.base - 1).next_power_of_two())
            BigInt::from(self.base) << 1
```

cjpatton: Alternative to `debug_assert_eq!()`: move this to unit tests.

src/fp32.rs
```diff
@@ -0,0 +1,296 @@
...
+    /// elements of `roots`.
+    pub g: u32,
+    /// The number of principal roots of unity in `roots`.
+    pub num_roots: usize,
+    /// Equal to `2^b - 1`, where `b` is the length of `p` in bits.
+    pub bit_mask: u32,
+    /// `roots[l]` is the `2^l`-th principal root of unity, i.e., `roots[l]` has order `2^l` in the
+    /// multiplicative group. `roots[0]` is equal to one by definition.
+    pub roots: [u32; MAX_ROOTS + 1],
+}
```
cjpatton, line 30: Instead of duplicating `FieldParameters`, consider making it generic.

```rust
pub(crate) struct FieldParameters<W> {
    pub p: W,
    pub base: u128,
    pub mu: W,
    pub r2: W,
    pub g: W,
    pub num_roots: usize,
    pub bit_mask: W,
    pub roots: [W; MAX_ROOTS + 1],
}
```

src/fp32.rs
```diff
@@ -0,0 +1,296 @@
...
+        // -   0, p
+        // ========
+        // b1,s1,s0
+        let (s0, b0) = z.overflowing_sub(self.p);
+        let (_s1, b1) = (carry as u32).overflowing_sub(b0 as u32);
+        // if b1 == 1: return z
+        // else:       return s0
+        let m = 0u32.wrapping_sub(b1 as u32);
+        (z & m) | (s0 & !m)
+    }
```
cjpatton, line 51: This function appears to be identical to the method with the same name on `FieldParameters`. If we make `FieldParameters` generic, as recommended above, then we can deduplicate this. Perhaps there are other methods that can be de-duplicated in the same way? Definitely not `mul()`.

cjpatton: To specialize the implementation of mul() you would do something like
```rust
impl FieldParameters<u32> {
   pub fn mul(blah) -> blah {
      blah
   }
}
```

src/fp.rs
armfazh: on a second though I removed the field.

src/fp.rs
armfazh: good point, added one test

src/fp32.rs
armfazh: This is a good idea, but I propose to work on that version in a different PR.

cjpatton: Why not now? I don't think we're in a hurry to land this.

armfazh: See for the implementation that unifies both implementations https://github.com/divviup/libprio-rs/pull/1099

armfazh:
Closing in favor of https://github.com/divviup/libprio-rs/pull/1099

PR #1099, opened 2024-07-19
fp: Unifying implementations of prime fields that fit in a single word.

armfazh:
It adds a new generic structure that implements field operations specialized for prime moduli that fit in a single word. Implementation is generic for basic datatypes.

Field32 is now 1.8x faster. No performance changes for **Field64** and **Field128**.

**Field32**

| poly_mul | N | Diff| New | Old |
|--|--|--|--|--|
| direct | 1   | 1.42  |   38.6±0.69 ns |   55.0±2.16ns  |
| direct | 120 | 1.88  |   40.1±0.28 µs |   75.3±0.74µs  |
| direct | 150 | 1.88  |  158.4±1.13 µs |  298.3±2.48µs  |
| direct | 255 | 1.89  |  158.3±1.82 µs |  298.9±3.48µs  |
| direct | 30  | 1.87  |    2.5±0.07 µs |    4.7±0.07µs  |
| direct | 60  | 1.85  |   10.3±0.20 µs |   19.1±0.15µs  |
| direct | 90  | 1.89  |   40.1±0.20 µs |   75.8±3.02µs  |
| fft    | 1   | 1.26  |  118.3±1.34 ns |  149.2±2.33ns  |
| fft    | 120 | 1.77  |   13.7±0.08 µs |   24.3±0.09µs  |
| fft    | 150 | 1.82  |   30.0±0.59 µs |   54.7±0.91µs  |
| fft    | 255 | 1.80  |   29.9±0.11 µs |   53.8±0.71µs  |
| fft    | 30  | 1.69  |    2.8±0.07 µs |    4.8±0.02µs  |
| fft    | 60  | 1.73  |    6.3±0.10 µs |   10.9±0.05µs  |
| fft    | 90  | 1.78  |   13.7±0.08 µs |   24.4±0.32µs  |


src/fp/single_word.rs
```diff
@@ -0,0 +1,260 @@
...
+    (
+        /// $struct_name is the name of the structure containing field
+        /// parameters and implementing field operations.
+        $struct_name:ident,
+        /// $W is the datatype of `W` bits, such that `p < 2^W`, used for
+        /// arithmetic operations.
+        $W:ty,
+        /// $W2 is the datatype of `2*W` bits used for double-length
+        /// integer operations. It must hold that `W2::BITS == 2*W::BITS`.
+        $W2:ty
```
cjpatton, line 18: nit: I'm fairly certain these are not docucomments, but I could be wrong
```suggestion
// impl_field_single_word implements field operations for prime modulus
// that fits in one word of `W` bits.
macro_rules! impl_field_single_word {
    (
        // $struct_name is the name of the structure containing field
        // parameters and implementing field operations.
        $struct_name:ident,
        // $W is the datatype of `W` bits, such that `p < 2^W`, used for
        // arithmetic operations.
        $W:ty,
        // $W2 is the datatype of `2*W` bits used for double-length
        // integer operations. It must hold that `W2::BITS == 2*W::BITS`.
        $W2:ty
```

src/fp/single_word.rs
```diff
@@ -0,0 +1,260 @@
...
+    ) => {
+        const _: () = assert!(<$W2>::BITS == 2 * <$W>::BITS);
```
cjpatton, line 20: Rust :)

src/fp/single_word.rs
```diff
@@ -0,0 +1,260 @@
+// SPDX-License-Identifier: MPL-2.0
+
+//! Finite field arithmetic for any field GF(p) for which `p < 2^W`,
+//! where `W` is a specified word size.
+
+/// impl_field_single_word implements field operations for prime modulus
```
cjpatton, line 6:
```suggestion
/// impl_field_single_word implements field operations for a prime modulus
```

src/fp/single_word.rs
```diff
@@ -0,0 +1,260 @@
...
+/// that fits in one word of `W` bits.
```
cjpatton, line 7: Is this true if `W` is `u128`? The `mul()` looks to be tailored for `u64` (and maybe `u32`).

src/fp/single_word.rs
```diff
@@ -0,0 +1,260 @@
...
+
+            #[inline(always)]
+            fn modp(&self, x: $W) -> $W {
+                let (z, carry) = x.overflowing_sub(self.p);
+                let m = (0 as $W).wrapping_sub(carry as $W);
+                z.wrapping_add(m & self.p)
+            }
+        }
+
+        #[cfg(test)]
```
cjpatton, line 189: Instead of making this code match the API of the tests, I'd suggest making the tests generic in the word size.

cjpatton (commented):
> It uses a macro since the implementation with generic types introduced a non-small performance overhead. 

Interesting. What benchmarks did you see regress? Did it make Field64 slower, or just Field32?

src/fp/single_word.rs
armfazh: I have updated the code and tests to rely on a generic implementation.

armfazh:
Hi folks, this is ready for another round of reviews.

cjpatton (approved):

src/fp/ops.rs
```diff
@@ -0,0 +1,417 @@
...
+
+        // Montgomery Reduction
+        // z = z + p * mu*(z mod 2^64), where mu = (-p)^(-1) mod 2^64.
+
+        // z3,z2,z1,z0
+        // +     p1,p0
+        // *         w = mu*z0
+        // ===========
+        // z3,z2,z1, 0
+        let mut w = low(Self::MU.wrapping_mul(&z0));
```
divergentdave, line 306: Previously, `FieldParameters::mu` was a `u64` field for 128-bit primes. Now, the `FieldParameters::MU` associated constant has type `W`, meaning it's a `u128` for 128-bit primes. The previous code had a `as u64` cast on the right multiplicand. For one thing, this means the wrapping multiplication operations will wrap at different powers of two. Thus, I'm worried there could be correctness issues downstream if `w` is not in the range it's supposed to be. Secondly, from a performance perspective, this would now be a u128 * u128 multiplication where we expect the top half of both multiplicands to be zero, so the compiler would emit a built-in with three extra u64 * u64 multiplications we're not truly using.

To solve this, we may be able to pass both inputs through `low()`, and then pass the result through `low()` as well to wrap at the right place. It's possible the compiler will see all the masking operations and emit the wrapping u64 * u64 multiplication we want. Alternately, the more involved and more complete solution would be to add another type parameter or an associated for unsigned integers with LOG2_BASE bits, as opposed to `W` which has LOG2_RADIX bits. Then we could make MU smaller, and add the truncating cast back here. (one implementation with 'words' and 'double words', another with 'half words' and 'words')

src/fp/ops.rs
```diff
@@ -0,0 +1,417 @@
...
+
+    /// Negation, i.e., `-x (mod p)` where `p` is the modulus.
+    #[inline(always)]
+    fn neg(x: W) -> W {
+        Self::sub(W::ZERO, x)
+    }
+
+    #[inline(always)]
+    fn modp(x: W) -> W {
+        Self::sub(x, Self::PRIME)
```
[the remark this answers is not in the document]
armfazh: Correct, this works since prime modulus has the form `p=2^W-x`, where `W` is the word size. 
I was thinking how to make this operation more general, if this is ever needed.

src/fp/ops.rs
```diff
@@ -0,0 +1,417 @@
...
+    /// Multiplication. The result will be in [0, p), so long as both x
+    /// and y are as well.
+    fn mul(x: W, y: W) -> W;
+
+    /// Modular exponentiation, i.e., `x^exp (mod p)` where `p` is
+    /// the modulus. Note that the runtime of this algorithm is
+    /// linear in the bit length of `exp`.
+    fn pow(x: W, exp: W) -> W {
+        let mut t = Self::ROOTS[0];
+        for i in (0..W::BITS).rev() {
```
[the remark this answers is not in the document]
armfazh: done,

src/fp/ops.rs
armfazh: Good catch, I've added an auxiliary type and add a compile-time check that tests the size of MU.

divergentdave (approved):

src/fp/ops.rs
```diff
@@ -0,0 +1,451 @@
...
+        let (s0, b0) = z.overflowing_sub(&Self::PRIME);
+        let (_s1, b1) = cc.overflowing_sub(&W::Carry::from(b0).as_());
+        // if b1 == 1: return z
+        // else:       return s0
+        let mask = W::ZERO.wrapping_sub(&W::Carry::from(b1).as_());
+        (z & mask) | (s0 & !mask)
+    }
+}
+
+/// `FieldMulOpsTwoWords` implements prime field multiplication.
```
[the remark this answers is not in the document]
armfazh: done.

src/fp/ops.rs
```diff
@@ -0,0 +1,451 @@
...
+///
+/// The implementation assumes that the modulus `p` fits in one word of
+/// 'W' bits, but the product of two integers does not fit. Thus,
```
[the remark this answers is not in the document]
armfazh: done.

src/fp/ops.rs
```diff
@@ -0,0 +1,451 @@
...
+{
+    type DoubleWord: Word + AsPrimitive<W>;
+
+    /// Multiplication of field elements in the Montgomery domain.
+    /// This uses the Montgomery's [REDC algorithm][montgomery].
+    /// The result will be in [0, p).
+    ///
+    /// [montgomery]: https://www.ams.org/journals/mcom/1985-44-170/S0025-5718-1985-0777282-X/S0025-5718-1985-0777282-X.pdf
+    fn mul(x: W, y: W) -> W {
+        let hi_lo = |v: Self::DoubleWord| ((v >> W::BITS).as_(), v.as_());
```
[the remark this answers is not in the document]
armfazh: done.

src/fp/ops.rs
```diff
@@ -0,0 +1,451 @@
...
+    + WrappingSub
+    + WrappingMul
+    + ConstZero
+    + ConstOne
+{
+    /// Number of bits of word size.
+    const BITS: u32;
+    /// Carry type allows the conversion of carry/borrow values
+    /// produced by adding/subtracting to Word values.
+    type Carry: AsPrimitive<Self> + From<bool>;
```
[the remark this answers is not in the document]
armfazh: good call,

armfazh:
I have included the latest review comments and squashed all commits.

divergentdave:
Thanks!

PR #1107, opened 2024-08-13
Mastic aggregator and collector implementation

hannahdaviscrypto:
Implements preparation, aggregation, and unsharding for the Mastic VDAF, and adjusts the VIDPF and SZK modules in support of this development.


src/vdaf/mastic.rs
```diff
@@ -379,32 +391,485 @@ where
...
+///
+/// Broadcast message from an aggregator between rounds of Mastic. Includes the
+/// hashed VIDPF proofs for every prefix in the aggregation parameter, and optionally
+/// the verification message for Szk.
+#[derive(Clone, Debug)]
+pub enum MasticPrepareShare<F: FieldElement, const SEED_SIZE: usize> {
+    /// Includes a batched VIDPF proof and an SZK verification message for the root weight.
+    FirstRound(Seed<SEED_SIZE>, SzkQueryShare<F, SEED_SIZE>),
+    /// Includes only a batched VIDPF proof as SZK will not be run.
+    LaterRound(Seed<SEED_SIZE>),
```
cjpatton, line 417: Same comment as for prep state: "has-a" is probably clearer than "is-a".

src/vdaf/mastic.rs
```diff
@@ -379,32 +391,485 @@ where
...
+}
+
+impl<F: FieldElement, const SEED_SIZE: usize> Encode for MasticPrepareShare<F, SEED_SIZE> {
+    fn encode(&self, bytes: &mut Vec<u8>) -> Result<(), CodecError> {
+        match self {
+            MasticPrepareShare::FirstRound(seed, query_share) => {
+                seed.encode(bytes).and_then(|_| query_share.encode(bytes))
```
cjpatton, line 424: Fancy!!! However the following is easier to read in my opinion:
```suggestion
                seed.encode(bytes)?;
                query_share.encode(bytes))?;
```

src/vdaf/mastic.rs
```diff
@@ -379,32 +391,485 @@ where
...
+    for MasticPrepareShare<F, SEED_SIZE>
+{
+    fn decode_with_param(
+        prep_state: &MasticPrepareState<F, SEED_SIZE>,
+        bytes: &mut Cursor<&[u8]>,
+    ) -> Result<Self, CodecError> {
+        match prep_state {
+            MasticPrepareState::FirstRound(_, _, verifier_len, requires_joint_rand) => {
+                Ok(MasticPrepareShare::FirstRound(
+                    Seed::<SEED_SIZE>::decode(bytes)?,
```
cjpatton, line 450: I haven't checked, but the compiler probably can infer this type. Not specifying the type is preferable I think. Here and below.
```suggestion
                    Seed::decode(bytes)?,
```

src/vdaf/mastic.rs
```diff
@@ -379,32 +391,485 @@ where
...
+            MasticPrepareState::LaterRound(_, _) => {
+                Ok(MasticPrepareShare::LaterRound(Seed::<SEED_SIZE>::decode(
+                    bytes,
+                )?))
+            }
+        }
+    }
+}
+
+/// Mastic prepare message
```
cjpatton, line 466:
```suggestion
/// Mastic prepare message.
```

src/vdaf/mastic.rs
```diff
@@ -379,32 +391,485 @@ where
...
+///
+/// Result of preprocessing the broadcast messages of all parties during the
```
cjpatton, line 468:
```suggestion
/// Result of preprocessing the broadcast messages of both aggregators during the
```

src/vdaf/mastic.rs
```diff
@@ -379,32 +391,485 @@ where
...
+/// preparation phase.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub enum MasticPrepareMessage<F: FieldElement, const SEED_SIZE: usize> {
+    ///If Szk is being run, all SzkQueryShares have been summed
```
cjpatton, line 472:
```suggestion
    /// If Szk is being run, all SzkQueryShares have been summed
```

src/vdaf/mastic.rs
```diff
@@ -379,32 +391,485 @@ where
...
+        }
+    }
+}
+
+/// Mastic prepare message
+///
+/// Result of preprocessing the broadcast messages of all parties during the
+/// preparation phase.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub enum MasticPrepareMessage<F: FieldElement, const SEED_SIZE: usize> {
```
cjpatton, line 471: Like above, this might just be an alias for `Option<SzkVerifier<F, SEED_SIZE>>`.

src/vdaf/mastic.rs
```diff
@@ -379,32 +391,485 @@ where
...
+where
+    T: Type,
+    P: Xof<SEED_SIZE>,
+{
+    type PrepareState = MasticPrepareState<T::Field, SEED_SIZE>;
+
+    /// The type of messages sent by each aggregator at each round of the Prepare Process.
+    ///
+    /// Decoding takes a [`Self::PrepareState`] as a parameter; this [`Self::PrepareState`] may be
+    /// associated with any aggregator involved in the execution of the VDAF.
```
cjpatton, line 524: No need to include this comment since this is a method required by a trait we're implementing. Here and below.
```suggestion
```

src/vdaf/mastic.rs
```diff
@@ -379,32 +391,485 @@ where
...
+        public_share: &MasticPublicShare<VidpfWeight<T::Field>>,
+        input_share: &MasticInputShare<T::Field, SEED_SIZE>,
+    ) -> Result<
+        (
+            MasticPrepareState<T::Field, SEED_SIZE>,
+            MasticPrepareShare<T::Field, SEED_SIZE>,
+        ),
+        VdafError,
+    > {
+        let mut xof = P::init(
```
cjpatton, line 555: We probably want to give this variable a more distinctive name.

src/vdaf/mastic.rs
```diff
@@ -379,32 +391,485 @@ where
...
+            verify_key,
+            &self.domain_separation_tag(DST_PATH_CHECK_BATCH),
+        );
+        let mut output_shares = Vec::<T::Field>::with_capacity(
+            self.vidpf.weight_parameter * agg_param.level_and_prefixes.prefixes().len(),
```
cjpatton, line 560: Hmm, should `weight_parameter` just be `weight_len` or something?

src/vdaf/mastic.rs
```diff
@@ -379,32 +391,485 @@ where
...
+        );
+        for prefix in agg_param.level_and_prefixes.prefixes() {
+            let mut value_share =
+                self.vidpf
+                    .eval(&input_share.vidpf_key, public_share, prefix, nonce)?;
+            xof.update(&value_share.proof);
+            output_shares.append(&mut value_share.share.0);
```
cjpatton, line 567: We should do tree caching here: If we traverse two paths in the tree that share a path in common, then we only need to traverse that sub-path once. I don't think that's happening here?

cjpatton (commented):
Comments through `prep_init()`. Great start!

src/flp/szk.rs
```diff
@@ -209,18 +208,59 @@ impl<F: FieldElement + Decode, const SEED_SIZE: usize> ParameterizedDecode<(bool
...
+                },
+        )
+    }
+}
+
+impl<F: FieldElement + Decode, const SEED_SIZE: usize> ParameterizedDecode<(bool, usize)>
+    for SzkQueryShare<F, SEED_SIZE>
+{
+    fn decode_with_param(
+        (requires_joint_rand, verifier_len): &(bool, usize),
```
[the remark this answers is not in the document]
hannahdaviscrypto: we occasionally say has_joint_rand, have fixed that.

src/vdaf/mastic.rs
```diff
@@ -379,32 +391,485 @@ where
     }
 }
 
+/// Mastic prepare state
+///
+///  State held by an aggregator between rounds of Mastic. Includes intermediate
```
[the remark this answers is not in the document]
hannahdaviscrypto: No, I do mean preparation

src/vdaf/mastic.rs
hannahdaviscrypto: Do we? It's mostly just passed into submodules that call it xof as well; I think changing the name would obscure that connection.

src/vdaf/mastic.rs
hannahdaviscrypto: Terminology is inherited from the previous instantiation of VIDPF; I'm not sure what the other purposes of it were.

src/vidpf.rs
```diff
@@ -226,6 +228,67 @@ impl<W: VidpfValue, const NONCE_SIZE: usize> Vidpf<W, NONCE_SIZE> {
         })
     }
 
+    /// [`Vidpf::eval_cached`] evaluates the entire `input` and produces a share of the
+    /// input's weight. It reuses computation from previous levels available in the
+    /// cache
+    pub fn eval_cached(
```
cjpatton, line 234: More rusty (here and below)
```suggestion
    pub fn eval_with_cache(
```

src/flp/szk.rs
```diff
@@ -209,17 +207,54 @@ impl<F: FieldElement + Decode, const SEED_SIZE: usize> ParameterizedDecode<(bool
...
+            Ok(SzkQueryShare {
+                joint_rand_part_opt: Some(Seed::<SEED_SIZE>::decode(bytes)?),
+                flp_verifier: decode_fieldvec(*verifier_len, bytes)?,
+            })
+        } else {
+            Ok(SzkQueryShare {
+                joint_rand_part_opt: None,
+                flp_verifier: decode_fieldvec(*verifier_len, bytes)?,
+            })
+        }
```
cjpatton, line 254: Don't repeat code if you don't need to.
```suggestion
        Ok(SzkQueryShare {
            joint_rand_part_opt: (*requires_joint_rand)
                .then(|| Seed::<SEED_SIZE>::decode(bytes))
                .transpose()?,
            flp_verifier: decode_fieldvec(*verifier_len, bytes)?,
        })
```

src/idpf.rs
```diff
@@ -80,6 +80,11 @@ impl IdpfInput {
         }
     }
 
+    /// Create a new empty IDPF input.
+    pub fn empty_input() -> IdpfInput {
```
cjpatton, line 84: More rusty. However since the implementation is a one liner and is pretty readable on its own (`IdfpInput::from_bools(&[])`) I probably would just inline it.
```suggestion
    pub fn new_empty() -> IdpfInput {
```

src/vdaf/mastic.rs
```diff
@@ -379,32 +389,501 @@ where
...
+            verify_key,
+            &self.domain_separation_tag(DST_PATH_CHECK_BATCH),
+        );
+        let mut output_shares = Vec::<T::Field>::with_capacity(
+            self.vidpf.weight_parameter * agg_param.level_and_prefixes.prefixes().len(),
+        );
+        for prefix in agg_param.level_and_prefixes.prefixes() {
+            let mut value_share =
+                self.vidpf
+                    .eval(&input_share.vidpf_key, public_share, prefix, nonce)?;
```
cjpatton, line 555: Use `eval_cached()`.

src/vidpf.rs
```diff
@@ -226,6 +228,67 @@ impl<W: VidpfValue, const NONCE_SIZE: usize> Vidpf<W, NONCE_SIZE> {
...
+        let path = input;
+        match cache_tree.get(input.prefix(0).index.as_bitslice()) {
+            Some(_) => (),
+            None => cache_tree
+                .insert(
+                    IdpfInput::empty_input().index.as_bitslice(),
+                    self.eval_next_cached(key.id, public, input, 0, &state, nonce)?,
+                )
+                .expect("inserting into top of empty tree"),
+        };
```
cjpatton, line 257: Instead of calling `get()` and then `insert()` you should be able to call `insert()` and handle the error if the data already exists.

src/vidpf.rs
```diff
@@ -226,6 +228,67 @@ impl<W: VidpfValue, const NONCE_SIZE: usize> Vidpf<W, NONCE_SIZE> {
...
+
+        let mut cache_node = cache_tree
+            .get_node(IdpfInput::empty_input().index.as_bitslice())
+            .expect("previous match statement ensures initialization");
```
cjpatton, line 261: I don't think this is true unless `IdpfInput::empty_input()` is equal to `prefix`.

src/vidpf.rs
```diff
@@ -226,6 +228,67 @@ impl<W: VidpfValue, const NONCE_SIZE: usize> Vidpf<W, NONCE_SIZE> {
         })
     }
 
+    /// [`Vidpf::eval_cached`] evaluates the entire `input` and produces a share of the
```
cjpatton, line 231: Make sure this function is exercised by unit tests.

src/vidpf.rs
hannahdaviscrypto: I'm not following--the previous statement inserted at the root, and we're now reading from the root. Prefix isn't used?

cjpatton: Ok, I think I understand. In any case, addressing the coment above might make the flow clearer.

src/vidpf.rs
```diff
@@ -226,6 +232,65 @@ impl<W: VidpfValue, const NONCE_SIZE: usize> Vidpf<W, NONCE_SIZE> {
...
+    /// input's weight. It reuses computation from previous levels available in the
+    /// cache
+    pub fn eval_with_cache(
+        &self,
+        key: &VidpfKey,
+        public: &VidpfPublicShare<W>,
+        input: &VidpfInput,
+        cache_tree: &mut BinaryTree<VidpfEvalCache<W>>,
+        nonce: &[u8; NONCE_SIZE],
+    ) -> Result<VidpfValueShare<W>, VidpfError> {
```
cjpatton, line 245: This implementation looks correct, but it's a bit messy in ways that reveal some sharp edges in the `BinaryTree` API. Namely, there are a bunch of points where we unwrap and `Option`. Also, it's awkward that we have to re-construct an index each time we want to insert a node.

I think it would be cleaner to just construct the tree ourselves. Here's what I came up with: this requires some refactoring of the non-public API, but we're of course free to make changes to this as needed.

```rust
let n = input.len();
if n > public.cw.len() {
    return Err(VidpfError::InvalidAttributeLength);
}

if cache_tree.root.is_none() {
    cache_tree.root = Some(Box::new(Node::new(VidpfEvalCache {
        state: VidpfEvalState::init_from_key(key),
        share: W::zero(&self.weight_parameter), // not used
    })));
}

let mut sub_tree = cache_tree.root.as_mut().expect("root was visited");
for (level, bit) in input.iter().enumerate() {
    sub_tree = if !bit {
        if sub_tree.left.is_none() {
            let (new_state, new_share) =
                self.eval_next(key.id, public, input, level, &sub_tree.value.state, nonce)?;
            sub_tree.left = Some(Box::new(Node::new(VidpfEvalCache {
                state: new_state,
                share: new_share,
            })));
        }

        sub_tree.left.as_mut().expect("left child was visited")
    } else {
        if sub_tree.right.is_none() {
            let (new_state, new_share) =
                self.eval_next(key.id, public, input, level, &sub_tree.value.state, nonce)?;
            sub_tree.right = Some(Box::new(Node::new(VidpfEvalCache {
                state: new_state,
                share: new_share,
            })));
        }

        sub_tree.right.as_mut().expect("right child was visited")
    }
}

Ok(sub_tree.value.to_share())
```

src/vidpf.rs
```diff
@@ -46,6 +48,10 @@ pub enum VidpfError {
     #[error("level index out of bounds")]
     IndexLevel,
 
+    /// Error during VIDPF evaluation: writing to the evaluation cache did not succeed.
+    #[error("failed to insert node into an empty cache")]
+    CacheError,
```
cjpatton, line 53: This should be a situation we can always prevent.

src/vdaf/mastic.rs
```diff
@@ -75,6 +81,16 @@ pub struct MasticAggregationParam {
     require_check_flag: bool,
 }
 
+#[cfg(test)]
+impl MasticAggregationParam {
+    fn new(prefixes: Vec<VidpfInput>, require_check_flag: bool) -> Result<Self, VdafError> {
```
cjpatton, line 86: Which check? We call this the "weight check" in the spec.

src/vdaf/mastic.rs
```diff
@@ -379,32 +391,500 @@ where
     }
 }
 
+/// Mastic prepare state.
+///
+/// State held by an aggregator between rounds of Mastic preparation. Includes intermediate
+/// state for Szk verification, the output shares currently being validated, and
+/// parameters of Mastic used for encoding.
```
cjpatton, line 398: This needs rewording: there is just one round of preparation for Mastic.

src/vdaf/mastic.rs
```diff
@@ -379,32 +391,500 @@ where
     }
 }
 
+/// Mastic prepare state.
+///
+/// State held by an aggregator between rounds of Mastic preparation. Includes intermediate
+/// state for Szk verification, the output shares currently being validated, and
```
cjpatton, line 397: Wherever you refer to a struct defined in this crate, you should make it a link. Likewise below.
```suggestion
/// state for [`Szk`] verification, the output shares currently being validated, and
```

src/vdaf/mastic.rs
```diff
@@ -379,32 +391,500 @@ where
...
+}
+
+/// Mastic prepare share.
+///
+/// Broadcast message from an aggregator between rounds of Mastic. Includes the
+/// hashed VIDPF proofs for every prefix in the aggregation parameter, and optionally
+/// the verification message for Szk.
+#[derive(Clone, Debug)]
+pub struct MasticPrepareShare<F: FieldElement, const SEED_SIZE: usize> {
+    /// Batched VIDPF proofs
```
cjpatton, line 415: What does "batched" mean in this context? I think you mean this is the hash of the node proofs and the path checks, but we should make the language match the spec here.

src/vdaf/mastic.rs
```diff
@@ -379,32 +391,500 @@ where
...
+    /// Includes output shares for eventual aggregation.
+    output_shares: MasticOutputShare<F>,
+    /// If SZK verification is being performed, we also store the relevant state for that operation.
+    szk_query_state: SzkQueryState<SEED_SIZE>,
+    verifier_len: Option<usize>,
+}
+
+/// Mastic prepare share.
+///
+/// Broadcast message from an aggregator between rounds of Mastic. Includes the
```
cjpatton, line 410: As above, let "round" refer to a round of preparation, of which there is one.

src/vdaf/mastic.rs
```diff
@@ -379,32 +391,500 @@ where
...
+/// hashed VIDPF proofs for every prefix in the aggregation parameter, and optionally
```
cjpatton, line 411: We call this the "evaluation proof" or "VIDPF evaluation proof" in the spec.

src/vdaf/mastic.rs
```diff
@@ -379,32 +391,500 @@ where
...
+/// the verification message for Szk.
+#[derive(Clone, Debug)]
+pub struct MasticPrepareShare<F: FieldElement, const SEED_SIZE: usize> {
+    /// Batched VIDPF proofs
+    vidpf_proofs: Seed<SEED_SIZE>,
+    /// If Szk verification of the root weight is needed, an SZK verification message.
```
cjpatton, line 417: nit: I would normally leave a space between documented fields of a struct.
```suggestion
    vidpf_proofs: Seed<SEED_SIZE>, 
   
    /// If Szk verification of the root weight is needed, an SZK verification message.
```

src/vdaf/mastic.rs
```diff
@@ -379,32 +391,500 @@ where
...
+                szk_query_share_opt: Some(SzkQueryShare::<F, SEED_SIZE>::decode_with_param(
+                    &(false, verifier_len),
+                    bytes,
+                )?),
+            }),
+            (None, None) => Ok(MasticPrepareShare {
+                vidpf_proofs: Seed::<SEED_SIZE>::decode(bytes)?,
+                szk_query_share_opt: None,
+            }),
+            (Some(_), None) => Err(CodecError::UnexpectedValue),
```
cjpatton, line 467: I don't think this this is parsing error, since we can only match this branch if somehow our prep state is incorrect.

src/vdaf/mastic.rs
```diff
@@ -379,32 +391,500 @@ where
...
+        }
```
cjpatton, line 468: I think this code would be simpler (that is, have less duplication) if we split this big match statement into two steps, one for each optional component:

```rust
let vidpf_proofs = Seed::decode(bytes)?;
let szk_query_share_opt = match { ... };
Ok(Self { vidpf_proofs, szk_query_share_opt })
```

src/vdaf/mastic.rs
```diff
@@ -379,32 +391,500 @@ where
...
+        match (&prep_state.szk_query_state, prep_state.verifier_len) {
+            (Some(_), Some(verifier_len)) => Ok(Some(
+                SzkVerifier::<F, SEED_SIZE>::decode_with_param(&(true, verifier_len), bytes)?,
+            )),
+            (None, Some(verifier_len)) => Ok(Some(SzkVerifier::<F, SEED_SIZE>::decode_with_param(
+                &(false, verifier_len),
+                bytes,
+            )?)),
+            (None, None) => Ok(None),
+            (Some(_), None) => Err(CodecError::UnexpectedValue),
```
cjpatton, line 510: Likewise here: this is an issue with the host's state, not an issue with parsing the peer's message.

src/vdaf/mastic.rs
```diff
@@ -379,32 +391,500 @@ where
...
+
+impl<T, P, const SEED_SIZE: usize> Aggregator<SEED_SIZE, NONCE_SIZE> for Mastic<T, P, SEED_SIZE>
+where
+    T: Type,
+    P: Xof<SEED_SIZE>,
+{
+    type PrepareState = MasticPrepareState<T::Field, SEED_SIZE>;
+
+    type PrepareShare = MasticPrepareShare<T::Field, SEED_SIZE>;
+
```
cjpatton, line 523:
```suggestion
    type PrepareShare = MasticPrepareShare<T::Field, SEED_SIZE>;
```

src/vdaf/mastic.rs
```diff
@@ -379,32 +391,500 @@ where
...
+    type PrepareMessage = MasticPrepareMessage<T::Field, SEED_SIZE>;
+
+    /// Begins the Prepare process with the other Aggregators. The [`Self::PrepareState`] returned
+    /// is passed to [`Self::prepare_next`] to get this aggregator's first-round prepare message.
+    ///
+    /// Implements `Vdaf.prep_init` from [VDAF].
+    ///
+    /// [VDAF]: https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-vdaf-08#section-5.2
```
cjpatton, line 531: No need to copy documentation from the trait. Here and below.
```suggestion
```

src/vdaf/mastic.rs
```diff
@@ -379,32 +391,500 @@ where
...
+    fn prepare_init(
+        &self,
+        verify_key: &[u8; SEED_SIZE],
+        _agg_id: usize,
```
cjpatton, line 535: Hmm, this is odd. I guess this is because the aggregator ID is encoded by the `VidfpKey` in the `input_share`, so we don't need it here?

Is the aggregator ID encoded on the wire, as part of `input_share`?
* If so, this shouldn't be happening. We should instead get the aggregator ID from here.
* If not, where do we get the aggregator ID from? Regardless, we should enforce here that `agg_id` matches the value we already have in the `VidpfKey`.

My preference would be to remove the aggregator ID from the `input_share` entirely. Perhaps change the type of `input_share.vidfp_key` to `Seed`.

cjpatton (commented):
Comments up to `prepare_init()`.

src/vdaf/mastic.rs
hannahdaviscrypto: Yes, it's the weight check.

src/vdaf/mastic.rs
hannahdaviscrypto: By likewise do you mean we should be able to prevent this situation ever happening? I think that's true, but the match statement here still needs to be exhaustive. I've tried using a different error type.

cjpatton: I mean that "unexpected value" is for cases where the peer sent an unexpected value. Here we get this error because the decoding parameter is incorrect. This is our mistake, our peer's.

src/vdaf/mastic.rs
hannahdaviscrypto: It's encoded on the wire within the vidpfkey within the input share, but not separately. Happy to take it out.

hannahdaviscrypto: Yes, it's being encoded within the VidpfKeys. Changing this.

src/vdaf/mastic.rs
```diff
@@ -379,32 +382,477 @@ where
...
+                root_share.as_ref(),
+                input_share.proof_share.clone(),
+                verify_key,
+                nonce,
+            )?)
+        } else {
+            None
+        };
+        let (prep_share, prep_state) =
+            if let Some((szk_query_share, szk_query_state)) = szk_verify_opt {
```
cjpatton, line 582: Suggestion: It would improve readability if we could remove this `if`-`else` block. To do this we would need to compute `szk_query_state_opt`, `verifier_len`, and `szk_query_share_opt` before we get to this point. I believe we might be able to accomplish this with a couple changes:

1. move `verifier_len` into `szk_query_state` (this may also fix the `match` statements above)
2. replace `let szk_verify_opt = ...` above with `let (szk_query_state_opt, szk_query_share_opt) = ...`

src/vdaf/mastic.rs
```diff
@@ -379,32 +382,477 @@ where
...
+        VdafError,
+    > {
+        let id = match agg_id {
+            0 => Ok(VidpfServerId::S0),
+            1 => Ok(VidpfServerId::S1),
+            _ => Err(VdafError::Uncategorized(
+                "Invalid aggregator ID".to_string(),
+            )),
+        }?;
+        let mut xof = P::init(
```
cjpatton, line 531: This includes the onehot proof, but if I'm reading the code correctly, we're missing the path and counter consistency checks: https://jimouris.github.io/draft-mouris-cfrg-mastic/draft-mouris-cfrg-mastic.html#section-3-9

If you prefer we can do this in a future PR.

src/vdaf/mastic.rs
```diff
@@ -379,32 +382,477 @@ where
...
+        let mut inputs_iter = inputs.into_iter();
+        let leader_share = inputs_iter.next().ok_or(VdafError::Uncategorized(
+            "No leader share received".to_string(),
+        ))?;
+        let helper_share = inputs_iter.next().ok_or(VdafError::Uncategorized(
+            "No helper share received".to_string(),
+        ))?;
+        if inputs_iter.next().is_some() {
+            return Err(VdafError::Uncategorized(
+                "more than 2 prepare shares".to_string(),
```
cjpatton, line 627:
```suggestion
                "more than two prepare shares".to_string(),
```

cjpatton, line 627: nit: try to be consistent about punctuation in error messages. Here you begin with a lower case letter, elsewhere you start with an upper case letter.

src/vdaf/mastic.rs
```diff
@@ -379,32 +382,477 @@ where
...
+                if leader_vidpf_proof == helper_vidpf_proof {
+                    Ok(Some(SzkQueryShare::merge_verifiers(
+                        leader_query_share,
+                        helper_query_share,
+                    )))
+                } else {
+                    Err(VdafError::Uncategorized(
+                        "Vidpf proof verification failed".to_string(),
+                    ))
+                }
```
cjpatton, line 651: Cleaner control flow:
```rust
if leader_vidfp_proof != helper_vidpf_proof {
   return Err(...);
}

Ok(Some(SzkQueryShare::merge_verifiers(...)))
```

src/vdaf/mastic.rs
```diff
@@ -379,32 +382,477 @@ where
...
+                    szk_query_share_opt: None,
+                },
+            ) => {
+                if leader_vidpf_proof == helper_vidpf_proof {
+                    Ok(None)
+                } else {
+                    Err(VdafError::Uncategorized(
+                        "Vidpf proof verification failed".to_string(),
+                    ))
+                }
```
cjpatton, line 669: Control flow: we check `leader_vidpf_proof != helper_vidpf_proof` twice in this function; it should be possible to refactor so that we only call it once. This is better because there is one and only one place where this step can fail.

src/vdaf/mastic.rs
```diff
@@ -379,32 +382,477 @@ where
...
+                MasticPrepareShare {
+                    vidpf_proof: helper_vidpf_proof,
+                    szk_query_share_opt: Some(helper_query_share),
+                },
+            ) => {
+                if leader_vidpf_proof == helper_vidpf_proof {
+                    Ok(Some(SzkQueryShare::merge_verifiers(
+                        leader_query_share,
+                        helper_query_share,
+                    )))
```
cjpatton, line 646: Hmm, it looks like we pass the FLP verifier to `prep_next()` below. This is not what the spec says to do; instead we call FLP decide here and delay the joint randomness check to the next step: https://jimouris.github.io/draft-mouris-cfrg-mastic/draft-mouris-cfrg-mastic.html#section-4.2-13

src/vdaf/mastic.rs
```diff
@@ -379,32 +382,418 @@ where
...
+        for prefix in agg_param.level_and_prefixes.prefixes() {
+            let mut value_share = self.vidpf.eval_with_cache(
+                id,
+                &input_share.vidpf_key,
+                public_share,
+                prefix,
+                &mut cache_tree,
+                nonce,
+            )?;
+            eval_proof.update(&value_share.proof);
```
cjpatton, line 534: The eval_proof includes the node proofs, but it also needs the counter and path consistency checks. At this point, let's not try to add this to the current PR. Instead, please note this as a TODO in #947. Also, we may want to add this feature only after we've addressed https://github.com/jimouris/draft-mouris-cfrg-mastic/issues/92, as this should make the implementation easier.

src/vdaf/mastic.rs
```diff
@@ -379,32 +382,418 @@ where
...
+            let root_share = self.vidpf.eval_root_with_cache(
+                id,
+                &input_share.vidpf_key,
+                public_share,
+                &mut cache_tree,
+                nonce,
+            )?;
+            Some(self.szk.query(
+                root_share.as_ref(),
+                input_share.proof_share.clone(),
```
cjpatton, line 548: We should be able to avoid this clone by passing a reference to `query()` rather than the proof share itself. This may be a lot of savings, as the proof share may be 100s of bytes.

src/vdaf/mastic.rs
```diff
@@ -379,32 +382,418 @@ where
...
+            ));
+        };
+        match (
+            leader_share.szk_query_share_opt,
+            helper_share.szk_query_share_opt,
+        ) {
+            (Some(leader_query_share), Some(helper_query_share)) => Ok(self
+                .szk
+                .merge_verifiers(leader_query_share, helper_query_share)?),
+            (None, None) => Ok(None),
```
cjpatton, line 617: I'd suggest noting there that we only check the SZK once.

src/vdaf/mastic.rs
```diff
@@ -379,32 +382,418 @@ where
...
+                    szk_query_state,
+                    verifier_len: _,
+                },
+                Some(ref joint_share),
+            ) => {
+                self.szk.decide(szk_query_state, joint_share)?;
+                Ok(PrepareTransition::Finish(output_shares))
+            }
+        }
+    }
```
cjpatton, line 650: * Avoiding the match statement here will massively simplify control flow.
* Also, a cool Rust tip: You can unpack a struct in the parameters of the function!
```suggestion
    fn prepare_next(
        &self,
        MasticPrepareState {
            output_shares,
            szk_query_state,
            verifier_len: _,
        }: MasticPrepareState<T::Field, SEED_SIZE>,
        input: MasticPrepareMessage<SEED_SIZE>,
    ) -> Result<PrepareTransition<Self, SEED_SIZE, NONCE_SIZE>, VdafError> {
        if let Some(ref joint_share) = input {
            self.szk.decide(szk_query_state, joint_share)?;
        }

        Ok(PrepareTransition::Finish(output_shares))
    }
```

src/vdaf/mastic.rs
```diff
@@ -379,32 +382,418 @@ where
...
+
+    fn aggregate<M: IntoIterator<Item = Self::OutputShare>>(
```
cjpatton, line 652: Convention: Use the concrete type here
```suggestion
    fn aggregate<M: IntoIterator<Item = MasticOutputShare<T::Field>>>(
```

src/vdaf/mastic.rs
```diff
@@ -379,32 +382,418 @@ where
...
+    T: Type,
+    P: Xof<SEED_SIZE>,
+{
+    fn unshard<M: IntoIterator<Item = Self::AggregateShare>>(
+        &self,
+        agg_param: &MasticAggregationParam,
+        agg_shares: M,
+        _num_measurements: usize,
+    ) -> Result<Self::AggregateResult, VdafError> {
+        let n = agg_param.level_and_prefixes.prefixes().len();
```
cjpatton, line 683:
```suggestion
        let num_prefixes = agg_param.level_and_prefixes.prefixes().len();
```

src/vdaf/mastic.rs
```diff
@@ -379,32 +382,418 @@ where
...
+            self.vidpf.weight_parameter
+                * n
+        ]);
+        for agg_share in agg_shares.into_iter() {
+            agg_final.merge(&agg_share)?;
+        }
+        let mut result = Vec::<T::AggregateResult>::with_capacity(n);
+        for i in 0..n {
+            let encoded_result = &agg_final.0
+                [i * self.vidpf.weight_parameter..(i + 1) * self.vidpf.weight_parameter];
```
cjpatton, line 695: Use `chunks()` to get an iterator instead: https://doc.rust-lang.org/std/slice/struct.Chunks.html#example

This is easier to read and allows the compiler to do some more optimization.

src/vdaf/mastic.rs
```diff
@@ -379,32 +382,418 @@ where
...
+            result.push(
+                self.szk
+                    .typ
+                    .decode_result(&self.szk.typ.truncate(encoded_result.to_vec())?[..], 1)?,
```
cjpatton, line 699: You've hard coded the number of measurements to be `1`. In fact, this is supposed to be the number of measurements that have the prefix in common. See https://github.com/jimouris/draft-mouris-cfrg-mastic/blob/482903a99a52cfb61ec06d979ade2fc0512e8c3f/poc/mastic.py#L351.

Rather than address it in this PR, I suggest leaving it as a TODO for #947 and taking care of it later.

src/vdaf/mastic.rs
```diff
@@ -379,32 +382,418 @@ where
...
-        let first_input = VidpfInput::from_bytes(&[15u8, 0u8, 1u8, 4u8][..]);
-
+        let first_input = VidpfInput::from_bytes(&[240u8, 0u8, 1u8, 4u8][..]);
+        let second_input = VidpfInput::from_bytes(&[112u8, 0u8, 1u8, 4u8][..]);
+        let third_input = VidpfInput::from_bytes(&[48u8, 0u8, 1u8, 4u8][..]);
+        let fourth_input = VidpfInput::from_bytes(&[32u8, 0u8, 1u8, 4u8][..]);
+        let fifth_input = VidpfInput::from_bytes(&[0u8, 0u8, 1u8, 4u8][..]);
+        let first_prefix = VidpfInput::from_bools(&[false, false, true]);
+        let second_prefix = VidpfInput::from_bools(&[false]);
+        let third_prefix = VidpfInput::from_bools(&[true]);
```
cjpatton, line 735: Put this in an array and reference the indices of the array below.
```suggestion
        let prefixes = [
            VidpfInput::from_bytes(&[240u8, 0u8, 1u8, 4u8][..]),
            VidpfInput::from_bytes(&[112u8, 0u8, 1u8, 4u8][..]),
            ... and so on
        ];
```

src/vdaf/mastic.rs
```diff
@@ -447,18 +859,125 @@ mod tests {
...
+        let (public_share, input_shares) = mastic.shard(&(first_input, true), &nonce).unwrap();
+        run_vdaf_prepare(
+            &mastic,
+            &verify_key,
+            &first_agg_param,
+            &nonce,
+            public_share,
+            input_shares,
+        )
+        .unwrap();
```
cjpatton, line 928: I think we should use `run_vdaf()` to run the whole thing end-to-end, including aggregation and collection: https://github.com/divviup/libprio-rs/blob/d2fe428cadaf690a56a74f4c682f62326f959f06/src/vdaf.rs#L486

src/vidpf.rs
```diff
@@ -493,43 +531,128 @@ pub struct VidpfPublicShare<W: VidpfValue> {
...
+        for proof in &self.cs {
+            bytes.extend_from_slice(proof);
+        }
+        Ok(())
     }
 
     fn encoded_len(&self) -> Option<usize> {
-        todo!()
+        let control_bits_count = (self.cw.len()) * 2;
+        let mut len = (control_bits_count + 7) / 8 + (self.cw.len()) * 16;
```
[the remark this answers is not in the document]
hannahdaviscrypto: I'm 99% sure I did not write this but it makes sense to me.

src/flp/szk.rs
```diff
@@ -276,24 +240,81 @@ impl<F: FieldElement, const SEED_SIZE: usize> Encode for SzkVerifier<F, SEED_SIZ
...
+impl<F: FieldElement + Decode> ParameterizedDecode<usize> for SzkVerifier<F> {
+    fn decode_with_param(
+        verifier_len: &usize,
+        bytes: &mut Cursor<&[u8]>,
+    ) -> Result<Self, CodecError> {
+        decode_fieldvec(*verifier_len, bytes)
+    }
+}
+
+/// Joint share type for the SZK proof.
```
divergentdave, line 286: I think this needs some further explanation, since this is a novel term.
```suggestion
/// Joint share type for the SZK proof.
///
/// This is produced by [`Szk::merge_verifiers`] as the result of combining two query shares.
/// It contains the joint randomness parts from each aggregator, if applicable. It is consumed by [`Szk::decide`].
```

Or, taking https://github.com/divviup/libprio-rs/pull/1107/files#r1823527498 into account:
```suggestion
/// Joint share type for the SZK proof.
///
/// This is produced by [`Szk::merge_verifiers`] as the result of combining two query shares.
/// It contains the re-computed joint randomness seed, if applicable. It is consumed by [`Szk::decide`].
```

src/flp/szk.rs
```diff
@@ -591,34 +610,56 @@ where
...
         &self,
-        verifier: SzkVerifier<T::Field, SEED_SIZE>,
         query_state: SzkQueryState<SEED_SIZE>,
-    ) -> Result<bool, SzkError> {
-        // Check if underlying FLP proof validates
-        let check_flp_proof = self.typ.decide(&verifier.flp_verifier)?;
-        if !check_flp_proof {
-            return Ok(false);
-        }
+        joint_share: &[Seed<SEED_SIZE>; 2],
```
divergentdave, line 645: I think we could better encapsulate the SZK logic if this took in a `&SzkJointShare` instead. This would let us remove the match block from `Mastic::prepare_next()`.

src/flp/szk.rs
```diff
@@ -591,34 +610,56 @@ where
...
+            *x += y;
+        }
+        if self.typ.decide(&leader_share.flp_verifier)? {
+            match (
+                leader_share.joint_rand_part_opt,
+                helper_share.joint_rand_part_opt,
+            ) {
+                (Some(leader_part), Some(helper_part)) => Ok(Some([leader_part, helper_part])),
+                (None, None) => Ok(None),
+                _ => Err(SzkError::InvalidState(
```
divergentdave, line 632: As mentioned above, this error case is no longer about the prepare state, since this just depends on query shares.

src/vdaf/mastic.rs
```diff
@@ -5,19 +5,22 @@
 //! [draft-mouris-cfrg-mastic-01]: https://www.ietf.org/archive/id/draft-mouris-cfrg-mastic-01.html
 
 use crate::{
+    bt::{BinaryTree, Path},
```
divergentdave, line 8: Can we remove `#![allow(dead_code)]` etc. from the `bt` module now that this is using it?

divergentdave (commented):
Partial review, of the SZK module:

divergentdave:
FYI there are now merge conflicts in the `mastic.rs` file, all in unit tests, following merging #1145. It should be sufficient to drop the incoming chunks, and then add context string parameters back in to the `.shard()` and `.prepare_init()` calls.

src/vdaf/mastic.rs
cjpatton: @divergentdave  there are other discrepancies as well. Our goal for this PR won't be to interop with the spec, but just have an MVP done. We'd follow up with PRs for aligning with the spec. Does this sound OK?

divergentdave: Sure, sounds good.

src/vdaf/mastic.rs
hannahdaviscrypto: I believe we're leaving this to a future PR

src/vidpf.rs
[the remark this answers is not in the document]
hannahdaviscrypto: Actually, no, it's just hardcoded throughout and needs to be 16 for use with both AES and TurboShake. I tried parameterizing it but this would involve rewriting some of the xof code and I think that's best left out of this PR

src/vdaf/mastic.rs
hannahdaviscrypto: No, we're not using the main BT API so it's still dead.

cjpatton: > Can we remove `#![allow(dead_code)]` etc. from the `bt` module now that this is using it?

We ended up not needing some of what @armfazh prepared. We should prune the code we're not using, but we should wait until until Mastic is complete, just in case we end up wanting it.

> Relatedly, I think we ought to make the `bt` module public now, since `BinaryTree` appears in the arguments of the public `Vidpf::eval_with_cache()` method.

I'd suggest we make `eval_with_cache()` `pub(crate)` for now.

src/flp/szk.rs
```diff
@@ -276,24 +240,81 @@ impl<F: FieldElement, const SEED_SIZE: usize> Encode for SzkVerifier<F, SEED_SIZ
...
+    fn decode_with_param(
+        verifier_len: &usize,
+        bytes: &mut Cursor<&[u8]>,
+    ) -> Result<Self, CodecError> {
+        decode_fieldvec(*verifier_len, bytes)
+    }
+}
+
+/// Joint share type for the SZK proof.
+pub type SzkJointShare<const SEED_SIZE: usize> = Option<[Seed<SEED_SIZE>; 2]>;
```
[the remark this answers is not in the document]
cjpatton: I think this is what you mean, but can you confirm?
```rust
pub struct SzkJointShare<const SEED_SIZE: usize>(Option<[Seed<SEED_SIZE>; 2]>);
```

src/vdaf/mastic.rs
```diff
@@ -91,11 +107,11 @@ impl Encode for MasticAggregationParam {
 impl Decode for MasticAggregationParam {
```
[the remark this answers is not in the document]
cjpatton: @hannahdaviscrypto he means have a test that encodes an agg param, decodes it, and confirms that the decoded agg param matches the original.

cjpatton (approved):
There's one substantive change to make that I saw, but the rest is nits.

src/flp/szk.rs
```diff
@@ -547,13 +537,13 @@ where
...
-                        &proof_share_seed_and_blind,
+                        proof_share_seed_and_blind,
                         input_share,
                         nonce,
                     ) {
                         Ok(helper_joint_rand_part) => (
                             self.derive_joint_rand_and_seed(
-                                &leader_joint_rand_part,
+                                leader_joint_rand_part,
                                 &helper_joint_rand_part,
```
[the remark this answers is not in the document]
hannahdaviscrypto: The compiler says no

src/vdaf/mastic.rs
```diff
@@ -380,19 +382,334 @@ where
...
+            return false;
+        };
+
+        // Unpack this agg param and the last one in the list
+        let cur_poplar_agg_param = &cur.level_and_prefixes;
+        let prev_poplar_agg_param = &prev.last().as_ref().unwrap().level_and_prefixes;
+        Poplar1::<P, SEED_SIZE>::is_agg_param_valid(
+            cur_poplar_agg_param,
+            &[prev_poplar_agg_param.clone()],
+        )
```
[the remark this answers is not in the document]
hannahdaviscrypto: Tried that, it doesn't work bc the variables underlying the Poplar1AggregationParam are private.  Should I be making changes to Poplar1 in support of this?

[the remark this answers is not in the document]
divergentdave: You could pass in `std::slice::from_ref(prev_poplar_agg_param)` to avoid the clone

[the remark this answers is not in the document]
cjpatton: > Tried that, it doesn't work bc the variables underlying the Poplar1AggregationParam are private. Should I be making changes to Poplar1 in support of this?

Yes, feel free to change as needed.

src/flp/szk.rs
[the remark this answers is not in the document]
divergentdave: Yes, that's right.

divergentdave (commented):
This is looking good. My only outstanding comment is regarding changing `SzkJointShare` from a type alias to a struct.

FWIW I'd like to merge this before we implement XOF seed size changes from draft-12, to avoid messy merge conflicts.

divergentdave (approved):
Thank you for your contributions, and your persistence!

Issue #1122, opened 2024-10-08
Implement draft-irtf-cfrg-vdaf-13


divergentdave:
Here's a summary of what we'll need to implement, based on the changelog.

- [x] Set up a new release branch for draft-08. Update the README, Dependabot, and cargo-vet configuration (to import from the main branch). (#1130)
- [x] Poplar1 `is_valid()` now checks that the current candidate prefixes are descendents of previous candidate prefixes. (draft-09)
- [x] Add Prio3MultiHotCountVec. (draft-10) (#1123)
- [x] FLP circuit output may be a vector. If so, it is reduced using query randomness. (draft-10) (#1132)
- [x] Poplar1: change handling of an edge case in the first round preparation message. (it's possible we may not need to do anything here) (draft-10)
- [x] Add an application context input, in addition to the nonce from the application. (draft-12) (#1145)
- [x] Prio3: Soundness improvement in places where we use the Schwartz-Zippel lemma, both reducing vectors inside various circuits and reducing the vector output of the circuit. (draft-12) (#1132)
- [x] Prio3: Reuse one seed for the helper's measurement share and proof share (draft-12) (#1167)
- [x] Prio3Sum: Change the circuit to allow specifying a maximum measurement, instead of the number of bits, and remove the use of joint randomness (and change the field). (draft-12) (#1150)
- [x] Prio3Histogram and Prio3MultihotCountVec: Move the last random linear combination of two values from inside the circuit (using joint randomness) to outside the circuit (using query randomness). (draft-12) (#1132)
- [x] Add the application context string to XOF calls inside the IDPF as well. (draft-12) (#1146)
- [x] IDPF: Use XofTurboShake128 on the last level, instead of XofFixedKeyAes128. (draft-12) (#1158)
- [x] Changes to IDPF public share encoding and Poplar1 aggregation parameter encoding. Closes #508. (draft-12) (#1171)
  - [x] We should double check that all `IdpfInput` methods make sense when compared to new draft. (draft-12)
- [x] Double the SEED_SIZE of XofTurboShake128, and allow variable-size seeds in the Xof trait. (draft-12) (#1165)
- [x] XofTurboShake128 and XofFixedKeyAes128: Change the length prefix for the domain separation tag from one to two bytes. (draft-12) (#1146 and #1165)
- [x] All algorithm IDs were reassigned. (draft-12) (#1152)
[...]

PR #1123, opened 2024-10-17
Implement `Prio3MutlihotCountVec`

rozbb:
This implements `Prio3MultiHotCountVec` (introduced in draft 10).

Any feedback is welcome. One thing I need: I have some TODOs in here regarding some constants whose origin I don't know or understand. Any pointers would be helpful.

Partially addresses #1122.

cc @divergentdave @cjpatton

### DIFF
--- a/src/flp/types.rs
+++ b/src/flp/types.rs
@@ -472,15 +472,23 @@ where
     }
 }
 
-/// The multihot counter data type. Each measurement is a set of integers in `[0, length)`, of size
-/// at most `max_weight`, and the aggregate is a histogram counting the number of occurrences of
-/// each integer across all measurements.
+/// The multihot counter data type. Each measurement is a list of booleans of length `length`, with
+/// at most `max_weight` true values, and the aggregate is a histogram counting the number of true
+/// values at each position across all measurements.
 #[derive(PartialEq, Eq)]
 pub struct MultihotCountVec<F, S> {
+    // Parameters
+    /// The number of elements in the list of booleans
     length: usize,
+    /// The max number of permissible `true` values in the list of booleans
     max_weight: usize,
+    /// The size of the chunks fed into our gadget calls
     chunk_length: usize,
+
+    // Calculated from parameters
     gadget_calls: usize,
+    bits_for_weight: usize,
+    offset: usize,
     phantom: PhantomData<(F, S)>,
 }
 
@@ -524,19 +532,21 @@ impl<F: FftFriendlyFieldElement, S: ParallelSumGadget<F, Mul<F>>> MultihotCountV
 
         // The bitlength of a measurement is the number of buckets plus the bitlength of the max
         // weight
-        let meas_length = {
-            let bits_for_weight = ilog2(max_weight) as usize + 1;
-            num_buckets + bits_for_weight
-        };
+        let bits_for_weight = ilog2(max_weight) as usize + 1;
+        let meas_length = num_buckets + bits_for_weight;
 
         // Gadget calls is ⌈meas_length / chunk_length⌉
         let gadget_calls = (meas_length + chunk_length - 1) / chunk_length;
+        // Offset is 2^max_weight.bitlen() - 1 - max_weight
+        let offset = (1 << bits_for_weight) - 1 - max_weight;
 
         Ok(Self {
             length: num_buckets,
             max_weight,
             chunk_length,
             gadget_calls,
+            bits_for_weight,
+            offset,
             phantom: PhantomData,
         })
     }
@@ -549,6 +559,8 @@ impl<F, S> Clone for MultihotCountVec<F, S> {
             length: self.length,
             max_weight: self.max_weight,
             chunk_length: self.chunk_length,
+            bits_for_weight: self.bits_for_weight,
+            offset: self.offset,
             gadget_calls: self.gadget_calls,
             phantom: self.phantom,
         }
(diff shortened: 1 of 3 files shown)


src/flp/types.rs
```diff
@@ -471,6 +472,234 @@ where
...
+
+    fn prove_rand_len(&self) -> usize {
+        // TODO: where does this calculation come from? I copied it from Histogram
+        self.chunk_length * 2
+    }
+
+    fn query_rand_len(&self) -> usize {
+        // TODO: where does this calculation come from? I copied it from Histogram
+        1
+    }
```
divergentdave, line 700: These are similarly derived from the same block of methods, with the gadget details substituted in an simplified.

However, note that we can't yet implement Prio3MultihotCountVec exactly as in draft-12 yet, since this codebase is still on draft-08, and hasn't adopted the `EVAL_OUTPUT_LEN` change yet, so we have to do a bit of backporting. Thus, we'll keep `query_rand_len` equal to 1, for the one evaluation point we need for one gadget polynomial test. Once we get up to draft-10, then this and other VDAFs will need more query randomness values, in order to compute a random linear combination of their circuit outputs.

src/flp/types.rs
```diff
@@ -471,6 +472,234 @@ where
...
+
+    fn proof_len(&self) -> usize {
+        // TODO: where does this calculation come from? I copied it from Histogram
+        (self.chunk_length * 2) + 2 * ((1 + self.gadget_calls).next_power_of_two() - 1) + 1
+    }
+
+    fn verifier_len(&self) -> usize {
+        // TODO: where does this calculation come from? I copied it from Histogram
+        2 + self.chunk_length * 2
+    }
```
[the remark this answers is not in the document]
rozbb: I see, ok. It might be nice to have an automated way to do this in the future, e.g., a default fn impl that does the computation in the spec. I'll mark resolved here

src/flp/types.rs
rozbb: Gotcha I'll make a note

rozbb:
Addressed comments

src/flp/types.rs
```diff
@@ -471,6 +472,232 @@ where
...
+        let multihot_vec: Vec<F> = measurement
+            .iter()
+            // We can unwrap because any Integer type can cast from bool
+            .map(|bit| F::from(F::valid_integer_try_from(*bit as usize).unwrap()))
+            .collect();
+
+        // Encode the measurement weight in binary (actually, the weight plus some offset)
+        let offset_weight_bits = {
+            let bits_for_weight = ilog2(self.max_weight) as usize + 1;
+            let offset = (1 << bits_for_weight) - 1 - self.max_weight;
```
divergentdave, line 594: nit: Since we compute this offset in a few different, I think it might make sense to lift it into the constructor, and store it in the struct. (we can omit it from the Debug impl, as with `gadget_calls`)

src/flp/types.rs
```diff
@@ -957,6 +1186,119 @@ mod tests {
...
+        );
+        // Weight too high. This actually panics because weight + offset cannot fit into a bitvector
+        // of the correct length. In other words, being out-of-range requires the prover to lie
+        // about their weight, which is tested above
+        /*
+        FlpTest::expect_invalid::<NUM_SHARES>(
+            &multihot_instance,
+            &[&[one, one, one], &*encoded_weight_plus_offset(3)].concat(),
+        );
+        */
```
divergentdave, line 1294: nit: We generally try to avoid commented-out code, so I suggest deleting this bit and adjusting the comment above to match, mentioning that there's no way to encode a too-high offset weight in the auxiliary bits.

src/vdaf/prio3.rs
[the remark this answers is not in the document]
rozbb: Sorry, what line does this refer to?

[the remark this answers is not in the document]
divergentdave: See https://github.com/divviup/libprio-rs/blob/45ce6c3576b9923c0463de527f61d79139d7515c/src/vdaf/prio3.rs#L235-L283 for the relevant lines for Prio3Histogram

rozbb:
Ok, ready for another pass. Thank you for your comments so far!

src/flp/types.rs
```diff
@@ -471,6 +472,236 @@ where
...
+        Ok([multihot_vec, offset_weight_bits].concat())
+    }
+
+    fn decode_result(
+        &self,
+        data: &[Self::Field],
+        _num_measurements: usize,
+    ) -> Result<Self::AggregateResult, FlpError> {
+        // The aggregate is the same as the decoded result. Just convert to integers
+        Ok(data.iter().map(|f| F::Integer::from(*f)).collect())
```
divergentdave, line 619: We can use this helper method here. This will additionally do a sanity check of the vector length.
```suggestion
        decode_result_vec(data, self.length)
```

src/flp/types.rs
```diff
@@ -471,6 +472,236 @@ where
...
+
+        // Take a random linear combination of both checks.
+        let out = joint_rand[1] * range_check + (joint_rand[1] * joint_rand[1]) * weight_check;
+        Ok(out)
+    }
+
+    // Truncates the measurement, removing extra data that was necessary for validity (here, the
+    // encoded weight), but not important for aggregation
+    fn truncate(&self, input: Vec<Self::Field>) -> Result<Vec<Self::Field>, FlpError> {
+        // Cut off the encoded weight
```
divergentdave, line 667: This sanity check will keep us from panicking on a too-small input.
```suggestion
    fn truncate(&self, input: Vec<Self::Field>) -> Result<Vec<Self::Field>, FlpError> {
        self.truncate_call_check(&input)?;
        // Cut off the encoded weight
```

divergentdave (approved):

PR #1130, opened 2024-11-04
Setup for new release branch

divergentdave:
This updates Dependabot configuration, updates the README, and bumps the crate version on main to avoid confusion.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "prio"
-version = "0.16.7"
+version = "0.17.0-alpha.0"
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "prio"
-version = "0.16.7"
+version = "0.17.0-alpha.0"
```
branlwyd, line 3: Don't we want to change this version number after branching? Otherwise, the new release branch will have a 0.17 version number.

branlwyd (approved):

Cargo.toml
divergentdave: I already created the release/0.16 branch with the then-current tip of main, so it won't be affected.

PR #1139, opened 2024-11-13
Added `Aggregator::is_agg_param_valid` method

rozbb:
Copied the method verbatim from [here](https://www.ietf.org/archive/id/draft-irtf-cfrg-vdaf-12.html#section-8.2.3), which is effectively unchanged from draft-09.

One point: this doesn't have any unit tests. I've read and re-read the draft and I still don't quite understand what this is supposed to be testing. A unit test that tests the error case we care about would be much appreciated.

### DIFF
--- a/src/vdaf/poplar1.rs
+++ b/src/vdaf/poplar1.rs
@@ -1249,27 +1249,19 @@ impl<P: Xof<SEED_SIZE>, const SEED_SIZE: usize> Aggregator<SEED_SIZE, 16>
 
     /// Validates that no aggregation parameter with the same level as `cur` has been used with the
     /// same input share before. `prev` contains the aggregation parameters used for the same input.
-    ///
-    /// # Panics
-    /// Panics if `prev.len() > 1`
+    /// `prev` MUST be sorted from least to most recently used.
     fn is_agg_param_valid(cur: &Poplar1AggregationParam, prev: &[Poplar1AggregationParam]) -> bool {
-        assert!(
-            prev.len() <= 1,
-            "list of previous aggregation parameters must be of size 0 or 1"
-        );
-
-        // Helper function to determine the prefix of `input` at `last_level`.
-        fn get_ancestor(input: &IdpfInput, this_level: u16, last_level: u16) -> IdpfInput {
-            input.prefix((this_level - last_level) as usize)
-        }
-
-        // Exit early if this is the first time
+        // Exit early if there are no previous aggregation params to compare to, i.e., this is the
+        // first time the input share has been processed
         if prev.is_empty() {
             return true;
         }
 
         // Unpack this agg param and the last one in the list
-        let Poplar1AggregationParam { level, prefixes } = cur;
+        let Poplar1AggregationParam {
+            level: cur_level,
+            prefixes: cur_prefixes,
+        } = cur;
         let Poplar1AggregationParam {
             level: last_level,
             prefixes: last_prefixes,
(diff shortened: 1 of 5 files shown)


src/vdaf/poplar1.rs
```diff
@@ -1245,6 +1246,52 @@ impl<P: Xof<SEED_SIZE>, const SEED_SIZE: usize> Aggregator<SEED_SIZE, 16>
...
+    /// same input share before. `prev` contains the aggregation parameters used for the same input.
+    ///
+    /// # Panics
+    /// Panics if `prev.len() > 1`
+    fn is_agg_param_valid(cur: &Poplar1AggregationParam, prev: &[Poplar1AggregationParam]) -> bool {
+        assert!(
+            prev.len() <= 1,
+            "list of previous aggregation parameters must be of size 0 or 1"
+        );
+
```
divergentdave, line 1260: For heavy hitters applications using Poplar1, we'd want the full list of previous aggregation parameters to be passed in here. Thus, we shouldn't have a cap on the length of this list.

While Poplar1 only cares about the most recent previous aggregation parameter, other VDAFs may care about more, so the contract should be that they all get passed in. In particular, Mastic will look at all of them, and check that exactly one aggregation parameter has the `do_weight_check` flag set.
```suggestion
    fn is_agg_param_valid(cur: &Poplar1AggregationParam, prev: &[Poplar1AggregationParam]) -> bool {
```

src/vdaf/poplar1.rs
```diff
@@ -1245,6 +1246,52 @@ impl<P: Xof<SEED_SIZE>, const SEED_SIZE: usize> Aggregator<SEED_SIZE, 16>
...
+        // Helper function to determine the prefix of `input` at `last_level`.
+        fn get_ancestor(input: &IdpfInput, this_level: u16, last_level: u16) -> IdpfInput {
+            input.prefix((this_level - last_level) as usize)
```
divergentdave, line 1263: This got cleaned up a lot in the spec in cfrg/draft-irtf-cfrg-vdaf#433. The `this_level - last_level` subtraction was only necessary for bit manipulation when we represented all IDPF indices as bigints, so passing that to `prefix()` is incorrect.
```suggestion
        // Helper function to determine the prefix of `input` at `level`.
        fn get_ancestor(input: &IdpfInput, level: u16) -> IdpfInput {
            input.prefix(level as usize + 1)
```

src/vdaf/poplar1.rs
```diff
@@ -1245,6 +1246,52 @@ impl<P: Xof<SEED_SIZE>, const SEED_SIZE: usize> Aggregator<SEED_SIZE, 16>
...
+            prefixes: last_prefixes,
+        } = prev.last().as_ref().unwrap();
+        let last_prefixes_set = BTreeSet::from_iter(last_prefixes);
+
+        // Check that the level increased.
+        if level <= last_level {
+            return false;
+        }
+
+        // Check that prefixes are suffixes of the last level's prefixes.
```
divergentdave, line 1284: Neither will be the suffix of the other generally, but the last parameter's prefixes should be prefixes of the current parameter's prefixes.
```suggestion
        // Check that prefixes are extensions of the last level's prefixes.
```

divergentdave (changes requested):
We have some unit tests in the proof of concept implementation here: https://github.com/cfrg/draft-irtf-cfrg-vdaf/blob/a4874547794818573acd8734874c9784043b1140/poc/tests/test_vdaf_poplar1.py#L187

The most important case to test is that you can't aggregate the same level of the tree twice, since that would double-dip the Beaver multiplication triple generated by the client.

src/vdaf/poplar1.rs
rozbb: Agreed this should be general. This panic condition and check is only for Poplar1, though. Isn't that the appropriate amount of specialization?

divergentdave: I don't think this is correct for Poplar1. Poplar1 expects to be run many times on the same reports, not just twice, in order to carry out the heavy hitters algorithm.

rozbb: I see, ok. Why does the validity check only look at the last elem then?

rozbb: Oh, I think I understand. There's an implicit requirement that the agg params are all sorted in least to most recently used, and that `is_valid()` was called at every new invocation. I've written that assumption in the comments

rozbb:
Alright, should be ready for another pass

src/vdaf/poplar1.rs
```diff
@@ -1979,6 +2013,57 @@ mod tests {
         assert_matches!(err, CodecError::Other(_));
     }
 
+    // Tests Poplar1::is_valid() functionality. This unit tests is translated from
```
divergentdave, line 2016: nit: typo
```suggestion
    // Tests Poplar1::is_valid() functionality. This unit test is translated from
```

divergentdave (approved):

PR #1145, opened 2024-11-19
Add context string to VDAF

rozbb:
Adds the `ctx` string to all the methods that require it. Importantly, this requires updates to the `Aggregator` trait, among others. My methodology was basically to add `ctx` to the args of `domain_separation_tag()` and propagate the changes out. It's a lot.

I kept an empty context string for the benchmarks to minimize the affect on XOF computation.

One note: clippy isn't very happy with how many arguments `Aggregator::prepare_init` and `PingPongTopology::helper_initialized` have. Open to suggestions here.

Completes a subtask of #1122.

### DIFF
--- a/src/vdaf/dummy.rs
+++ b/src/vdaf/dummy.rs
@@ -123,6 +123,7 @@ impl vdaf::Aggregator<0, 16> for Vdaf {
     fn prepare_init(
         &self,
         _verify_key: &[u8; 0],
+        _ctx: &[u8],
         _: usize,
         aggregation_param: &Self::AggregationParam,
         _nonce: &[u8; 16],
@@ -175,6 +176,7 @@ impl vdaf::Aggregator<0, 16> for Vdaf {
 impl vdaf::Client<16> for Vdaf {
     fn shard(
         &self,
+        _ctx: &[u8],
         measurement: &Self::Measurement,
         _nonce: &[u8; 16],
     ) -> Result<(Self::PublicShare, Vec<Self::InputShare>), VdafError> {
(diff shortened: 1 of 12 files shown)


src/vdaf/prio3_test.rs
```diff
@@ -163,11 +176,12 @@ where
...
     P: Xof<SEED_SIZE>,
 {
-    let verify_key = t.verify_key.as_ref().try_into().unwrap();
+    let verify_key = t.verify_key.as_ref();
+    let ctx = t.ctx.as_ref().try_into().unwrap();
 
     let mut all_output_shares = vec![Vec::new(); prio3.num_aggregators()];
     for (test_num, p) in t.prep.iter().enumerate() {
-        let output_shares = check_prep_test_vec(prio3, verify_key, test_num, p);
+        let output_shares = check_prep_test_vec(prio3, ctx, verify_key, test_num, p);
```
divergentdave, line 184: These arguments are in the wrong order. With this fix, the definition of `verify_key` above can be changed back.
```suggestion
        let output_shares = check_prep_test_vec(prio3, verify_key, ctx, test_num, p);
```

src/vdaf/dummy.rs
```diff
@@ -123,6 +123,7 @@ impl vdaf::Aggregator<0, 16> for Vdaf {
     fn prepare_init(
         &self,
         _verify_key: &[u8; 0],
+        _ctx: &[u8],
```
divergentdave, line 126: I think we should just apply `#[allow(clippy::too_many_arguments)]` to this method and the other one with a lint. This is just the most complicated part of the protocol, and I don't think encapsulating subsets of these parameters would be a net benefit.

In this case, we have the following:

- VDAF instance: what we're doing
- Verification key: shared secret
- Application context: what we're doing, more specifically
- Aggregator ID: who am I
- Aggregation parameter: what the Collector said
- Nonce: Identity of the Client's message
- Public share: What the Client said to everybody
- Input share: What the Client said to us

Maybe we could throw the last three references in a struct, but I know that callers would just construct that immediately before calling `prepare_init()`, and then never use it again, so it would be a false simplification. Perhaps we could push more into `Self`, but then we would have to start caring about how `Self` is reused.

I think we should just ignore this instance and call it good. FWIW, we have a couple precedents for this in the IDPF and Poplar1 modules.

rozbb: Ok, this makes sense. Will do

rozbb:
Should be addressed

divergentdave (approved):

PR #1146, opened 2024-11-26
Added `ctx` string to DPF computation

rozbb:
This adds a `ctx` parameter to the IDPF's `gen`, `gen_with_random`, `eval`, and `eval_from_node` methods. Addresses a subtask of #1122. Unfortunately I had to `#[ignore]` even more known-answer tests with this one.

Also: I tried to do more with this PR but failed each time. I wanted to do the other update which uses `XofTuboShake128` for leaf nodes, but could not find a good way of abstracting away the key sizes, the reuse of the XOF for all levels but the leaf, the use of the XOF as an RngCore, etc. I think I'll leave that headache for Chris :)

### DIFF
--- a/src/vdaf/xof.rs
+++ b/src/vdaf/xof.rs
@@ -283,15 +283,24 @@ pub struct XofFixedKeyAes128Key {
 #[cfg(all(feature = "crypto-dependencies", feature = "experimental"))]
 impl XofFixedKeyAes128Key {
     /// Derive the fixed key from the domain separation tag and binder string.
-    pub fn new(dst: &[u8], binder: &[u8]) -> Self {
+    ///
+    /// # Panics
+    /// Panics if `fixed_dst.len() + ctx.len() > u16::MAX`
+    pub fn new(fixed_dst: &'static [u8], ctx: &[u8], binder: &[u8]) -> Self {
         let mut fixed_key_deriver = TurboShake128::from_core(TurboShake128Core::new(
             XOF_FIXED_KEY_AES_128_DOMAIN_SEPARATION,
         ));
+        // We feed the fixed part of dst, followed by ctx. The combined length of both is what we
+        // compute here
         Update::update(
             &mut fixed_key_deriver,
-            &[dst.len().try_into().expect("dst must be at most 255 bytes")],
+            u16::try_from(fixed_dst.len() + ctx.len())
+                .expect("dst must be at most 65536 bytes")
+                .to_le_bytes()
+                .as_slice(),
         );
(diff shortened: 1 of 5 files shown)


src/vdaf/xof.rs
```diff
@@ -330,6 +350,9 @@ pub struct XofFixedKeyAes128 {
     base_block: Block,
 }
 
+// This impl is ONLY used by Mastic. It does not have to comply with the VDAF spec.
+// TODO: try to remove the duplicated code below. init() It's mostly the same as
```
divergentdave, line 354: I filed #1147 for this.

src/vdaf/xof.rs
```diff
@@ -330,6 +350,9 @@ pub struct XofFixedKeyAes128 {
     base_block: Block,
 }
 
+// This impl is ONLY used by Mastic. It does not have to comply with the VDAF spec.
```
divergentdave, line 353: FWIW, I think Mastic will want to move to the `XofFixedKeyAes128Key` interface eventually, once we turn from implementation to optimization. Computation in IDPF/VIDPF-based protocols is almost all XOF invocations, so we'll definitely want to compute the fixed keys just once.

I wouldn't say that this doesn't have to comply with the VDAF spec -- in fact, we make use of this implementation to run common XOF tests, and then we also have a test (`xof_fixed_key_aes128_alternate_apis()`) to confirm that the `Xof` constructor pathway and the key reuse pathway produce the same results. Rather, I'd think about this implementation as a direct reflection of satisfying the `Xof` interface in the spec, and `XofFixedKeyAes128Key` is an implementation-specific optimization, producing the same results. (and in fact, the spec has an implementation note suggesting caching the fixed key)

rozbb: Makes sense. I've toned down the comment a bit. Agreed that having a unified API would be nice. It'll be a little bit of a lift

rozbb:
Should be ready

src/vdaf/xof.rs
```diff
@@ -330,6 +350,10 @@ pub struct XofFixedKeyAes128 {
     base_block: Block,
 }
 
+// This impl is only used by Mastic right now. The XofFixedKeyAes128Key impl is used in cases where
+// the base XOF can be reused with different contexts. This is the case in VDAF IDPF computation.
+// TODO: try to remove the duplicated code below. init() It's mostly the same as
```
divergentdave, line 355: Let's add an issue reference to this
```suggestion
// TODO(#1147): try to remove the duplicated code below. init() It's mostly the same as
```

divergentdave (approved):

Issue #912, opened 2024-01-19
Revisit decoding parameters of associated types


divergentdave:
From discussion with @cjpatton, here's an idea for a possible future trait change. The `Vdaf` and `Aggregator` traits require several associated types for different messages, and most of them are required to implement either `Decode` or `ParameterizedDecode`.

* `Vdaf::AggregationParam: Decode`
* `Vdaf::PublicShare: ParameterizedDecode<Self>`
* `Vdaf::InputShare: for<'a> ParameterizedDecode<(&'a Self, usize)>`
* `Vdaf::OutputShare: for<'a> ParameterizedDecode<(&'a Self, &'a Self::AggregationParam)>`
* `Vdaf::AggregateShare: for<'a> ParameterizedDecode<(&'a Self, &'a Self::AggregationParam)>`
* `Aggregator::PrepareShare: ParameterizedDecode<Self::PrepareState>`
* `Aggregator::PrepareMessage: ParameterizedDecode<Self::PrepareState>`

The `Aggregator::PrepareState` associated type has no `Decode`/`ParameterizedDecode` trait bounds, but it must be deserializable one way or another in order for an aggregator that's implemented as a distributed system to use the VDAF.

Lastly, `Vdaf::Measurement` and `Vdaf::AggregateResult` have no such trait bound, and don't need to traverse network connections or be written to disk.

These trait boundaries were arrived at through iteration, and notably had to be changed in order to implement Poplar1. It would be better if we redesigned this in a more principled way, starting from what set of information participants will have a priori when they need to decode different messages. This would prevent the need for future changes motivated by other VDAFs (or, in the alternative, avoid us having to use partially-self-describing serialization if we don't make trait changes for a new VDAF that needs more information).

The various existing decoding parameters include the instantiated VDAF, the aggregator's ID (as a usize, though now the spec requires that it fit in one byte), the aggregation parameter, and the prepare state. Note that the prepare state implicitly provides the current aggregation round. Here's a first cut of a maximal set of decoding parameters:

| Message | Instantiated VDAF | Aggregator ID | Aggregation parameter | Prepare state or round |
| -- | -- | -- | -- | -- |
| `PublicShare` | ✓ | ✗ | ✗ | ✗ |
| `InputShare` | ✓ | ✓ | ✗ | ✗ |
| `AggregationParam` | ✓ | ✗ | ✗ | ✗ |
| `PrepareState`[^1] | ✓ | ✓ | ✓ | ? |
[...]

PR #1151, opened 2024-12-04
Change decoding parameter for PrepareShare/PrepareMessage

divergentdave:
This is a prototype of a change to the decoding parameters used for `PrepareShare` and `PrepareMessage`. See #912. The handling of round numbers could use more attention, both in terms of documenting the semantics of round numbers for prepare shares and prepare messages, and testing to catch off-by-one errors. Currently anonymous tuples are used for the decoding parameters, but since we have two integer types next to each other now, I think those need to become named structs instead. I think the impact on the ping pong API is worth looking at. Since it does encoding and decoding internally, I had to add some round number arguments. Those, too, need better documentation of their semantics, especially since some transitions may involve two VDAF preparations with different round numbers.

It's not strictly necessary that the whole aggregation parameter be present for decoding. It would be sufficient to project out the level, or whether we are at the last level. That would add another associated type and more complication, though.

One other idea: what if we had two decoding implementations, one using the more convenient `ParameterizedDecode<Self::PrepareState>`, and the other using a more complicated struct assembled from public information? Would other use cases with an extra intermediary be able to use this easily? Which should be required in the type bounds in the `Aggregator` trait?

### DIFF
--- a/src/vdaf.rs
+++ b/src/vdaf.rs
@@ -239,7 +239,10 @@ pub trait Aggregator<const VERIFY_KEY_SIZE: usize, const NONCE_SIZE: usize>: Vda
     ///
     /// Decoding takes a [`Self::PrepareState`] as a parameter; this [`Self::PrepareState`] may be
     /// associated with any aggregator involved in the execution of the VDAF.
-    type PrepareShare: Clone + Debug + ParameterizedDecode<Self::PrepareState> + Encode;
+    type PrepareShare: Clone
+        + Debug
+        + for<'a> ParameterizedDecode<(&'a Self, &'a Self::AggregationParam, usize, u16)>
+        + Encode;
 
     /// Result of preprocessing a round of preparation shares. This is used by all aggregators as an
     /// input to the next round of the Prepare Process.
@@ -250,7 +253,7 @@ pub trait Aggregator<const VERIFY_KEY_SIZE: usize, const NONCE_SIZE: usize>: Vda
         + Debug
         + PartialEq
         + Eq
-        + ParameterizedDecode<Self::PrepareState>
+        + for<'a> ParameterizedDecode<(&'a Self, &'a Self::AggregationParam, u16)>
         + Encode;
 
     /// Begins the Prepare process with the other Aggregators. The [`Self::PrepareState`] returned
@@ -629,8 +632,8 @@ pub mod test_utils {
             .prepare_shares_to_prepare_message(
                 ctx,
                 agg_param,
-                outbound.iter().map(|encoded| {
-                    V::PrepareShare::get_decoded_with_param(&states[0], encoded)
+                outbound.iter().enumerate().map(|(agg_id, encoded)| {
+                    V::PrepareShare::get_decoded_with_param(&(vdaf, agg_param, agg_id, 0), encoded)
                         .expect("failed to decode prep share")
                 }),
             )?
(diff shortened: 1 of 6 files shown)


src/vdaf.rs
```diff
@@ -239,7 +239,10 @@ pub trait Aggregator<const VERIFY_KEY_SIZE: usize, const NONCE_SIZE: usize>: Vda
     ///
     /// Decoding takes a [`Self::PrepareState`] as a parameter; this [`Self::PrepareState`] may be
     /// associated with any aggregator involved in the execution of the VDAF.
-    type PrepareShare: Clone + Debug + ParameterizedDecode<Self::PrepareState> + Encode;
+    type PrepareShare: Clone
+        + Debug
+        + for<'a> ParameterizedDecode<(&'a Self, &'a Self::AggregationParam, usize, u16)>
```
cjpatton, line 244: Type aliases for the aggregator ID (`usize`) and round number (`u16`) might make this clearer.

cjpatton (commented):
Thanks for putting this together, @divergentdave!

Taking a step back for a sec, I think there are two ends of a spectrum for deciding what to decode with:
1. Everything needed to decode is in the `PrepareState`
2. Everything needed to decode is provided by the application (`Vdaf`, round number, aggregator ID, agg param)

This PR does 2. Currently we're near 1, with the exception of the aggregator ID. 1 Is nice because the API is the simplest; 2 is nice because we are able to remove all redundant information in the PrepareState, which may help a bit if you're processing a lot of reports at once. Does that sound about right to you?

After looking at this PR, I think 2 makes the API a little too complex. The ping-pong changes in particular look a little hairy.

In fact, my motivation for this change was simpler: I don't want to have copy information from `Vdaf` to `PreparesState`, like we do here:
https://github.com/divviup/libprio-rs/blob/3c1aeb30c661d373566749a81589fc0a4045f89a/src/vdaf/prio3.rs#L1110

This could be resolve simply by adding `Vdaf` to the decoding parameter.

Incidentally, we also store the agg_id in the `PrepareState` for `Prio3`. Is this necessary? Might we codify this requirement by removing the agg ID from the decoding parameter?



> This is a prototype of a change to the decoding parameters used for `PrepareShare` and `PrepareMessage`. See #912. The handling of round numbers could use more attention, both in terms of documenting the semantics of round numbers for prepare shares and prepare messages, and testing to catch off-by-one errors. Currently anonymous tuples are used for the decoding parameters, but since we have two integer types next to each other now, I think those need to become named structs instead. I think the impact on the ping pong API is worth looking at. Since it does encoding and decoding internally, I had to add some round number arguments. Those, too, need better documentation of their semantics, especially since some transitions may involve two VDAF preparations with different round numbers.

Exposing the round number in the ping pong is likely to cause confusion, for the reason you describe here.

> It's not strictly necessary that the whole aggregation parameter be present for decoding. It would be sufficient to project out the level, or whether we are at the last level. That would add another associated type and more complication, though.

Not worth it IMO.

> One other idea: what if we had two decoding implementations, one using the more convenient `ParameterizedDecode<Self::PrepareState>`, and the other using a more complicated struct assembled from public information? Would other use cases with an extra intermediary be able to use this easily? Which should be required in the type bounds in the `Aggregator` trait?

I'm not sure if there's a use case that would need this functionality. One potential advantage of decoding with public information is that you don't have any redundancy in PrepareState, but so far the overhead is pretty small. It's also possible to encode/decode this state wi/o the redundant information.

src/vdaf.rs
divergentdave: Type aliases would just help with self-documentation, while newtypes would catch confusion.

PR #1153, opened 2024-12-04
Update test vector files

divergentdave:
This updates the test vector files and makes some preparatory changes to related tests. This includes the new files from cfrg/draft-irtf-cfrg-vdaf#526.


src/vdaf/poplar1.rs
```diff
@@ -2391,25 +2406,37 @@ mod tests {
     #[ignore]
     #[test]
     fn test_vec_poplar1_0() {
-        check_test_vec(include_str!("test_vec/08/Poplar1_0.json"));
+        check_test_vec(include_str!("test_vec/13/Poplar1_0.json"));
```
cjpatton, line 2409: FWIW we ended up dropping the version number from the path in the spec repo: https://github.com/cfrg/draft-irtf-cfrg-vdaf/tree/main/test_vec/vdaf

It might help make updating easier if we mirror the directory structure there. Though TBH I think it's fine as is.

cjpatton (approved):

inahga (approved):

PR #1158, opened 2024-12-09
idpf: Use XofTurboShake128 for the leaves

cjpatton:
Partially addresses #1122.

As required by the latest draft, modify IDPF by using XofTurboShake128 for extension/conversion at the leaves of the tree. To accomplish this, encapsulate in the fixed AES key -- now use only for inner nodes -- with a type that has the content appropriate for the level of the tree. For the leaf nodes, this is the context string and nonce.

While at it, rename "binder" to "nonce", as has been done in the latest draft.

### DIFF
--- a/src/idpf.rs
+++ b/src/idpf.rs
@@ -7,7 +7,7 @@ use crate::{
     codec::{CodecError, Decode, Encode, ParameterizedDecode},
     field::{FieldElement, FieldElementExt},
     vdaf::{
-        xof::{Seed, XofFixedKeyAes128Key},
+        xof::{Seed, Xof, XofFixedKeyAes128Key, XofTurboShake128},
         VdafError, VERSION,
     },
 };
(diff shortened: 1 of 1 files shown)


cjpatton (commented):
Alternatives considered:

1. My first thought was to overload `ValueParameter` to indicate the level of the tree being generated. That would require a bound on the type of `ValueParameter` in order to invoke `convert()` or `extend()`, which would mean we could only use `idpf` for `Poplar1`. In practice, this doesn't seem to matter, as `Mastic`, and hopefully future DPF constructions, won't have the special field constraint that `Poplar1` has.

2. I also thought about just plumbing in an `is_leaf` flag to `convert()` and `extend()`. However, we also need the context string and nonce. And since the fixed keys are only used for inner nodes, it made sense to unify all of this configuration into a single type,`XofMode`.

divergentdave (approved):

PR #1165, opened 2024-12-16
Align XofTurboShake128 with VDAF-13

cjpatton:
Partially addresses #1122.

* Bump seed size for XofTurboShake128 to 32 bytes
* Allow variable length seeds for compatibility with IDPF.
* Make the dst length prefix 2 bytes


tests/discrete_gauss.rs
```diff
@@ -20,12 +20,13 @@ use serde::Deserialize;
...
 pub struct DiscreteGaussTestVector {
     #[serde(with = "hex")]
-    seed: [u8; 16],
+    seed: [u8; 32],
     std_num: u128,
     std_denom: u128,
     samples: Vec<i128>,
 }
 
+#[ignore = "need to upgrade the test vector code to VDAF-13"]
```
cjpatton, line 29: @divergentdave do you wanna upgrade your gist and re-generate test vectors real quick?

src/dp/distributions.rs
```diff
@@ -570,7 +570,7 @@ mod tests {
             .unwrap();
 
         // collect that number of samples
-        let mut rng = SeedStreamTurboShake128::from_seed([0u8; 16]);
+        let mut rng = SeedStreamTurboShake128::from_seed([1u8; 32]);
```
cjpatton, line 573: @divergentdave this test seems to be pretty sensitive to the seed's value. Originally I had the seed set to `[0u8; 32]`, but the test fails with that value. In fact the probability was significantly lower than `alpha`.

This code is only called by `empirical_test_gauss`. What's the value of this test?

tests/discrete_gauss.rs
divergentdave: Updated, the test passes with fresh test vectors

src/dp/distributions.rs
divergentdave: It looks like you just got unlucky with the seed, so changing it is fine. I re-ran the test 1000 times with random seeds, and got a 96% success rate. We specify a significance level (`alpha`) of 0.05, so this shows our test is well-calibrated. This function performs a chi-squared statistical test which compares a histogram drawn from our discrete Gaussian sampler, and an expected histogram produced by evaluating the discrete Gaussian's cumulative distribution function on the same buckets. The test tells us that the sampler produces a distribution that is approximately correct. (We don't want to do more trials/samples to get a more powerful statistical test because this unit test will typically be running in debug mode, and its runtime is already pushing it)

tests/discrete_gauss.rs
cjpatton: Done.

src/flp/szk.rs
```diff
@@ -690,18 +690,16 @@ mod tests {
...
-        let leader_seed_opt = if szk_typ.requires_joint_rand() {
-            Some(Seed::<16>::generate().unwrap())
-        } else {
-            None
-        };
+        let prove_rand_seed = Seed::generate().unwrap();
+        let helper_seed = Seed::generate().unwrap();
+        let leader_seed_opt = szk_typ
+            .requires_joint_rand()
+            .then_some(Seed::generate().unwrap());
```
divergentdave, line 702: I feel like Clippy is going to add a lint for this in the future.
```suggestion
            .then(|| Seed::generate().unwrap());
```

src/vdaf/poplar1.rs
```diff
@@ -2147,12 +2147,21 @@ mod tests {
...
+        {
+            output.copy_from_slice(input);
+        }
+
+        let mut poplar_random = [[0u8; 32]; 3];
+        for (input, output) in prep
+            .rand
+            .as_ref()
+            .chunks_exact(32)
+            .zip(poplar_random.iter_mut())
```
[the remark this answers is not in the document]
cjpatton: I split the slice into two chunks, one for IDPF and another for the rest.

src/vdaf/xof.rs
```diff
@@ -206,21 +206,34 @@ impl Debug for SeedStreamAes128 {
 #[derive(Clone, Debug)]
 pub struct XofTurboShake128(TurboShake128);
 
-impl Xof<16> for XofTurboShake128 {
-    type SeedStream = SeedStreamTurboShake128;
-
-    fn init(seed_bytes: &[u8; 16], dst: &[u8]) -> Self {
+impl XofTurboShake128 {
+    pub(crate) fn init_with_variable_length_seed(seed_bytes: &[u8], dst: &[u8]) -> Self {
```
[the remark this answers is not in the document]
cjpatton: Works for me!

divergentdave (approved):

PR #1167, opened 2024-12-17
Combined helper measurement and helper proof seeds

rozbb:
Before draft 12, the Prio3 sharding algorithm used different PRG seeds for the non-leader measurement share and the non-leader proof share. Since draft 12 ([spec diff](https://github.com/cfrg/draft-irtf-cfrg-vdaf/commit/ea39dccccc83988029fd667555aa45f6589195b2#diff-6500cd0dbb0c97511d7a78ab7412d0b0ca834ff9c003ae1ad0a40edb6734e75c)), these shares are generated from the same seed (with different domain separators). This PR updates the Prio3 sharding algorithm, and also has a few cleanups.

Closes a subtask of #1122.


src/vdaf/prio3.rs
```diff
@@ -554,18 +554,11 @@ where
 
     fn random_size(&self) -> usize {
         if self.typ.joint_rand_len() == 0 {
-            // Two seeds per helper for measurement and proof shares, plus one seed for proving
-            // randomness.
-            (usize::from(self.num_aggregators - 1) * 2 + 1) * SEED_SIZE
+            // One seed per (share, proof) pair
```
divergentdave, line 557: I think it would be worthwhile to account for how the different seeds are used in this comment, even if we cancel out the -1 and +1 in the formula.
```suggestion
            // One seed per helper (share, proof) pair, plus one seed for proving randomness.
```

src/vdaf/prio3.rs
```diff
@@ -554,18 +554,11 @@ where
...
         } else {
-            (
-                // Two seeds per helper for measurement and proof shares
-                usize::from(self.num_aggregators - 1) * 2
-                // One seed for proving randomness
-                + 1
-                // One seed per aggregator for joint randomness blinds
-                + usize::from(self.num_aggregators)
-            ) * SEED_SIZE
+            // Two seed per (share, proof) pair
```
divergentdave, line 560: Same as above
```suggestion
            // One seed per helper (share, proof) pair, plus one seed for proving randomness, plus
            // one seed per aggregator for joint randomness blinds.
```

src/vdaf/prio3.rs
```diff
@@ -854,7 +840,7 @@ where
 /// Message sent by the [`Client`] to each [`Aggregator`] during the Sharding phase.
 #[derive(Clone, Debug)]
 pub struct Prio3InputShare<F, const SEED_SIZE: usize> {
-    /// The measurement share.
+    /// The share for measurement and proof
     measurement_share: Share<F, SEED_SIZE>,
 
     /// The proof share.
```
divergentdave, line 846: The serialization format for helper input shares is either one seed (without joint randomness) or two (with joint randomness), so the `Encode` and `Decode` implementations will need updating too.

Another idea: we could change this to the following:

```rust
pub enum Prio3InputShare<F, const SEED_SIZE: usize> {
    Leader {
        measurement_share: Vec<F>,
        proofs_share: Vec<F>,
        joint_rand_blind: Option<Seed<SEED_SIZE>>,
    },
    Helper {
        meas_and_proofs_share: Seed<SEED_SIZE>,
        joint_rand_blind: Option<Seed<SEED_SIZE>>,
    },
}
```

This would eliminate the redundancy in the helper share.

rozbb: The new commit makes this enum. I didn't modify how the encoding works. Let me know if I should change that. There's no enum tag necessary because it's given in the decoding params.

src/vdaf/prio3.rs
```diff
@@ -1335,7 +1464,7 @@ where
 
         Ok((
             Prio3PrepareState {
-                measurement_share: msg.measurement_share.clone(),
+                measurement_share: msg.measurement_share().clone(),
```
divergentdave, line 1467: This method already clones the vector/seed, so we can remove a clone here.
```suggestion
                measurement_share: msg.measurement_share(),
```

divergentdave (approved):
Looks good, save for merge conflicts. Once those are resolved, you might be able to un-ignore some of the Prio3 tests.

rozbb:
Looks like un-ignoring might be nontrivial. Wanna land this and I'll start another PR?

divergentdave:
Sure sounds good. I think the remaining failure will be fixed by #1170.

rozbb:
Actually nvm I got it all working

divergentdave (approved):

PR #462, opened 2023-02-16
IDPF: Change byte order of packed control bits

divergentdave:
This implements the change proposed in cfrg/draft-irtf-cfrg-vdaf#151.

PR #1168, opened 2024-12-18
Align IDPF public share encoding with VDAF-13

cjpatton:
Partially addresses #1122.

To simplify the spec, we've coalesced the seeds and payloads into continuous chunks.

### DIFF
--- a/src/idpf.rs
+++ b/src/idpf.rs
@@ -690,10 +690,16 @@ where
     VL: Encode,
 {
     fn encode(&self, bytes: &mut Vec<u8>) -> Result<(), CodecError> {
-        // Control bits need to be written within each byte in LSB-to-MSB order, and assigned into
-        // bytes in big-endian order. Thus, the first four levels will have their control bits
-        // encoded in the last byte, and the last levels will have their control bits encoded in the
-        // first byte.
+        // draft-irtf-cfrg-vdaf-13, Section 8.2.6.1:
+        //
+        // struct {
+        //     opaque packed_control_bits[packed_len];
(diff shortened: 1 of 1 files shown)


src/idpf.rs
```diff
@@ -690,10 +690,19 @@ where
...
+        //     opaque packed_control_bits[packed_len];
+        //     opaque seed[poplar1.idpf.KEY_SIZE*B];
+        //     Poplar1FieldInner payload_inner[Fi*poplar1.idpf.VALUE_LEN*(B-1)];
+        //     Poplar1FieldLeaf payload_leaf[Fl*poplar1.idpf.VALUE_LEN];
+        // } Poplar1PublicShare;
+        //
+        // Control bits. These need to be written within each byte in LSB-to-MSB order, and
+        // assigned into bytes in big-endian order. Thus, the first four levels will have their
+        // control bits encoded in the last byte, and the last levels will have their control bits
+        // encoded in the first byte.
```
divergentdave, line 705: This comment, and the control bit packing and unpacking, need to be updated as well. We now put the bytes in the opposite order, with the first levels in the first byte.

cjpatton: I'm confused ... is it just the comment that's out of date, or do you think the bit packing/unpacking code is incorrect? FWIW this is code is currently passing the test vectors on my development branch. (I'm working on a PR for updating the agg param encoding.)

divergentdave: Ah nevermind, just the comment is wrong, and it's been out of sync with the implementation since #462.

cjpatton: Removed the comment!

divergentdave (approved):

PR #1171, opened 2024-12-18
Align agg param encoding for Poplar1 with VDAF-13

cjpatton:
Partially addresses #1122.

The encoding of the prefixes has been simplified by avoiding splitting a prefix across a byte boundary. Each prefix is now zero-padded to the next byte.

Also, re-enable test vectors for Poplar1 and IDPF.

### DIFF
--- a/src/vdaf/poplar1.rs
+++ b/src/vdaf/poplar1.rs
@@ -14,7 +14,6 @@ use crate::{
         Aggregatable, Aggregator, Client, Collector, PrepareTransition, Vdaf, VdafError,
     },
 };
-use bitvec::{prelude::Lsb0, vec::BitVec};
 use rand_core::RngCore;
 use std::{
     collections::BTreeSet,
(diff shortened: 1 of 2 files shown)


src/vdaf/poplar1.rs
```diff
@@ -2023,6 +2050,17 @@ mod tests {
...
+        let encoded = [
+            0, 11, // level
+            0, 0, 0, 1, // number of prefixes
+            1, 7, // encoded prefix with trailing bits
+        ];
+        let err = Poplar1AggregationParam::get_decoded(&encoded).unwrap_err();
+        assert_matches!(err, CodecError::UnexpectedValue);
+    }
+
     #[test]
```
cjpatton, line 2064: @divergentdave this test passes, but I'm not sure if it's for the reason we expect. Do you recall what "wrong unused bit" means?

src/vdaf/poplar1.rs
```diff
@@ -776,69 +775,68 @@ impl Poplar1AggregationParam {
...
+                let mut mask = 0;
+                for bit_position in 8 - num_bits..8 {
+                    mask |= 1 << bit_position;
+                }
+                mask ^ 255
+            }
+        };
+        for _ in 0..num_prefixes {
+            bytes.read_exact(&mut buf)?;
+            // Ensure there are no trailing bits in the encoded prefix.
```
cjpatton, line 834: @divergentdave do you think we should specify this case in the draft?

src/vdaf/poplar1.rs
```diff
@@ -776,69 +775,68 @@ impl Poplar1AggregationParam {
...
-        bytes.append(&mut packed);
+
+        // Number of prefixes
+        let num_prefixes =
+            u32::try_from(self.prefixes.len()).map_err(|e| CodecError::Other(e.into()))?;
+        num_prefixes.encode(bytes)?;
+
+        // Encoded prefixes
+        for prefix in self.prefixes.iter() {
+            bytes.append(&mut prefix.to_bytes());
```
cjpatton, line 796: @divergentdave I wonder if it would be useful to have unit tests enforce the bit order within each byte.

src/vdaf/poplar1.rs
divergentdave: This comes from 6f253b3b90f3448f6cbd09ed6c1f254e4e9409e2, so it's similar to the new test above, confirming that unused bits of encoded prefixes are cleared.

src/vdaf/poplar1.rs
divergentdave: I think it would make sense. It can only help catch faulty implementations.

src/vdaf/poplar1.rs
divergentdave: I think the `round_trip_agg_param` test covers this, but a more focused unit test wouldn't hurt.

src/vdaf/poplar1.rs
cjpatton: https://github.com/cfrg/draft-irtf-cfrg-vdaf/issues/529

src/vdaf/poplar1.rs
cjpatton: I agree, let's not bother.

src/vdaf/poplar1.rs
cjpatton: Ok, that's what I thought ... Just wanted to make sure. I've merged the two test cases into one test.

divergentdave (approved):

PR #1172, opened 2024-12-19
Add aggegation initialization method

### LINKED ISSUE #569: Expose method to retrieve `AggregateShare` identity element
It would be useful for users of a VDAF to be able to retrieve the identity element (or "zero") of the VDAF's `AggregateShare` -- for example, this will simplify applications' aggregation logic which might otherwise need to use `Option<AggregateShare>` to represent "an aggregate share, possibly with no aggregated output shares yet."

I suggest a method on the `Vdaf` trait, named something like `aggregate_share_identity`. (This method needs to go on `Vdaf` rather than on the aggregate share type directly, since -- without some fairly significant refactoring -- the aggregate share type may not know how to produce an identity value for all of the VDAFs that use it. In particular, several VDAFs use a variable-length vector of field elements as their aggregate share; each VDAF has a specific expected length for their aggregate shares, but the aggregate share type itself does not know this length.)

We should take this chance to clean up the aggregation API, as well.

divergentdave:
This adds an `Aggregator::aggregate_init()` method to match `vdaf.agg_init()` from the spec. When combined with the existing `Aggregatable` trait, this completes the streaming aggregation API from draft-irtf-cfrg-vdaf-13. A default implementation is provided for `Aggregator::aggregate()`, making use of the streaming API.

Closes #569.

### DIFF
--- a/src/vdaf.rs
+++ b/src/vdaf.rs
@@ -306,7 +306,13 @@ pub trait Aggregator<const VERIFY_KEY_SIZE: usize, const NONCE_SIZE: usize>: Vda
         &self,
         agg_param: &Self::AggregationParam,
         output_shares: M,
-    ) -> Result<Self::AggregateShare, VdafError>;
+    ) -> Result<Self::AggregateShare, VdafError> {
+        let mut share = self.aggregate_init(agg_param);
+        for output_share in output_shares {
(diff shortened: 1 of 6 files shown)


cjpatton (approved):
👍

inahga (approved):

PR #1177, opened 2024-12-27
Various improvements to VIDPF

cjpatton:
Partially addresses #947.

I'm working on completing Mastic. This first PR cleans up several aspects of VIDPF that should make it easier to bring it up to spec.

It's best to review this PR commit-by-commit.

### DIFF
--- a/src/vidpf.rs
+++ b/src/vidpf.rs
@@ -63,12 +63,12 @@ pub trait VidpfValue: IdpfValue + Clone + Debug + PartialEq + ConstantTimeEq {}
 
 #[derive(Clone, Debug)]
 /// An instance of the VIDPF.
-pub struct Vidpf<W: VidpfValue, const NONCE_SIZE: usize> {
+pub struct Vidpf<W: VidpfValue> {
     /// Any parameters required to instantiate a weight value.
     pub(crate) weight_parameter: W::ValueParameter,
 }
 
-impl<W: VidpfValue, const NONCE_SIZE: usize> Vidpf<W, NONCE_SIZE> {
+impl<W: VidpfValue> Vidpf<W> {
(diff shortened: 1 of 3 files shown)


src/vidpf.rs
```diff
@@ -799,6 +776,28 @@ impl<F: FieldElement> ParameterizedDecode<<Self as IdpfValue>::ValueParameter> f
...
+struct VidpfIndex<'a> {
+    bit: Choice,
+    input: &'a VidpfInput,
+    level: u16,
+}
+
+impl VidpfInput {
+    fn index_iter<'a>(
+        &'a self,
+    ) -> Result<Box<dyn Iterator<Item = VidpfIndex<'a>> + 'a>, VidpfError> {
```
cjpatton, line 756: I used `Box<dyn ...>` here simply because I didn't wanna try to figure out the type of the iterator :D

divergentdave: FWIW, since this uses a map combinator with a closure, the only options are to use an opaque `impl Iterator<...>` type or a trait object like this. (because the type of the closure cannot be written down)

src/vidpf.rs
```diff
@@ -581,75 +537,57 @@ impl<W: VidpfValue> ParameterizedDecode<(usize, W::ValueParameter)> for VidpfPub
...
-            proof: VidpfProof::default(),
+            node_proof: VidpfProof::default(),
         }
     }
 }
 
-/// Vidpf evaluation cache
-///
-/// Contains the values produced during input evaluation at a given level.
+/// VIDPF evaluation cache.
```
divergentdave, line 578: nit: This should be updated along with the new name.

divergentdave (approved):

PR #1178, opened 2024-12-29
Make Mastic feature complete

cjpatton:
Stacked on #1177.
Partially addresses #947.

Fix the onehot check, add the payload check, add the prefix counter to $\beta$, and clean up the Mastic and VIDPF APIs a bit.

This PR should make Mastic more ore less feature complete. However, there is still some work remaining before it is spec compliant. We still need to align the XOF computations and add test vectors.

The changes are split into several commits, which should be reviewed one by one.


divergentdave (approved):

cjpatton:
Rebased.

PR #1181, opened 2024-12-31
xof: Pass domain separation tag in parts

cjpatton:
Stacked on #1178.

Modify the `Xof` trait by allowing the user to pass the domain separation tag in parts. This saves us from allocating a `Vec` in many cases.


divergentdave (approved):

cjpatton:
Rebased.

PR #1182, opened 2024-12-31
mastic: Align XOF computations with the spec

cjpatton:
Stacked on #1181.
Partially addresses #947.

The last step is to add the test vectors and fix any bugs they catch. Note that we need to make a few minor adjustments to the Mastic spec and regenerate test vectors.

### DIFF
--- a/src/flp/szk.rs
+++ b/src/flp/szk.rs
@@ -16,7 +16,10 @@ use crate::{
     field::{decode_fieldvec, encode_fieldvec, FieldElement},
     flp::{FlpError, Type},
     prng::{Prng, PrngError},
-    vdaf::xof::{IntoFieldVec, Seed, Xof, XofTurboShake128},
+    vdaf::{
+        mastic::{self, USAGE_PROOF_SHARE},
+        xof::{IntoFieldVec, Seed, Xof, XofTurboShake128},
+    },
 };
 use std::borrow::Cow;
 use std::ops::BitAnd;
@@ -24,14 +27,6 @@ use std::{io::Cursor, marker::PhantomData};
 use subtle::{Choice, ConstantTimeEq};
 
 // Domain separation tags
-const DST_PROVE_RANDOMNESS: u16 = 0;
-const DST_PROOF_SHARE: u16 = 1;
-const DST_QUERY_RANDOMNESS: u16 = 2;
-const DST_JOINT_RAND_SEED: u16 = 3;
-const DST_JOINT_RAND_PART: u16 = 4;
-const DST_JOINT_RANDOMNESS: u16 = 5;
-
-const MASTIC_VERSION: u8 = 0;
 
 /// Errors propagated by methods in this module.
 #[derive(Debug, thiserror::Error)]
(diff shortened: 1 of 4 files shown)


src/flp/szk.rs
```diff
@@ -308,14 +303,13 @@ where
 {
     /// The Type representing the specific FLP system used to prove validity of an input.
     pub(crate) typ: T,
-    algorithm_id: u32,
```
cjpatton, line 311: We currently don't bind XOF computations to this, like we do in Prio3. See https://github.com/jimouris/draft-mouris-cfrg-mastic/issues/99.

src/flp/szk.rs
```diff
@@ -498,12 +501,14 @@ where
 
     pub(crate) fn query(
         &self,
+        ctx: &[u8],
+        level: u16, // level of the prefix tree
```
cjpatton, line 505: This is per the spec. It's slightly awkward since a detail of VIDPF is bleeding into SZK. However, I think this feature of the spec is useful because it ensures independence of query randomness at different levels of the tree. This isn't strictly necessary given that the SZK is queried once, but is useful for defense in depth in case an aggregator inadvertently allows multiple queries.

src/flp/szk.rs
cjpatton, line 1: Details of Mastic are bleeding into this code. My plan is to move this module to a sub-module of `vdaf::mastic`.

src/vdaf/mastic.rs
```diff
@@ -19,19 +19,38 @@ use crate::{
...
+
 const NONCE_SIZE: usize = 16;
 
+pub(crate) const USAGE_PROVE_RAND: u8 = 0;
+pub(crate) const USAGE_PROOF_SHARE: u8 = 1;
+pub(crate) const USAGE_QUERY_RAND: u8 = 2;
+pub(crate) const USAGE_JOINT_RAND_SEED: u8 = 3;
+pub(crate) const USAGE_JOINT_RAND_PART: u8 = 4;
+pub(crate) const USAGE_JOINT_RAND: u8 = 5;
+pub(crate) const USAGE_ONEHOT_PROOF_HASH: u8 = 7;
```
cjpatton, line 44: `6` is used to compute `USAGE_ONEHOT_PROOF_INIT`, which we hardcode below.

src/vdaf/mastic.rs
```diff
@@ -286,29 +305,44 @@ where
 {
     fn shard_with_random(
         &self,
-        alpha: &VidpfInput,
-        weight: &T::Measurement,
+        ctx: &[u8],
+        (alpha, weight): &(VidpfInput, T::Measurement),
```
cjpatton, line 309: this change prevents a clippy warning (too many arguments).

src/vdaf/mastic.rs
```diff
@@ -580,7 +607,8 @@ where
         };
 
         let eval_proof = {
-            let mut eval_proof_xof = P::init(&[0; SEED_SIZE], &[]);
+            // TODO spec: Use a zero seed.
+            let mut eval_proof_xof = P::init(&[0; SEED_SIZE], &[&dst_usage(USAGE_EVAL_PROOF), ctx]);
```
cjpatton, line 611: Or we could use an empty seed per the current spec. However that would require making `P` concrete (`XofTurboShake128`)

src/vdaf/mastic.rs
divergentdave: `ONEHOT_PROOF_INIT` lives in `src/vidpf.rs` currently, should it be moved to this file as well?

divergentdave (approved):

cjpatton:
Rebased.

PR #1186, opened 2025-01-06
Replace use of decode_items() with decode_fieldvec()

divergentdave:
This is a follow-up to #1167 to optimize Prio3 input share decoding.


cjpatton (approved):
Nice!

inahga (approved):

PR #1185, opened 2025-01-06
mastic: Add encoding for prep state

cjpatton:
Partially addresses #947.

PR #1187, opened 2025-01-07
Mastic clean up

cjpatton:
Stacked on #1185.
Partially addresses #947.

Various clean up things in the wake of finishing Mastic. Review commit-by-commit.

### DIFF
--- a/src/vidpf.rs
+++ b/src/vidpf.rs
@@ -528,11 +528,17 @@ impl<W: VidpfValue> Encode for VidpfPublicShare<W> {
     }
 
     fn encoded_len(&self) -> Option<usize> {
-        let control_bits_count = self.cw.len() * 2;
-        let mut len = (control_bits_count + 7) / 8 + self.cw.len() * 48;
-        for cw in self.cw.iter() {
-            len += cw.weight.encoded_len()?;
-        }
+        // We assume the weight has the same length at each level of the tree.
+        let weight_parameter = self
+            .cw
+            .first()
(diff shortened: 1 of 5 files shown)


src/vidpf.rs
```diff
@@ -528,11 +528,17 @@ impl<W: VidpfValue> Encode for VidpfPublicShare<W> {
     }
 
     fn encoded_len(&self) -> Option<usize> {
-        let control_bits_count = self.cw.len() * 2;
-        let mut len = (control_bits_count + 7) / 8 + self.cw.len() * 48;
-        for cw in self.cw.iter() {
-            len += cw.weight.encoded_len()?;
-        }
+        // We assume the weight has the same length at each level of the tree.
+        let weight_parameter = self
```
divergentdave, line 530: nit: This name suggests a relationship to the `ValueParameter` associated type, which counts field elements, but this counts bytes instead. How about `weight_len` or `weight_size`?

divergentdave (approved):

src/vidpf.rs
cjpatton: Done. We might also consider doing away with the `WeightParameter` associated type in the future. Want to do that here, or can it wait?

divergentdave: Are you thinking of hardcoding `ValueParameter` to be a `usize`? I think the flexibility still might be worth it. For example, you may want to generate a tensor of field elements instead of just a vector. It could also be useful for switching between two field sizes, like in Poplar1, but without relying on the aggregation parameter as much.

PR #1190, opened 2025-01-09
Remove `getrandom()`, clean up error handling

### LINKED ISSUE #955: Do we still need `getrandom`?

cjpatton:
Closes #955.

Use `thread_rng().gen::<[u8; SEED_SIZE]>()` to generate seeds rather than `getrandom()`. Since this method is infallible, we no longer need to propagate errors from `getrandom()`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -358,10 +358,8 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c4567c8db10ae91089c99af84c68c38da3ec2f087c3f82960bcdbf3656b6f4d7"
 dependencies = [
  "cfg-if",
- "js-sys",
  "libc",
  "wasi",
- "wasm-bindgen",
 ]
 
 [[package]]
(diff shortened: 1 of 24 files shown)


divergentdave (approved):
FWIW, `gen()` and `fill()` are technically infallible because they're unwrapping a Result, but I think that we can ignore such failures on modern mainstream systems. Note that `rand::thread_rng()` also unwraps a Result during per-thread initialization -- this Result comes from `getrandom`.

cjpatton:
Rebased.

PR #1195, opened 2025-01-13
mastic: Add VDAF algorithm ID to domain separation

cjpatton:
Stacked on #1189.
Partially addresses #947.

See https://github.com/jimouris/draft-mouris-cfrg-mastic/pull/105.


src/flp/szk.rs
```diff
@@ -320,9 +321,10 @@ where
 {
     /// Construct an instance of this sharedZK proof system with the underlying
     /// FLP.
-    pub fn new(typ: T) -> Self {
+    pub fn new(typ: T, algorithm_id: u32) -> Self {
         Self {
             typ,
+            id: algorithm_id.to_le_bytes(),
```
divergentdave, line 327: Note that Prio3 and Poplar1 are using big-endian encoding for the algorithm ID in their binder strings. It might be nice to change the spec, so that the byte orders match.

src/vdaf/mastic.rs
```diff
@@ -86,9 +87,9 @@ where
     /// Creates a new instance of Mastic, with a specific attribute length and weight type.
     pub fn new(algorithm_id: u32, typ: T, bits: usize) -> Result<Self, VdafError> {
         let vidpf = Vidpf::new(bits, typ.input_len() + 1)?;
-        let szk = Szk::new(typ);
+        let szk = Szk::new(typ, algorithm_id);
         Ok(Self {
-            algorithm_id,
+            id: algorithm_id.to_le_bytes(),
```
divergentdave, line 92: Same here

divergentdave (approved):

src/flp/szk.rs
cjpatton: Good call: https://github.com/jimouris/draft-mouris-cfrg-mastic/pull/112

I'll add this to the top of the stack.

cjpatton:
Rebased after merging #1189.

PR #1196, opened 2025-01-13
mastic: Use empty seeds instead of zero seeds

cjpatton:
Stacked on #1195.
Partially addresses #947.

See https://github.com/jimouris/draft-mouris-cfrg-mastic/pull/106.

To allow using empty seeds, we hardcode the XOF to `XofTurboShake128`. We also considered API changes to accommodate different instantiations of the XOF, but opted to hardcode in order to nudge users towards the standardized variant. Note that this also means that Mastic requires the "crypto-dependencies" feature.


cjpatton:
Rebased on main.

divergentdave (approved):

PR #1198, opened 2025-01-13
mastic: Make the evaluation proof pseudorandom

cjpatton:
Stacked on #1197.
Partially addresses #947.

See https://github.com/jimouris/draft-mouris-cfrg-mastic/pull/108.

PR #1199, opened 2025-01-13
mastic: Let beta be the counter plus the encoded, truncated weight

### LINKED ISSUE #947: Implement the Mastic VDAF

cjpatton:
Stacked on #1198.
Closes #947. (Just a bit of clean up to follow.)

See https://github.com/jimouris/draft-mouris-cfrg-mastic/pull/109.

This required shifting the SZK abstraction boundary to include FLP input shares as well as the proof shares.


cjpatton:
We ultimately decided against this change to the spec: https://github.com/jimouris/draft-mouris-cfrg-mastic/issues/78#issuecomment-2593627448

PR #1205, opened 2025-01-18
Add methods for adding/subtracting vectors

### LINKED ISSUE #301: Implement std::ops::Add for vectors of field elements
There are many places in the codebase where we zip or index through two vectors, adding one to the other. We could write blanket implementations of `Add` and `AddAssign` on vectors of field elements to get rid of some boilerplate. These could also check that the lengths of the addend vectors are equal, and panic otherwise to be defensive.

See: https://github.com/divviup/libprio-rs/pull/300#discussion_r966213248

cjpatton:
Closes #301: see https://github.com/divviup/libprio-rs/issues/301#issuecomment-2599407256.

Add a method `add_assign_vector` that replaces the following zip pattern:

```
for (x, y) in a.iter_mut().zip(y) {
    *x += *y;
}
```

The method also panics if `b` is shorter than `a`.

Replace the zip pattern with this method wherever we add up vectors. This helps us ensure that we're not accidentally adding a vector that's shorter than the other.

Likewise for subtracting vectors.

### DIFF
--- a/src/field.rs
+++ b/src/field.rs
@@ -867,9 +867,7 @@ pub(crate) fn merge_vector<F: FieldElement>(
     if accumulator.len() != other_vector.len() {
         return Err(FieldError::InputSizeMismatch);
     }
-    for (a, o) in accumulator.iter_mut().zip(other_vector.iter()) {
-        *a += *o;
-    }
+    add_assign_vector(accumulator, other_vector.iter().copied());
 
     Ok(())
 }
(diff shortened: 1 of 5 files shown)


jcjones (commented):
I've gone through these changes and they look fine and correct to me.

divergentdave (approved):

src/field.rs
```diff
@@ -886,15 +884,37 @@ pub(crate) fn split_vector<F: FieldElement>(inp: &[F], num_shares: usize) -> Vec
...
     outp
 }
 
+pub(crate) fn sub_assign_vector<F: FieldElement>(a: &mut [F], b: impl IntoIterator<Item = F>) {
+    let mut count = 0;
+    for (x, y) in a.iter_mut().zip(b) {
+        *x -= y;
+        count += 1;
+    }
+    assert_eq!(a.len(), count);
```
jcjones, line 900: If we wanted to be real clever, we could add a debug assertion that a.len == b.len using `#[cfg(debug_assertions)]` to conditionally compile it.

divergentdave: At this point, we only know `b` produces an iterator, so we'd have to do so in a more specialized method that takes two slices

PR #1208, opened 2025-01-23
mastic: Implement the heavy hitters sketch from ia.cr/2024/666

cjpatton:
This commit adds an initial implementation of the sketching technique from ia.cr/2024/666 for approximate heavy hitters. This main advantage is that the computation requires just one round of aggregation rather than many.

### DIFF
--- a/src/vdaf/mastic/hh_sketch.rs
+++ b/src/vdaf/mastic/hh_sketch.rs
@@ -1,0 +1,359 @@
+// SPDX-License-Identifier: MPL-2.0
+
+//! The approximate heavy hitters protocol of [[RZCGP24], Section 5].
+//!
+//! [RZCGP24]: https://eprint.iacr.org/2024/666
+
+use crate::{
+    field::{Field64, FieldElement, FieldElementWithInteger},
+    flp::{gadgets::PolyEval, Flp, FlpError, Gadget, Type},
+    fp::log2,
+    vdaf::VdafError,
+    vidpf::VidpfInput,
+};
+
+use sha3::{Digest, Sha3_256};
+use std::iter::once;
+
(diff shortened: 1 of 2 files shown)


src/vdaf/mastic/hh_sketch.rs
```diff
@@ -0,0 +1,357 @@
...
+                panic!("non-heavy hitter output by Collector: {heavy_hitter:?}");
+            }
+        }
+
+        // Make sure all heavy hitters are accounted for.
+        let num_heavy_hitters = count
+            .iter()
+            .filter(|(_weight, count)| **count >= mastic.szk.typ.threshold)
+            .count();
+        assert_eq!(heavy_hitters.len(), num_heavy_hitters);
```
cjpatton, line 355: This test may flake, since bucket collisions can garble heavy hitters.

src/vdaf/mastic/hh_sketch.rs
```diff
@@ -0,0 +1,359 @@
...
+        _num_shares: usize,
+    ) -> Result<Vec<Field64>, FlpError> {
+        self.valid_call_check(input, joint_rand)?;
+
+        // Check that each input is either a `1` or a `-1`.
+        input
+            .iter()
+            .copied()
+            .map(|x| gadgets[0].call(&[x]))
+            .collect()
```
cjpatton, line 208: A typical input length will probably be something like `num_bits == 256`. I wonder if we'd be better off (in terms of proof size and computation) with `ParallelSum` here.

That would mean we would need joint randomness and thus a larger field (`Field128`). I'm not sure if a larger field would pay off.

cjpatton:
Hi @divergentdave, @tgeoghegan - I'm going to close this PR and move the code to a personal repo.

Issue #1222, opened 2025-02-18
Move to `rand_distr::Zipf` since `zipf` crate is deprecated


jcjones:
As noted in https://github.com/divviup/libprio-rs/pull/1217 and https://crates.io/crates/zipf, the `zipf` crate "is deprecated. Prefer [rand_distr::Zipf](https://docs.rs/rand_distr/latest/rand_distr/struct.Zipf.html)."

jcjones:
Completed in https://github.com/divviup/libprio-rs/pull/1223

PR #1223, opened 2025-02-18
Move to rand_distr::Zipf since zipf crate is deprecated

jcjones:
Fixes issue #1222.
Also supersedes PR https://github.com/divviup/libprio-rs/pull/1217.

### DIFF
--- a/src/idpf.rs
+++ b/src/idpf.rs
@@ -1036,7 +1036,7 @@ enum XofMode<'a> {
 pub mod test_utils {
     use super::*;
 
-    use zipf::ZipfDistribution;
+    use rand_distr::Zipf;
 
     /// Generate a set of IDPF inputs with the given bit length `bits`. They are sampled according
     /// to the Zipf distribution with parameters `zipf_support` and `zipf_exponent`. Return the
@@ -1064,9 +1064,9 @@ pub mod test_utils {
 
         // Sample a number of inputs according to the Zipf distribution.
         let mut samples = Vec::with_capacity(measurement_count);
-        let zipf = ZipfDistribution::new(zipf_support, zipf_exponent).unwrap();
+        let zipf = Zipf::new(zipf_support as u64, zipf_exponent).unwrap();
         for _ in 0..measurement_count {
-            samples.push(inputs[zipf.sample(rng) - 1].clone());
+            samples.push(inputs[zipf.sample(rng) as usize - 1].clone());
         }
 
         // Compute the prefix tree for the desired threshold.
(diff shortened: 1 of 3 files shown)


src/idpf.rs
```diff
@@ -1064,9 +1064,9 @@ pub mod test_utils {
 
         // Sample a number of inputs according to the Zipf distribution.
         let mut samples = Vec::with_capacity(measurement_count);
-        let zipf = ZipfDistribution::new(zipf_support, zipf_exponent).unwrap();
+        let zipf = Zipf::new(zipf_support as u64, zipf_exponent).unwrap();
         for _ in 0..measurement_count {
-            samples.push(inputs[zipf.sample(rng) - 1].clone());
+            samples.push(inputs[zipf.sample(rng) as usize - 1].clone());
```
divergentdave, line 1069: Following the logic here, it looks like we should be rounding. https://github.com/foyer-rs/foyer/pull/854#discussion_r1940399740
```suggestion
            samples.push(inputs[zipf.sample(rng).round() as usize - 1].clone());
```

jcjones: Not real sure how to write a test for this case, given the use of this within a test framework. Any ideas?

divergentdave: Since we're dealing with statistical distributions, testing would be computationally intensive. The first route is to draw a large number of samples using both implementations, compare the results, and do a chi-squared test or Kolmogorov-Smirnov test. The other route is to draw a large number of samples from just the `rand-dist` implementation, and check if the mean is close enough to what we get from closed-form formulas for the distribution's mean. You could also draw many large samples, and then the sample means would be normally distributed, by the central limit theorem. Then you could do a statistical test to see if the sample means are coming from a normal distribution with the correct mean or not.

This is the sort of thing I'd only want to run in release mode, so it wouldn't be a good fit for the test suite. I'm kinda concerned about whether my original suggestion was correct, because the `rand-distr` docs do not mention rounding. https://docs.rs/rand_distr/latest/rand_distr/struct.Zipf.html#integer-vs-fp-return-type I might do a one-off experiment to see which gets us the correct results.

divergentdave (approved):

src/idpf.rs
jcjones: That matches the documentation for https://docs.rs/rand_distr/latest/rand_distr/struct.Zipf.html#method.new, which was why I first truncated -- but this morning I didn't understand how we were using this distribution. Now I more or less get it.

Truncation should be more efficient; do we care?

divergentdave: No, I don't think it matters much either way. The result will be the same, and any performance difference is swamped by how expensive the Zipf distribution sampler is.

Issue #82, opened 2021-09-22
Make `field::random_vector` a method on `FieldElement`


cjpatton:
As discussed during review of #81.

cjpatton:
This has been done for a while.

PR #1224, opened 2025-03-03
Make random_vector() an associated method

divergentdave:
This makes `random_vector()` an associated method of the `FieldElement` trait, instead of a free function. Implements #82.


jcjones (approved):
Nice cleanup.

Obviously I'm still working on my ability to tell what's what from diffs - I had to go actually try removing `#[cfg(feature = "multithreaded")]` from the current `main` of gadgets.rs:465 to see that it is covering how `FieldElement` is unused in non-`multithreaded` tests right now, a problem no longer present in the later version.

Anyway, good cleanup.

tgeoghegan (approved):

PR #1238, opened 2025-03-06
Upgrade to rand 0.9

### LINKED ISSUE #1216: Upgrade to rand 0.9
Version 0.9.0 of the rand crate was recently released. Since we expose `rand_core::RngCore`, via the `Xof` trait, this upgrade will be a breaking change for us. Since I expect we'll be using the next release of this crate for a while, it would probably be good to do this upgrade before we cut a release.

There is currently work ongoing to remove some additional dependencies that were added in 0.9.0, most notably `zerocopy`. We should wait for a release that includes these changes before we dive into dependency audits.

The `zipf` crate, used in tests, has not been upgraded, and it is no longer maintained. We can replace this with `rand_distr::Zipf`, which is maintained, and almost identical except some integer types are replaced with floating point types.

The `num-bigint` crate, used in differential privacy samplers, has not yet been upgraded, but I expect it will be in the coming weeks.

Beyond breaking changes, other ecosystem crates, and audits, we may need to address reproducibility issues, especially related to the differential privacy samplers. See https://github.com/rust-random/rand/blob/master/CHANGELOG.md#reproducibility-breaking-changes.

I also noticed that the new version of rand-distr has added support for Kolmogorov-Smirnov tests, so we could remove our dependency on `statrs` (and, transitively, `nalgebra` and `simba`).

divergentdave:
This closes #1216. Arbitrary precision integer sampling routines from num-bigint were vendored to allow us to proceed without a new version of that crate. All new versions of dependencies have been audited. Note that version 0.9.1 of the `rand` crate and 0.9.3 of the `rand_core` crate do not depend on `zerocopy`, unlike earlier 0.9 versions.

There is still a transitive dependency on rand 0.8 via `statrs` and `nalgebra`, but we can take care of that separately, as it only affects some tests.

### DIFF
--- a/src/dp/rand_bigint.rs
+++ b/src/dp/rand_bigint.rs
@@ -23,23 +23,25 @@
 //!
 //! These were vendored from num-bigint. Unused functionality has been deleted, trait
 //! implementations have been transformed into free functions, use of the private
-//! `biguint_from_vec()` function has been replaced, use of the `SampleBorrow`
-//! trait has been removed, and documentation has been updated.
+//! `biguint_from_vec()` function has been replaced, use of the `SampleBorrow` trait has been
+//! removed, method names and signatures have been updated to align with rand 0.9, and documentation
+//! has been updated.
 
 use num_bigint::BigUint;
 use num_integer::Integer;
 use num_traits::{ToPrimitive, Zero};
+use rand::distr::uniform::Error;
 use rand::Rng;
 
 /// Use [`Rng::fill`] to generate random bits.
 ///
 /// If `rem` is greater than zero, than only the lowest `rem` bits of the last u32 are filled with
 /// random data.
-fn gen_bits<R>(rng: &mut R, data: &mut [u32], rem: u64)
+fn random_bits<R>(rng: &mut R, data: &mut [u32], rem: u64)
 where
     R: Rng + ?Sized,
 {
-    // `fill` is faster than many `gen::<u32>` calls
+    // `fill` is faster than many `random::<u32>` calls
     rng.fill(data);
     if rem > 0 {
         let last = data.len() - 1;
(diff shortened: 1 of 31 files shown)


divergentdave:
I updated this to vendor the BigUint uniform sampler and bump crates to the newest versions. The `zerocopy` dependency is no longer necessary with rand 0.9.1 and rand-core 0.9.3.

cjpatton:
How close is this to landing? This would be nice to have :)

divergentdave:
I mainly have to vet ~3 more crate diffs, then do a bit of cleanup.

divergentdave:
This is finally ready for review.

LICENSE-MIT
```diff
@@ -0,0 +1,25 @@
+Copyright (c) 2014 The Rust Project Developers
```
jcjones, line 1: Given this is embedded in Janus as if it's our license, it seems like the copyright here should reflect now and ISRG. We probably should check on this.

src/dp/rand_bigint.rs
```diff
@@ -0,0 +1,101 @@
...
+use rand::Rng;
+
+/// Uniform distribution producing [`BigUint`].
+pub(super) struct UniformBigUint {
+    len: BigUint,
+    base: BigUint,
+}
+
+impl UniformBigUint {
+    pub(super) fn new(low: &BigUint, high: &BigUint) -> Result<Self, Error> {
```
tgeoghegan, line 32: Since the internal representation of this is `base, len` anyway, I think it'd be better for the arguments here to match that, because then construction would be infallible. However, since we're vendoring this code and already presumably have a bunch of usage of this pattern elsewhere, we should stick with what's already done.

tgeoghegan (approved):

PR #1292, opened 2025-07-08
Implement Prio3L1BoundSum

tgeoghegan:
Adds an implementation of Prio3L1BoundSum ([1]). Test vectors are copied from [2].

[1]: https://martinthomson.github.io/prio-l1-bound-sum/draft-thomson-ppm-l1-bound-sum.html
[2]: https://github.com/martinthomson/prio-l1-bound-sum/pull/12

Part of #1293


tgeoghegan:
A couple things to discuss:

- I put new modules under `flp::types` and `vdaf::prio3` for the L1BoundSum FLP and VDAF, respectively. We could just chuck them both in among the existing Prio3 variants. Since it comes from a different document, it made sense to me to keep them apart.
- Relatedly, we might want to gate the availability of this VDAF on a feature like `experimental`, but I think we have reasonable confidence that L1BoundSum is sound.
-  Finally, I've argued in the past that new VDAFs should be implemented outside of this crate. However this one is so closely related to SumVec, and uses many of the same internals, so we'd have to change the visibility of several items in `prio` if we moved `L1BoundSum` out of the crate, which seems like a hassle.

src/vdaf/prio3/l1boundsum.rs
```diff
@@ -0,0 +1,67 @@
...
+        num_aggregators: u8,
+        bits: usize,
+        len: usize,
+        chunk_length: usize,
+    ) -> Result<Self, VdafError> {
+        Prio3::new(
+            num_aggregators,
+            1,
+            // TODO: use real codepoint once assigned
+            0xFFFF0000,
```
divergentdave, line 35: The private use codepoint might be reason enough to put this behind the `experimental` feature for now.

src/flp/types/l1boundsum.rs
```diff
@@ -0,0 +1,402 @@
...
+    chunk_length: usize,
+    /// Number of gadget calls needed to evaluate validity.
+    gadget_calls: usize,
+
+    phantom: PhantomData<(S, F)>,
+}
+
+impl<F: NttFriendlyFieldElement, S> Debug for L1BoundSum<F, S> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        // TODO(timg): more fields
```
divergentdave, line 39: nit: cleanup
```suggestion
```

src/flp/types/l1boundsum.rs
```diff
@@ -0,0 +1,402 @@
...
+
+        let range_check = parallel_sum_range_checks(
+            &mut gadget[0],
+            input,
+            joint_randomness,
+            self.chunk_length,
+            num_shares,
+        )?;
+
+        // The encoded input consists of the measurement with a single field element appended,
```
divergentdave, line 154: This comment is a bit misleading, because there will be `bits` many field elements appended. Rather, the claimed weight is also encoded with one field element per bit, like the numbers in the original measurement.

src/flp/types/l1boundsum.rs
```diff
@@ -0,0 +1,402 @@
...
+        self.chunk_length * 2
+    }
+}
+
+impl<F, S> Type for L1BoundSum<F, S>
+where
+    F: NttFriendlyFieldElement,
+    S: ParallelSumGadget<F, Mul<F>> + Eq + 'static,
+{
+    /// The measurement is the input with the L1 norm appended.
```
divergentdave, line 202: This is incorrect, because the `Measurement` type comes before the AFE encoding. This is just the original vector of integers.

divergentdave (approved):

PR #1299, opened 2025-07-25
Remove recursive NTT implementation

divergentdave:
This removes the recursive NTT implementation, and ports over part of one test, comparing the output of the fast NTT implementation versus the summation definition of the NTT. We previously only tested the recursive implementation in this way, and then tested that the recursive and iterative implementations were equivalent.


src/ntt.rs
```diff
@@ -214,4 +196,27 @@ mod tests {
 
         assert_eq!(got, want);
     }
+
+    #[test]
+    fn test_ntt_interpolation() {
+        let count = 128;
+        let points = repeat_with(random).take(count).collect::<Vec<Field128>>();
```
tgeoghegan, line 203: nit: clippy always yells at me about doing `std::iter::repeat_n` instead of chaining `repeat_with` and `take`

tgeoghegan (approved):

src/ntt.rs
divergentdave: Actually, we have a `random_vector()` for just this

Issue #571, opened 2023-05-07
Prio2 optimization: use barycentric interpolation formula in `poly_interpret_eval()`


divergentdave:
## Background

For comparison, Prio3 proof messages include shares of gadget polynomial coefficients. (i.e., the gadget polynomial is sent in its monomial basis representation)

Prio2 proof messages include shares of evaluations of the gadget polynomial at odd 2n-th roots of unity, and separately the zeroth root of unity.  Verifiers assume that the gadget polynomial evaluates to zero on all other even 2n-th roots of unity, which holds for all valid inputs. Thus servers/verifiers represent the gadget polynomial with this list of evaluations on all 2n-th roots of unity. (i.e., it is in the Lagrange basis)

As discussed in cfrg/draft-irtf-cfrg-vdaf#63, polynomials can be directly evaluated at any point from a list of evaluations at other points (that is long enough to define the polynomial), using a 'barycentric interpolation formula'.

## Optimization

`poly_interpret_eval()` takes the evaluations of the gadget polynomial on the 2n-th roots of unity as input, along with a field element at which to evaluate the polynomial. Currently, it reconstructs the coefficients of the gadget polynomial first (i.e., does a change of basis), and then evaluates the polynomial at the given point using those coefficients. The inverse FFT operation to recover the coefficients should run in O(n log n) time, while both `poly_eval()` and the barycentric interpolation formula should both run in O(n) time. Thus, using the barycentric interpolation formula should be asymptotically better, and I'm curious to see how the two approaches compare in terms of concrete runtime at typical input lengths.

While Prio2 in ENPA is winding down, this could still be useful for the VDAF form of Prio2, plus it would be good to gain experience with this technique for further work on cfrg/draft-irtf-cfrg-vdaf#63.

divergentdave:
I prototyped this out really quick in commit 93e5acbf815e5d6cab22fb7381a6e0e4f65e1fd4 (motivated by #650) but unfortunately the concrete performance is 4x-5x slower on the benchmarks, asymptotic behavior aside. The runtime is dominated by the inverse needed for the division inside the sum, in $P(x)=\frac{x^N-1}{N} \sum_i \frac{y_i*\omega^i}{x-\omega^i}$. That inverse is implemented with a big exponentiation, and so we wind up doing way more multiplies than we would have done during the FFTs, at least for input sizes from 10 to 1000.

[...]

PR #1340, opened 2025-09-23
Faster polynomial evaluation in Lagrange basis with rhizomes algorithm.

armfazh:
Use of faster polynomial evaluation in the Lagrange basis from the rhizomes paper.
- It is applied to FLP.Query algorithm to evaluate several polynomials at a random value.
- In the paper, Algorithm 6 evaluates a single polynomial, while Alg. 7 is its batched version that is implemented.
- Savings are around 2x faster for PrepInit in all VDAFs.
- Unlike #571, our method uses no field inverses.

Reference: 
[link](https://eprint.iacr.org/2025/1727) Rhizomes and the Roots of Efficiency, Latincrypt 2025.

## Running Benchmarks

```sh
git checkout main
cargo bench --features experimental --bench speed_tests -- "prio3.*prepare_init" --quiet --save-baseline baseline

git checkout pr
cargo bench --features experimental --bench speed_tests -- "prio3.*prepare_init" --quiet --save-baseline rhizomes
critcmp baseline rhizomes
```

## Timings

Plot of the timings for PrepInit step.

comparison

### DIFF
--- a/src/polynomial.rs
+++ b/src/polynomial.rs
@@ -87,8 +87,10 @@ pub(crate) fn poly_eval_batched<F: FieldElement>(
     x: F,
 ) -> Vec<F> {
     let mut l = F::one();
-    let mut u = Vec::with_capacity(polynomials.len());
-    u.extend(polynomials.iter().map(|poly| poly[0]));
+    let mut u: Vec<F> = polynomials
+        .iter()
+        .map(|p| p.first().copied().unwrap_or_else(F::zero))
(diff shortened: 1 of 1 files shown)


divergentdave (commented):
This is looking good, just noted one edge case. Thanks so much for this work, the performance improvement is amazing!

jcjones (commented):
I've gone back through this and am good to merge with the existing review fixes in place.

PR #1349, opened 2025-10-28
Implement new range check technique

divergentdave:
This implements the technique from cfrg/draft-irtf-cfrg-vdaf#549 in Prio3Sum, Prio3SumVec, and Prio3MultihotCountVec. See spec PRs cfrg/draft-irtf-cfrg-vdaf#576 and cfrg/draft-irtf-cfrg-vdaf#581.

I updated the test vectors with files from those PRs, and this implementation matches (though I had to unwind some field name changes that haven't been adopted here yet). I implemented the common encoding/decoding step in a helper function, and used that in the implementation of each FLP type.

Note that I had to add some new trait bounds from the subtle crate to our `Integer` associated type, in order to make encoding constant time. I added them here instead of on the related `Integer` trait because they each imply a `Copy` bound. We previously left the `Copy` bound on this associated type, and not the `Integer` trait, because we reuse the trait in tests where we implement it for a non-`Copy` arbitrary precision integer type, to help with `GF(2^255-19)` tests.

I'm opening this as a draft for now until we arrive at a decision on the spec changes.

### DIFF
--- a/src/field.rs
+++ b/src/field.rs
@@ -71,6 +71,10 @@ pub enum FieldError {
     /// if the number of bits is larger than the bit length of the field's modulus.
     #[error("bit vector length exceeds modulus bit length")]
     BitVectorTooLong,
+    /// Returned when encoding an integer to "bitvector representation" if the value is greater than
+    /// the allowed maximum value.
+    #[error("value exceeds maximum allowed by bit vector representation")]
+    BitVectorOverflow,
 }
 
 /// Objects with this trait represent an element of `GF(p)` for some prime `p`.
@@ -254,7 +258,7 @@ pub trait FieldElementWithInteger: FieldElement + From<Self::Integer> {
         // it. The above check on `bits` ensures this shift won't panic due to the shift width
         // being too large.
         if input >> bits != Self::Integer::zero() {
-            return Err(FieldError::InputSizeMismatch);
+            return Err(FieldError::BitVectorOverflow);
         }
 
         Ok(BitvectorRepresentationIter {


tgeoghegan (approved):
This change LGTM compared to the spec changes in https://github.com/cfrg/draft-irtf-cfrg-vdaf/pull/576 and https://github.com/cfrg/draft-irtf-cfrg-vdaf/pull/581.

The awkward thing is that the test vectors don't correspond to draft-irtf-cfrg-vdaf-18, but those vectors also account for the Rhizomes work to evaluate polynomials in the Lagrange basis, which we don't have yet. So I think we should merge this as-is, and we'll sync up with the VDAF test vectors in the next change we take.

It's also unfortunate to have to disable the Mastic test vector tests, but we'd need that specification to update before we can do anything about it, and I don't want to hold up shipping draft-irtf-cfrg-vdaf-18's Prio3 on that.

src/flp/types.rs
```diff
@@ -943,13 +966,7 @@ where
...
         let mut flattened = Vec::with_capacity(self.flattened_len);
         for summand in measurement.iter() {
-            if summand > &self.max {
-                return Err(FlpError::Encode(format!(
-                    "summand exceeds maximum of 2^{}-1",
-                    self.bits
-                )));
-            }
-            flattened.extend(F::encode_as_bitvector(*summand, self.bits)?);
+            encode_range_checked_int(*summand, self.bits, self.last_weight, &mut flattened)?;
```
jcjones: nit: kind of a shame to trade the real clear `"summand exceeds maximum of 2^{}-1"` for a `FieldError::InputSizeMismatch`.

We could put back in an explicit bounds check, similar to how `Sum::encode_measurement` checks at line 245.

PR #1434, opened 2026-04-27
Upgrade RustCrypto dependencies

divergentdave:
This upgrades all the RustCrypto dependencies together, and adjusts to some API changes and removals.

### DIFF
--- a/src/prng.rs
+++ b/src/prng.rs
@@ -39,7 +39,11 @@ impl<F: FieldElement> Prng<F, SeedStreamAes128> {
     /// bytes of the seed are used, respectively, for the key and initialization vector for AES128
     /// in CTR mode.
     pub(crate) fn from_prio2_seed(seed: &[u8; 32]) -> Self {
-        let seed_stream = SeedStreamAes128::new(&seed[..16], &seed[16..]);
+        // Unwrap safety: these slices are of fixed size, so these conversions always succeed.
+        let seed_stream = SeedStreamAes128::new(
+            (&seed[..16]).try_into().unwrap(),
+            (&seed[16..]).try_into().unwrap(),
+        );
         Self::from_seed_stream(seed_stream)
     }
 }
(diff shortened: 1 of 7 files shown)


src/vdaf/xof.rs
```diff
@@ -172,7 +172,7 @@ pub struct SeedStreamAes128(Ctr64BE<Aes128>);
 impl SeedStreamAes128 {
     /// Construct an instance of the seed stream with the given AES key `key` and initialization
     /// vector `iv`.
-    pub fn new(key: &[u8], iv: &[u8]) -> Self {
+    pub fn new(key: &[u8; 16], iv: &[u8; 16]) -> Self {
```
tgeoghegan, line 175: Could these be `[u8; 16]` (values, not references)? That type is `Copy` meaning the ergonomics are better, if anything, for the caller, and I suspect `SeedStreamAes128::new` will copy the value anyway.

src/prng.rs
```diff
@@ -39,7 +39,11 @@ impl<F: FieldElement> Prng<F, SeedStreamAes128> {
     /// bytes of the seed are used, respectively, for the key and initialization vector for AES128
     /// in CTR mode.
     pub(crate) fn from_prio2_seed(seed: &[u8; 32]) -> Self {
```
tgeoghegan, line 41: As above, I wonder if this is better as a `[u8; 32]` value

tgeoghegan (approved):

jcjones (approved):
I've nothing to add.

src/vdaf/xof.rs
divergentdave: `KeyIvInit::new()` takes a couple references, `&Key<Self>` and `&Iv<Self>`, both of which are effectively `&hybrid_array::Array<u8, N>`. Making a change to owned arrays here would require more adaptations throughout the code base, so I'd prefer to keep it this way. FWIW, I dove into the AES-128 initialization, and the reference to the key array here actually gets passed all the way down to an `expand_key()` implementation, where it is cast into a pointer and then used in a SIMD load instruction (depending on the architecture). The IV does get copied into an array of `u64`s, with special endianness handling.

tgeoghegan: Ah, OK. I thought those `into()` calls were copying `&[u8; 16]` to `[u8; 16]`.

PR #1437, opened 2026-05-12
Remove "Vet Dependencies" workflow

tgeoghegan:
The audit backlog from `cargo vet` has become unsustainable and this process is largely obsolete given the advent of cheap, high quality vulnerability scanning in tools like Claude. In order to maintain forward velocity, crate `prio` will not require all dependencies to be audited on new release branches.

Pull requests targeting `release/0.16` will still require vetted dependencies.

### DIFF
--- a/.github/workflows/supply-chain.yml
+++ b/.github/workflows/supply-chain.yml
@@ -3,12 +3,10 @@ name: supply-chain
 on:
   push:
     branches:
-     - main
-     - release/**
+     - release/0.16
   pull_request:
     branches:
-     - main
-     - release/**
+     - release/0.16
   workflow_dispatch:
 
 jobs:


tgeoghegan:
Besides taking this change, we'll need to update the list of required checks in the branch protection rule on this repository. CI is going to fail here because this PR targets `main` and so "Vet Dependencies" will never run.

divergentdave (approved):

jcjones (approved):

tgeoghegan:
The branch protection rule has been updated. Note that since we have a single rule that applies to both `main` and `release/**`, "Vet Dependencies" will run and its status will be shown in the PR view, but success won't be required on that branch. I felt it wasn't worth defining a duplicate branch protection rule.

PR #1451, opened 2026-06-16
Update dependabot configuration

tgeoghegan:
Two changes to how we use dependabot in this repository:

- Since we no longer block merges to `main` on `cargo vet`, we can now enable the `dependabot-automerge` workflow we use elsewhere to reduce oncall toil. `DIVVIUP_GITHUB_AUTOMATION_DEPENDABOT_APPROVER_PAT` is an organization level secret, so it's already accessible by workflows in this repository.
- Configure dependabot cooldowns ([1]) to make it less likely we'll get caught out by a supply chain attack. Seven days seems like a reasonable starting value.

[1]: https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#cooldown-

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,6 +5,8 @@ updates:
     schedule:
       interval: "weekly"
     open-pull-requests-limit: 20
+    cooldown:
+      default-days: 7
     groups:
       rand:
         patterns:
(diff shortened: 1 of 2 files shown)


divergentdave (approved):

tgeoghegan:
Annoyingly, I think the only way to test this will be to merge it and see whether subsequent dependabot PRs go through. I also want to look into auditing our dependabot ignores, so hopefully if I cancel some of those, then that will cause some PRs to get created.

tgeoghegan:
This worked as expected in https://github.com/divviup/libprio-rs/pull/1452!

